PR #45, opened 2022-06-13
Add Tilt gen-api and gen-dashboard-client

jraddaoui:
Local resources to build the API generated code and the dashboard TS
client, triggered automatically on file changes.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -34,15 +34,50 @@ k8s_resource("opensearch-dashboards", port_forwards="7500:5601")
 
 k8s_resource("temporal-ui", port_forwards="7440:8080")
 
+local_resource(
+  "gen-api",
+  cmd="docker run \
+    --rm \
+    --user $(id -u):$(id -g) \
+    --volume $(pwd):/src \
+    fixl/goagen:3.7.5 gen \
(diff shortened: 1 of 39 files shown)


jraddaoui:
I kept the Makefile implementations in case they need to be executed manually without setting up Tilt.

Questions, should they be triggered automatically? If so, are they watching the correct paths?

Tiltfile
```diff
@@ -34,6 +34,25 @@ k8s_resource("opensearch-dashboards", port_forwards="7500:5601")
...
+    openapitools/openapi-generator-cli:v6.0.0 generate \
+      --input-spec /local/internal/api/gen/http/openapi.json \
+      --generator-name typescript-fetch \
+      --output /local/dashboard/src/openapi-generator/ \
+      --skip-validate-spec \
+      -p 'generateAliasAsModel=false' \
+      -p 'withInterfaces=true' \
+      -p 'supportsES6=true' \
+  ",
+  deps=["internal/api/gen/http/openapi.json"],
```
jraddaoui, line 52: Do we need to watch for changes in other files or just in `http/openapi.json` file?

PR #73, opened 2022-07-06
Small fixes

sevein:


jraddaoui (approved):

internal/api/api.go
```diff
@@ -121,6 +125,34 @@ func versionHeaderMiddleware(version string) func(http.Handler) http.Handler {
...
+func recoverMiddleware(logger logr.Logger) func(http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			defer func() {
+				if r := recover(); r != nil {
+					var msg string
+					switch x := r.(type) {
+					case string:
+						msg = fmt.Sprintf("panic: %s", x)
+					case error:
```
jraddaoui: Can we put both in the same case?

sevein: Fixed, since fmt.Sprintf is happy taking an `interface{}`.

PR #83, opened 2022-07-11
Add more dashboard tests and CI workflow

sevein:


dashboard/src/stores/package.ts
```diff
@@ -197,3 +202,13 @@ function handlePackageLocationUpdated(
...
+
+class UIRequest {
+  inner: number = 0;
+  request() {
+    this.inner++;
+  }
+  watch() {
+    return this.inner;
+  }
+}
```
jraddaoui: I think we could use arrow functions here:

```js
class UIRequest {
  inner: number = 0;
  request = () => this.inner++;
  watch = () => this.inner;
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#arrow_functions_used_as_methods
https://stackoverflow.com/questions/51464318/arrow-function-vs-class-methods-memory-footprint

dashboard/src/stores/package.ts
```diff
@@ -17,6 +17,11 @@ export const usePackageStore = defineStore("package", {
 
     // A list of packages shown during searches.
     packages: [] as Array<api.EnduroStoredPackageResponseBody>,
+
+    // User-interface interactions between components.
+    ui: {
+      download: new UIRequest(),
+    },
```
jraddaoui, line 25: I guess we could use and watch a property directly, what's the reason for a class with a property and different methods?

dashboard/src/components/PackageDetailsCard.vue
```diff
@@ -15,6 +15,11 @@ const download = () => {
 const stored = computed(() => {
   return packageStore.current?.aipId?.length;
 });
+
+watch(
+  () => packageStore.ui.download.watch(),
+  () => download()
+);
```
jraddaoui: Does the source need to be a function? Could it be `packageStore.ui.download.watch()` directly?

jraddaoui (approved):
Really nice! Added a couple of questions and a suggestion, but LGTM!

PR #84, opened 2022-07-12
Update preservation actions API endpoint

replaceafill:


internal/workflow/processing.go
```diff
@@ -503,6 +509,20 @@ func (w *ProcessingWorkflow) waitForReview(ctx temporalsdk_workflow.Context) (*p
 }
 
 func (w *ProcessingWorkflow) transferA3m(sessCtx temporalsdk_workflow.Context, tinfo *TransferInfo) error {
+	var paID uint
+	{
+		ctx := withLocalActivityOpts(sessCtx)
+		err := temporalsdk_workflow.ExecuteLocalActivity(ctx, createPreservationActionLocalActivity, w.pkgsvc, &createPreservationActionLocalActivityParams{
+			Name:       "Create AIP", // XXX: move to a translatable constant?
+			WorkflowID: tinfo.WorkflowID,
```
sevein: Should we load the identifier using `temporalsdk_workflow.GetInfo(ctx)`?

internal/workflow/local_activities.go
```diff
@@ -76,14 +76,63 @@ func setLocationLocalActivity(ctx context.Context, pkgsvc package_.Service, pkgI
...
+	Location    string
+	WorkflowID  string
+	Status      package_.PreservationTaskStatus
+	StartedAt   time.Time
+	CompletedAt time.Time
+}
+
+func saveLocationMovePreservationActionLocalActivity(ctx context.Context, pkgsvc package_.Service, params *saveLocationMovePreservationActionLocalActivityParams) error {
+	paID, err := createPreservationActionLocalActivity(ctx, pkgsvc, &createPreservationActionLocalActivityParams{
+		Name:        "Move package", // XXX: move to a translatable constant?
```
sevein, line 90: Can we use `TODO` instead of `XXX`?

internal/workflow/processing.go
```diff
@@ -503,6 +509,20 @@ func (w *ProcessingWorkflow) waitForReview(ctx temporalsdk_workflow.Context) (*p
 }
 
 func (w *ProcessingWorkflow) transferA3m(sessCtx temporalsdk_workflow.Context, tinfo *TransferInfo) error {
+	var paID uint
+	{
+		ctx := withLocalActivityOpts(sessCtx)
+		err := temporalsdk_workflow.ExecuteLocalActivity(ctx, createPreservationActionLocalActivity, w.pkgsvc, &createPreservationActionLocalActivityParams{
+			Name:       "Create AIP", // XXX: move to a translatable constant?
```
sevein, line 515: We could also use an application enum.

PR #93, opened 2022-07-12
Unmount components after testing

sevein:
I've learned the hard way that `unmount` is necessary when testing components, otherwise you may be mounting the same component from multiple tests and start seeing some weird results in your assertions. This pull request also adds a test that was flaky because unmount was not being used.

### DIFF
--- a/dashboard/test/components/PackageDetailsCard.test.ts
+++ b/dashboard/test/components/PackageDetailsCard.test.ts
@@ -3,12 +3,12 @@ import PackageDetailsCard from "../../src/components/PackageDetailsCard.vue";
 import { usePackageStore } from "../../src/stores/package";
 import { createTestingPinia } from "@pinia/testing";
 import { render } from "@testing-library/vue";
-import { flushPromises } from "@vue/test-utils";
 import { expect, describe, it, vi } from "vitest";
+import { nextTick } from "vue";
 
 describe("PackageDetailsCard.vue", () => {
   it("watches download requests from the store", async () => {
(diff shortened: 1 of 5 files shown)


dashboard/test/components/PackageDetailsCard.test.ts
```diff
@@ -31,13 +31,15 @@ describe("PackageDetailsCard.vue", () => {
     // Someone requests the download of the AIP via the package store.
     const packageStore = usePackageStore();
     packageStore.ui.download.request();
-    await flushPromises();
+    await nextTick();
```
sevein, line 34: We'll need nextTick or flushPromises depending on the situation, see https://test-utils.vuejs.org/guide/advanced/async-suspense.html for more.

dashboard/test/components/PackageLocationCard.test.ts
```diff
@@ -35,10 +36,57 @@ describe("PackageLocationCard.vue", () => {
...
+    });
+
+    vi.mock("../../src/dialogs", () => {
+      return {
+        openPackageLocationDialog: () => "perma-aips-2",
+      };
+    });
+
+    const button = getByText("Choose storage location");
+    await fireEvent.click(button);
```
sevein, line 81: Using await here ensures that we wait for the next tick, see https://github.com/testing-library/vue-testing-library/blob/4a663b9a265c4e2a51e0b550c8d1809169bd1daf/src/fire-event.js#L4-L12.

jraddaoui (approved):

PR #95, opened 2022-07-13
Add package status legend

sevein:
This pull request adds a new PackageListLegend component that uses [`v-model (form input binding)`](https://vuejs.org/guide/essentials/forms.html) so the state of the legend being shown is bound to a property that the parent can control.

Peek 2022-07-13 19-10


dashboard/src/pages/packages/index.vue
```diff
@@ -29,7 +35,10 @@ const openPackage = (id: number) => {
           <th scope="col">UUID</th>
           <th scope="col">Started</th>
           <th scope="col">Location</th>
-          <th scope="col">Status</th>
+          <th scope="col">
```
jraddaoui: We could use `class="text-nowrap"` to avoid the line break in smaller widths. We may want to do it for all table headers too ... 

```css
th {
  white-space: nowrap;
}
```

jraddaoui (approved):

dashboard/src/pages/packages/index.vue
sevein: :bulb: thanks!

PR #139, opened 2022-07-22
Cache Go modules in CI

sevein:


sevein:
From ~2'30" to ~1'10", pretty good!

image

PR #140, opened 2022-07-22
Use posva/unplugin-vue-router

sevein:
This pull request replaces [vite-plugin-pages](https://github.com/hannoeru/vite-plugin-pages) with [unplugin-vue-router](https://github.com/posva/unplugin-vue-router). The latter is better integrated with vue-router and developed by the same team.

This plugin dynamically builds a router based on the filesystem structre sitting under `src/pages`. It generates types for the routes added to the filesystem so they can be consumed when creating router links and similar operations with the router.


dashboard/src/components/Breadcrumb.vue
```diff
@@ -10,12 +10,12 @@ const layoutStore = useLayoutStore();
...
         class="breadcrumb-item"
-        :class="!item.routeName ? 'active' : ''"
-        :aria-current="!item.routeName ? 'page' : undefined"
+        :class="!item.route ? 'active' : ''"
+        :aria-current="!item.route ? 'page' : undefined"
       >
         <router-link
-          :to="{ name: item.routeName }"
-          v-if="item.routeName"
+          :to="item.route"
```
Diogenesoftoronto, line 17: should this be { name: 'item.route' }?

dashboard/src/components/Sidebar.vue
```diff
@@ -51,7 +54,7 @@ onMounted(() => {
               class="d-block py-3 text-decoration-none sidebar-link"
               active-class="active"
               exact-active-class="exact-active"
-              :to="{ name: item.routeName }"
+              :to="item.route"
```
Diogenesoftoronto, line 57: same here

Diogenesoftoronto (commented):
Looks good just had some questions

dashboard/src/components/Breadcrumb.vue
sevein: I changed the breadcrumb store to contain the actual route object as opposed to the route name (string) so I think this is fine. The breadcrumb seems to be working propery.

dashboard/src/components/Sidebar.vue
sevein: Similarly, `menuItems` now contains the actual route object so we don't need to build a new one after the string.

Diogenesoftoronto (approved):

PR #141, opened 2022-07-25
Add k8s cluster data flush

jraddaoui:
Add k8s jobs to recreate the MySQL databases and MinIO buckets and
delete the Opensearch index. Add a Tilt UI button to trigger them and
use them in the GHA workflow to update the Pulumi stack.


sevein:
Noice!

PR #209, opened 2022-08-21
Goa updates

sevein:
I've updated Goa to v3.8.3. Raphael kindly fixed the issues that we reported - we can now use OneOf in the event streaming method and all nested types are generated successfully, without build errors nor missing views!
Unfortunately, the fact that the OneOf maps to JSON just as a nested JSON-encoded string means that the event types are not described in the OpenAPI spec hence we lose all the TypeScript interfaces produced by OpenAPI Generator like EnduroPackageLocationUpdatedEventResponseBody, etc... [`package.ts`](https://github.com/artefactual-sdps/enduro/blob/33585aab3490f1681237878a99ec8fb9cd22d69f/dashboard/src/stores/package.ts#L186-L197) relies on them, so that's unfortunate. Not sure if there's a good solution, will see!

Unrelated, I've also used `"type:generate:force"` to generate unused types like Package or Location. This allows us to use them in the persistence layer. I've tried and it works but it's also not great because we have to convert back to *string in most cases (uuid.UUID, enums, etc...). Not sure if this is what we want, maybe it's good enough until we adopt protobuf types.


sevein:
A significant improvement could be to adopt [`Meta("struct:field:name")`](https://github.com/goadesign/goa/blob/v3.8.3/dsl/meta.go#L207) where we want custom types, that'd be :100: if it works.

Update: using uuid.UUID in Goa works!

PR #277, opened 2022-09-12
Use golangci-lint binary

sevein:
This pull request updates the use of the golangci-lint tool so it uses the official binary instead of building it from the sources. The latter method has been working inconsistently for us and it's not recommended by the developers.

See https://golangci-lint.run/usage/install/#install-from-source.

image

In CI, I've used golangci-lint-action which is doing some heavy caching for better performance.
Locally, I've used our Makefile to download the binary when needed.


sevein:
TODO: Don't use go install, use curl instead.

sevein:
I've changed the target to make it download the binary from GitHub. Things to test:

* `make lint` still works
* `ls ~/.cache/enduro-sdps/Linux/x86_64/bin/golangci-lint` finds the symlink

replaceafill (approved):
Seems to work for me :+1:

sbreker (approved):
Nice improvement!

PR #279, opened 2022-09-13
Disable Dependabot version updates

sevein:
This pull request suggests that we disable [Dependabot](https://docs.github.com/en/code-security/dependabot) alerts generated after version updates by setting `open-pull-requests-limit` to zero, mostly because they're too frequent and distracting. Dependabot will still let us know about security updates.

I've added a document that describes the manual process, which could use more refinements but it gives you an idea. This document includes a section describing a potential automated process where we run dependabot-core in some other way where version updates are reported differently, e.g. weekly via Slack. A good starting point could be this workflow: https://github.com/dependabot/dependabot-script/blob/main/manual-github-actions.yaml.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,19 +4,24 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    open-pull-requests-limit: 0
   - package-ecosystem: "gomod"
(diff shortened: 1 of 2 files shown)


sevein:
Provide automated way to report version updates.

PR #281, opened 2022-09-14
Use OneOf in package events

sevein:
This is a long-standing pull request where I attempted to make use of OneOf in Goa but I've been running into various blockers that were reported to Goa and releases were made as a result. I still think that there may be another bug but it's not a blocker, read more below.

##### How to test

1. Confirm that the UI continuous receiving updates asynchronously, i.e. you don't need to refresh the page to receive data changes.
2. Open the Network tab in your browser and confirm that the WS channel is open and you see messages being received.

##### Description

This pull request introduces the use of [OneOf](https://pkg.go.dev/goa.design/goa/v3@v3.14.0/dsl#OneOf) for a more efficient way of expressing the monitor event data. In our current design of the type [type](https://github.com/artefactual-sdps/enduro/blob/e68a7c750e27f27a6a1fc5e0056230c626a3635a/internal/api/design/package_monitor.go#L7-L56) we simply list a set of nullable attributes. In other words, we have a message with many fields where at most one field will be set at the same time. Enforcing this behavior with OneOf is a good practice.

When connected to the WebSocket, messages look as follows:

```json
{"event":{"Type":"monitor_ping_event","Value":"{\"Message\":\"Ping\"}"}}
```

It is more succint if you compare it with the previous message:

```json
{"MonitorPingEvent":{"Message":"Ping"},"PackageCreatedEvent":null,"PackageUpdatedEvent":null,"PackageStatusUpdatedEvent":null,"PackageLocationUpdatedEvent":null,"PreservationActionCreatedEvent":null,"PreservationActionUpdatedEvent":null,"PreservationTaskCreatedEvent":null,"PreservationTaskUpdatedEvent":null}
```

What's interesting about the new structure generated by Goa:
1. The event is JSON-encoded but also the inner value is encoded again,
2. Instead of using snake_case like in other areas of the API, this uses SnakeCase, e.g. I would expect "Message" to read as "message". I'm solving that problem on the client for now as follows:

```ts
handleEvent(event: api.MonitorEventEvent) {
  const json = JSON.parse(event.value);
  const value = mapKeys(json, (value, key) => snakeCase(key));
  handlers[event.type](value);
}
```

mapKeys is used to convert attributes from snakeCase to snake_case. I suspect this is a bug in Goa, but I have started a conversation with the maintainers to confirm whether that's the case or I'm just doing something wrong. Generally, Goa encodes JSON using snake_case for attributes, so this is definitely an anomaly. It's easy for me to do the conversion from the client, but the generated spec reads that should be snake_case. This is good enough until we figure out if we can address it on the backend.


sevein:
I had previously paused work on this pull request, but I am planning to revisit it. A big change in Goa's OneOf is that it uses JSON within JSON to encode the payloads. This will require additional development on the frontend to handle it.

Diogenesoftoronto:
Taking a look now

dashboard/src/client.ts
```diff
@@ -52,8 +51,10 @@ function connectPackageMonitor() {
   const socket = new WebSocket(url);
   socket.onmessage = (event: MessageEvent) => {
     const body = JSON.parse(event.data);
-    const data = api.MonitorResponseBodyFromJSON(body);
-    store.handleEvent(data);
+    const data = api.MonitorEventFromJSON(body);
+    if (data.event) {
+      store.handleEvent(data.event);
```
Diogenesoftoronto, line 56: handle event seems a bit dangerous. Should we not handle the failure case?

dashboard/src/stores/package.ts
```diff
@@ -66,15 +68,10 @@ export const usePackageStore = defineStore("package", {
...
-      let key: keyof api.MonitorResponseBody;
-      for (key in event) {
-        const payload: any = event[key];
-        if (!payload) continue;
-        const handler = handlers[key];
-        handler(payload);
-        break;
-      }
+    handleEvent(event: api.MonitorEventEvent) {
+      const json = JSON.parse(event.value);
```
Diogenesoftoronto, line 72: Parse can return errors if event.value is not valid json.

Diogenesoftoronto (approved):
Given that the blob is checked for errors before being sent to the front end I leave it to you to decide whether or not to do check whether bullet proofing a the parse is okay. There does seem to be a small risk with no client side check that the json is valid.

dashboard/src/client.ts
sevein: Currently, I'm relying on the browser's default behavior: halting the event processing and logging the error when an issue arises. If we were to introduce a try-catch block, what additional actions should we take to handle the error? Since I had no plans beyond deferring to the browser's default behavior, I thought it wasn't necessary to include it.

sevein: I'm going to merge, but if you come up with ideas let me know!

Diogenesoftoronto: I was just considering what we should do for the user experience. Error logs in the devtools isn't exactly ideal. But I guess this is good enough for now especially given how unlikely this is.

sevein: Yeah that makes sense, maybe at some point we could build some kind of indicator or dialog raising connectivity issues. Let's keep that in mind.

PR #463, opened 2023-05-29
Update Go alpine image version to 1.20.4

djjuhasz:


jraddaoui:
We should update the version used in CI and maybe the one in go.mod.

PR #464, opened 2023-05-30
Fix ent client tests for `created_at` fields

djjuhasz:
Use UTC date/time for fake `created_at` field tests to match expected values.


jraddaoui (approved):
But the tests passed in https://github.com/artefactual-sdps/enduro/actions/runs/5116366560/jobs/9198461911, why is this needed?

djjuhasz:
> But the tests passed in https://github.com/artefactual-sdps/enduro/actions/runs/5116366560/jobs/9198461911, why is this needed?

@jraddaoui the tests were failing in my local dev environment. I don't know if the different results are due to me  running Go 1.20.4 locally vs. Go 1.19.9 in the job, or because of some other difference in the test environment. :confused:

PR #126, opened 2022-07-19
Bring defaults from the vuejs/create-vue project 

sevein:
Already included in previous PR: 66f8dfb43cbce62bb6af6a4d21886f08df21852a.

* [x] @vue/tsconfig
  * openapi-generator issues in stric tmode
* [x] eslint
  * eslint-plugin-vue not fully compatible with `<script setup>` and Vue 3.2
* [ ] cypress

PR #475, opened 2023-06-01
Bump dashboard deps

sevein:
Backward-incompatible changes:
- In Vue 3.3 reactivity transformations are deprecated, now using `@vue-macros/reactivity-transform` instead.
- In Bootstrap 5.3, needed to import `variables-dark.scss` because it's needed by `_maps.scss`.
- In TypeScript 5.0, a small assert broke in PreservationActionCollapse.test.ts, not sure why - just deleted it for now.


jraddaoui (approved):
Things move fast, thanks @sevein! Some linting issues but it looks good to me.

sevein:
Prettier doens't like the double $$ we use in reactivity transforms and replaces it with a single $.

I've tried not upgrading prettier but the problem persists.

Maybe we need to stop using prettier and just move to eslint instead for linting.

sevein:
```
$ npx prettier --write .
$ git diff .
diff --git a/dashboard/src/components/PreservationActionCollapse.vue b/dashboard/src/components/PreservationActionCollapse.vue
index 0a9785a9..391f5ed8 100644
--- a/dashboard/src/components/PreservationActionCollapse.vue
+++ b/dashboard/src/components/PreservationActionCollapse.vue
@@ -42,13 +42,13 @@ const hide = () => {
   shown = false;
 };
 
-watch($$(toggleAll), () => {
+watch($(toggleAll), () => {
   if (toggleAll === null) return;
   toggleAll ? show() : hide();
 });
 
 let expandCounter = $ref<number>(0);
-watch($$(expandCounter), () => show());
+watch($(expandCounter), () => show());
 </script>
 
 <template>
diff --git a/dashboard/src/components/UUID.vue b/dashboard/src/components/UUID.vue
index c1adf0b1..3e9a27e9 100644
--- a/dashboard/src/components/UUID.vue
+++ b/dashboard/src/components/UUID.vue
@@ -15,7 +15,7 @@ const { copy, copied, isSupported } = useClipboard({ source });
 const el = $ref<HTMLElement | null>(null);
 let tooltip: Tooltip | null = null;
 
-watch($$(el), () => {
+watch($(el), () => {
   if (el) tooltip = new Tooltip(el);
 });
```

Diogenesoftoronto:
After a thorough review of the Prettier configuration options and Vue's reactivity transform documentation, I have found no setting that addresses this particular issue. The relevant resources I consulted are listed below:

- [Prettier configuration documentation](https://prettier.io/docs/en/configuration.html#configuration-schema)
- [Vue 3 reactivity transform documentation](https://v3.vuejs.org/guide/reactivity.html#declarative-reactivity-with-watch)

I also searched through Github but couldn't find a solution for this specific issue.

As a workaround, we could potentially set Prettier to ignore Vue files and subsequently run them through ESLint. However, my attempt at selectively ignoring the $$ expressions with Prettier failed. Here is the snippet where I tried to implement it:

```js
// prettier-ignore
watch($(toggleAll), () => {
  if (toggleAll === null) return;
  toggleAll ? show() : hide();
});
```
Despite the `// prettier-ignore` comment, the code still ended up formatted. 

After further research, I found [eslint-plugin-prettier-vue](https://github.com/meteorlxy/eslint-plugin-prettier-vue) which uses ESLint specifically for Vue tags. This could potentially be a viable solution. However, the last time it had commits was June 2022 and it looks unmaintained. Though it might work regardless it may break if vue changes and we need to update our dependencies again.

However, we are presented with a conundrum. We risk losing the strictness and benefits that Prettier provides in other files if we decide to remove it entirely. We have two alternatives:

1. We could use the [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) plugin instead and configure the exceptions ourselves. This might be a superior choice considering it is under active development and maintained by the Prettier team.

2. Continue to use Prettier, but manage the issues with Vue on a case by case basis. This involves either accepting the formatting or using the ignore comments where needed.

Diogenesoftoronto:
I just realized that prettier ignore is failing because the ast in a vue file works differently and uses the `<!-- prettier-ignore -->` html comments. This works but ignores the entire script tag. This seems like a bad option.
```vue
<!-- prettier-ignore -->
<script setup lang="ts">
//...
watch($$(el), () => {
  if (el) tooltip = new Tooltip(el);
});
</script>
```

^ works

sevein:
Thanks, @Diogenesoftoronto. I've updated the `npm run lint` step so its status doesn't affect the dashboard job in CI, so I can merge this as soon as possible. We can ignore the linter until we move to eslint as you're suggesting.

Ideally we would be using the default configs provided by [create-vue](https://github.com/vuejs/create-vue).

E.g. I've started a new project with `npm create vue@3` with eslint enabled and prettier disabled. These are the contents of `.eslintrc.cjs`:

```js
/* eslint-env node */
require('@rushstack/eslint-patch/modern-module-resolution')

module.exports = {
  root: true,
  'extends': [
    'plugin:vue/vue3-essential',
    'eslint:recommended',
    '@vue/eslint-config-typescript'
  ],
  parserOptions: {
    ecmaVersion: 'latest'
  }
}
```

The `scripts.lint` section in `package.json` is a also a little different:

    "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix --ignore-path .gitignore"
    
I started a branch a while ago to bring some other defaults from create-vue (https://github.com/artefactual-sdps/enduro/pull/126), but I think we should just focus on eslint for now.

PR #492, opened 2023-06-07
Make default Tilt trigger "manual"

djjuhasz:
Make the Tilt "trigger mode" default "manual", with the option to override it by setting a TRIGGER_MODE_AUTO environment variable.


Diogenesoftoronto (approved):

sevein (approved):
I agree "manual" is a better default. I think it was only @jraddaoui that had a pretty beefy machine to the point that he didn't notice slowdowns during the builds. It was quite the opposite for me, I could feel it just in the overall desktop responsiveness.

PR #493, opened 2023-06-07
Fix swagger.json in the Tilt HTTP server

djjuhasz:
- In the Dockerfile copy the Goa generated `openapi.json` file to the `/home/enduro/static/` directory in the Enduro image
- Update `design/swagger.go` to proxy the `/swagger/swagger.json` URL to the `openapi.json` file in the Docker image


Diogenesoftoronto (approved):

sevein (approved):
I'm not sure exactly what you mean by "Tilt HTTP server", would that be the equivalent to the Enduro API server?

Thanks for the fix! I think down the road is probably better to [embed](https://pkg.go.dev/embed) the schema so the binary that we built self-contained and doesn't depend on a specific filesystem layout.

djjuhasz:
> I'm not sure exactly what you mean by "Tilt HTTP server", would that be the equivalent to the Enduro API server?

Yes, I meant getting `/swagger/swagger.json` to work in the Enduro API server in the Tilt dev environment.  I had a hard time trying to summarize the change in 52 chars and it sounds like the final result isn't very clear. :laughing:  I'll rewrite the commit message.
 
> Thanks for the fix! I think down the road is probably better to [embed](https://pkg.go.dev/embed) the schema so the binary that we built self-contained and doesn't depend on a specific filesystem layout.

Yeah, that's a good idea.  :thinking: I'll close this PR and work on the embedded solution instead.

PR #494, opened 2023-06-07
Evaluating Redis Pop Operation and Dead Letter Queue Utility

Diogenesoftoronto:
@djjuhasz and I made improvements to the Redis pop operation to transform it into a more 'reliable' queue. Events are now moved to a secondary list (dead-letter queue or DLQ) that can be monitored. However, I have identified a few concerns we should address:

1. **Logging DLQ Size**: I attempted to use `logger.v2.info` to log the number of items in the DLQ, but this approach consistently caused test failures. For now, I've removed this logging to prevent the panics, but this feels like a band-aid rather than a solution. We should explore why this is happening and how we can log the DLQ size without causing test failures.

2. **Evaluating the Utility of the DLQ**: The introduction of the DLQ has added a considerable amount of dependency injection to our code, particularly for monitoring purposes. This has complicated our codebase somewhat, and I'm not sure the benefits we're getting from the DLQ justify this added complexity. We aren't currently using the DLQ in our preprocessing or in Enduro, so we should reconsider if it's necessary.

3. **Handling Errors in Pop Operation**: The `pop` function currently moves an event from the main list to the DLQ, tries to process it, and removes it from the DLQ if the processing succeeds. However, we're not handling the scenario where processing fails:

```go
event, err := w.event(val)
if err != nil {
    // How should we handle this scenario? 
    return nil, fmt.Errorf("error processing item received: %w", err)
}
```
Instead we are just logging it as an error, but I have a suggestion that might prove more useful but I am unsure how to procede with it. Should we consider throwing a custom error that cause us to retry proccessing of items in the DLQ? 

The second suggestion I have to make the DLQ more meaningful in enduro is to send the list to the frontend or for it to be passed to the application to be retried on something like a better network connection etc. This second suggestion is more wonky because I am not sure exactly what we might do with the DLQ which might mean that it isn't worth going through with.

Thanks!

### DIFF
--- a/internal/watcher/config.go
+++ b/internal/watcher/config.go
@@ -41,17 +41,18 @@ type FilesystemConfig struct {
 
 // See minio.go for more.
 type MinioConfig struct {
-	Name         string
-	RedisAddress string
-	RedisList    string
-	Region       string
-	Endpoint     string
-	PathStyle    bool
-	Profile      string
-	Key          string
-	Secret       string
-	Token        string
-	Bucket       string
+	Name            string
+	RedisAddress    string
+	RedisList       string
+	RedisFailedList string
+	Region          string
+	Endpoint        string
+	PathStyle       bool
+	Profile         string
+	Key             string
+	Secret          string
+	Token           string
+	Bucket          string
 
 	RetentionPeriod  *time.Duration
 	StripTopLevelDir bool
--- a/internal/watcher/minio.go
+++ b/internal/watcher/minio.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/go-logr/logr"
 	"github.com/go-redis/redis/v8"
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/s3blob"
@@ -17,10 +18,12 @@ import (
 
 // minioWatcher implements a Watcher for watching lists in Redis.
 type minioWatcher struct {
-	client   *redis.Client
-	sess     *session.Session
-	listName string
-	bucket   string
+	client     *redis.Client
+	sess       *session.Session
+	logger     logr.Logger
+	listName   string
+	failedList string
+	bucket     string
 	*commonWatcherImpl
 }
 
@@ -33,7 +36,7 @@ var _ Watcher = (*minioWatcher)(nil)
 
 const redisPopTimeout = time.Second * 2
 
-func NewMinioWatcher(ctx context.Context, config *MinioConfig) (*minioWatcher, error) {
+func NewMinioWatcher(ctx context.Context, logger logr.Logger, config *MinioConfig) (*minioWatcher, error) {
 	opts, err := redis.ParseURL(config.RedisAddress)
 	if err != nil {
 		return nil, err
@@ -59,12 +62,17 @@ func NewMinioWatcher(ctx context.Context, config *MinioConfig) (*minioWatcher, e
 	if err != nil {
 		return nil, err
 	}
+	if config.RedisFailedList == "" {
+		config.RedisFailedList = config.RedisList + "-dead"
+	}
 
 	return &minioWatcher{
-		client:   client,
-		sess:     sess,
-		listName: config.RedisList,
-		bucket:   config.Bucket,
+		client:     client,
+		sess:       sess,
+		listName:   config.RedisList,
+		failedList: config.RedisFailedList,
+		logger:     logger,
+		bucket:     config.Bucket,
 		commonWatcherImpl: &commonWatcherImpl{
 			name:             config.Name,
 			retentionPeriod:  config.RetentionPeriod,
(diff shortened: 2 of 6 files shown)


internal/watcher/config.go
```diff
@@ -41,17 +41,18 @@ type FilesystemConfig struct {
...
-	PathStyle    bool
-	Profile      string
-	Key          string
-	Secret       string
-	Token        string
-	Bucket       string
+	Name           string
+	RedisAddress   string
+	RedisList      string
+	RedisDeadQueue string
```
sevein: For consistency with `RedisList`, should we call it `RedisDLQList`, or `RedisFailedList` or similar?

internal/watcher/minio_test.go
```diff
@@ -40,7 +41,8 @@ func newWatcher(t *testing.T) (*miniredis.Miniredis, watcher.Watcher) {
 	}
 
 	var w watcher.Watcher
-	w, err = watcher.NewMinioWatcher(context.Background(), &config)
+	var logger logr.Logger
+	w, err = watcher.NewMinioWatcher(context.Background(), &config, logger)
```
sevein: > Logging DLQ Size: I attempted to use logger.v2.info to log the number of items in the DLQ, but this approach consistently caused test failures. For now, I've removed this logging to prevent the panics, but this feels like a band-aid rather than a solution. We should explore why this is happening and how we can log the DLQ size without causing test failures.

The zero value of `loggger` panics when it's used. You should instead use `logger := logr.Discard()` or update logr to v1.2.4 where I believe they've changed the zero value so it doesn't panic.

internal/watcher/watcher.go
```diff
@@ -80,13 +81,13 @@ type serviceImpl struct {
 
 var _ Service = (*serviceImpl)(nil)
 
-func New(ctx context.Context, c *Config) (*serviceImpl, error) {
+func New(ctx context.Context, c *Config, logger logr.Logger) (*serviceImpl, error) {
```
sevein: I think that there is a pattern in this project where the logger is listed first, or second when there's a context, then the rest of parameters. That said, I think it should be: `New(ctx context.Context, logger logr.Logger, c *Config)`.

internal/watcher/minio.go
```diff
@@ -91,15 +99,19 @@ func (w *minioWatcher) Path() string {
...
-	event, err := w.event(val[1])
+	event, err := w.event(val)
 	if err != nil {
-		return nil, fmt.Errorf("ereror processing item received: %w", err)
+		return nil, fmt.Errorf("error processing item received: %w", err)
+	}
+
+	if _, err := w.client.LRem(ctx, w.deadQueue, 1, val).Result(); err != nil {
+		w.logger.Error(err, "Error removing message from failed list.", "list", w.deadQueue)
 	}
```
sevein: The problem with this approach is that `w.event` is only decoding the message, it's not triggering the processing workflow. That's happening [here](https://github.com/artefactual-sdps/enduro/blob/b76841622bcb0e6b7aa9c91efa1a5ccf73210540/main.go#L272C9-L274).
This means that you're removing the message from the DLQ before you have confirmed that Temporal has started our workflow - many things could go wrong, e.g. Temporal is down, unreachable, etc... Ideally LREM is deferred for after the the workflow is started.

One possible solution would be to return a callable so the caller can call it immediately after the fact, e.g.:

```go
type ACK func() error

type Watcher interface { // ... in watcher.go
    Watch(ctx context.Context) (*BlobEvent, ACK, error)
}
```

Another alternative is what we did for STE where the watcher receives the Temporal client and knows how to start the workflow.

sevein (changes requested):
Looking good. I've suggested some changes.

internal/watcher/config.go
Diogenesoftoronto: I have decided to call it failedlist to keep it consistent with preprocessing

internal/watcher/minio_test.go
Diogenesoftoronto: For now I have decided to use discard and will update logger in a separate pr

internal/watcher/minio.go
sevein: @Diogenesoftoronto this is something you may want to address in a separate pull request too, after all moving away from blpop is already a step forward since that redis command is deprecated. Ensureing that the message isn't lost and is processed at least once can be done later.

Diogenesoftoronto: Okay. I was trying to tackle this. I will merge this if you think it is ready and open a separate pr for what I have.

internal/watcher/minio.go
```diff
@@ -91,15 +99,21 @@ func (w *minioWatcher) Path() string {
...
+	event, err := w.event(val)
 	if err != nil {
-		return nil, fmt.Errorf("ereror processing item received: %w", err)
+		return nil, fmt.Errorf("error processing item received: %w", err)
+	}
+
+	if c, err := w.client.LRem(ctx, w.failedList, 1, val).Result(); err != nil {
+		w.logger.Error(err, "Error removing message from failed list.", "list", w.failedList)
+	} else {
+		w.logger.V(2).Info("Successfully removed %d messages", "amount", c)
```
sevein: I've been logging full sentences with the ending period and all, does that make sense? When I see the ending period missing then I can only think it's trying to be a heading. Not sure if the plural form is useful, it's always going to be `1` unless it errors.

sevein (approved):
Just one more minor comment but approving now.

PR #495, opened 2023-06-08
Embed generated openapi.json in Go binary

djjuhasz:
Embed `openapi.json` in the Go binary so it can be served via the
Enduro API as `/swagger/swagger.json`.


djjuhasz:
@sevein I tried embedding the OpenAPI files generated by Goa, and then serving the "openapi.json" file from the "/swagger/swagger.json" URL, but I get a "404 page not found" response. :( 

I can't see what I'm doing wrong. Could you please have a look?

internal/api/api.go
```diff
@@ -33,6 +34,9 @@ import (
 	intupload "github.com/artefactual-sdps/enduro/internal/upload"
 )
 
+// go:embed gen/http/*.*
```
sevein: I think that the main issue is in that extra space after `//`. It should read:

```go
//go:embed gen/http/*.*
```

But you probably want something like this to embed just that file:

```go
//go:embed gen/http/openapi.json
```

internal/api/design/swagger.go
```diff
@@ -9,7 +9,7 @@ var _ = Service("swagger", func() {
 	HTTP(func() {
 		Path("/swagger")
 	})
-	Files("/swagger.json", "internal/api/gen/http/openapi.json", func() {
+	Files("/swagger.json", "openapi.json", func() {
```
sevein: I think that the second parameter should read: `gen/http/openapi.json`.

That would be the path used to access the embedded filesystem.

internal/api/api.go
djjuhasz: Done. :)

internal/api/design/swagger.go
djjuhasz: Ah! That did the trick, thanks!! :relieved:

sevein (approved):

PR #507, opened 2023-06-12
Triggering the cleanup of the dead letter queue

Diogenesoftoronto:
This is a continuation of the work in the previous merge request #494.

In this merge request I am implementing the solution @sevein  suggested [here](https://github.com/artefactual-sdps/enduro/pull/494#discussion_r1221357065).

Instead of ACK, I named it cleanup since that seemed more like what I am doing and provides greater clarity. I also talked with @djjuhasz and a good next step would be implementing a better cleanup of the filesystem function.
Ideally I would have liked to implement this in the alternative way that aligns with what preprocessing is doing but I found that this was a lot simpler for me to grasp.

### DIFF
--- a/internal/watcher/minio.go
+++ b/internal/watcher/minio.go
@@ -63,7 +63,7 @@ func NewMinioWatcher(ctx context.Context, logger logr.Logger, config *MinioConfi
 		return nil, err
 	}
 	if config.RedisFailedList == "" {
-		config.RedisFailedList = config.RedisList + "-dead"
+		config.RedisFailedList = config.RedisList + "-failed"
 	}
 
 	return &minioWatcher{
@@ -81,42 +81,53 @@ func NewMinioWatcher(ctx context.Context, logger logr.Logger, config *MinioConfi
 	}, nil
 }
 
-func (w *minioWatcher) Watch(ctx context.Context) (*BlobEvent, error) {
-	event, err := w.pop(ctx)
+func (w *minioWatcher) Watch(ctx context.Context) (*BlobEvent, Cleanup, error) {
+	event, val, err := w.pop(ctx)
 	if err != nil {
 		if errors.Is(err, redis.Nil) {
 			err = ErrWatchTimeout
 		}
-		return nil, err
+		return nil, noopCleanup, err
 	}
 	if event.Bucket != w.bucket {
-		return nil, ErrBucketMismatch
+		return nil, noopCleanup, ErrBucketMismatch
 	}
-	return event, nil
+	cleanup := w.rem(val)
+
+	return event, cleanup, nil
 }
 
 func (w *minioWatcher) Path() string {
 	return ""
 }
 
-func (w *minioWatcher) pop(ctx context.Context) (*BlobEvent, error) {
+// rem return a function that allows items from the failed list to be safely removed
+// in the event of a workflow or some other type of redis error.
+func (w *minioWatcher) rem(val string) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		logger := w.logger.WithValues("list", w.failedList)
+		if _, err := w.client.LRem(ctx, w.failedList, 1, val).Result(); err != nil {
+			logger.Error(err, "Error removing message from failed list.")
+			return err
+		}
+		logger.V(2).Info("Successfully removed message(s).")
+
+		return nil
+	}
+}
+
+func (w *minioWatcher) pop(ctx context.Context) (*BlobEvent, string, error) {
 	val, err := w.client.BLMove(ctx, w.listName, w.failedList, "RIGHT", "LEFT", redisPopTimeout).Result()
 	if err != nil {
-		return nil, fmt.Errorf("error retrieving from Redis list: %w", err)
+		return nil, "", fmt.Errorf("error retrieving from Redis list: %w", err)
 	}
 
 	event, err := w.event(val)
 	if err != nil {
-		return nil, fmt.Errorf("error processing item received: %w", err)
+		return nil, "", fmt.Errorf("error processing item received: %w", err)
 	}
 
-	if _, err := w.client.LRem(ctx, w.failedList, 1, val).Result(); err != nil {
-		w.logger.Error(err, "Error removing message from failed list.", "list", w.failedList)
-	} else {
-		w.logger.V(2).Info("Successfully removed messages.")
-	}
-
-	return event, nil
+	return event, val, nil
 }
 
 // event processes Minio-specific events delivered via Redis. We expect a
--- a/internal/watcher/watcher.go
+++ b/internal/watcher/watcher.go
@@ -18,9 +18,18 @@ var (
 	ErrBucketMismatch = errors.New("bucket mismatch")
 )
 
+type Cleanup func(ctx context.Context) error
+
+func noopCleanup(ctx context.Context) error {
+	return nil
+}
+
 type Watcher interface {
 	// Watch waits until a blob is dispatched.
-	Watch(ctx context.Context) (*BlobEvent, error)
+	// After the event is successfully processed by the receiver, the returned
+	// Cleanup function must be executed to mark the event as processed.
+	// Implementors must not return a nil function.
+	Watch(ctx context.Context) (*BlobEvent, Cleanup, error)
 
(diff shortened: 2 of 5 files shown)


internal/watcher/watcher.go
```diff
@@ -18,9 +18,11 @@ var (
 	ErrBucketMismatch = errors.New("bucket mismatch")
 )
 
+type Cleanup func() error
```
sevein: We should have a `noopCleanup` that we can return from filesystemWatcher.

    var noopCleanup = func() error { return nil }

If we ensure that implementors always return noopCleanup we can also avoid runtime panics, e.g. https://go.dev/play/p/XIsgt8V_ZDB.

internal/watcher/watcher.go
```diff
@@ -18,9 +18,11 @@ var (
...
+
 type Watcher interface {
 	// Watch waits until a blob is dispatched.
-	Watch(ctx context.Context) (*BlobEvent, error)
+	Watch(ctx context.Context) (*BlobEvent, Cleanup, error)
```
sevein, line 32: We can describe how we expect the interface and/or methods to be used. It can also describe some details for implementors if necessary (filesystemWatcher, minioWatcher, etc...). See for example the [http.Handler interface](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/net/http/server.go;l=62-88).

Given the new Cleanup function, perhaps the Watch method can be documented as follows:

```go
// Watch blocks until a *BlobEvent is received. After the event is successfully
// processed by the receiver, the returned Cleanup function must be executed to
// mark the event as processed. Implementors must not return a nil function.
```

internal/watcher/watcher.go
sevein: I would suggest one more change in the signature of this function:

    type Cleanup func(context.Context) error

By accepting a context with give the user the control to establish new deadlines or cancellation management for the cleanup routine specifically. Implicitly using the context provided elsewhere is limiting and uncommon.

internal/watcher/minio.go
```diff
@@ -81,42 +81,45 @@ func NewMinioWatcher(ctx context.Context, logger logr.Logger, config *MinioConfi
...
-	if _, err := w.client.LRem(ctx, w.failedList, 1, val).Result(); err != nil {
-		w.logger.Error(err, "Error removing message from failed list.", "list", w.failedList)
-	} else {
-		w.logger.V(2).Info("Successfully removed messages.")
+		return nil, "", fmt.Errorf("error processing item received: %w", err)
 	}
 
-	return event, nil
+	return event, val, nil
 }
```
sevein: I haven't tested but if I remember correctly you could also have a method match the Cleanup function signature, so the cleanup can be a `w.rem` method just like `w.pop` is a method too:

```go
func (w *minioWatcher) Watch(ctx context.Context) (*BlobEvent, Cleanup, error) {
	event, val, err := w.pop(ctx)
	if err != nil {
		if errors.Is(err, redis.Nil) {
			err = ErrWatchTimeout
		}
		return nil, noopCleanup, err
	}

	if event.Bucket != w.bucket {
		return nil, noopCleanup, ErrBucketMismatch
	}

	return event, w.rem, nil
}


func (w *minioWatcher) rem(ctx context.Context, val string) error {
	logger := w.logger.WithValues("list", w.failedList)

	_, err := w.client.LRem(ctx, w.failedList, 1, val).Result()
	if err != nil {
		logger.Error(err, "Error removing message from failed list.")
		return err
	}
	logger.V(2).Info("Successfully removed message from failed list.")

	return nil
}
```

sevein (changes requested):
Looking good! I've suggested some minor changes on top of what you've done already:
1. Change signature of Cleanup to accept context,
2. Have minioWatcher provide the cleanup fn as a `w.rem` method,
3. Update documentation of the interface, and
4. Define a `noopCleanup` to avoid code repetition, consider to always return `noopCleanup` instead of `nil` to avoid runtime panics if the user accidentally tries to execute it when unavailable.

internal/watcher/watcher.go
Diogenesoftoronto: I was about to abject but I noticed that LREM takes a context. Yeah that is weird thanks for catching this.

internal/watcher/watcher.go
Diogenesoftoronto: I think I will create an issue to improve the comments here like in the link that you mentioned. I have also add this.

internal/watcher/minio.go
Diogenesoftoronto: I return a closure instead now

internal/watcher/minio.go
```diff
@@ -81,42 +81,52 @@ func NewMinioWatcher(ctx context.Context, logger logr.Logger, config *MinioConfi
 	}, nil
 }
 
-func (w *minioWatcher) Watch(ctx context.Context) (*BlobEvent, error) {
-	event, err := w.pop(ctx)
+func (w *minioWatcher) Watch(ctx context.Context) (*BlobEvent, Cleanup, error) {
+	event, val, err := w.pop(ctx)
+	cleanup := w.rem(val)
```
sevein: I think that `cleanup` should be declared after you're done checking the errors.

internal/watcher/minio.go
```diff
@@ -81,42 +81,52 @@ func NewMinioWatcher(ctx context.Context, logger logr.Logger, config *MinioConfi
...
 }
 
-func (w *minioWatcher) pop(ctx context.Context) (*BlobEvent, error) {
+// rem return a function that allows items from the failed list to be safely removed
+// in the event of a workflow or some other type of redis error.
+func (w *minioWatcher) rem(val string) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		logger := w.logger.WithValues("list", w.failedList)
+		if _, err := w.client.LRem(ctx, w.failedList, 1, val).Result(); err != nil {
+			logger.Error(err, "Error removing message from failed list.", "list", w.failedList)
```
sevein: Notice that the new logger in L107 has already defined the `list` value, so you don't have to redefine it here.

internal/watcher/minio.go
```diff
@@ -81,42 +81,52 @@ func NewMinioWatcher(ctx context.Context, logger logr.Logger, config *MinioConfi
...
+			return err
+		}
+		logger.V(2).Info("Successfully removed messages.")
```
sevein: Would it make more sense to use the singular form? i.e. `message`.

internal/watcher/watcher.go
```diff
@@ -18,9 +18,18 @@ var (
...
+func noopCleanup(ctx context.Context) error {
+	return nil
+}
+
 type Watcher interface {
 	// Watch waits until a blob is dispatched.
-	Watch(ctx context.Context) (*BlobEvent, error)
+	// After the event is successfully
+	// processed by the receiver, the returned Cleanup function must be executed to
+	// mark the event as processed. Implementors must not return a nil function.
```
sevein: The way you're wrapping the comment is weird, why is the line break right after successfully? I think it should be as follows:

```go
	// After the event is successfully processed by the receiver, the returned
	// Cleanup function must be executed to mark the event as processed.
	// Implementors must not return a nil function.
```

sevein (changes requested):
Almost ready, just a few minor changes requested, mostly cosmetics.

internal/watcher/minio.go
Diogenesoftoronto: ah makes sense

internal/watcher/minio.go
Diogenesoftoronto: okay that makes sense.

internal/watcher/minio.go
Diogenesoftoronto: got it. I thought you were talking about a method on logger for a second.

internal/watcher/watcher.go
Diogenesoftoronto: make sense.

internal/watcher/minio.go
Diogenesoftoronto: I thought about this but there could be more than one message in the failed list right? We don't know how many actually live there so, messages makes sense. We could also do this: "Successfully removed message(s)."

sevein (approved):
Nice job, Keith!

PR #508, opened 2023-06-12
Set CORS Allow-Origin header by config

djjuhasz:
Allow configuration of the CORS "Access-Control-Allow-Origin" header in the API from a config file, or environment variable.


sevein (approved):

internal/config/config.go
```diff
@@ -87,5 +88,23 @@ func Read(config *Configuration, configFile string) (found bool, configFileUsed
...
+			"failed to set CORS Origin environment variable: %w", err,
+		)
+	}
+
 	return found, configFileUsed, nil
 }
+
+// setCORSOriginEnv sets the CORS Origin environment variable needed by
+// Goa-generated code for the API.
+func setCORSOriginEnv(config *Configuration) error {
```
jraddaoui, line 102: Why returning an error that will never happen?

sevein: I missed that! Should broadcast the error returned by os.Setenv.

djjuhasz: @jraddaoui Oops! There are two reasons for the never used error return:

1. The function used to return an error if `config.API.CORSOrigin` was empty, but I decided it would be better not to panic when CORS Origin was not set, especially when upgrading from an earlier version of Enduro.  I forgot to change the function signature when I changed that behaviour.
2. I was really tired yesterday :sleepy: :laughing: 

@sevein yeah, good idea to broadcast the os.Setenv error.

PR #511, opened 2023-06-15
Generate deterministic Goa sample data

djjuhasz:
Avoid inconsequential changes to Goa generated file sample data that can hide signficant changes.

- Use a deterministic randomizer
- Generate deterministic UUIDs
- Generate deterministic content_types


internal/api/design/uuid.go
```diff
@@ -0,0 +1,21 @@
+package design
+
+// Generate consistent UUIDs in Goa sample data.
```
sevein: I feel like this comment needs to be moved near Example() or needs to be extended, because it doesn't describe the purpose of the file, e.g. to provide functions to declare UUID-based attributes.

I was curious how this comment would render in the generated docs and came up with a couple of commands. This first one below prints the docs right in the terminal:

    $ go doc -all ./internal/api/design | less

If you install `godoc` (naming is confusing for sure), then you can bring up a web server just like [pkg.go.dev](https://pkg.go.dev/):

    $ go install golang.org/x/tools/cmd/godoc@latest
    $ godoc -http localhost:6060

Then visit http://127.0.0.1:6060/pkg/github.com/artefactual-sdps/enduro/internal/api/design/?m=all - `m=all` asks godoc to include internal packages and unexported symbols.

In both cases, the comment is not displayed. TIL! The [standard library](https://cs.opensource.google/go/go/+/master:src/net/http/client.go;l=5-8)) seems to follow a similar approach. When describing the contents in a specific file, the comment goes before the package declaration, and it's the empty line what hides them from the generated docs:

```go
// UUID attributes providing consistent sample data.

package design
```

Without the empty line then it's included in the global package comments. But the result is misleading in my opinion:

```go
// UUID attributes providing consistent sample data.
package design
```

They've added more formatting features a few releases ago, like [heading support](https://tip.golang.org/doc/comment#headings) and [links](https://tip.golang.org/doc/comment#links). Here's an alternative that will still show in the globa package section of the generated docs but with its own heading.

```go
// # UUID attributes
//
// Use [AttributeUUID] or [TypedAttributeUUID] to declare UUID attributes.
// They produce consistent sample data.
package design
```

sevein (approved):
Nice! :+1:

internal/api/design/uuid.go
djjuhasz: Cool! Thanks for the research @sevein. I'll have to check out the godoc tool. :star_struck:

PR #513, opened 2023-06-15
Reorganising the documentation to decrease scatter

### LINKED ISSUE #512: Reorganise documentation
The documentation is a bit scattered, we need to reorganise it to make it more intuitive to find.

Diogenesoftoronto:
There is now a docs folder and the top level README includes links to
other files.

Closes #512


docs/deps.md
```diff
@@ -0,0 +1,84 @@
...
+
+    ENT_VERSION ?= 0.11.10
+
+Now you can generate the code with:
+
+    make gen-ent
+
+### `goa.design/goa/v3`
+
+Update the dependency:
```
djjuhasz, line 69: Goa now has a `hack/make/dep_goa.mk` file like ent does, so it should now be updated in the same way as ent.

docs/release.md
```diff
@@ -0,0 +1,18 @@
+# Release
```
djjuhasz: Enduro doesn't have a `VERSION.txt` file or versioning system yet, so I think this file should be deleted for now.

djjuhasz (changes requested):
@Diogenesoftoronto looks good overall, just a couple of change requests.

docs/release.md
Diogenesoftoronto: BTW on the release stuff it might make sense to suggest using github's release system. I can even add something like that in the docs with a guide for how to use gh cli to create release versions. Secondly, this repo lacks a contributor.md file this might be a good idea.

djjuhasz: I don't know anything about Github's release system, so I can't comment on whether we should use it for Enduro.

I agree that a `contributor.md` would be a good addition.

docs/devel.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Known issues
+
+#### Minio uploads don't trigger workflows
+
+The setup of the Minio buckets and the communication between Minio and Redis
+is sometimes not setup properly. To solve it, from the Tilt UI, restart the
+`minio` resource and then trigger the `minio-setup-buckets` resource.
+
+[national health archive (nha)]: https://www.piql.com/norwegians-digital-health-data-to-be-preserved-for-future-generations/
```
djjuhasz: I don't think this link is used in the docs and should be removed.

docs/devel.md
```diff
@@ -0,0 +1,252 @@
...
+[national center for truth and reconciliation (nctr)]: https://nctr.ca/about/about-the-nctr/our-mandate/
```
djjuhasz: I don't think this link is used in the docs and should be removed.

djjuhasz (changes requested):
@Diogenesoftoronto I found a couple of links that I think are old and should be removed.

docs/release.md
Diogenesoftoronto: I will use a temporary contributor md and then add an issue to work on it later

CONTRIBUTOR.md
```diff
@@ -0,0 +1,3 @@
+# Contributing
+
+We welcome contributions from everyone. More details on how to contribute will be coming. For now simply submit a pull-request or create an issue.
```
djjuhasz: @Diogenesoftoronto we can't actually accept contributions without a signed contributors agreement (See https://github.com/artefactual/archivematica/blob/stable/1.14.x/CONTRIBUTING.md#contributors-agreement for an example).  I think it's better to omit the contributor file for now, and add one later when we have worked out the contents.

Diogenesoftoronto: Sensible

djjuhasz (approved):
LGTM @Diogenesoftoronto.  Don't forget to squash the commits before merging to main.

Issue #510, opened 2023-06-14
Problem: object storage requires a storage provider


djjuhasz:
**Please describe the problem you'd like to be solved**

The way Enduro currently manages storage objects like SIPs, AIPs, and DIPs requires access to an S3 compatible object storage API.  The reliance on a running an object storage provider complicates usage scenarios like integration testing where setting up a storage provider is onerous. 

**Describe the solution you'd like to see implemented**

Enduro uses the [Go Cloud blob](https://gocloud.dev/howto/blob/) package to abstract S3 storage actions (e.g. save, retrieve, delete). We could use GoCloud blob URLs (e.g. "s3://my-bucket", "file:///tmp/my-bucket") as a flexible way to specify the underlying storage driver in different scenarios.  For integration an unit tests the local filesystem could be used for storage, while a more robust cloud (public or private) storage provider can be used for staging and production.

**Describe alternatives you've considered**

**Additional context**

In the local development and Pulumi environments Enduro uses a MinIO Docker container to provide an S3 compatible object storage service.

PR #586, opened 2023-06-20
Allow configuring storage location by URL

djjuhasz:
Refs issue #510.

- Add `URLConfig` to Goa storage service
- Add `URLConfig` location type
- Add `LocationConfig.URL` field
- Configure location bucket by URL, when URL is set
- Add unit tests


internal/storage/location_test.go
```diff
@@ -0,0 +1,131 @@
...
+			errMsg: "invalid configuration",
+		},
+		{
+			name: "Errors when URL Config is empty",
+			location: &goastorage.Location{
+				UUID:   locationID,
+				Config: &goastorage.URLConfig{},
+			},
+			errMsg: "invalid configuration",
+		},
```
sevein: This case seems to be a copy of the previous one.

internal/storage/location_test.go
```diff
@@ -0,0 +1,131 @@
...
+			if tc.errMsg != "" {
+				assert.Error(t, err, tc.errMsg)
+				return
+			}
+
+			assert.NilError(t, err)
+			defer loc.Close()
+
+			assert.Equal(t, loc.UUID(), tc.uuid)
+			assert.Check(t, loc.Bucket() != nil)
```
sevein: I don't have a habit of using Check (https://pkg.go.dev/gotest.tools/assert#hdr-Assert_and_Check), I guess I always use Assert or its derivatives (Equal, NilError, etc...) because I feel it's easier for me to assume that the test will stop immediately, whereas when using Check the test may or may not fail but will continue running. Was that on purpose? I find the mechanics of Check a little harder to understand.

internal/storage/types/location_config_test.go
```diff
@@ -109,3 +111,63 @@ func TestLocationConfig(t *testing.T) {
...
+
+		cfg := types.LocationConfig{
+			Value: &types.URLConfig{
+				URL: "mem:///test-bucket",
+			},
+		}
+		blob, err := json.Marshal(cfg)
+
+		assert.NilError(t, err)
+		assert.DeepEqual(t, string(blob), `{"url":{"url":"mem:///test-bucket"}}`)
```
sevein: Maybe we don't need DeepEqual to compare strings, I can't remember off the top of my head.

sevein (approved):
Nicely done, just a few minor suggestions.

I'm thinking thtat maybe these changes enable us to eliminate the notion of location factories used by the storage service, since we designed them with the only purpose of injecting filesystem or memory-based buckets in our tests. `storage.NewService`  could now receive URL configs. Would that be right? I'm aware that'd be a significant change, perhaps in a future pull request.

djjuhasz:
> Nicely done, just a few minor suggestions.
> 
> I'm thinking thtat maybe these changes enable us to eliminate the notion of location factories used by the storage service, since we designed them with the only purpose of injecting filesystem or memory-based buckets in our tests. `storage.NewService` could now receive URL configs. Would that be right? I'm aware that'd be a significant change, perhaps in a future pull request.

@sevein I think it makes sense to get rid of the location factories and factory fakes, and the change will simplify the code significantly.  Right now I have a few questions that might get in the way:

1. If we get rid of the current S3config and replace it with URL based bucket config, do we lose any features of the S3config that are used in Enduro, like authentication?
2. Can we use URL config for SFTP locations, or do we need to maintain SFTPConfig as a separate config type? It seems like the SFTP location config is handled by a third party library, and I haven't dug into the particulars yet.

Can you answer either of these two questions, or do you have any suggestions for how to answer them?

internal/storage/location_test.go
djjuhasz: Ah, I didn't realize Check handled failure differently than Assert; I was just using Check as a way to fail the test based on a boolean condition.   I'll switch from Check to Assert.

internal/storage/types/location_config_test.go
djjuhasz: Right, probably not - I was originally comparing byte slices.

internal/storage/location_test.go
djjuhasz: Oops, I've changed this to an invalid schema test.

sevein:
I suspect the URL openers implemented by both sftp and s3 integrations can be somewhat limiting.

The sftpblob package seems to have a URLOpener but you can't pass keys for example, only passwords. The s3 integration is more flexible since you can pass a profile id that makes the underlying aws s3 sdk load config from the home directory, using the official aws config file which supports multiple authentication schemes.

I'm having a hard time imagining self-contained URL-only configs, but maybe I'm missing something.

djjuhasz:
> I suspect the URL openers implemented by both sftp and s3 integrations can be somewhat limiting.
> 
> The sftpblob package seems to have a URLOpener but you can't pass keys for example, only passwords. The s3 integration is more flexible since you can pass a profile id that makes the underlying aws s3 sdk load config from the home directory, using the official aws config file which supports multiple authentication schemes.
> 
> I'm having a hard time imagining self-contained URL-only configs, but maybe I'm missing something.

Oh, I probably misunderstood the scope of what you were suggesting with eliminating the config factories. I thought you were advocating for getting rid of the separate S3 location config (at a minimum) and *just* using URLs, but it sounds like that's not what you meant.

To clarify, are you suggesting we keep the option to specify a storage location by URL or S3 config or SFTP config, but eliminate the config factories?  I'm imagining an implementation like that would have a conditional statement to determine the type of configuration (e.g. if the URL field has a value use URL config), and would return a `*blob.Bucket`.  I guess without the factories that the unit and integration tests would just test the URL implementation and it would be up to end-to-end testing to make sure S3/Minio and SFTP config works?

sevein:
> To clarify, are you suggesting we keep the option to specify a storage location by URL or S3 config or SFTP config, but eliminate the config factories? I'm imagining an implementation like that would have a conditional statement to determine the type of configuration (e.g. if the URL field has a value use URL config), and would return a *blob.Bucket. I guess without the factories that the unit and integration tests would just test the URL implementation and it would be up to end-to-end testing to make sure S3/Minio and SFTP config works?

Exactly! Thanks.

PR #587, opened 2023-06-20
Update dependencies

sevein:
Including github.com/redis/go-redis v9 with some minor backward-incompatible changes.

I've used `gomajor list` to discover new updates in go.mod. Moved indirect dependencies to the second block, not sure why `go mod tidy` doesn't do that automatically but it's much nicer to read that file when the first block only shows direct dependencies.

In the dashboard/, if you run `npm install`, you can do the following:
- `npx ncu --target=minor --interactive` shows patch and minor updates available (safe to bump all)
- `npx ncu --interactive` will now show major updates (like from v1 to v2 where you want to be careful)


djjuhasz (approved):
LGTM. :)

Diogenesoftoronto (approved):

PR #588, opened 2023-06-20
Cosmetic changes to storage service

djjuhasz:
- Close a dangling `io.Reader` in a unit test
- Split long lines
- Add comments
- Wrap a unit test in a `t.Run()` method

### DIFF
--- a/internal/storage/service.go
+++ b/internal/storage/service.go
@@ -22,6 +22,7 @@ import (
 
 var SubmitURLExpirationTime = 15 * time.Minute
 
+// Service provides an interface for persisting storage data.
 type Service interface {
 	goastorage.Service
 
@@ -59,7 +60,15 @@ var _ Service = (*serviceImpl)(nil)
 
 var ErrInvalidToken error = goastorage.Unauthorized("invalid token")
 
-func NewService(logger logr.Logger, config Config, storagePersistence persistence.Storage, tc temporalsdk_client.Client, internalLocationFactory InternalLocationFactory, locationFactory LocationFactory, tokenVerifier auth.TokenVerifier) (s *serviceImpl, err error) {
+func NewService(
+	logger logr.Logger,
+	config Config,
+	storagePersistence persistence.Storage,
+	tc temporalsdk_client.Client,
+	internalLocationFactory InternalLocationFactory,
+	locationFactory LocationFactory,
+	tokenVerifier auth.TokenVerifier,
+) (s *serviceImpl, err error) {
 	s = &serviceImpl{
 		logger:             logger,
 		tc:                 tc,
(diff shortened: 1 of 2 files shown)


internal/storage/service.go
```diff
@@ -59,7 +60,15 @@ var _ Service = (*serviceImpl)(nil)
...
-func NewService(logger logr.Logger, config Config, storagePersistence persistence.Storage, tc temporalsdk_client.Client, internalLocationFactory InternalLocationFactory, locationFactory LocationFactory, tokenVerifier auth.TokenVerifier) (s *serviceImpl, err error) {
+func NewService(
+	logger logr.Logger,
+	config Config,
+	storagePersistence persistence.Storage,
+	tc temporalsdk_client.Client,
+	internalLocationFactory InternalLocationFactory,
+	locationFactory LocationFactory,
+	tokenVerifier auth.TokenVerifier,
+) (s *serviceImpl, err error) {
```
sevein, line 71: It'd be good to come up with the heuristic for this type of formatting style so we can all adopt it consistently. It reminds me to Temporal projects, they follow a similar style. I think that there are a couple of tools that can automate the transition to this new style: lll (available via golangci-lint) and [golines](https://github.com/segmentio/golines) (not compatible with golangci-lint), but I haven't tried them.

Many other projects, including the standard library or golang/tools (examples [1](https://github.com/golang/tools/blob/master/gopls/internal/lsp/protocol/tsclient.go) or [2](https://github.com/golang/tools/blob/master/gopls/internal/lsp/cache/analysis.go)) don't seem to care much unless the length is too long.  I personally feel that this style doesn't take into account the actual workspace that most developers use. I can comfortably accommodate 120 to 160 columns most of the time, long gone are the days of 80 columns. 

I was comparing this function from Temporal using both styles:

image

Ultimately a matter of taste of course, I think I'm more used to the second version. In any case, I am most concerned about consistency and reducing some noise in code reviews if possible. Maybe this is something we should investigate further.

sevein (approved):
:+1:

internal/storage/service.go
djjuhasz: @sevein yes, it would be nice to have a consistent style guide, and preferably a linter that can enforce it.  I miss the strong opinions of Black for Python. :laughing: 

I find it hard to parse a function definition with a lot of parameters if they aren't split into multiple lines, e.g.:

```
func NewService(logger logr.Logger, config Config, storagePersistence persistence.Storage, tc temporalsdk_client.Client, internalLocationFactory InternalLocationFactory, locationFactory LocationFactory, tokenVerifier auth.TokenVerifier) (s *serviceImpl, err error) {
```

In a definition like that I have a hard time reading the variable name + type pairs, and even with type hinting it can be hard me to figure out which parameter I'm entering when I'm writing a call to the function.  I agree it's a matter of taste, and I don't mind as much when a function definition is at 100 or 120 chars, but only has 2 or 3 parameters.

Issue #518, opened 2023-06-17
Writing commands using ffcli


Diogenesoftoronto:
## Is your feature request related to a problem? Please describe.
Compiling multiple binaries for different commands in Enduro can be cumbersome, particularly when it comes to integration testing. Managing and distributing these binaries separately adds complexity to the development workflow.

## Describe the solution you'd like
It would be beneficial to have the capability to include multiple commands within a single binary in Enduro. This feature would greatly improve the development experience and enhance the efficiency of integration testing. Packaging and delivering applications with multiple commands would be simplified, making distribution and deployment easier. This feature would save developers time and effort, contributing to a more streamlined development workflow. Considering the use case, we are exploring the adoption of `ffcli`, a command-line interface library that can serve as a front-end for managing commands, testing, and configuration within the application. Additionally, it would be advantageous to have a setup process in place.

## Describe alternatives you've considered
As an alternative to `ffcli`, we currently use `cobra` and `viper` in `SDPS-preprocessing`. However, `cobra` is more suitable for complex command-line applications and may not be the ideal solution for the Enduro project.

## Additional context
The `sdps-api-service` is deprecated, and `ffcli` is used there. However, in other non-deprecated services, `cobra` is used. Transitioning those services to `ffcli` would be challenging due to their larger size and feature completeness. Given the deprecation of `sdps-api-service` and the inconsistent tooling across projects, it may be more practical to adopt `cobra` instead of making changes to the preprocessing application. Standardizing the approach to packaging and delivering multiple commands within a single binary would simplify the deployment process and improve overall application maintainability.

sevein:
See https://github.com/artefactual-sdps/enduro/issues/574 for more details.

PR #590, opened 2023-06-22
Implement Robust comp time versioning of binaries

### LINKED ISSUE #520: Versioning and build scripts
## Is your feature request related to a problem? Please describe.
The current build system for `enduro` could be enhanced for better control over software versioning, as well as facilitating code modularity. The lack of a dedicated entry point outside of the root directory (main.go) can lead to complications when trying to modularize the project or adapt it for forked environments.

## Describe the solution you'd like
I suggest a revamp of the build system in a manner similar to the sdps_api_service. This would allow for the production of both binaries and Docker images, while injecting version and SCM details into the final artifact. Additionally, software versioning would be better tracked as part of the repo itself rather than just being an injectable attribute like a tag or build-time argument.

For code modularity, I propose moving the main.go file from the root directory to a dedicated cmd folder, mimicking the structure in the sdps_api_service. By moving the current entry point, scripts and references that currently point to main.go (like the genpkg script) would also need to be updated.

## Describe alternatives you've considered
Another option could be to leave the main.go file in the root directory and modify the build system and scripts to accommodate modularity and version tracking. However, this might not provide as clean a solution or align as well with commonly used project structures like that in sdps_api_service.

## Additional context
This feature request connects with issue #574 , and is in line with our ongoing efforts to enhance code modularity. I am already in the process of moving the main.go file and updating scripts and references accordingly, which will provide a solid foundation for this proposed build system improvement.

Diogenesoftoronto:
This PR is now fixed. It creates but does not actually implement the new version command via ffcli. This will be tackled in a follow up pr along with a new command server command. This will give enduro a similiar structure to our other projects. 

Refs #520, #518


internal/version/version.go
```diff
@@ -1,32 +1,68 @@
+// Package version provides the version that the binary was built at.
 package version
 
 import (
 	"fmt"
-	"runtime"
 	"runtime/debug"
+	"strings"
+
+	enduroroot "github.com/artefactual-sdps/enduro"
```
djjuhasz, line 9: Did you mean to import "github.com/artefactual-sdps/enduroroot" here?  I think importing github.com/artefactual-sdps/enduro will cause a recursive import?

Dockerfile
```diff
@@ -12,7 +12,7 @@ COPY --link . .
 FROM build-go AS build-enduro
 RUN --mount=type=cache,target=/go/pkg/mod \
 	--mount=type=cache,target=/root/.cache/go-build \
-	go build -o /out/enduro .
+	go build -o /out/enduro ./cmd/enduro-ctl
```
djjuhasz: I think ideally we want to build the binaries with the version flags, e.g. 
```
    go build -trimpath -o /out/enduro \
    -ldflags="-X '${VERSION_PATH}.Long=${VERSION_LONG}' -X '${VERSION_PATH}.Short=${VERSION_SHORT}' -X '${VERSION_PATH}.GitCommit=${VERSION_GIT_HASH}'" \
    ./cmd/enduro-ctl
```

This would require some more work to get the version info from `build_dist.sh shellvars`.

internal/version/version.go
```diff
@@ -1,32 +1,68 @@
...
-	return fmt.Sprintf("%s version %s (commit=%s) built on %s using %s",
-		appName, Version, GitCommit, BuildTime, GoVersion)
+
+	// Backup path, using Go 1.18's built-in git stamping.
+	Short = strings.TrimSpace(enduroroot.Version) + "-dev" + commitDate
+	Long = Short + "-t" + commitHashAbbrev + dirty
+}
+
+func Info(appName string) string {
+	return fmt.Sprintf("%s version %s",
```
djjuhasz, line 66: This seems short enough to be a one-liner.

internal/version/version.go
Diogenesoftoronto: This doesn't cause circular imports, I am pretty sure I would get circular import errors but I don't. Also, I modelled it the same way as sdps-api which also imports the root.

internal/version/version.go
Diogenesoftoronto: sensible

Dockerfile
Diogenesoftoronto: That is a good point I will work on that

go.work
```diff
@@ -2,5 +2,7 @@ go 1.20
 
 use (
 	.
+
 	./hack/pulumi
+
```
sevein: These blank lines you're adding I guess can be removed.

Dockerfile
```diff
@@ -3,16 +3,24 @@
...
+ARG VERSION_PATH
+ARG VERSION_LONG
+ARG VERSION_SHORT
+ARG VERSION_GIT_HASH
 RUN --mount=type=cache,target=/go/pkg/mod \
 	--mount=type=cache,target=/root/.cache/go-build \
-	go build -o /out/enduro .
+    go build -trimpath -o /out/enduro \
+    -ldflags="-X '${VERSION_PATH}.Long=${VERSION_LONG}' -X '${VERSION_PATH}.Short=${VERSION_SHORT}' -X '${VERSION_PATH}.GitCommit=${VERSION_GIT_HASH}'" \
+    ./cmd/enduro-ctl
```
sevein, line 23: You're not using tabs here.

sevein (commented):
@Diogenesoftoronto, would it be too hard at this point to address one new pull request addressing #520 only? Once merged, we could go back to this pull request to focus on the use of ffcli and subcommands.

Diogenesoftoronto:
> @Diogenesoftoronto, would it be too hard at this point to address one new pull request addressing #520 only? Once merged, we could go back to this pull request to focus on the use of ffcli and subcommands.

It does seem like more work than it is worth at this point. I am already working on stuff that is based on this pr

.bingo/Variables.mk
```diff
@@ -1,4 +1,4 @@
-# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.8. DO NOT EDIT.
+# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.7. DO NOT EDIT.
```
sevein, line 1: Some changes under the .bingo directory that I don't think are relevant to this pull request.

CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog], and this project adheres to [Semantic
+Versioning].
+
+[keep a changelog]: https://keepachangelog.com/en/1.0.0/
+[semantic versioning]: https://semver.org/spec/v2.0.0.html
```
sevein, line 9: A changelog is cool but isn't that outside the scope of this pull request?

Dockerfile
```diff
@@ -3,16 +3,24 @@
 ARG TARGET=enduro
 
 FROM golang:1.20.5-alpine AS build-go
+LABEL maintainer="Artefactual Systems Inc <info@artefactual.com>"
```
sevein, line 6: Question: is this label inherited by other images based on `build-go`?

Dockerfile
```diff
@@ -3,16 +3,24 @@
...
 
 FROM build-go AS build-enduro
+ARG VERSION_PATH
+ARG VERSION_LONG
+ARG VERSION_SHORT
+ARG VERSION_GIT_HASH
 RUN --mount=type=cache,target=/go/pkg/mod \
 	--mount=type=cache,target=/root/.cache/go-build \
-	go build -o /out/enduro .
+    go build -trimpath -o /out/enduro \
```
sevein, line 21: I've noticed that you're not passing the new version flags to enduro-a3m-worker, why is that?

internal/rootcmd/root.go
```diff
@@ -0,0 +1,37 @@
...
+	fs := flag.NewFlagSet("enduro-ctl", flag.ExitOnError)
+	cfg.RegisterFlags(fs)
+
+	return &ffcli.Command{
+		Name:       "enduro-ctl",
+		ShortUsage: "enduro-ctl [flags] <subcommand> [flags] [<arg>...]",
+		FlagSet:    fs,
+		Exec:       cfg.Exec,
+	}, &cfg
+}
```
sevein, line 28: Are you using this constructor? Looks unused at the moment.

sevein (changes requested):
> It does seem like more work than it is worth at this point.

The way I see it the versioning part of this work is well defined and workign smothly - could be merged right away and we'd be making progress. The subcommands, on the other hand, I believe we need to explore a bit more. Subcommands work nicely for command-line interfaces like tctl, gh or kubectl that group numerous tasks. But I'm wondering if this is the best pattern for our services, e.g. enduro, enduro-a3m-worker, a possible enduro-storage, enduro-gateway, etc. A single binary with everything inside is going to take longer to build, have you considered that?

Dockerfile
```diff
@@ -3,16 +3,24 @@
 ARG TARGET=enduro
 
 FROM golang:1.20.5-alpine AS build-go
+LABEL maintainer="Artefactual Systems Inc <info@artefactual.com>"
+USER root
```
sevein, line 7: Was this added for some reason? What is it fixing?

internal/rootcmd/root.go
Diogenesoftoronto: Yup it is unused

.bingo/Variables.mk
Diogenesoftoronto: will remove thanks

CHANGELOG.md
Diogenesoftoronto: eh not really this seems to be relevant to version stuff it fits well enough and its useful.

Diogenesoftoronto: I should add that in the final commit message and just change the PR Title to something like improve version project version management.

Dockerfile
Diogenesoftoronto: I should definitely do that nice catch

CHANGELOG.md
sevein: Or you can keep the scope of this PR small and focused on the original issue: version stamping. We're the authors of many projects where the log of changes is communicated differently, I thought it'd be of your interest to consider other practices followed at Artefactual.

sevein:
Superseded by https://github.com/artefactual-sdps/enduro/pull/623.

PR #593, opened 2023-06-23
Remove storage location factory methods

### LINKED ISSUE #510: Problem: object storage requires a storage provider

djjuhasz:
With URL configuration of buckets we can use local filesystem or in-memory buckets in test without using a factory method to inject different bucket drivers.

- Remove `InternalLocationFactory` and `LocationFactory` types and constructors
- Remove `internalLocationFactory` and `locationFactory` fields from the `serviceImpl` struct and constructor
- Rename `Location` interface to `Locationer`
- Rename `locationImpl` type to `Location` and export it
- Integrate `NewInternalLocation` logic for S3 and URL configs
- Don't open `*blob.Bucket` in `NewInternalLocation` or `NewLocation` constructors to avoid unnecessary opens and forgotten closes
- Update tests to use URL config for local filesytem or in-memory buckets


internal/storage/location.go
```diff
@@ -12,92 +12,47 @@ import (
...
-}
-
-type LocationFactory func(location *goastorage.Location) (Location, error)
-
-var DefaultLocationFactory = func(location *goastorage.Location) (Location, error) {
-	return NewLocation(location)
-}
-
-type Location interface {
+type Locationer interface {
```
sevein: The -er suffix is only used when the interface has a single method. E.g. Reader has a Read method alone. I think it's mentioned in Effective Go.

internal/storage/location.go
```diff
@@ -12,92 +12,47 @@ import (
...
 	UUID() uuid.UUID
-	Bucket() *blob.Bucket
-	Close() error
+	Bucket(ctx context.Context) (*blob.Bucket, error)
 }
 
-type locationImpl struct {
+type Location struct {
```
sevein: You can't control how your objects are created if you export them, that can be a source of bugs if the setup is complex. If unexported, users need to rely on your constructors so you fully own how they're built.

Users should be able to access its methods if they use the interface instead of the concrete types. The constructor can return an interface type, that I think is very common if the implementation is unexported, something I always forget to do.

sevein (commented):
Some early feedback. From my phone, haven't tested.

internal/storage/location.go
djjuhasz: Effective Go says:
> Interfaces with only one or two methods are common in Go code, and are usually given a name derived from the method, such as io.Writer for something that implements Write.

Ref: https://go.dev/doc/effective_go#interfaces

That doesn't sound to me like a prohibition using the "-er" suffix for interfaces with more than two methods.  I find it confusing when a concrete noun like "Location" can refer to an struct or an interface.  I think "Locator" would be better English, but proper English doesn't seem to be a requirement for interface names either. :laughing:

internal/storage/location.go
djjuhasz: Okay, thanks. It looks like I'm not accessing `Location` from outside the package, so making it an internal struct shouldn't be a problem.  I'll return an interface from the constructors too.

internal/storage/location.go
sevein: More on that topic here: https://go.dev/doc/effective_go#interface-names.

To be considered: Location is also the name used in Archivematica Storage Service.

djjuhasz: Nothing in https://go.dev/doc/effective_go#interface-names suggests to me that one shouldn't use the "-er" naming pattern for interfaces that have more than one method. :confused: 

To clarify, I assume you are saying Location the name of an _interface_ in the Archivematica Storage Service?

Another consideration is that in the Goa `storage` Service, `Location` is the name of a struct type:
https://github.com/artefactual-sdps/enduro/blob/main/internal/api/gen/storage/service.go#L89

djjuhasz: Also at this point `storage.location` is the only type that implements the `Locationer` interface.  We could just ditch the interface altogether and just use a `storage.Location` concrete type.  The `storage.location` fields aren't exported, so I think the worst thing you could do if it was `storage.Location` (exported) is create an empty struct. :shrug:

djjuhasz: Another option for the interface name is `LocationProvider`

internal/storage/service_test.go
```diff
@@ -363,36 +275,39 @@ func TestServiceSubmit(t *testing.T) {
...
-		assert.NilError(t, err)
-		b, err := fileblob.OpenBucket("/tmp", &fileblob.Options{URLSigner: fileblob.NewURLSignerHMAC(furl, []byte("1234"))})
-		assert.NilError(t, err)
+		td := tfs.NewDir(t, "enduro-service-test")
 
-		AIPID := uuid.MustParse("5ab42bc3-acc2-420b-bbd0-76efdef94828")
-		attrs := &setUpAttrs{
-			internalLocationFactory: fakeInternalLocationFactory(t, b),
+		attrs := setUpAttrs{
+			config: ref.New(storage.Config{
```
sevein: You can just use the & operator in Go.

`ref.New` was created for literals:

    config := &"foobar" // won't compile
    
    s := "foobar"
    config := &s // compiles but it's annoying

    config := ref.New("foobar")  // one liner

internal/storage/location.go
sevein: > Also at this point storage.location is the only type that implements the Locationer interface. 

The interface can be useful because you can inject a mock wherever is consumed, and the mock help us build better tests, allows us for testing complex things like retries, rare conditions, error paths, etc... Regardless, the choice is not critical at this point. As you wish! We can revert to the previous approach should it prove necessary in the future.

sevein: > RE: Locationer | LocationProvider

These two might suggest an entity that performs some action related to locations rather than an entity that represents a location itself, maybe? To be clear I'm indifferent to the final decision and I'd be fine using Locationer, but I must admit that I find enjoyable to think about it. I've been looking for interface names. I've found the `io` package in stdlib interesting: [`File`](https://pkg.go.dev/io/fs#File) or [`FS`](https://pkg.go.dev/io/fs#FS) are some of their choices for interface names. [Upspin](https://pkg.go.dev/upspin.io/upspin) has interfaces like Packer, KeyServer, Client, File, Config, Dialer or Service.

sevein (approved):

internal/storage/location.go
djjuhasz: Okay, I have come around on the `Location` interface name, and gone back to that and `locationImpl` for the concrete type.  Man, naming right? :face_with_spiral_eyes: :

PR #601, opened 2023-06-26
Use go.artefactual.dev/tools/log

sevein:
I've published a simple logging package to bring some consistency across our projects.

I've added a new configuration field (`Verbosity`) to define the V-Level of the logr. `0` is the least verbose (and the default, for prod).


djjuhasz (approved):
LGTM :+1:

PR #605, opened 2023-06-27
Set http server write timeout

### LINKED ISSUE #592: Problem: write server timeout is unlimited
It is now possible to establish write deadlines from handlers only where needed (e.g. download endpoints) using `ResponseController`, as opposed to setting it globally for the entire server.

For reference:  see this middleware - https://github.com/artefactual-labs/enduro/pull/532.

djjuhasz:
Fixes issue #592

- Set an http server write timeout to 5 seconds for most responses
- Add an unlimited write timeout for storage server downloads


sevein (approved):
LGTM

PR #607, opened 2023-06-28
Change namespace to enduro

scollazo:


djjuhasz:
@scollazo can you please provide some context for why you are changing the namespace to enduro?  What problem does this solve?

scollazo:
We currently have a k3s cluster for development/test purposes, but the namespace "sdps" is already in use there by other (internal) project, so I want to be able to deploy this project in the same cluster, but avoid the namespace clash.

djjuhasz:
@scollazo looks fine to me.  @sevein do you have any concerns with changing the namespace? I don't know any of the history of why sdps was used.

sevein:
I don't mind the change! One alternative could be "enduro-sdps" so we can leave "enduro" reserved for https://github.com/artefactual-labs/enduro.

djjuhasz:
I like "enduro-sdps".  What do you think @scollazo?

scollazo:
Anything different from "sdps" works for me.

sevein (approved):
:+1:

PR #645, opened 2023-07-24
Make a3m transfers configurable

Diogenesoftoronto:
The a3m submissions were previously hardcoded. Now defaults are set and the configuration is with the rest of a3m.
This has not been tested.


sevein:
> This has not been tested.

Is this still true? Should I wait until you have a chance to test?

scollazo:
Could you add some examples about how to use the configuration settings to enduro.toml ?

scollazo:
The processing configuration works part works as expected, could you remove the changes regarding aip naming, and leave them as part of #656 ?

enduro.toml
```diff
@@ -64,6 +64,24 @@ bucket = "aips"
 [a3m]
 address = "127.0.0.1:7000"
 shareDir = "/home/a3m/.local/share/a3m/share"
+name = "mytest"
```
djjuhasz: I think this should be removed, as the app name doesn't need to be configurable.

internal/a3m/a3m.go
```diff
@@ -81,28 +82,49 @@ func (a *CreateAIPActivity) Execute(ctx context.Context, opts *CreateAIPActivity
 				if err != nil {
 					return err
 				}
-
+				if reflect.ValueOf(a.cfg.Name).IsZero() {
```
djjuhasz: Remove this block too please.

internal/a3m/a3m.go
```diff
@@ -81,28 +82,49 @@ func (a *CreateAIPActivity) Execute(ctx context.Context, opts *CreateAIPActivity
...
+						PerformPolicyChecksOnPreservationDerivatives: true,
+						AipCompressionLevel:                          1,
+						AipCompressionAlgorithm:                      a3m_transferservice.ProcessingConfig_AIP_COMPRESSION_ALGORITHM_S7_BZIP2,
+					}
+				}
 				submitResp, err := c.TransferClient.Submit(
 					ctx,
 					&a3m_transferservice.SubmitRequest{
-						Name: "enduro",
+						Name: a.cfg.Name,
```
djjuhasz: Revert this please.

internal/a3m/a3m.go
```diff
@@ -132,7 +154,7 @@ func (a *CreateAIPActivity) Execute(ctx context.Context, opts *CreateAIPActivity
 						return errors.New("package failed or rejected")
 					}
 
-					result.Path = fmt.Sprintf("%s/completed/%s-%s.7z", a.cfg.ShareDir, "enduro", result.UUID)
+					result.Path = fmt.Sprintf("%s/completed/%s-%s.7z", a.cfg.ShareDir, a.cfg.Name, result.UUID)
```
djjuhasz: Revert

internal/a3m/config.go
```diff
@@ -1,6 +1,28 @@
 package a3m
 
+import transferservicev1beta1 "go.buf.build/grpc/go/artefactual/a3m/a3m/api/transferservice/v1beta1"
```
djjuhasz: Nitpick: I would label this import "transferService", a3m_transferService" or "a3mTS" - something that avoids coding the API version into the label.  If the transferservice API is updated at some point, then the label won't need to be changed to match.

djjuhasz (changes requested):
@Diogenesoftoronto I flagged the occurrences of the customizable app name that I think should be removed, as per @scollazo request.  I also noted one little naming nitpick.

internal/a3m/config.go
sevein: It's not uncommon to do it for some reason, e.g. see [gc_controller.go](https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/pkg/controlplane/controller/apiserverleasegc/gc_controller.go#L26) (many other examples available in that project). Since the version number relates to the `major` component (SemVer), maybe it's a way to signal that if it changes it's backward-incompatible so a new name is the best way to reflect that.
If you go with just the name I'd suggest `transferservice` since that's the original name of the package, `transferService` may be understood as an instance of something.

enduro.toml
```diff
@@ -64,6 +64,24 @@ bucket = "aips"
...
+
+[a3m.Processing]
```
sevein: For consistency, should be `[a3m.processing]` (e.g. see `[api.auth]`, etc).
I don't think Viper cares.

internal/a3m/a3m.go
```diff
@@ -81,28 +82,49 @@ func (a *CreateAIPActivity) Execute(ctx context.Context, opts *CreateAIPActivity
...
+						IdentifySubmissionAndMetadata:                true,
+						IdentifyBeforeNormalization:                  true,
+						Normalize:                                    true,
+						TranscribeFiles:                              true,
+						PerformPolicyChecksOnOriginals:               true,
+						PerformPolicyChecksOnPreservationDerivatives: true,
+						AipCompressionLevel:                          1,
+						AipCompressionAlgorithm:                      a3m_transferservice.ProcessingConfig_AIP_COMPRESSION_ALGORITHM_S7_BZIP2,
+					}
+				}
```
sevein: To be clear, the zero value of a type such Processing is an instance where all its fields are set to their respective zero values, e.g. `Normalize: false`, `AipCompressionLevel: 0`, etc... but if someone purposely sets all fields to zero this block would overwrite the user's config with the defaults you're proposing, that would be misleading.

If you want to provide configuration defaults I think you'd need to [establish defaults using Viper's API](https://pkg.go.dev/github.com/spf13/viper#readme-establishing-defaults). You could try to declare a new `ProcessingDefaults` in the `a3m` package and refer to it from `internal/config/config.go` (`Read`), e.g.:

```go
v.SetDefault("a3m.processing", a3m.ProcessingDefaults)
```

internal/a3m/config.go
```diff
@@ -1,6 +1,28 @@
...
+	IdentifyTransfer                             bool
+	IdentifySubmissionAndMetadata                bool
+	IdentifyBeforeNormalization                  bool
+	Normalize                                    bool
+	TranscribeFiles                              bool
+	PerformPolicyChecksOnOriginals               bool
+	PerformPolicyChecksOnPreservationDerivatives bool
+	AipCompressionLevel                          int
+	AipCompressionAlgorithm                      transferservicev1beta1.ProcessingConfig_AIPCompressionAlgorithm
+}
```
sevein: It would be nice to add a comment in Processing to clarify the source of all these fields, e.g.:

    // Processing mirrors all the processing configuration fields in transferservicev1beta1.ProcessingConfig.
    // The goal is to reflect all the fields available in the upstream API.

Viper provides its own extensible [unmarshal](https://pkg.go.dev/github.com/spf13/viper#readme-unmarshaling) mechanism, maybe something we could explore in the future.

sevein (changes requested):
A few suggestions.

internal/a3m/config.go
Diogenesoftoronto: This is what the lsp suggests. I was just using that.

djjuhasz (commented):
@Diogenesoftoronto looks good to me, but I think we should wait on merging it until the fix for #666 is merged and we can confirm all the tests pass.

djjuhasz:
@Diogenesoftoronto looks good to me.  Please make sure to squash all of the commits before rebasing and merging this to main.

Issue #608, opened 2023-06-30
Problem: dashboard configuration is set in Dockerfile


scollazo:
**Describe the bug**

Dashboard container includes [configuration values 
](https://github.com/artefactual-sdps/enduro/blob/main/dashboard/Dockerfile#L4)

**To Reproduce**

- Build dashboard container
- Pass variables pointing to a different dex auth server

**Expected behavior**

Dex auth can be configured at runtime, avoiding the need of rebuilding the container for each environment.

**Screenshots**

**Additional context**

jraddaoui:
We could use Nginx to pass those configuration values with environment variables: https://developers.redhat.com/blog/2021/03/04/making-environment-variables-accessible-in-front-end-containers#.

djjuhasz:
@jraddaoui the Vite OIDC environment variables are set in the dashboard kube config: 
https://github.com/artefactual-sdps/enduro/blob/main/hack/kube/base/enduro-dashboard.yaml

My local dev environment works fine after simply removing the OIDC environment variables from the Dockerfile:
https://github.com/artefactual-sdps/enduro/pull/646/commits/9437020ee23c4231f467caf4d06e6756471740b1

Do you think it is necessary to set the OIDC environment variables in other contexts, or do you think the commit above is sufficient?

jraddaoui:
Hi @djjuhasz, welcome back!

In [development](https://github.com/artefactual-sdps/enduro/blob/main/Tiltfile#L20) we use the [builder](https://github.com/artefactual-sdps/enduro/blob/main/dashboard/Dockerfile#L3) target from that Dockerfile. That target builds the assets for the final Nginx based image but it also runs Vite's dev server. The dev server is able to read those variables from the environment and pass it to the frontend, it also works with [.env files](https://vitejs.dev/guide/env-and-mode.html#env-files) where we have [the same defaults](https://github.com/artefactual-sdps/enduro/blob/main/dashboard/.env).

However, for production we build the final assets and serve them with Nginx and, with the current implementation, those env vars need to be set at build time, so having them in a k8s manifest has no effect. What I linked above explains a way to pass those env vars with Nginx to the frontend assets after build, which then will allow to configure them in the k8s manifest using the same Docker image and frontend assets.

djjuhasz:
[...]

PR #646, opened 2023-07-26
Remove dex config from dashboard Dockerfile

djjuhasz:
Fixes issue #608


dashboard/hack/start_nginx.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+inject_vite_envs.sh
+
```
scollazo:
```suggestion
envsubst '${ENDURO_API_ADDRESS}' <  /etc/nginx/templates/default.conf.template > /etc/nginx/conf.d/default.conf
```

We need to inject the address for the enduro api

PR #654, opened 2023-08-02
Allow runtime dashboard config for OIDC

djjuhasz:
Fixes issue #608

- Inject environment vars into the vue.js distribution files when starting Nginx to allow changing the values without rebuilding the docker image
- Remove Dex config vars from dashboard Dockerfile
- Standardize the use of VITE_OIDC_AUDIENCE_CLIENT_ID and VITE_OIDC_CLIENT_ID vars
- Add `.env.production` Vite config file with OIDC var placeholders
- Use port 8080:80 for the Vite and Nginx servers to simplify switching between the two in testing


scollazo (approved):
Works for me , LGTM!

Nitpick: could we use .env as placeholder, and use .env.dev for the development settings?

scollazo (approved):
LGTM!

hack/kube/overlays/dev/dex-secret.yaml
```diff
@@ -18,7 +18,7 @@ stringData:
           mode: "false"
     web:
       http: 0.0.0.0:5556
-      allowedOrigins: ["http://localhost:3000", "http://localhost:7440"]
+      allowedOrigins: ["http://localhost:8080", "http://localhost:7440"]
```
jraddaoui: Unrelated to this changes, I missed the addition of the Temporal URL in the other PR (https://github.com/artefactual-sdps/enduro/pull/653/files#diff-6de9eaa13ba8a389401b40b8466d17aa08141bb6989344b35489724f28e1cb62R21). I think Temporal UI uses a BFF (backend for frontend) auth workflow and it may not be needed here.

jraddaoui (approved):
Nice!

hack/kube/overlays/dev/dex-secret.yaml
djjuhasz: @jraddaoui huh, interesting. I'll give it a try without the Temporal URL.

djjuhasz: The Temporal UI login works without this, but for some reason I have to login twice.  When I log out, then log back in with the Dex prompt I get an "unauthorized" message.

image

Then if I log in again, it works. :confused:

djjuhasz: Hmm, when I log out the Temporal UI redirects me to:
http://localhost:7440/login?returnUrl=http%3A%2F%2Flocalhost%3A7440%2Fnamespaces%2Fdefault%2Fworkflows

Then when I try to log in again, URL for the above "unauthorized" page is: http://localhost:7440/login?returnUrl=http%3A%2F%2Flocalhost%3A7440%2Fnamespaces%2Fdefault%2Fworkflows&error=unauthorized

If I remove the "returnUrl" from the first URL, then the login works fine :man_shrugging:

PR #697, opened 2023-09-06
Add `CreatePackage` persistence method

djjuhasz:
- Add `CreatePackage` method to persistence service interface
- Add entclient implementation of `CreatePackage`
- Add the `ConvertPkgToGoa()` function to convert ent `Pkg` data objects to Goa `EnduroStoredPackage` objects
- Add error functions to convert db specific errors to more general persistence errors to prevent leaking implementation details

### DIFF
--- a/internal/persistence/ent/client/client.go
+++ b/internal/persistence/ent/client/client.go
@@ -1,11 +1,18 @@
 package entclient
 
 import (
-	"github.com/go-logr/logr"
+	"context"
+	"time"
 
+	"github.com/go-logr/logr"
+	"github.com/google/uuid"
+
+	"github.com/artefactual-sdps/enduro/internal/package_"
 	"github.com/artefactual-sdps/enduro/internal/persistence"
 	"github.com/artefactual-sdps/enduro/internal/persistence/ent/db"
 )
+
+const TimeFormat = time.RFC3339
 
 type client struct {
 	logger logr.Logger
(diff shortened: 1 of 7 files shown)


internal/persistence/ent/client/client_test.go
```diff
@@ -24,30 +28,183 @@ func setUpClient(t *testing.T, logger logr.Logger) (*db.Client, persistence.Serv
...
+
+			id, err := svc.CreatePackage(ctx, tt.args.pkg)
+			if tt.wantErr != "" {
+				assert.Error(t, err, tt.wantErr)
+				return
+			}
+			assert.NilError(t, err)
+
+			p, err := entc.Pkg.Get(ctx, id)
+			fmt.Printf("p: %+v", p)
```
sevein: :boom:

internal/persistence/ent/client/client.go
```diff
@@ -1,22 +1,90 @@
...
 type client struct {
 	logger logr.Logger
+	evsvc  event.EventService
 	ent    *db.Client
 }
 
 var _ persistence.Service = (*client)(nil)
 
-func New(logger logr.Logger, ent *db.Client) persistence.Service {
+func New(logger logr.Logger, evsvc event.EventService, ent *db.Client) persistence.Service {
```
sevein: This could be a good use case for the decorator pattern to decouple the concerns of persisting the data and publishing the events. I'm curious to know if you think it would be a good fit. You'd have to create an additional implementation of `persistence.Service` that wraps the `client`. In each method you would invoke the corresponding method in `client` and emit the appropiate event based on the outcome.

```go
type eventManager struct {
	evsvc event.EventService
	inner Service
}

var _ Service = (*eventManager)(nil)

func WithEvents(evsvc event.EventService, inner Service) *eventManager {
	return &eventManager{evsvc, inner}
}

// You can always use goimpl to generate large interface stubs automatically.
func (m *eventManager) CreatePackage(ctx context.Context, pkg *package_.Package) (int, error) {
	// _, err := m.inner.CreatePackage(...)
	// ... publish event if err is nil
}
```

There is an example in sdps_preprocessing, see `pkg/persistence/wrapper.go`.

sevein (approved):

internal/persistence/ent/client/client_test.go
djjuhasz: Hehe, oops. :)

djjuhasz: Removed.

internal/persistence/ent/client/client.go
djjuhasz: Nice! I implemented the eventManager decorator and I am quite happy with the results. :)

sevein: That's awesome.

PR #698, opened 2023-09-07
Add ent `UpdatePackage` persistence method

### LINKED ISSUE #517: Problem: we don't have a good pattern to implement flexible update methods in persistence
A little trick we saw in Dex where the Update persistence method accepts an updater function where the user is given the existing object so it can be modified within a transaction.\ It's a flexible way to provide different update contexts within a single method safely.

djjuhasz:
- Add an `UpdatePackage` method to the persistence service interface
- Implement the `UpdatePackage` method in the ent client
- Add a `ConvertPkg()` method to the ent client to convert an ent package representation to a `package_.Package` representation
- Add a DB transaction "rollback()" helper function
- Add an error "newUpdaterError()" helper function

Closes https://github.com/artefactual-sdps/enduro/issues/517.


sevein (approved):

internal/persistence/fake/mock_client.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/artefactual-sdps/enduro/internal/package_"
+	"github.com/artefactual-sdps/enduro/internal/persistence"
+)
+
+var (
+	CreatedAt = time.Unix(1694213364, 0) // 2023-09-08T22:49:24+00:00
+	StartedAt = time.Unix(1694213435, 0) // 2023-09-08T22:50:35+00:00
+)
+
+type mockClient struct{}
```
sevein: I would use mockgen for consistency but also it reduces the maintenance overhead significantly.

sevein (approved):
:+1:

sevein (approved):

PR #701, opened 2023-09-08
Use SIP filename for AIP filename

### LINKED ISSUE #656: Problem: AIP names are not named after the transfer name

All the AIPs created by enduro have the name "enduro-\<UUID\>.zip", and when uncompressed, they have a folder named enduro-\<UUID\>

Add a configuration setting that allows for AIPs  to be created using the transfer name, without the ".zip" suffix, so the folder is named <transfer_name>-UUID, and the AIP <transfer_name>-UUID.zip

**Additional context**

Diogenesoftoronto:
Closes #656

### DIFF
--- a/internal/fsutil/fsutil.go
+++ b/internal/fsutil/fsutil.go
@@ -3,12 +3,22 @@ package fsutil
 import (
 	"errors"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/otiai10/copy"
 )
 
 // Used for testing.
 var Renamer = os.Rename
+
+func BaseNoExt(path string) string {
+	base := filepath.Base(path)
+	if idx := strings.IndexByte(base, '.'); idx != -1 {
+		return base[:idx]
+	}
+	return base
+}
 
 // Move moves files or directories. It copies the contents when the move op
 // failes because source and destination do not share the same filesystem.
--- a/internal/storage/download.go
+++ b/internal/storage/download.go
@@ -12,9 +12,12 @@ import (
 
 	"github.com/artefactual-sdps/enduro/internal/api/gen/http/storage/server"
 	goastorage "github.com/artefactual-sdps/enduro/internal/api/gen/storage"
+	"github.com/artefactual-sdps/enduro/internal/fsutil"
 )
 
-// Download returns an HTTP handler that se
+// Download returns an HTTP handler that sets the headers for the AIP when it is downloaded.
+// The headers are the Content Type, Content Length, and the Content Disposition.
+// If there is an error with the file download, it will return http-status not found (404).
 func Download(svc Service, mux goahttp.Muxer, dec func(r *http.Request) goahttp.Decoder) http.HandlerFunc {
 	return func(rw http.ResponseWriter, req *http.Request) {
 		// Decode request payload.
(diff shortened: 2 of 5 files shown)


internal/workflow/processing.go
```diff
@@ -123,6 +125,11 @@ type TransferInfo struct {
 	DefaultPermanentLocationID *uuid.UUID
 }
 
+func (t *TransferInfo) Name() string {
+	s := strings.Split(filepath.Base(t.Key), ".")
+	return s[0]
+}
```
sevein: Here's an alternative that seems to be faster and doesn't allocate the slice:

```go
func (t *TransferInfo) Name() string {
	base := filepath.Base(t.Key)
	if i := strings.Index(base, "."); i != -1 {
		return base[:i]
	}
	return base
}
```

https://gist.github.com/sevein/f071b4205170dd28482b085778d8ecff

```
$ go test -bench=. -benchmem .
goos: linux
goarch: amd64
pkg: foobar
cpu: Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
BenchmarkV1-6   	12694426	        85.29 ns/op	      48 B/op	       1 allocs/op
BenchmarkV2-6   	81417373	        12.79 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	foobar	2.239s
```

internal/storage/download.go
```diff
@@ -47,7 +51,7 @@ func Download(svc Service, mux goahttp.Muxer, dec func(r *http.Request) goahttp.
 		}
 		defer reader.Close()
 
-		filename := fmt.Sprintf("enduro-%s.7z", pkg.AipID)
+		filename := fmt.Sprintf("%s-%s.7z", strings.Split(filepath.Base(pkg.Name), ".")[0], pkg.AipID)
```
sevein: strings.Split(filepath.Base(pkg.Name), ".")[0]

Could you create a `fsutil.FilenameWithoutExt` function in the fsutil package?

sevein (changes requested):
I've suggested a couple of changes.

internal/workflow/processing.go
Diogenesoftoronto: great test bench example

internal/storage/download.go
Diogenesoftoronto: yes, I will take your edited function and use that in fsutil

Diogenesoftoronto: I modified the function a bit. I checked the definition of index and it just uses the indexbyte function if the string is one in len. so i decided to just do that.

internal/fsutil/fsutil.go
```diff
@@ -3,13 +3,23 @@ package fsutil
...
 
 	"github.com/otiai10/copy"
 )
 
 // Used for testing.
 var Renamer = os.Rename
 
+func BaseNoExt(path string) string {
+	base := filepath.Base(path)
+	if idx := strings.IndexByte(base, "."[0]); idx != -1 {
```
djjuhasz: I think it's cleaner to use single quotes to return the full stop character as a rune (e.g. '.'). Passing a rune to  `IndexByte` seems to work fine:

```
if idx := strings.IndexByte(base, '.'); idx != -1 {
```

djjuhasz (changes requested):
One minor change request @Diogenesoftoronto.

djjuhasz (approved):
LGTM :+1:

PR #736, opened 2023-10-14
Update use of #nosec directive

sevein:
We've seen that gosec v2.18.1 seems to introduce an issue in the way that the nosec directive is interpreted. This pull request introduces a syntax workaround until they can get it fixed.


Diogenesoftoronto:
This is great. I am going to merge.

PR #737, opened 2023-10-18
Add an SFTP package

djjuhasz:
- Add an SFTP service definition
- Implement an SFTP upload service method using native Go SSH and SFTP packages
- Add SFTP configuration with some default values
- Generate a service mock with mockgen
- Add tests against a test SFTP server


djjuhasz:
@sevein CI is passing now, so this is ready to review.

internal/sftp/config.go
```diff
@@ -0,0 +1,35 @@
...
+		c.Port = "22"
+	}
+
+	if c.KnownHostsFile == "" {
+		c.KnownHostsFile = filepath.Join("$HOME", ".ssh", "known_hosts")
+	}
+
+	if c.PrivateKey.Path == "" {
+		c.PrivateKey.Path = filepath.Join("$HOME", ".ssh", "id_rsa")
+	}
```
sevein: I'd probably make use of [`os.UserHomeDir`](https://pkg.go.dev/os#UserHomeDir) to build these defaults, it's cross-platform and saves you from expanding the environment strings indiscriminately in SSHConnect.

internal/sftp/goclient.go
```diff
@@ -0,0 +1,84 @@
...
+		return 0, fmt.Errorf("SFTP: couldn't create remote file %q: %w", dest, err)
+	}
+
+	bytes, err := io.Copy(w, src)
+	if err != nil {
+		return 0, fmt.Errorf("SFTP: failed to write to %q: %w", dest, err)
+	}
+
+	return bytes, nil
+}
```
sevein: Not critical, but one issue with this implementation is that we can't pass a cancellation signal via context, e.g. if the Temporal activity is cancelled, etc. Ideally you'd accept a `ctx context.Context` as the first parameter. `io.Copy` can't use context out of the box but https://pkg.go.dev/github.com/dolmen-go/contextio provides context-aware wrappers for readers that you could use. We used it successfully in sdps_preprocessing (e.g. `upload_dip.go` if you want to take a look).
Maybe a future improvement or a small change you can make as part of this pull request if time permits!
Related: https://pace.dev/blog/2020/02/03/context-aware-ioreader-for-golang-by-mat-ryer.html.

internal/sftp/goclient.go
```diff
@@ -0,0 +1,84 @@
...
+// Close closes the SFTP client first, then the SSH client.
+func (c *GoClient) close() error {
+	var errs error
+
+	if err := c.sftp.Close(); err != nil {
+		errs = errors.Join(err, errs)
+	}
+	if err := c.ssh.Close(); err != nil {
+		errs = errors.Join(err, errs)
+	}
```
sevein: If you did an extra nil check (c.sftp and c.ssh) then you could safely call close() even if c.dial() returns an error. I think that would guarantee that all resources are freed in an scenario where the SSH connection was established but the SFTP client couldn't connect?

internal/sftp/goclient.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/pkg/sftp"
+	"golang.org/x/crypto/ssh"
+)
+
+// GoClient implements the SFTP service using native Go SSH and SFTP packages.
+type GoClient struct {
```
sevein, line 16: Perhaps one mention I would make to this struct is that this client is not thread-safe, e.g. two concurrent actors wouldn't be able to perform uploads in the way this is implemented, which I think is okay but it's something commonly documented from what I see in other projects. I think the reason is that Upload calls dial() which overrides c.ssh and c.sftp which is probably okay for now but best to make it explicit.

internal/sftp/service.go
```diff
@@ -0,0 +1,9 @@
+package sftp
+
+import (
+	"io"
+)
+
+type Service interface {
+	Upload(src io.Reader, dest string) (int64, error)
+}
```
sevein: I've used `Service` a lot in other packages when I couldnt' find a better name. In this particular case maybe one name that could work better is `Client`.

sevein (changes requested):
Some initial feedback, I had to stop half-way through but will continue later.

internal/sftp/ssh.go
```diff
@@ -0,0 +1,63 @@
...
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/knownhosts"
+)
+
+// SSHConnect connects to an SSH server using the given configuration and
+// returns a client connection.
+//
+// Only private key authentication is currently supported, with or without a
+// passphrase.
+func SSHConnect(cfg Config) (*ssh.Client, error) {
```
sevein: Do we need to export it?

internal/sftp/goclient.go
djjuhasz: Yeah, that's a good idea considering these uploads could take a long time. :+1:

internal/sftp/goclient.go
```diff
@@ -0,0 +1,84 @@
...
+func (c *GoClient) Upload(src io.Reader, dest string) (int64, error) {
+	if err := c.dial(); err != nil {
+		return 0, err
+	}
+	defer c.close()
+
+	// Note: Some SFTP servers don't support O_RDWR mode.
+	w, err := c.sftp.OpenFile(dest, (os.O_WRONLY | os.O_CREATE | os.O_TRUNC))
+	if err != nil {
+		return 0, fmt.Errorf("SFTP: couldn't create remote file %q: %w", dest, err)
```
sevein: I see many `%w` instead of `%v` in this function but also in SSHConnect. I try to use `%v` unless I know I need to expose not just the error message but also the error type. As it is right now someone using the `Service` interface may start making use of errors like [ssh.PassphraseMissingError](https://pkg.go.dev/golang.org/x/crypto/ssh#PassphraseMissingError) but that's an implementation detail that only applies to *GoClient. You could use `%v` or define a set of sentinel errors next to the interface, e.g. ErrReadPrivateKey, ErrSSHConnectionFailed and so on, but only if you're planning to use them.

internal/sftp/goclient.go
djjuhasz: What do you see as the barrier to thread safety? I don't know much about writing thread-safe code in Go, so I'm curious to learn more.  Is it the risk of two threads trying to write to the same file without a mutex?

internal/sftp/service.go
sevein: Suggested docstring:

```go
// A Client manages the transmission of data over SFTP.
//
// Implementations of the Client interface handle the connection details,
// authentication, and other intricacies associated with different SFTP
// servers and protocols.
type Client interface {
	// Upload transfers data from the provided source reader to a specified
	// destination on the SFTP server.
	Upload(src io.Reader, dest string) (int64, error)
}
```

internal/sftp/ssh.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	// Configure the SSH client.
+	sshConfig := &ssh.ClientConfig{
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeys(signer),
+		},
+		HostKeyCallback: hostcallback,
+		Timeout:         5 * time.Second,
+	}
```
sevein, line 53: It's great that the ssh package provides reasonable defaults when ciphers etc are not described explicitly!

internal/sftp/service.go
djjuhasz: Man naming, right? :weary: 

TL;DR `Service` -> `Client` sounds fine. :)

I'm still debating whether it would be better to call this package `upload` and leave the upload protocol up to the implementation (e.g. SFTP, SCP, Blob?).  But then what do you call the upload method, `upload.Do()`?  Plus all of the code is implementing SSH/SFTP, so it's probably better to call a spade a spade.

It does some weird to me to have an sftp service with only an `Upload()` method. One advantage of the current design is that it could support other sftp methods like download, or public dial and close methods to use allow using a single SFTP session to upload multiple files. :shrug:

internal/sftp/goclient_test.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		return nil, err
+	}
+	srv.AddHostKey(signer)
+
+	go func() {
+		err = srv.ListenAndServe()
+	}()
+
+	return &srv, err
```
sevein: An alternative to ensure that the error is captured and that you can have the function block until the server is ready:

	go func() {
		if err := srv.ListenAndServe(); err != nil {
			t.Fatal(err)
		}
	}()

	// Wait for the server to be ready
	for {
		conn, err := net.DialTimeout("tcp", "127.0.0.1:2222", 1*time.Second)
		if err == nil {
			conn.Close()
			break
		}
		time.Sleep(10 * time.Millisecond)
	}

internal/sftp/goclient_test.go
```diff
@@ -0,0 +1,252 @@
...
+		fmt.Println("sftp server completed with error:", err)
+	}
+}
+
+// StartSFTPServer starts a test SFTP server, and returns a pointer to the
+// server. The caller must call Close() to shut down the server when done with
+// it.
+func startSFTPServer() (*ssh.Server, error) {
+	srv := ssh.Server{
+		Addr: "127.0.0.1:2222",
```
sevein: A future clenaup could be to provide our own net.Listener with:

    net.Listen("tcp", "127.0.0.1:0")

It findes a free port so we don't have to use a fixed number. [Serve](https://pkg.go.dev/github.com/gliderlabs/ssh#Serve) takes a listener, it's an alternative to ListenAndServe where you provide your own network listener.

internal/sftp/goclient_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+	type test struct {
+		name    string
+		cfg     sftp.Config
+		want    results
+		wantErr string
+	}
+
+	for _, tc := range []test{
+		{
```
sevein, line 171: Nice tests!

internal/sftp/goclient_test.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// HostKeySigner signs messages from the server to the client and allows the
+// client to confirm the host key signature.
+func hostKeySigner() (gossh.Signer, error) {
+	keyfile := "./testdata/serverkeys/test_rsa"
+
+	key, err := os.ReadFile(keyfile)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't read keyfile %q, %v\n", keyfile, err)
```
sevein: Remember you can accept *t.Testing and use t.Fatal etc everywhere in the context of a test.

sevein (changes requested):
More suggestions, all minor.

internal/sftp/goclient.go
djjuhasz: Ohhhh, I see - if `dial()` is called a second time it overwrites the `c.ssh` and `c.sftp` clients. :thinking:

djjuhasz: Assuming `ssh.Client` and `sftp.Client` are thread safe, In `dial()` I think I could just check if each client is not nil before creating a new one.  Do you think that would be sufficient for thread safety?

internal/sftp/ssh.go
djjuhasz: No, no we do not. :laughing:

internal/sftp/goclient.go
djjuhasz: Riiight.  My thought process was that it's better to allow unwrapping the errors (I recently did this to test if an temporal application error was retryable), but I forgot about hiding the implementation specific errors. :facepalm:

internal/sftp/ssh.go
djjuhasz: Yes! :sweat_smile:

internal/sftp/goclient_test.go
djjuhasz: Very nice! Thanks! :heart_eyes:

internal/sftp/goclient_test.go
djjuhasz: Oh, cool. :)

internal/sftp/goclient_test.go
djjuhasz: Thank you! :blush:

internal/sftp/goclient.go
sevein: I believe you're right!

internal/sftp/service.go
sevein: You're right. This is all reminding me to Douglas' work where he used sftpblob, a driver for [gocloud.dev/blob] for SFTP where the concept of blob is pushed beyond traditional S3 buckets, see https://github.com/artefactual-sdps/enduro/blob/735afd570567ab609519496d45822819d3daebd4/internal/storage/types/location_config.go.

djjuhasz: I considered using https://github.com/rukavina/sftpblob like in Douglas' code, but it looks like a personal project that was abandoned years ago.  I didn't look around for other SFTP drivers for gocloud.dev/blob but I sure didn't want to write one myself. :laughing: 

I don't know if your suggesting changing this code to something more like Douglas' blob implementation, but if you are I'm not sure how to get there from here. :thinking:

sevein: It could be worth the effort but definitely not in this iteration, it's a lot of work to write a blob Driver I think.

internal/sftp/goclient_test.go
djjuhasz: Turns out you can't call `t.Fatal(err)` from a goroutine: `call to (*T).Fatal from a non-test goroutine`

I'm not that concerned with capturing the ListenAndServe error though.

sevein: Fair enough! I'd try with an error channel instead if it doesn't take you a lot of time. It could be useful to see the error message, e.g. say the developer already had another process listening on `:2222`, it'd be hard to tell that's the problem if you don't capture the error.

I haven't tested, but something like this may work:

```go
errCh := make(chan error, 1)
go func() {
    // Non-blocking send because the channel is buffered.
    errCh <- srv.ListenAndServe()
}()

for {
    select {
    case err := <-errCh:
        if err != nil {
            t.Fatal(err)
        }
        return
    default:
        conn, err := net.DialTimeout("tcp", "127.0.0.1:2222", time.Second)
        if err == nil {
            conn.Close()
            return
        }
        time.Sleep(10 * time.Millisecond)
    }
}
```

internal/sftp/goclient_test.go
djjuhasz: Using a random port causes the known_hosts check to fail.  It seems like known_hosts can take a wildcard for an IP, but the port needs to specified. :(

internal/sftp/goclient_test.go
djjuhasz: What's the advantage of passing `*t.Testing` into setup functions that are not actually running tests, rather than just using the standard Go error return idiom?  If `t.Testing` isn't required in the function other than for stopping the tests, then it seems like an unnecessary argument to pass around.

internal/sftp/goclient.go
djjuhasz: Making this thread-safe turned out to be harder than I expected, so I've just added a comment that Upload is not thread safe for now.

internal/sftp/goclient_test.go
sevein: I think that t.Testing expects us to lower the bar when it comes to flow control in order to simplify our tests, e.g. we can write helpers that save the caller from looking up the errors:

```go
key, err := os.ReadFile(keyFile)
assert.NilError(t, err, "couldn't read keyfile %q", keyfile)
```

Or we can use `t.Cleanup` to save the caller from cleaning up resources, e.g. startSFTPServer, currently written as follows:

```go
func startSFTPServer(addr string) (*ssh.Server, error) { ... }

func TestGoClient(t *testing.T) {
	// ...
	srv, err := startSFTPServer(serverAddress)
	if err != nil {
		t.Fatalf("Couldn't start SFTP server: %v", err)
	}
	t.Cleanup(func() { srv.Close() })
	// ...
}
```

Could be simplified if accepted `*t.Testing`:

```go
func startSFTPServer(t *testing.T, addr string) *ssh.Server) {
	// srv := ...
	t.Cleanup(func() { srv.Close() })  
}

func TestGoClient(t *testing.T) {
	// Cleanup not needed, the helper takes care.
	srv := startSFTPServer(serverAddress)
	// ...
}
```

djjuhasz: > ```
> key, err := os.ReadFile(keyFile)
> assert.NilError(t, err, "couldn't read keyfile %q", keyfile)
> ```

It seems weird to me to use `assert` for code that is not actually part of the tests.  The SFTP server is an unfortunate test set up requirement in my opinion, if I could test the client code without it I'd be much happier.  I also like that I could theoretically copy and paste the code as-is outside of the test context and it would run without having to remove the testing library dependencies.

sevein: I understand, and it's valid to optimize for code reuse outside the testing context, even if it makes your tests a bit heavier.

djjuhasz: > ```
> func startSFTPServer(t *testing.T, addr string) *ssh.Server) {
>	 // srv := ...
>	 t.Cleanup(func() { srv.Close() })  
> }
> ```

Setting up the `t.Cleanup()` inside of `startSFTPServer` makes sense to me to make it easier to add more tests without having to worry about teardown each time.

djjuhasz: I don't know if there's a better way to break out of the `srvwait` loop then what I've done.  I could move the wait loop to a separate function, but I'm not sure if that's any better. :shrug:

djjuhasz: I like func() better. :)

internal/sftp/goclient.go
```diff
@@ -0,0 +1,96 @@
...
+	var err error
+
+	c.ssh, err = sshConnect(c.cfg)
+	if err != nil {
+		return fmt.Errorf("SSH: %v", err)
+	}
+
+	c.sftp, err = sftp.NewClient(c.ssh)
+	if err != nil {
+		return fmt.Errorf("Unable to start SFTP subsystem: %v", err)
```
sevein: Should read "unable to start..."

internal/sftp/goclient_test.go
```diff
@@ -0,0 +1,288 @@
...
+			log.Fatalln("SSH: couldn't parse authorized key.")
+		}
+		if ssh.KeysEqual(key, allowed) {
+			return true
+		}
+	}
+
+	log.Println("SSH: unknown key provided.")
+	return false
+}
```
sevein: `log.Fatalln` uses `os.Exit(1)` which kills the test binary. I believe that will prevent other tests in the same package from running. Even if you use os.Exit(0) I believe the test is marked as failed. Maybe we should return false instead?

internal/sftp/goclient_test.go
```diff
@@ -0,0 +1,288 @@
...
+			cfg: sftp.Config{
+				Host:           host,
+				Port:           port,
+				KnownHostsFile: "./testdata/known_hosts",
+				PrivateKey: sftp.PrivateKey{
+					Path:       "./testdata/clientkeys/test_pass_rsa",
+					Passphrase: "wrong",
+				},
+			},
+			wantErr: "SSH: couldn't parse private key with passphrase: x509: decryption password incorrect",
```
sevein: Just sharing a thought for future consideration – when writing/wrapping error messages, I often find myself trying to decide whether to use phrases like "couldn't..." or "failed...", and I sometimes hesitate about whether to prefix the error with the method name or leave that context to the caller.

Anyways, I've found this today: https://github.com/uber-go/guide/blob/master/style.md#error-wrapping. It seems useful, or if we come up with our own rules I'd be happy to write them somewhere.

sevein (approved):
A couple of minor observations that shouldn't block you for merging. :+1:

internal/sftp/goclient.go
djjuhasz: I removed "Unable to " altogether, based on the Uber style guide you linked below.

internal/sftp/goclient_test.go
djjuhasz: This is not testing the code, it's just set up for the tests.  I believe the tests should not run if server setup fails because otherwise you will get false positives from the client tests (e.g. "ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain")

internal/sftp/goclient_test.go
djjuhasz: I like the Uber style guide on error wrapping.  I'll try to follow it in the future. :+1:

internal/sftp/goclient_test.go
sevein: To be clear, this is killing the test binary that Go builds for the entire package. It may prevent other tests from running even if they're defined in other files, e.g. tests in another_impl_test.go.

djjuhasz: Ohhh, yeah that sounds bad. :(  

I don't think returning false is a good alternative though, because the error message will suggest that the private key provided is unknown, instead of reporting that the authorized_keys file is missing or invalid. :(  Maybe the best thing to do is move all of the authorized_keys file processing out of the pubkeyHandler and handle those errors with the `testing.T` exit methods. :shrug:

sevein: It turns out that Uber's guide also mentions this issue.

https://github.com/uber-go/guide/blob/master/style.md#exit-in-main

PR #743, opened 2023-10-27
Update go.work.sum

djjuhasz:


sevein (approved):
I've updated the mod workflow hoping that "go mod download" helps catching these issues in the future.

PR #744, opened 2023-10-30
Use HTTP clients with better defaults

sevein:
Our existing HTTP clients are built the right way, i.e. we're creating new clients with `&http.Client{}` and they have a timeout set, as opposed to relying on `http.DefaultClient` which is shared and mutable. But in this pull request I'm proposing to use [clienthttp](https://pkg.go.dev/github.com/hashicorp/go-cleanhttp) because in addition to promoting non-global clients they come with good defaults for a couple of different scenarios: [DefaultClient](https://pkg.go.dev/github.com/hashicorp/go-cleanhttp@v0.5.2#DefaultClient) (for short-lived conns against arbitrary hosts) and [DefaultPooledClient](https://pkg.go.dev/github.com/hashicorp/go-cleanhttp@v0.5.2#DefaultPooledClient) (for connection reuse against same hosts).

### DIFF
--- a/cmd/enduro-a3m-worker/main.go
+++ b/cmd/enduro-a3m-worker/main.go
@@ -10,6 +10,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/hashicorp/go-cleanhttp"
 	"github.com/oklog/run"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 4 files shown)


Diogenesoftoronto:
> Our existing HTTP clients are built the right way, i.e. we're creating new clients with `&http.Client{}` and they have a timeout set, as opposed to relying on `http.DefaultClient` which is shared and mutable. But in this pull request I'm proposing to use [clienthttp](https://pkg.go.dev/github.com/hashicorp/go-cleanhttp) because in addition to promoting non-global clients they come with good defaults for a couple of different scenarios: [DefaultClient](https://pkg.go.dev/github.com/hashicorp/go-cleanhttp@v0.5.2#DefaultClient) (for short-lived conns against arbitrary hosts) and [DefaultPooledClient](https://pkg.go.dev/github.com/hashicorp/go-cleanhttp@v0.5.2#DefaultPooledClient) (for connection reuse against same hosts).

This looks great. It always makes sense to remove globals like this, and have better defaults.

Diogenesoftoronto (approved):

PR #749, opened 2023-11-02
Use mkdocs to build the docs

sevein:
This pull request builds a documentation site using [MkDocs](https://www.mkdocs.org/) which can consume the current docs with minimal changes. [ReadTheDocs](https://readthedocs.io/) is used for the hosting and build automation pieces. Here's the result: https://enduro.readthedocs.io/. 

This could serve well for the short term until we have an opportunity to explore other alternatives, e.g. Sphinx, Hugo, etc. If merged, https://enduro.readthedocs.io/ should be updated automatically when code is merged into main, and pull requests should include a new status check with a link to a preview site so you can take a look at what you're building before merging.  Ideally you'd also run MkDocs locally during the writing process. The installation process is similar to Sphinx: needs Python and the installation of dependencies in the requirements.txt file.


sallain:
LGTM! Thanks @sevein!

PR #755, opened 2023-11-03
Make preservation system configurable

Diogenesoftoronto:
This is change exists to allow for the taskqueue to determine whether or not to use archivematica or not. This could be done in a different way than the current option which is to set the preservation system variable to the a3m taskqueue. I think that simply using the taskqueue variable was a little awkward since it was tide to the a3m configuration struct. I think perhaps changing the a3m struct to a more general purpose preservation system struct or creating a new preservation struct may be more useful.

### DIFF
--- a/cmd/enduro-a3m-worker/main.go
+++ b/cmd/enduro-a3m-worker/main.go
@@ -127,7 +127,7 @@ func main() {
 			MaxConcurrentSessionExecutionSize:  1000,
 			MaxConcurrentActivityExecutionSize: 1,
 		}
-		w := temporalsdk_worker.New(temporalClient, cfg.A3m.TaskQueue, workerOpts)
+		w := temporalsdk_worker.New(temporalClient, temporal.A3mWorkerTaskQueue, workerOpts)
 		if err != nil {
 			logger.Error(err, "Error creating Temporal worker.")
 			os.Exit(1)
--- a/enduro.toml
+++ b/enduro.toml
@@ -3,7 +3,6 @@
 debug = true
 debugListen = "127.0.0.1:9001"
 verbosity = 2
-useArchivematica = false
 
 [temporal]
 namespace = "default"
@@ -63,6 +62,10 @@ key = "minio"
 secret = "minio123"
 region = "us-west-1"
 bucket = "aips"
+
+# Change the taskqueue setting to your prefered preservation system, by default it is a3m.
+[preservation]
+taskqueue = "a3m"
 
 [a3m]
 address = "127.0.0.1:7000"
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -12,6 +12,7 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/api"
 	"github.com/artefactual-sdps/enduro/internal/db"
 	"github.com/artefactual-sdps/enduro/internal/event"
+	"github.com/artefactual-sdps/enduro/internal/pres"
 	"github.com/artefactual-sdps/enduro/internal/storage"
 	"github.com/artefactual-sdps/enduro/internal/temporal"
 	"github.com/artefactual-sdps/enduro/internal/upload"
@@ -23,19 +24,19 @@ type ConfigurationValidator interface {
 }
 
 type Configuration struct {
-	Verbosity        int
-	Debug            bool
-	DebugListen      string
-	UseArchivematica bool
-	API              api.Config
-	Event            event.Config
-	Database         db.Config
-	Temporal         temporal.Config
-	Watcher          watcher.Config
-	Storage          storage.Config
-	Upload           upload.Config
-	A3m              a3m.Config
-	Am               am.Config
+	Verbosity    int
+	Debug        bool
+	DebugListen  string
+	API          api.Config
+	Event        event.Config
+	Database     db.Config
+	Temporal     temporal.Config
+	Watcher      watcher.Config
+	Storage      storage.Config
+	Upload       upload.Config
+	A3m          a3m.Config
+	Am           am.Config
+	Preservation pres.Config
 }
 
 func (c Configuration) Validate() error {
@@ -63,7 +64,7 @@ func Read(config *Configuration, configFile string) (found bool, configFileUsed 
 	v.AddConfigPath("/etc")
 	v.SetConfigName("enduro")
 	v.SetDefault("api.processing", a3m.ProcessingDefault)
-	v.SetDefault("a3m.taskqueue", temporal.A3mWorkerTaskQueue)
+	v.SetDefault("preservation.taskqueue", temporal.A3mWorkerTaskQueue)
 	v.SetDefault("temporal.taskqueue", temporal.GlobalTaskQueue)
 	v.SetDefault("debugListen", "127.0.0.1:9001")
 	v.SetDefault("api.listen", "127.0.0.1:9000")
--- a/internal/pres/config.go
+++ b/internal/pres/config.go
@@ -1,0 +1,10 @@
+package pres
+
+type Config struct {
+	// TaskQueue sets the Temporal task queue to use for the processing workflow
+	// (e.g. ` temporal.A3mWorkerTaskQueue`, ` temporal.AMWorkerTaskQueue`).
+	// The task queue determines which processing worker will run the processing
+	// workflow - a3m or Archivematica, and is used to branch the processing
+	// workflow logic.
+	TaskQueue string
+}
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -26,18 +26,18 @@ import (
 )
 
 type ProcessingWorkflow struct {
-	logger           logr.Logger
-	pkgsvc           package_.Service
-	wsvc             watcher.Service
-	useArchivematica bool
+	logger                logr.Logger
+	pkgsvc                package_.Service
+	wsvc                  watcher.Service
+	preservationTaskQueue string
 }
 
-func NewProcessingWorkflow(logger logr.Logger, pkgsvc package_.Service, wsvc watcher.Service, useAm bool) *ProcessingWorkflow {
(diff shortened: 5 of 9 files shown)


internal/config/config.go
```diff
@@ -63,6 +63,8 @@ func Read(config *Configuration, configFile string) (found bool, configFileUsed
 	v.AddConfigPath("/etc")
 	v.SetConfigName("enduro")
 	v.SetDefault("api.processing", a3m.ProcessingDefault)
+	// TODO: Change from a3m.taskqueue to pres.taskqueue, and add other preservation configuration settings
```
djjuhasz: Why not replace `a3m.taskqueue` with `preservation.taskqueue` now? I don't understand why we'd want to have `perservationSystem` and `a3m.taskqueue` in the config.

main.go
```diff
@@ -76,6 +76,8 @@ func main() {
 		fmt.Printf("Failed to read configuration: %v\n", err)
 		os.Exit(1)
 	}
+	// Determine the preservation system
+	useArchivematica := cfg.PreservationSystem == temporal.AmWorkerTaskQueue
```
djjuhasz: I think we should get rid of `useArchivematica` altogether, and pass `cfg.PreservationSystem` directly as a Workflow argument.

djjuhasz (changes requested):
@Diogenesoftoronto I requested a couple of changes.

cmd/enduro-a3m-worker/main.go
```diff
@@ -127,7 +127,7 @@ func main() {
 			MaxConcurrentSessionExecutionSize:  1000,
 			MaxConcurrentActivityExecutionSize: 1,
 		}
-		w := temporalsdk_worker.New(temporalClient, cfg.A3m.TaskQueue, workerOpts)
+		w := temporalsdk_worker.New(temporalClient, cfg.Pres.TaskQueue, workerOpts)
```
djjuhasz: This worker should always pull from the "a3m" taskqueue.  The `enduro-am-worker` should always pull from the "AM" taskqueue.

enduro.toml
```diff
@@ -64,6 +64,10 @@ secret = "minio123"
 region = "us-west-1"
 bucket = "aips"
 
+# Change the taskqueue setting to your prefered preservation system, by default it is a3m.
+# [pres]
```
djjuhasz: I'd use "preservation" here because I don't think "pres" is a well known abbreviation.

djjuhasz: I think this section should be un-commented.

internal/config/config.go
```diff
@@ -23,19 +24,19 @@ type ConfigurationValidator interface {
...
+	API         api.Config
+	Event       event.Config
+	Database    db.Config
+	Temporal    temporal.Config
+	Watcher     watcher.Config
+	Storage     storage.Config
+	Upload      upload.Config
+	A3m         a3m.Config
+	Am          am.Config
+	Pres        pres.Config
```
djjuhasz: I think "Preservation" is clearer.

internal/workflow/processing.go
```diff
@@ -26,18 +26,18 @@ import (
...
 
 type ProcessingWorkflow struct {
-	logger           logr.Logger
-	pkgsvc           package_.Service
-	wsvc             watcher.Service
-	useArchivematica bool
+	logger  logr.Logger
+	pkgsvc  package_.Service
+	wsvc    watcher.Service
+	presSys string
```
djjuhasz: I think this should be `preservationTaskQueue` or `presTaskQueue`.

internal/workflow/processing.go
```diff
@@ -26,18 +26,18 @@ import (
...
 }
 
-func NewProcessingWorkflow(logger logr.Logger, pkgsvc package_.Service, wsvc watcher.Service, useAm bool) *ProcessingWorkflow {
+func NewProcessingWorkflow(logger logr.Logger, pkgsvc package_.Service, wsvc watcher.Service, presSys string) *ProcessingWorkflow {
```
djjuhasz: `preservationTaskQueue` or `presTaskQueue` instead of `presSys`.

internal/workflow/processing.go
```diff
@@ -175,7 +175,7 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *pack
 		var taskQueue string
 		maxAttempts := 5
 
-		if w.useArchivematica {
+		if w.presSys == temporal.AmWorkerTaskQueue {
```
djjuhasz: This conditional is unnecessary if we have `w.presTaskQueue`.

internal/workflow/processing_test.go
```diff
@@ -71,8 +71,11 @@ func (s *ProcessingWorkflowTestSuite) SetupWorkflowTest(useAm bool) {
 	s.env.RegisterActivityWithOptions(am.NewPollIngestActivity(
 		logger, &am.Config{}, amclienttest.NewMockIngestService(ctrl),
 		dur).Execute, temporalsdk_activity.RegisterOptions{Name: am.PollIngestActivityName})
-
-	s.workflow = NewProcessingWorkflow(logger, pkgsvc, wsvc, useAm)
+	if useAm {
```
djjuhasz: I think we should pass the taskqueue name directly to this method, e.g. 

```
func (s *ProcessingWorkflowTestSuite) SetupWorkflowTest(presTaskQueue string) {
```

Then this conditional block can be reduced to:
```
s.workflow = NewProcessingWorkflow(logger, pkgsvc, wsvc, presTaskQueue)
```

djjuhasz (changes requested):
Some more change requests @Diogenesoftoronto.

cmd/enduro-a3m-worker/main.go
djjuhasz: Or we should get rid of `enduro-am-worker` altogether and rename this `enduro-preservation-worker` (or similar).

Diogenesoftoronto: I am not sure that renaming enduro-am-worker makes sense. there is no longer a a3m taskqueue. There is only a preservation taskqueue. Perhaps hard coding this since this should only use "a3m" would be a solution here. But I think this is mostly out of bounds with this PR. Doing it this way changes no functionality in the worker. A problem I do see if you want to use multipe types of workers (an am and a3m at the same time).

cmd/enduro-a3m-worker/main.go
```diff
@@ -127,7 +127,7 @@ func main() {
 			MaxConcurrentSessionExecutionSize:  1000,
 			MaxConcurrentActivityExecutionSize: 1,
 		}
-		w := temporalsdk_worker.New(temporalClient, cfg.A3m.TaskQueue, workerOpts)
+		w := temporalsdk_worker.New(temporalClient, cfg.Preservation.TaskQueue, workerOpts)
```
djjuhasz: This needs to be:

```
w := temporalsdk_worker.New(temporalClient, temporal.A3mWorkerTaskQueue, workerOpts)
```

I think we should stick to using separate workers for a3m and AM due to the significant differences required in the processing workflow.

internal/pres/config.go
```diff
@@ -0,0 +1,6 @@
+package pres
+
+type Config struct {
+	// Preservation System TaskQueue.
```
djjuhasz: Suggested comment:

```
// TaskQueue sets the Temporal task queue to use for the processing workflow
// (e.g. ` temporal.A3mWorkerTaskQueue`, ` temporal.AMWorkerTaskQueue`).  
// The task queue determines which processing worker will run the processing
// workflow - a3m or Archivematica, and is used to branch the processing 
// workflow logic.
```

internal/workflow/processing.go
```diff
@@ -175,11 +175,7 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *pack
 		var taskQueue string
 		maxAttempts := 5
 
-		if w.useArchivematica {
-			taskQueue = temporal.AmWorkerTaskQueue
-		} else {
-			taskQueue = temporal.A3mWorkerTaskQueue
-		}
+		taskQueue = w.preservationTaskQueue
```
djjuhasz: The `taskQueue` variable seems unnecessary, is there any reason not to use `w.preservationTaskQueue` directly?

djjuhasz (changes requested):
A few more change request @Diogenesoftoronto.  Almost there. :)

cmd/enduro-a3m-worker/main.go
Diogenesoftoronto: These are still separate workers. This preserves the behavior of keeping the task queue name configurable.

internal/workflow/processing.go
Diogenesoftoronto: no reason! good catch

Diogenesoftoronto:
the last two commits add comment and remove variable are the only important changes. I just did some rebasing on main.

cmd/enduro-a3m-worker/main.go
djjuhasz: The a3m worker should not be pulling tasks from the AM task queue when `cfg.Preservation.TaskQueue` = "am"

djjuhasz (changes requested):
@Diogenesoftoronto the a3m worker task queue still needs to be changed.

cmd/enduro-a3m-worker/main.go
Diogenesoftoronto: I don't think the A3m worker runs if the task queue is AM. It should run otherwise.

djjuhasz: @Diogenesoftoronto you should test it with the Tilt environment, but I guarantee you the a3m worker runs unless you remove the deployment code that runs it.

djjuhasz: @Diogenesoftoronto this might be helpful: https://docs.temporal.io/workers#task-queue

Diogenesoftoronto: Thanks for the resources. Change has been made.

djjuhasz (approved):
@Diogenesoftoronto looks good :+1:

PR #757, opened 2023-11-08
Avoid the use of os.Exit in tests

sevein:
I'm suggesting to pass *testing.T to pubKeyHandler using a closure so we can interrupt the test without using os.Exit.

Related: https://github.com/uber-go/guide/blob/master/style.md#exit-in-main.

### DIFF
--- a/internal/sftp/goclient_test.go
+++ b/internal/sftp/goclient_test.go
@@ -24,27 +24,28 @@ import (
 // ServerAddress is the test SFTP server address.
 const serverAddress = "127.0.0.1:2222"
 
-// PubkeyHandler handles checking the client's public key against the keys in
-// the authorized_keys file.
-func pubkeyHandler(ctx ssh.Context, key ssh.PublicKey) bool {
+// pubkeyHandler returns a handler that checks the client's public key against
+// the keys in the authorized_keys file.
+func pubKeyHandler(t *testing.T, ctx ssh.Context, key ssh.PublicKey) bool {
 	file, err := os.Open("./testdata/authorized_keys")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


djjuhasz:
> I'm suggesting to pass *testing.T to pubKeyHandler using a closure so we can interrupt the test without using os.Exit.
> 
> Related: https://github.com/uber-go/guide/blob/master/style.md#exit-in-main.

How can I pass *testing.T to pubKeyHandler? it has to match the PublicKeyHandler signature: https://pkg.go.dev/github.com/gliderlabs/ssh#PublicKeyHandler

Are you suggesting I use a closure or something?

sevein:
Yeah exatly, a closure helps us to meet the signature while we run another function:

```go
		PublicKeyHandler: func(ctx ssh.Context, key ssh.PublicKey) bool {
			return pubKeyHandler(t, ctx, key)
		}
```

internal/sftp/goclient_test.go
```diff
@@ -24,27 +24,28 @@ import (
...
 	}
+
 	defer file.Close()
 
 	scanner := bufio.NewScanner(file)
 	for scanner.Scan() {
 		allowed, _, _, _, err := ssh.ParseAuthorizedKey([]byte(scanner.Text()))
 		if err != nil {
-			log.Fatalf("SFTP server: couldn't parse authorized keys: %s", err)
+			t.Fatalf("SFTP server: couldn't open authorized_keys file: %s", err)
```
djjuhasz: I think this is a copy-paste error. Should be:

```
t.Fatalf("SFTP server: couldn't parse authorized keys: %s", err)
```

djjuhasz (changes requested):
Looks good overall @sevein , I think there is just one copy-paste mistake.

Issue #759, opened 2023-11-09
Problem: Upload activity error


djjuhasz:
**Describe the bug**

The upload activity is failing with a cryptic error: "cannot perform operation".

**To Reproduce**

1. Start processing a new transfer by adding it to the Minio SIP bucket
2. The "upload-activity" fails with the error message "cannot perform operation".

**Expected behavior**

The AIP should successfully upload to the Minio "aips" bucket, and the workflow should complete successfully.

**Screenshots**

![image](https://github.com/artefactual-sdps/enduro/assets/626135/27dd62c7-d150-4fef-9ce1-40d1b4694fe9)

**Additional context**

- Tested on my local Tilt dev environment
- I can reliably reproduce this error on main (@ 01af95dd4f5763aa25d7617496df66423040d38a) and commit 07f2497068889ea77acfc53c0f0df86089d49dfc
- Upload works on commit 7f592447135afc9ee6a7d4b06885c64d1b2648e4 (the commit before 07f2497068889ea77acfc53c0f0df86089d49dfc)

Full Temporal output:

```
{
  "message": "activity error",
  "source": "GoSDK",
  "stackTrace": "",
  "encodedAttributes": null,
  "cause": {
    "message": "cannot perform operation",
    "source": "GoSDK",
    "stackTrace": "",
    "encodedAttributes": null,
    "cause": null,
    "applicationFailureInfo": {
      "type": "ServiceError",
      "nonRetryable": false,
      "details": null
    }
  },
  "activityFailureInfo": {
    "scheduledEventId": "33",
    "startedEventId": "34",
    "identity": "1@enduro-a3m-0@",
    "activityType": {
      "name": "upload-activity"
    },
    "activityId": "33",
    "retryState": "MaximumAttemptsReached"
  }
}
```

sevein:
Fixed in https://github.com/artefactual-sdps/enduro/pull/765.

PR #765, opened 2023-11-09
Set default value for storage taskqueue

djjuhasz:
Fixes issue #759.

- Also fix SetDefault key for setting a3m.processing defaults
- Add config test to confirm default values are set
- Reorder embedded configs alphabetically
- Log detailed error when storage.InitStorageUploadWorkflow fails

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -25,14 +25,15 @@ type Configuration struct {
 	Verbosity   int
 	Debug       bool
 	DebugListen string
-	API         api.Config
-	Event       event.Config
-	Database    db.Config
-	Temporal    temporal.Config
-	Watcher     watcher.Config
-	Storage     storage.Config
-	Upload      upload.Config
-	A3m         a3m.Config
+
+	A3m      a3m.Config
+	API      api.Config
+	Database db.Config
+	Event    event.Config
+	Storage  storage.Config
+	Temporal temporal.Config
+	Upload   upload.Config
+	Watcher  watcher.Config
 }
 
 func (c Configuration) Validate() error {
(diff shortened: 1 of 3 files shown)


Diogenesoftoronto:
I was confused for a second because there were differently profile pictures haha.

internal/config/config.go
```diff
@@ -25,14 +25,15 @@ type Configuration struct {
...
-	Watcher     watcher.Config
-	Storage     storage.Config
-	Upload      upload.Config
-	A3m         a3m.Config
+
+	A3m      a3m.Config
+	API      api.Config
+	Database db.Config
+	Event    event.Config
+	Storage  storage.Config
```
Diogenesoftoronto, line 33: Why reorder here?

Diogenesoftoronto (approved):
It looks good. Just have a question about the reordering.

internal/config/config.go
djjuhasz: I was writing the tests and and find it easier to check for completeness in an ordered list than an unordered one.

sevein (approved):
Nice! I have a couple of thoughts:

1. Was it hard to identify the root issue, David? Yesterday I only had time to reproduce the issue before I had to move to something else, but I remember seeing a cryptic uninmplemented gRPC error code, I guess because the worker was not listening on the expected task queue. If there is a learning here on how to diagnose these kind of issues I'd be happy to capture it in our docs.
2. @Diogenesoftoronto now I understand what was going on behind this line change... :laughing:
    image
    Hopefully we can build some high-level end-to-end tests soon to identify this kind of regression before the code is merged.

djjuhasz:
@sevein the biggest barrier to identifying the root problem was that I was getting a very generic "cannot perform operation" from the `storage.Service` here: https://github.com/artefactual-sdps/enduro/blob/main/internal/storage/service.go#L126

Which is why I added error logging here: https://github.com/artefactual-sdps/enduro/blob/main/internal/storage/service.go#L125

I don't know if the error message is very general on purpose to avoid leaking implementation details to the caller, or if it's just a poorly written error message.  I assumed the error message was intentionally vague and added the logging to provide additional context to the system adminstrators.

PR #766, opened 2023-11-10
Use shfmt to format shell programs

sevein:
[shfmt](https://github.com/mvdan/sh) provides linting and formatting of shell scripts.


djjuhasz (approved):
LGTM :+1:

sevein:
Apparently shfmt only uses exit codes to signal errors when combined with the `--diff` flag, so I added it to the make target.

PR #772, opened 2023-11-14
Add delete method to sftp client

Diogenesoftoronto:


internal/sftp/client.go
```diff
@@ -14,4 +14,7 @@ type Client interface {
 	// Upload transfers data from the provided source reader to a specified
 	// destination on the SFTP server.
 	Upload(ctx context.Context, src io.Reader, dest string) (bytes int64, err error)
+	// Delete removes data from the provided dest to a specified
```
djjuhasz: Suggested comment: 
```
// Delete deletes the dest file from the SFTP server.
```

internal/sftp/client.go
```diff
@@ -14,4 +14,7 @@ type Client interface {
...
+	// destination on the SFTP server.
+	Delete(ctx context.Context, dest string) (bytes int64, err error)
```
djjuhasz: I can't think of any reason we'd need to know the size of the file after deleting it - I think the `bytes` return value should be removed.

internal/sftp/goclient.go
```diff
@@ -57,6 +57,32 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
 	return bytes, nil
 }
 
+// Delete removes the data from dest and returns the
```
djjuhasz: See above suggestion for updating this comment. Also the comment text should be re-wrapped at 80 cols.

internal/sftp/goclient.go
```diff
@@ -57,6 +57,32 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
...
+//
+// Delete is not thread safe.
+func (c *GoClient) Delete(ctx context.Context, dest string) (int64, error) {
+	if err := c.dial(); err != nil {
+		return 0, err
+	}
+	defer c.close()
+
+	// Get file information to find out its size
+	fileInfo, err := c.sftp.Lstat(dest)
```
djjuhasz: I think even without returning `bytes` the lstat is a good check that the file exists before deleting it, so we can provide a better error message.

internal/sftp/goclient.go
```diff
@@ -57,6 +57,32 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
...
+	if err != nil {
+		return 0, fmt.Errorf("SFTP: unable to get file info for %q: %v", dest, err)
```
djjuhasz: Suggested error message: `fmt.Errorf("SFTP: remote file doesn't exist or is unreadable: %s", dest)`

internal/sftp/goclient_test.go
```diff
@@ -287,4 +286,60 @@ func TestGoClient(t *testing.T) {
 			assert.Assert(t, tfs.Equal(dest.Path(), tfs.Expected(t, tc.want.Paths...)))
 		})
 	}
+
+	type deleteTest struct {
```
djjuhasz: I think it's better to move this to it's own test, e.g.

```
func TestDelete(t *testing.T) {
```

And rename the first test `TestUpload` (or similar).

internal/sftp/goclient_test.go
```diff
@@ -287,4 +286,60 @@ func TestGoClient(t *testing.T) {
...
+	} {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			sftpc := sftp.NewGoClient(tc.cfg)
+			src := strings.NewReader("Testing 1-2-3")
+			dest := tfs.NewDir(t, "sftp_test")
+
+			sftpc.Upload(context.Background(), src, dest.Join("test.txt"))
```
djjuhasz: This test will fail if the `Upload()` method fails, which isn't ideal.  Because the SFTP server is using the same filesystem as the client, you can create a file directly in `dest` then call `sftpc.Delete()` to delete it.

djjuhasz (changes requested):
@Diogenesoftoronto overall this looks good, but I've requested a few changes.

internal/sftp/goclient.go
sevein: An alternative where we do it in one go to save us the extra round-trip:

```go
func (c *GoClient) Delete(ctx context.Context, dest string) error {
	if err := c.dial(); err != nil {
		return fmt.Errorf("SFTP: unable to dial: %v", err)
	}
	defer c.close()

	err := c.sftp.Remove(dest)
	if err != nil {
		// Wrap only non-exist and permission errors to avoid leaking.
		if os.IsNotExist(err) || os.IsPermission(err) {
			return fmt.Errorf("SFTP: unable to remove file %q: %w", dest, err)
		}
		return fmt.Errorf("SFTP: unable to remove file %q: %v", dest, err)
	}

	return nil
}
```

Notice that I'm only using `%w` when I want to expose `os.ErrNotExist` or `os.ErrPermission`, but using `%v` to avoid other errors that could lead to implementation leaking.

internal/sftp/goclient.go
```diff
@@ -57,6 +57,32 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
 	return bytes, nil
 }
 
+// Delete removes the data from dest and returns the
+// number of bytes removed. A new SFTP connection is opened before removing the file, and
+// closed when the delete is complete or cancelled.
+//
+// Delete is not thread safe.
+func (c *GoClient) Delete(ctx context.Context, dest string) (int64, error) {
```
sevein: I'm realizing that we can't use ctx for cancellation, but I suggest to leave it in case we find a way to use it in the future. I think that both the ssh and sftp packages that we're using predate the context package so they were not designed to accept a context. One alternative that we could consider is to pass the context to the dial method and create our own dialer with [net.DialContext](https://pkg.go.dev/net#Dialer.DialContext), but I don't see it as a blocker.

sevein: rclone is doing it [here](https://github.com/rclone/rclone/blob/e1b0417c286bdcf5e866d70d51cb75be179c25e6/backend/sftp/ssh_internal.go#L26-L48), but it looks complicated because they support edge cases like proxy settings, etc. It also looks like the ssh package is going to have a DialContext sometime soon: https://go-review.googlesource.com/c/crypto/+/504735.

djjuhasz: Huh, interesting @sevein.  I thought ctx would be unnecessary for delete because the actual `sftp.Remove()` operation should be very fast, AFAIK.  But now I realize that dialing the server could take a while if there are communications issues, so I agree ctx is still worthwhile for future use.

internal/sftp/goclient.go
djjuhasz: I like this alternative, but I would make a small tweak to the non-exist/permission error message:
```
return fmt.Errorf("SFTP: unable to remove file: %w", err)
```
I find that most (maybe all?) of the filesystem error messages include the name of the file, so adding it in the wrapper causes duplication.  That said, it's easy to test for `os.IsNotExist(err)` in the unit tests so it's probably worth checking what the actual error message looks like.

sevein: You're right. I had a quick look at the code and it looks like it only returns a sentinel error but it's better to test it.

internal/sftp/goclient.go
Diogenesoftoronto: One thing I have considered is doing something similar to what contextio does, or wwe can do @sevein suggestion. I think including context on the actual client struct makes sense and then we can context is equal for operations that normally do not have a context like in contextio

internal/sftp/goclient_test.go
Diogenesoftoronto: Yes this is a better idea.

internal/sftp/goclient_test.go
Diogenesoftoronto: Yes I was considering this. I will go through with this.

internal/sftp/client.go
Diogenesoftoronto: destination can be a directory.

internal/sftp/goclient_test.go
```diff
@@ -288,3 +287,75 @@ func TestGoClient(t *testing.T) {
...
+	host, port, err := net.SplitHostPort(serverAddress)
+	if err != nil {
+		t.Fatalf("Bad server address: %s", serverAddress)
+	}
+
+	_ = startSFTPServer(t, serverAddress)
+
+	// Start a listener on an open port and use the address to test a bad SFTP
+	// server address.
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
```
djjuhasz: `listener` is unused.

internal/sftp/goclient_test.go
```diff
@@ -288,3 +287,75 @@ func TestGoClient(t *testing.T) {
...
+			name: "Deletes a file",
+			cfg: sftp.Config{
+				Host:           host,
+				Port:           port,
+				KnownHostsFile: "./testdata/known_hosts",
+				PrivateKey: sftp.PrivateKey{
+					Path: "./testdata/clientkeys/test_ed25519",
+				},
+			},
+			bytes:    13,
```
djjuhasz: `bytes` is unused.

internal/sftp/goclient_test.go
```diff
@@ -288,3 +287,75 @@ func TestGoClient(t *testing.T) {
...
+			wantErr:  "SFTP: unable to get file info for",
+		},
+	} {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			sftpc := sftp.NewGoClient(tc.cfg)
+			dest := tfs.NewDir(t, "sftp_test")
+			pre := dest.Join("test.txt")
```
djjuhasz: This can be combined with the `tfs.NewDir()` method:

```
dest := tfs.NewDir(t, "sftp_test", tfs.WithFile("test.txt", ""))
```

internal/sftp/goclient_test.go
```diff
@@ -227,8 +227,7 @@ func TestGoClient(t *testing.T) {
 			},
 			wantErr: fmt.Sprintf(
 				"SSH: connect: dial tcp %s:%s: connect: connection refused",
-				badHost, badPort,
-			),
+				badHost, badPort),
```
djjuhasz: Why is this change necessary?

internal/sftp/goclient.go
```diff
@@ -57,6 +57,26 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
...
+		return fmt.Errorf("SFTP: unable to dial: %w", err)
+	}
+	defer c.close()
+
+	// Delete the file
+	if err := c.sftp.Remove(dest); err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("SFTP: file does not exist: %w", err)
+		} else if os.IsPermission(err) {
+			return fmt.Errorf("SFTP: insufficient conditions to delete file: %w", err)
```
djjuhasz: I think "permissions" is clearer than "conditions".  Also, as you pointed out, dest could be a dir.

Suggested:
```
return fmt.Errorf("SFTP: insufficient permissions to delete: %w", err)
```

djjuhasz (changes requested):
@Diogenesoftoronto a few more change requests.

internal/sftp/goclient.go
Diogenesoftoronto: good catch and good point

internal/sftp/goclient_test.go
Diogenesoftoronto: its a result of lint i think

internal/sftp/goclient_test.go
Diogenesoftoronto: it can be but I think the way it is clearer

Diogenesoftoronto: actually nope i agree with you now. just had to see it.

internal/sftp/goclient_test.go
djjuhasz: I just ran the linter on main and it didn't change this. :confused:

djjuhasz (approved):
Looks good. Don't forget to squash the commits.

PR #773, opened 2023-11-16
Upload transfer to Archivematica

djjuhasz:
- Create an enduro-am-worker and add an "am" Temporal Taskqueue
- Branch the processing workflow logic for AM processing
- Add a zip package
- Add a ZipActivity to zip the bundled transfer before transmission
- Add an UploadTransferActivity to upload the zipped transfer to the AM Storage Service transfer source directory
- Add SFTP configuration for the upload
- Add an SSH username and a logger to the SFTP package
- Add the enduro-am-worker build to the Dockerfile
- Add an enduro-am-worker to the Tilt dev deployment
- Add a Tilt environment variable to select a kube config on startup
- Add the enduro-am-worker container and configuration to the kube manifests
- Add a "dev-am" kube overlay


hack/kube/base/enduro-am.yaml
```diff
@@ -0,0 +1,103 @@
+apiVersion: apps/v1
+kind: StatefulSet
```
sevein: If enduro-am doesn't need access to `/home/a3m/.local/share/a3m/share` then maybe this could be a Deployment instead of a StatefulSet. I think the former has some benefits in terms of scheduling or allocation performance.

hack/kube/base/enduro-am.yaml
```diff
@@ -0,0 +1,103 @@
...
+                  key: password
+            - name: ENDURO_DATABASE_DSN
+              value: $(MYSQL_USER):$(MYSQL_PASSWORD)@tcp(mysql.enduro-sdps:3306)/enduro
+            - name: ENDURO_STORAGE_DATABASE_DSN
+              value: $(MYSQL_USER):$(MYSQL_PASSWORD)@tcp(mysql.enduro-sdps:3306)/enduro_storage
+            - name: ENDURO_WATCHER_EMBEDDED_KEY
+              value: $(MINIO_USER)
+            - name: ENDURO_WATCHER_EMBEDDED_SECRET
+              value: $(MINIO_PASSWORD)
+            - name: ENDURO_USEARCHIVEMATICA
```
sevein, line 62: If Viper can read both `ENDURO_USEARCHIVEMATICA` and `ENDURO_USE_ARCHIVEMATICA` I suggest to adhere to the latter for consistency.

hack/kube/overlays/dev-am/kustomization.yaml
```diff
@@ -0,0 +1,35 @@
...
+bases:
+  - ../../base
+resources:
+  - dex-secret.yaml
+  - enduro-secret.yaml
+  - enduro-am-secret.yaml
+  - ldap.yaml
+  - minio-secret.yaml
+  - mysql-secret.yaml
+  - temporal-ui-secret.yaml
```
sevein: I don't know enough about kustomize but it looks like with little effort we could avoid some duplication, e.g. given that the following files are exact copies of what we have in the "dev" overlay:

- dev-am/mysql-secret.yaml
- dev-am/ldap.yaml
- dev-am/enduro-secret.yaml
- dev-am/dex-secret.yaml
- dev-am/temporal-ui-secret.yaml
- dev-am/minio-secret.yaml

Perhaps we could have a new directory with shared components and use it as follows:

```yaml
bases:
  - ../../base
  - ../../components/dev/mysql-secret.yaml
  - ../../components/dev/ldap.yaml
  - ../../components/dev/enduro-secret.yaml
  - ../../components/dev/dex-secret.yaml
  - ../../components/dev/temporal-ui-secret.yaml
  - ../../components/dev/minio-secret.yaml
resources:
  - am.yaml
```

internal/sftp/ssh.go
```diff
@@ -50,12 +51,14 @@ func sshConnect(cfg Config) (*ssh.Client, error) {
...
 		HostKeyCallback: hostcallback,
 		Timeout:         5 * time.Second,
+		User:            cfg.User,
 	}
 
 	// Connect to the server.
 	address := net.JoinHostPort(cfg.Host, cfg.Port)
 	conn, err := ssh.Dial("tcp", address, sshConfig)
 	if err != nil {
+		logger.Error(err, "SSH dial:", "address", address, "user", cfg.User)
```
sevein: One issue with logging an error here is that we may also be doing the same above in the stack. That's extra work for the system administrator if we're producing two errors for the same event. An alternative could be to use `logger.Info("...", "err", err)` or `logger.V(2).Info("...", "err", err)` to avoid the pollution.

sevein (approved):
Nicely done, just a few minor observations / future improvements.

hack/kube/base/enduro-am.yaml
djjuhasz: Viper does not read `ENDURO_USE_ARCHIVEMATICA` :cry:

hack/kube/overlays/dev-am/kustomization.yaml
djjuhasz: kustomize is pretty finicky:

- the `bases` directive is deprecated, we should be putting `../../base` in the `resources` section instead
- included files have to be in the same directory as the kustomization.yaml file or a directory below it
- you can include *directories* from a parallel or higher directory, but they must include a kustomization.yaml file

So we could create a `../../components/dev/` directory like:
```
  - dex-secret.yaml
  - enduro-secret.yaml
  - kustomization.yaml
  - ldap.yaml
  - minio-secret.yaml
  - mysql-secret.yaml
  - temporal-ui-secret.yaml
```

Then include it like:
```
resources:
  - ../../base
  - ../../components/dev
  - enduro-am-secret.yaml
```

PR #774, opened 2023-11-17
Toggle dev preservation worker deployment

djjuhasz:
Add an $ENDURO_PRES_SYSTEM Tilt environment variable to toggle the
deployment of Enduro with an a3m or Archivematica preservation worker.


Tiltfile
```diff
@@ -3,22 +3,31 @@ secret_settings(disable_scrub=True)
 load("ext://uibutton", "cmd_button", "text_input")
 load('ext://dotenv', 'dotenv')
 
+# Set preservation system (default: 'a3m')
+PRES_SYS = os.environ.get('ENDURO_PRES_SYSTEM', 'a3m')
+os.putenv('ENDURO_PRES_SYSTEM', PRES_SYS)
```
sevein: Do we need this line (os.putenv)?

jraddaoui: I think just for this: https://github.com/artefactual-sdps/enduro/pull/774/files#diff-c2ee8653e1d6b85f0aadf87cd438a9250806c052877248442be4d434cbc52425R134

sevein: Oh I see, it's used in L134.

sevein (approved):
LGTM

Tiltfile
djjuhasz: Yeah, it's used in L134 but it's pretty hacky. :(  I realized this morning that Starlark can probably do standard Python variable expansion in a string, which would be much better than relying on env var expansion.

PR #776, opened 2023-11-21
Fix sftp delete implementation

sevein:
Quick follow-up for https://github.com/artefactual-sdps/enduro/pull/772.

### DIFF
--- a/internal/sftp/goclient.go
+++ b/internal/sftp/goclient.go
@@ -41,7 +41,7 @@ func NewGoClient(logger logr.Logger, cfg Config) *GoClient {
 //
 // Upload is not thread safe.
 func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int64, string, error) {
-	if err := c.dial(); err != nil {
+	if err := c.dial(ctx); err != nil {
 		return 0, "", err
 	}
 	defer c.close()
@@ -70,7 +70,7 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
 // Delete removes the data from dest. A new SFTP connection is opened before
 // removing the file, and closed when the delete is complete.
 func (c *GoClient) Delete(ctx context.Context, dest string) error {
-	if err := c.dial(); err != nil {
+	if err := c.dial(ctx); err != nil {
 		return fmt.Errorf("SFTP: unable to dial: %w", err)
 	}
 	defer c.close()
@@ -96,10 +96,10 @@ func (c *GoClient) Delete(ctx context.Context, dest string) error {
 // Dial connects to an SSH host then creates an SFTP client on the connection.
 // When the clients are no longer needed, close() must be called to prevent
 // leaks.
-func (c *GoClient) dial() error {
+func (c *GoClient) dial(ctx context.Context) error {
 	var err error
 
-	c.ssh, err = sshConnect(c.logger, c.cfg)
+	c.ssh, err = sshConnect(ctx, c.logger, c.cfg)
 	if err != nil {
 		return fmt.Errorf("SSH: %v", err)
 	}
--- a/internal/sftp/goclient.go
+++ b/internal/sftp/goclient.go
@@ -5,7 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/dolmen-go/contextio"
@@ -64,7 +67,7 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
 	return bytes, remotePath, nil
 }
 
-// Delete removes the data from dest, a new SFTP connection is opened before
+// Delete removes the data from dest. A new SFTP connection is opened before
 // removing the file, and closed when the delete is complete.
 func (c *GoClient) Delete(ctx context.Context, dest string) error {
 	if err := c.dial(); err != nil {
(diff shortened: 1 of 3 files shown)


internal/sftp/goclient.go
```diff
@@ -64,23 +64,29 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
...
-			return fmt.Errorf("SFTP: insufficient permissions to delete file: %w", err)
+	// SFTP assumes that "/" is used as the directory separator. See:
+	// https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-6.2
+	remotePath := strings.TrimSuffix(c.cfg.RemoteDir, "/") + "/" + dest
+
+	if err := c.sftp.Remove(remotePath); err != nil {
+		head := fmt.Sprintf("SFTP: unable to remove file %q", dest)
+		if os.IsNotExist(err) || os.IsPermission(err) {
+			return fmt.Errorf("%s: %w", head, err)
+		}
```
sevein:
```
SFTP: unable to remove file "test.txt": permission denied
SFTP: unable to remove file "test.txt": file does not exist
```

internal/sftp/goclient.go
```diff
@@ -64,23 +64,29 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
...
+		if status, ok := err.(*sftp.StatusError); ok {
+			return fmt.Errorf("%s: %v (status code: %d)", head, status.FxCode(), status.Code)
```
sevein:
```
SFTP: unable to remove file "restricted/test.txt": failure (status code: 4)
```

internal/sftp/goclient.go
```diff
@@ -64,23 +64,29 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
...
 		}
-		return fmt.Errorf("SFTP: unable to remove file %q: %v", dest, err)
+		return fmt.Errorf("%s: %v", dest, err)
```
sevein: This shouldn't happen:

    SFTP: unable to remove file "test.txt": ...

internal/sftp/goclient_test.go
```diff
@@ -314,78 +314,93 @@ func UploadTest(t *testing.T) {
...
 	if err != nil {
 		t.Fatalf("Bad server address: %s", serverAddress)
 	}
 
 	_ = startSFTPServer(t, serverAddress)
 
+	type params struct {
+		fsOps       []tfs.PathOp // The state of the filesystem served by the SFTP server.
+		restrictDir string       // Set 0o555 on dir to reproduce permission issues.
+		delete      string       // The file that we will delete.
```
Diogenesoftoronto: we could just call it file

internal/sftp/goclient.go
```diff
@@ -64,23 +64,29 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
...
 	}
+
```
Diogenesoftoronto, line 92: Why add this change?

internal/sftp/goclient.go
```diff
@@ -64,23 +64,29 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (int6
...
 
-	// Delete the file
-	if err := c.sftp.Remove(dest); err != nil {
-		if os.IsNotExist(err) {
-			return fmt.Errorf("SFTP: file does not exist: %w", err)
-		} else if os.IsPermission(err) {
-			return fmt.Errorf("SFTP: insufficient permissions to delete file: %w", err)
+	// SFTP assumes that "/" is used as the directory separator. See:
+	// https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-6.2
+	remotePath := strings.TrimSuffix(c.cfg.RemoteDir, "/") + "/" + dest
```
Diogenesoftoronto, line 80: Should we not just use os path separator?

Diogenesoftoronto (changes requested):

internal/sftp/goclient.go
sevein: For consistency with the other final returns in this file.

internal/sftp/goclient.go
sevein: The comment suggests that it's a slash only, i.e. not a back-slash.

internal/sftp/goclient.go
djjuhasz: It *shouldn't* happen, but if it did happen I think `return fmt.Errorf("SFTP: remove: %s: %v", dest, err)` (or something similar) would provide helpful context for tracking down where the error came from in the code.

internal/sftp/goclient.go
djjuhasz: The "failure (status code: 4)" message here isn't very helpful - maybe `status.Error()` would have more information about the cause of the error?

djjuhasz (approved):
Looks good to me @sevein. I added a couple of comments about the error reporting, but they aren't blockers.

internal/sftp/goclient.go
sevein: Yes, good point, fixed in d89087b2.

internal/sftp/goclient_test.go
sevein: Thanks, fixed in 4816cad7.

internal/sftp/goclient.go
sevein: Tried to improve it in 0a613cb1, let me know.

djjuhasz: @sevein I think there is something weird going on with the error handling in the sftp package. :(

The full `Error()` message returned from your "Errors when there are insufficient permissions" test is:

```
"remove /tmp/sftp_test_remote-767072262/restricted/test.txt: permission denied" (SSH_FX_FAILURE)
```

But I would expect a permission denied error to be caught by the `os.IsPermission(err)` check and not make it to this line. :confused: I think something weird is happening at https://github.com/pkg/sftp/blob/v1.13.6/client.go#L741 but their tests don't cover the `RemoveDirectory()` method at all. :(

In any case, I think "failure", "status code: 4", and "SSH_FX_FAILURE" are all generic errors that are not very helpful, but maybe it's the best we can do right now.  I don't really have a preference between "(status code: 4)" or "(SSH_FX_FAILURE)".

djjuhasz: Looks like error handling in `sftp.Remove()` is a thorny issue: https://github.com/pkg/sftp/issues/468

djjuhasz: Side note: the help text for `os.IsPermission(err)` recommends using `errors.Is(err, fs.ErrPermission)` instead:

```
This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrPermission).
```

`os.IsNotExist(err)` likewise recommends `errors.Is(err, fs.ErrNotExist)`.

Diogenesoftoronto (approved):

internal/sftp/goclient.go
sevein: Yeah I noticed that too, that's why I ended up working on looking up the StatusError, so weird.

Ok, I'll use errors.Is, good catch!

PR #777, opened 2023-11-21
Add start transfer activity to AM workflow

djjuhasz:
Add a start transfer activity to send an "/api/v2beta/package" POST request to the Archivematica REST API to start creating a SIP from a transfer.

Add filepaths to AM workflow in `processing_test.go` to catch path mistakes when moving transfer files around.

### DIFF
--- a/internal/am/start_transfer_test.go
+++ b/internal/am/start_transfer_test.go
@@ -1,0 +1,126 @@
+package am_test
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/google/uuid"
+	"go.artefactual.dev/amclient"
+	"go.artefactual.dev/amclient/amclienttest"
+	"go.artefactual.dev/tools/mockutil"
+	"go.artefactual.dev/tools/temporal"
+	temporalsdk_activity "go.temporal.io/sdk/activity"
+	temporalsdk_testsuite "go.temporal.io/sdk/testsuite"
+	"go.uber.org/mock/gomock"
+	"gotest.tools/v3/assert"
+
+	"github.com/artefactual-sdps/enduro/internal/am"
+)
+
+func TestStartTransferActivity(t *testing.T) {
+	transferID := uuid.New().String()
+	opts := am.StartTransferActivityParams{
+		Name: "Testing",
+		Path: "/tmp",
+	}
+
+	amcrDefault := func(m *amclienttest.MockPackageServiceMockRecorder, st http.Response) {
+		m.Create(
+			mockutil.Context(),
+			&amclient.PackageCreateRequest{
+				Name:             opts.Name,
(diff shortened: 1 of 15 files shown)


internal/am/start_transfer_test.go
```diff
@@ -0,0 +1,132 @@
...
+			)
+
+			future, err := env.ExecuteActivity(am.StartTransferActivityName, opts)
+			if tt.errMsg != "" {
+				assert.ErrorContains(t, err, tt.errMsg)
+
+				var e *temporalsdk_temporal.ApplicationError
+				if ok := errors.As(err, &e); ok {
+					assert.Assert(t, e.NonRetryable() != tt.retryableErr)
+				}
```
sevein: This could use https://pkg.go.dev/go.artefactual.dev/tools/temporal#NonRetryableError if you think it's convenient.

internal/workflow/processing.go
```diff
@@ -368,9 +368,8 @@ func (w *ProcessingWorkflow) SessionHandler(sessCtx temporalsdk_workflow.Context
 		}).Get(activityOpts, nil)
 	}
 
-	// Stop here for the the Archivematica workflow. AIP creation, review, and
-	// storage are handled entirely by Archivematica and the AM Storage Service.
-	if w.preservationTaskQueue == temporal.AmWorkerTaskQueue {
+	// Stop here for the Archivematica workflow.
+	if w.taskQueue == temporal.AmWorkerTaskQueue {
 		return nil
 	}
```
sevein, line 374: I'm not sure if I'm understanding this correctly, but it seems we are assuming that taskQueue will be either "am" or "a3m". This assumption might inadvertently prevent multiple instances of Enduro from sharing the same Temporal namespace, as other deployments could have workers on the same queue. Perhaps this isn't an issue, since namespaces should not be shared, but I wanted to raise this point in case it could be a concern. Just like you wouldn't use the same MySQL database for two different apps.

sevein (approved):
Nicely done!

internal/workflow/processing.go
djjuhasz: Yes, we are assuming taskQueue will be either "am" or "a3m".  

I guess the alternative would be to have a cfg variable for the name of each task queue, e.g. `cfg.AMTaskQueue` and `cfg.A3mTaskQueue`?  The down side I see to this approach is config variable proliferation - in addition to the these two config vars, we need to add a new config variable every time we add another preservation worker that needs it's own task queue.  Maybe the chance of adding a new preservation worker is small enough that this isn't a real problem.

Are there other alternatives I'm missing?

sevein: Yeah that'd be it, but agree not a deal breaker.

Diogenesoftoronto (approved):

PR #778, opened 2023-11-22
Fix creation and upload of zip file

djjuhasz:
- Don't include parent dirs when creating a zip file from a dir
- Use the zip file name as the remote file name in the upload transfer activity
- Fix a bug in the zip activity that set the destination file name to
  an empty string when `DestPath` was set


sevein (approved):
:+1:

PR #781, opened 2023-11-23
Add an activity to poll AM transfer status

### LINKED ISSUE #786: Problem: when Archivematica is waiting for manual input, this does not generate an error in Enduro

When Archivematica is being used with Enduro, it is intended to be used in a fully automated way. The processing configuration should not be set to wait for manual input at any point. Currently, when this occurs, no error is reported in Enduro.

If the Archivematica workflow pauses to wait for manual input, this should be reported as an error in Enduro.

None

**Additional context**

This is an assertion we are making about an Enduro + Archivematica solution. I will add it in to the documentation.

djjuhasz:
- Add a PollTransferActivity to poll AM until transfer processing is complete or fails
- Add a PollInterval config option to control frequency of polls
- Send a heartbeat after each poll, and set a maximum time between heartbeats
- Add a TransferDeadline config option that limits the total duration of the PollTransferActivity in case Archivematica processing stalls
- Improve error handling for AM HTTP errors based on the legacy Enduro error handling
- Use go.artefactual.dev/tools/temporal for creating and checking non-retryable temporal errors

### DIFF
--- a/cmd/enduro-am-worker/main.go
+++ b/cmd/enduro-am-worker/main.go
@@ -139,6 +139,10 @@ func main() {
 			temporalsdk_activity.RegisterOptions{Name: am.StartTransferActivityName},
 		)
 		w.RegisterActivityWithOptions(
+			am.NewPollTransferActivity(logger, &cfg.AM, amc.Transfer).Execute,
+			temporalsdk_activity.RegisterOptions{Name: am.PollTransferActivityName},
+		)
+		w.RegisterActivityWithOptions(
 			activities.NewCleanUpActivity().Execute,
 			temporalsdk_activity.RegisterOptions{Name: activities.CleanUpActivityName},
 		)
--- a/internal/am/poll_transfer.go
+++ b/internal/am/poll_transfer.go
@@ -1,0 +1,144 @@
+package am
+
+import (
+	context "context"
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/go-logr/logr"
+	"go.artefactual.dev/amclient"
+	temporal_tools "go.artefactual.dev/tools/temporal"
+	temporalsdk_activity "go.temporal.io/sdk/activity"
+)
+
+const PollTransferActivityName = "poll-transfer-activity"
+
+var errWorkOngoing = errors.New("work ongoing")
+
+type PollTransferActivityParams struct {
+	TransferID string
+}
+
+type PollTransferActivity struct {
+	logger logr.Logger
+	cfg    *Config
+	amts   amclient.TransferService
+}
+
+type PollTransferActivityResult struct {
+	SIPID string
+	Path  string
+}
+
+func NewPollTransferActivity(logger logr.Logger, cfg *Config, amts amclient.TransferService) *PollTransferActivity {
+	return &PollTransferActivity{logger: logger, cfg: cfg, amts: amts}
+}
+
+// Execute polls Archivematica for the status of a transfer and returns when
+// the transfer is complete or returns an error status. Execute sends an
+// activity heartbeat after each poll.
+//
+// A transfer status of "REJECTED", "FAILED", "USER_INPUT", or "BACKLOG" returns
+// a temporal.NonRetryableApplicationError to indicate that processing can not
+// continue.
+func (a *PollTransferActivity) Execute(ctx context.Context, params *PollTransferActivityParams) (*PollTransferActivityResult, error) {
+	a.logger.V(1).Info("Executing PollTransferActivity", "TransferID", params.TransferID)
+
+	ticker := time.NewTicker(a.cfg.PollInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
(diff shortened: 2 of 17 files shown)


cmd/enduro-am-worker/main.go
```diff
@@ -138,6 +138,10 @@ func main() {
 			am.NewStartTransferActivity(logger, &cfg.AM, amc.Package).Execute,
 			temporalsdk_activity.RegisterOptions{Name: am.StartTransferActivityName},
 		)
+		w.RegisterActivityWithOptions(
+			am.NewPollTransferActivity(logger, &cfg.AM, amc.Transfer).Execute,
```
sevein, line 142: I like how you're passing services as opposed to the whole client, it's much easier to test.

internal/am/poll_transfer.go
```diff
@@ -0,0 +1,115 @@
...
+
+// transferFailedError checks an amclient error to determine if the transfer has
+// failed, or if it is still processing (which returns a 400 status code). If an
+// error is returned the activity should return the error, which may or may not
+// be a non-retryable error.
+func transferFailedError(r *amclient.Response, err error) error {
+	if err == nil {
+		return nil
+	}
+
```
sevein, line 110: If I remember correctly, it is possible to have situations where r is nil and err is non-nil, e.g. when something goes wrong below the application layer, e.g. dns resolution issue, connection timeout, context deadline, etc... it may also be possible if there is a JSON-decoding error.

Something broad like `if r == nil && err != nil { ... }` or more specific like `if errors.Is(err, context.DeadlineExceeded) { ... }` or `if netErr, ok := err.(net.Error); ok && netErr.Timeout() { ... }` if you plan different outcomes.

internal/am/poll_transfer.go
```diff
@@ -0,0 +1,115 @@
...
+			return &PollTransferActivityResult{SIPID: resp.SIPID, Path: resp.Path}, nil
+		}
+
+		// Send a heartbeat then continue polling after the poll interval.
+		temporalsdk_activity.RecordHeartbeat(ctx,
+			fmt.Sprintf("PollTransferActivity: last HTTP response: %s", httpResp.Status),
+		)
+		time.Sleep(a.cfg.PollInterval)
+	}
+}
```
sevein: A problem with `time.Sleep` is that it blocks and cannot be cancelled via context. It's not a huge deal but if you wanted to react to cancellation immediately then I'd suggest a different approach that is quite common where you use a `select` in combination with a timer, while you're also receiving from the context channel. I haven't tested but I think you'll get the idea. I'm also suggesting to pass a new context with a timeout to amclient which can be important to make sure that you're never missing a heartbeat deadline, in case the global http.Client timeouts set up in main don't meet our particular expecations in this activity.

```go
func (a *PollTransferActivity) Execute(ctx context.Context, params *PollTransferActivityParams) (*PollTransferActivityResult, error) {
	ticker := time.NewTicker(a.cfg.PollInterval)
	defer ticker.Stop()

	for {
		select {
		case <-ctx.Done():
			return nil, ctx.Err()
		case <-ticker.C:
			reqCtx, cancel := context.WithTimeout(ctx, a.cfg.PollInterval/2)
			r, err := a.poll(reqCtx, params.TransferID, cancel)
			if err == errWorkOngoing {
				// Send a heartbeat then continue polling after the poll interval.
				temporalsdk_activity.RecordHeartbeat(ctx, fmt.Sprintf("Last HTTP response: %v", errWorkOngoing))
				continue
			}
			if err != nil {
				return nil, err
			}

			return r, nil
		}
	}
}
```

I chose to pass the cancel function to a.poll so you can use defer. Otherwise you'd need to call it from the three exit paths: ongoing, error and happy path, i.e. defer stacks on the function scope, not in the scope of the for loop.

sevein (approved):
I didn't have time to look at the tests but overall looks great.
I've added a couple of suggestions.

internal/am/poll_transfer.go
Diogenesoftoronto: Does it block even if the activity itself is cancelled?

sevein: I believe so. Temporal can only signal cancellation via ctx (only true for heartbeating activities). It can't interrupt the execution without collaboration from the activity. 

That's not necessarily true in every SDK, but I suspect the root cause is that the Go SDK uses goroutines which need some coordination with the goroutine itself to be interrupted, e.g. a channel, context, etc... Temporal adopted context because it's the adopted idiom in the community.

David's activity eventually stops hence is not a leak, but it's making use of a goroutine unnecessarily for a period of time.

Related: https://dave.cheney.net/2016/12/22/never-start-a-goroutine-without-knowing-how-it-will-stop.

Diogenesoftoronto: Thanks. Something to test later as well.

internal/am/poll_transfer.go
djjuhasz: @sevein I added a nil pointer check on `r` to the 400 response code condition.  I think that should be sufficient to cover the case where r is nil and err is non-nil - `convertAMClientError()` will return `err` unchanged when `r` is nil.

PR #787, opened 2023-11-27
Implement delete activity

### LINKED ISSUE #785: Problem: unneeded packages are being retained in the AM Storage Service's transfer source location

When Archivematica it used, transfers are copied from minIO to the Archivematica Storage Service's transfer storage location. These transfers are not automatically deleted and will eventually use up the storage space. These transfers are interim packages that do not need to be kept - the original transfer is kept and can be used as a fallback if something goes wrong. Once the AIP is created, the transfer in the AMSS can be deleted automatically.

Delete the interim transfer located in the AMSS once the AIP is created.

None.

**Additional context**

Diogenesoftoronto:
I have approached the deletion of transfers in a number of different ways, first by implementing a separate activity like in this approach. I also tried to add the capability of deleting transfers to the cleanup activity. The reason we did not go with this approach is that it was ultimately too complex for the benefits. The final approach was to not use an activity at all and just implement delete in the workflow directly, this also did not work as an approach because the workflow uses the 'internal.Context' interface and not the context.Context. It also meant spreading dependencies. The Am worker and the global worker needed to use the sftpclient. This can be a problem because worker clients can be different and there is no guarantee that they are the same.  

- Implement delete activity
- Add delete to am worker registration
Closes #785

### DIFF
--- a/internal/am/delete_transfer.go
+++ b/internal/am/delete_transfer.go
@@ -1,0 +1,38 @@
+package am
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+
+	"github.com/artefactual-sdps/enduro/internal/sftp"
+)
+
+const DeleteTransferActivityName = "DeleteTransferActivity"
+
+type DeleteTransferActivityParams struct {
+	Destination string
+}
+
+type DeleteTransferActivity struct {
+	client sftp.Client
+	logger logr.Logger
+}
+
+func NewDeleteTransferActivity(logger logr.Logger, client sftp.Client) *DeleteTransferActivity {
+	return &DeleteTransferActivity{client: client, logger: logger}
+}
+
+func (a *DeleteTransferActivity) Execute(ctx context.Context, params *DeleteTransferActivityParams) error {
+	a.logger.V(1).Info("Execute DeleteTransferActivity",
+		"destination", params.Destination,
+	)
+
+	err := a.client.Delete(ctx, params.Destination)
+	if err != nil {
+		return fmt.Errorf("delete transfer: path: %q: %v", params.Destination, err)
+	}
+
+	return nil
+}
--- a/internal/am/delete_transfer_test.go
+++ b/internal/am/delete_transfer_test.go
@@ -1,0 +1,107 @@
+package am_test
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"go.artefactual.dev/tools/mockutil"
+	temporalsdk_activity "go.temporal.io/sdk/activity"
+	temporalsdk_testsuite "go.temporal.io/sdk/testsuite"
+	"go.uber.org/mock/gomock"
+	"gotest.tools/v3/assert"
+	tfs "gotest.tools/v3/fs"
+
+	"github.com/artefactual-sdps/enduro/internal/am"
+	sftp_fake "github.com/artefactual-sdps/enduro/internal/sftp/fake"
+)
+
+func TestDeleteTransferActivity(t *testing.T) {
+	t.Parallel()
+
+	filename := "fake_bag"
+	activityErr := "activity error (type: DeleteTransferActivity, scheduledEventID: 0, startedEventID: 0, identity: ): "
+	td := tfs.NewDir(t, "enduro-delete-transfer-test",
+		tfs.WithFile(filename, "Testing 1-2-3!"),
+	)
+
+	type test struct {
+		name     string
+		params   am.DeleteTransferActivityParams
+		recorder func(*sftp_fake.MockClientMockRecorder, am.DeleteTransferActivityParams)
+		errMsg   string
+	}
+	for _, tt := range []test{
+		{
+			name: "Deletes transfer",
+			params: am.DeleteTransferActivityParams{
+				Destination: td.Path(),
+			},
+			recorder: func(m *sftp_fake.MockClientMockRecorder, params am.DeleteTransferActivityParams) {
+				m.Delete(
+					mockutil.Context(),
+					params.Destination,
+				).Return(nil)
+			},
+		},
+		{
+			name: "Errors when file does not exist",
+			params: am.DeleteTransferActivityParams{
+				Destination: td.Join("missing"),
+			},
+			recorder: func(m *sftp_fake.MockClientMockRecorder, params am.DeleteTransferActivityParams) {
+				m.Delete(
+					mockutil.Context(),
+					params.Destination,
+				).Return(
+					errors.New("SFTP: unable to remove file \"test.txt\": file does not exist"),
+				)
+			},
+			errMsg: fmt.Sprintf("delete transfer: path: %q: %v", td.Join("missing"), errors.New("SFTP: unable to remove file \"test.txt\": file does not exist")),
+		},
+		{
+			name: "Errors when Delete fails",
+			params: am.DeleteTransferActivityParams{
+				Destination: td.Join(filename),
+			},
+			recorder: func(m *sftp_fake.MockClientMockRecorder, params am.DeleteTransferActivityParams) {
+				m.Delete(
+					mockutil.Context(),
+					params.Destination,
+				).Return(
+					errors.New("SSH: failed to connect: dial tcp 127.0.0.1:2200: connect: connection refused"),
+				)
+			},
+			errMsg: fmt.Sprintf("delete transfer: path: %q: %v", td.Join(filename), errors.New("SSH: failed to connect: dial tcp 127.0.0.1:2200: connect: connection refused")),
+		},
+	} {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			ts := &temporalsdk_testsuite.WorkflowTestSuite{}
+			env := ts.NewTestActivityEnvironment()
+			msvc := sftp_fake.NewMockClient(gomock.NewController(t))
+
+			if tt.recorder != nil {
+				tt.recorder(msvc.EXPECT(), tt.params)
+			}
+
+			env.RegisterActivityWithOptions(
+				am.NewDeleteTransferActivity(logr.Discard(), msvc).Execute,
+				temporalsdk_activity.RegisterOptions{
+					Name: am.DeleteTransferActivityName,
+				},
+			)
+
+			_, err := env.ExecuteActivity(am.DeleteTransferActivityName, tt.params)
+			if tt.errMsg != "" {
+				assert.Error(t, err, activityErr+tt.errMsg)
+				return
+			}
+
+			assert.NilError(t, err)
+		})
+	}
+}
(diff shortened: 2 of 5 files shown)


internal/am/delete_transfer.go
```diff
@@ -0,0 +1,45 @@
...
+	logger logr.Logger
+}
+
+func NewDeleteTransferActivity(logger logr.Logger, client sftp.Client) *DeleteTransferActivity {
+	return &DeleteTransferActivity{client: client, logger: logger}
+}
+
+func (a *DeleteTransferActivity) Execute(ctx context.Context, params *DeleteTransferActivityParams) error {
+	a.logger.V(1).Info("Execute DeleteTransferActivity",
+		"SourcePath", params.Destination,
```
djjuhasz: Nitpick: "SourcePath" -> "Destination" to match the param name.

internal/am/delete_transfer.go
```diff
@@ -0,0 +1,45 @@
...
+	)
+
+	dest, err := os.Open(params.Destination)
```
djjuhasz: At one point we had a check for file existence in the client `GoClient.Delete()` method, but we removed it to avoid the need for an extra round trip server command.  I think the same logic applies here, especially since we have the extra overhead of creating the SSH connection.

internal/am/delete_transfer.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return fmt.Errorf("Delete transfer: %v", err)
+	}
+	defer dest.Close()
+
+	err = a.client.Delete(ctx, params.Destination)
+	if err != nil {
+		return fmt.Errorf("Delete transfer: %v", err)
```
djjuhasz: I think it's better to log this error message and not stop the workflow or report an activity error in the Temporal UI.

internal/workflow/processing.go
```diff
@@ -693,5 +693,13 @@ func (w *ProcessingWorkflow) transferAM(sessCtx temporalsdk_workflow.Context, ti
 		return err
 	}
 
+	// Delete Transfer
```
djjuhasz: `// Delete transfer.`

internal/am/delete_transfer.go
djjuhasz: I think it would be good to include the remote file path in the error message too, so an operator can manually find and delete the transfer if necessary.

djjuhasz (changes requested):
Looks good @Diogenesoftoronto, just a few change requests.

internal/am/delete_transfer_test.go
```diff
@@ -0,0 +1,93 @@
...
+			},
+			recorder: func(m *sftp_fake.MockClientMockRecorder, params am.DeleteTransferActivityParams) {
+				m.Delete(
+					mockutil.Context(),
+					params.Destination,
+				).Return(nil)
+			},
+		},
+		{
+			name: "Logs when local file can't be read",
```
djjuhasz: This test and the following one aren't actually testing anything.  I can see a few choices to fix this:

1. Delete both tests
2. Use https://github.com/tonglil/buflogr (or similar) to record the log then assert that the actual value meets expectations
3. Return an error from the activity. This could be instead of logging or in addition to logging.

I actually prefer option #3.  I think returning an error will make the activity show up as "failed" in the Temporal UI, but we can not propagate the error so the workflow continues anyway.  

I know I said I *didn't* want the error to show up in the Temporal UI, but after thinking about it I think it is better to notify the operator that the delete failed.  I doubt anybody will be checking the app logs unless they know an error has occurred from checking the Temporal UI, the AM Dashboard, or from some other red flag.

internal/am/delete_transfer.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (a *DeleteTransferActivity) Execute(ctx context.Context, params *DeleteTransferActivityParams) error {
+	a.logger.V(1).Info("Execute DeleteTransferActivity",
+		"Destination", params.Destination,
+	)
+
+	err := a.client.Delete(ctx, params.Destination)
+	if err != nil {
+		a.logger.V(1).Error(err, fmt.Sprintf("Delete transfer activity failed for %s", params.Destination))
```
djjuhasz: I think it's helpful to include the client error here (but use "%v" so it can't be unwrapped) to give more information about what went wrong.

djjuhasz (changes requested):
@Diogenesoftoronto I've requested yet another change to functionality, sorry. :(

It's up to you if you want to return an error from the delete activity or just update the tests, but I now think returning an error is the best option.

internal/workflow/processing.go
```diff
@@ -693,5 +693,10 @@ func (w *ProcessingWorkflow) transferAM(sessCtx temporalsdk_workflow.Context, ti
 		return err
 	}
 
+	// Delete Transfer.
+	_ = temporalsdk_workflow.ExecuteActivity(sessCtx, am.DeleteTransferActivityName, am.DeleteTransferActivityParams{
```
djjuhasz: I think the `sessCtx` activity options are too generous here, there are no timeouts and I'm not sure about the retry policy.  I would use the `withActivityOptsForRequest()` options, or something similar.

internal/am/delete_transfer.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	return nil
```
djjuhasz, line 37: I'd return a simple results struct here, just to confirm that the activity succeeded, e.g.

```
type DeleteTransferActivityResult struct {
	OK bool
}
```

internal/am/delete_transfer.go
Diogenesoftoronto: Client error is included in the error log. This is standard practice for the logger. Either way I will change this back to returning an error with fmt.Errorf()

internal/am/delete_transfer.go
Diogenesoftoronto: If the activity does not return an error than it succeeded. If we stick with it returning an error then this is unnecessary value.

sevein: We should probably return an empty struct, e.g.: in https://github.com/temporalio/background-checks/blob/main/activities/activities.go they have four empty structs used as result types. I believe they do that to ease schema evolution, e.g. adding a new field to an empty struct is a backward-compatible change.

internal/am/delete_transfer.go
```diff
@@ -0,0 +1,38 @@
...
+	logger logr.Logger
+}
+
+func NewDeleteTransferActivity(logger logr.Logger, client sftp.Client) *DeleteTransferActivity {
+	return &DeleteTransferActivity{client: client, logger: logger}
+}
+
+func (a *DeleteTransferActivity) Execute(ctx context.Context, params *DeleteTransferActivityParams) error {
+	a.logger.V(1).Info("Execute DeleteTransferActivity",
+		"Destination", params.Destination,
```
sevein: Keys are not capitalized, should read `destination` or `dest`.

internal/am/delete_transfer.go
```diff
@@ -0,0 +1,38 @@
...
+	)
+
+	err := a.client.Delete(ctx, params.Destination)
+	if err != nil {
+		return fmt.Errorf("delete transfer: path: %s. error: %v", params.Destination, err)
```
sevein: I'm suggesting a couple of changes: use `%q` and update wording to avoid using the word `error` (seems redundant and inconsistent with how we've been building error messages).

```suggestion
		return fmt.Errorf("delete transfer: path: %q: %v", params.Destination, err)
```

internal/am/delete_transfer_test.go
```diff
@@ -0,0 +1,104 @@
...
+	temporalsdk_testsuite "go.temporal.io/sdk/testsuite"
+	"go.uber.org/mock/gomock"
+	"gotest.tools/v3/assert"
+	tfs "gotest.tools/v3/fs"
+
+	"github.com/artefactual-sdps/enduro/internal/am"
+	sftp_fake "github.com/artefactual-sdps/enduro/internal/sftp/fake"
+)
+
+func TestDeleteTransferActivity(t *testing.T) {
```
sevein, line 20: We should be able to use t.Parallel() here too.

internal/am/delete_transfer_test.go
```diff
@@ -0,0 +1,104 @@
...
+					mockutil.Context(),
+					params.Destination,
+				).Return(
+					errors.New("SFTP: unable to remove file \"test.txt\": file does not exist"),
+				)
+			},
+			errMsg: fmt.Sprintf("delete transfer: path: %s", td.Path()),
+		},
+		{
+			name: "Errors When Delete fails",
```
sevein:
```suggestion
			name: "Errors when Delete fails",
```

internal/workflow/processing.go
```diff
@@ -693,5 +693,14 @@ func (w *ProcessingWorkflow) transferAM(sessCtx temporalsdk_workflow.Context, ti
 		return err
 	}
 
+	// Delete Transfer.
```
sevein:
```suggestion
	// Delete transfer.
```

internal/workflow/processing_test.go
```diff
@@ -72,6 +72,11 @@ func (s *ProcessingWorkflowTestSuite) SetupWorkflowTest(taskQueue string) {
 		am.NewUploadTransferActivity(logger, sftpc).Execute,
 		temporalsdk_activity.RegisterOptions{Name: am.UploadTransferActivityName},
 	)
+	s.env.RegisterActivityWithOptions(
+		am.NewDeleteTransferActivity(logger, sftpc).Execute,
+		temporalsdk_activity.RegisterOptions{Name: am.
+			DeleteTransferActivityName},
```
sevein:
```suggestion
		temporalsdk_activity.RegisterOptions{
			Name: am.DeleteTransferActivityName
		},
```

sevein (approved):
I haven't had a chance to test but LGTM.

internal/am/delete_transfer_test.go
```diff
@@ -0,0 +1,106 @@
...
+			env.RegisterActivityWithOptions(
+				am.NewDeleteTransferActivity(logr.Discard(), msvc).Execute,
+				temporalsdk_activity.RegisterOptions{
+					Name: am.DeleteTransferActivityName,
+				},
+			)
+
+			_, err := env.ExecuteActivity(am.DeleteTransferActivityName, tt.params)
+			if tt.errMsg != "" {
+				assert.ErrorContains(t, err, tt.errMsg)
```
sevein: I prefer Error over ErrorContains because is more explicity. Sometimes it is not possible, e.g. if the error contains the full path of a temporary directory that is different on each test run, but if that's not the case you could try to see if it works.

internal/am/delete_transfer_test.go
```diff
@@ -0,0 +1,106 @@
...
+			},
+			recorder: func(m *sftp_fake.MockClientMockRecorder, params am.DeleteTransferActivityParams) {
+				m.Delete(
+					mockutil.Context(),
+					params.Destination,
+				).Return(nil)
+			},
+		},
+		{
+			name: "Errors when local file can't be read",
```
djjuhasz: `name: "Errors when file doesn't exist",`

djjuhasz (approved):
I just requested one minor change to a comment.

PR #788, opened 2023-11-27
Add contributing and security markdown files

sallain:
In an attempt to rein in some bad habits, I've added a contributing document that describes how and when to file an issue. It's largely lifted from the Archivematica version. Because the contributing md mentions security, I have also included a security document that tells users how to report a vulnerability.


sallain:
@sevein I created these as standalone MD files, following the model from other repositories, but would be happy to stick them in docs or another location if that makes more sense!

sevein:
> @sevein I created these as standalone MD files, following the model from other repositories, but would be happy to stick them in docs or another location if that makes more sense!

That's perfect.

pre-commit complains because of the missing final newline, here's how to do it from VS Code: https://stackoverflow.com/a/44704969.

PR #789, opened 2023-11-28
Add poll ingest activity

djjuhasz:
Add a poll ingest activity to check the Archivematica ingest status and report completion or errors.


sevein (approved):
LGTM

PR #790, opened 2023-11-28
Allow running a local version of a3m

jraddaoui:
Also, allows setting preservation system in .tilt.env:

- Load `.tilt.env` before getting `ENDURO_PRES_SYSTEM` env. var.
- Fail if `ENDURO_PRES_SYSTEM` is not one of the allowed values.
- Document Tilt environment configuration.


sevein (approved):
LGTM

PR #792, opened 2023-11-29
Create AM secret with secret generator and document AM usage

jraddaoui:


enduro.toml
```diff
@@ -105,6 +105,7 @@ transferDeadline = "1h"
 
 [am.sftp]
 host = "" # The Archivematica Storage Service hostname.
+port = ""
```
jraddaoui, line 116: Without this, the environment variable was not being considered.

hack/kube/overlays/dev-am/enduro-am.yaml
```diff
@@ -99,9 +114,9 @@ spec:
           secret:
             secretName: enduro-am-secret
             items:
-              - key: id_ed25519
+              - key: .id_ed25519.secret
```
jraddaoui, line 117: This keys are set by the secret generator based on the file names. I updated a single rule in the `.gitignore` file to ignore all `*.secret` files in the repo, but we could change that if we care about this keys.

djjuhasz (approved):
I like this solution @jraddaoui, thanks for figuring it out.

Tilt does fail with an error if the secret files are not present, e.g.:
```
stderr: "error: loading KV pairs: env source files: [.am.secret]: evalsymlink failure on '/home/.../enduro/hack/kube/overlays/dev-am/.am.secret' : lstat /home/.../enduro/hack/kube/overlays/dev-am/.am.secret: no such file or directory\n"
```

I think this is okay because it's only in the dev environment, but it might catch some devs by surprise.

jraddaoui:
Thanks @djjuhasz!

Not the best error message, but I think it provides the needed info. I thought that was better than not failing and having to figure out the issue by looking at the AM worker logs, as it did initially for me.

PR #795, opened 2023-11-30
Add worker capacity config option

### LINKED ISSUE #793: Feature: Add configuration for the number of transfers a worker can process concurrently

When more than one transfers are started in a short time Enduro will attempt to process all of them (max 1,000) simultaneously, which can cause resource starvation in the worker environment or in the preservation system (a3m or Archivematica).  I would like a way to limit the number of transfers that each preservation worker can process concurrently without changing the Enduro code.

Add a preservation worker configuration option (e.g. "capacity") to limit the number of transfers a worker can concurrently work on.

None

**Additional context**

- I believe we can use the [MaxConcurrentWorkflowTaskExecutionSize](https://docs.temporal.io/dev-guide/go/foundations#maxconcurrentworkflowtaskexecutionsize) worker option to limit the number of workflow executions a single worker can take on at once

djjuhasz:
Add a configuration option for the a3m and am workers to allow setting the maximum concurrent sessions the worker can handle, which effectively throttles the number of transfers the worker can process concurrently.


cmd/enduro-a3m-worker/main.go
```diff
@@ -120,11 +120,13 @@ func main() {
 
 	// Activity worker.
 	{
+		logger.V(1).Info("A3m worker config", "capacity", cfg.A3m.Capacity)
```
sevein:
```suggestion
		logger.V(1).Info("a3m worker config", "capacity", cfg.A3m.Capacity)
```

sevein (approved):
Perhaps we should define a sensible default, e.g. `1` could be a good start. The default proposed by the Temporal SDK is unlikely to be useful in our context.

sevein:
We should probably user the Viper facility to set the defaults.

https://github.com/artefactual-sdps/enduro/blob/68f1e1e9c3e682d1ee0676c52f0173b76531bd8c/internal/config/config.go#L68-L74

PR #801, opened 2023-12-04
Fix AM transfer deletion

jraddaoui:
The SFTP client uses a `remoteDir` config value that is used as prefix in the `Upload()` and `Delete()` functions. Using the remote path from the upload transfer result duplicates that prefix and makes the deletion to fail with a "file does not exist" error.

Return also the remote relative path from the upload activity and use it as destination for the delete activity.

Refs #785.

### DIFF
--- a/internal/am/upload_transfer.go
+++ b/internal/am/upload_transfer.go
@@ -19,7 +19,10 @@ type UploadTransferActivityParams struct {
 
 type UploadTransferActivityResult struct {
 	BytesCopied int64
-	RemotePath  string
+	// Full path including `remoteDir` config path.
+	RemoteFullPath string
+	// Relative path to the `remoteDir` config path.
+	RemoteRelativePath string
 }
 
 type UploadTransferActivity struct {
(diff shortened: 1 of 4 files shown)


internal/am/upload_transfer.go
```diff
@@ -40,13 +41,15 @@ func (a *UploadTransferActivity) Execute(ctx context.Context, params *UploadTran
 	}
 	defer src.Close()
 
-	bytes, path, err := a.client.Upload(ctx, src, filepath.Base(params.SourcePath))
+	filename := filepath.Base(params.SourcePath)
```
Diogenesoftoronto, line 46: I think we should keep the Upload activity unchanged and just take the filepath.Base of the return value of the RemotePath in processing.go.

internal/workflow/processing.go
```diff
@@ -738,7 +738,7 @@ func (w *ProcessingWorkflow) transferAM(sessCtx temporalsdk_workflow.Context, ti
 	// Delete transfer.
 	activityOpts = withActivityOptsForRequest(sessCtx)
 	err = temporalsdk_workflow.ExecuteActivity(activityOpts, am.DeleteTransferActivityName, am.DeleteTransferActivityParams{
-		Destination: uploadResult.RemotePath,
+		Destination: uploadResult.RemoteFilename,
```
Diogenesoftoronto: This is probably not a good idea. There's the problem here, the filename might be inside a child directory of the the remote directory instead of at the root which it currently is. We probably will add a child directory in the future to avoid something like name conflicts(though you could also just do this with a uuid for filenames) Either way just using the base here makes more sense in the workflow rather than the activity as it require far less changes.

Diogenesoftoronto (changes requested):
Nice catch. I should have remembered this was a problem I addressed this in a past pr but that got lost during the changes.

jraddaoui:
Thanks for the feedback @Diogenesoftoronto!

Good point about this having a child directory at some point, `RemoteRelativePath` and `RemoteFullPath` would make it a bit clearer later on. But I think it's a good idea to do it in the upload activity precisely for that reason, if we use a child directory in the future it will probably happen there. I don't want to make assumptions of what happened in the activity in workflow, if we change the relative path in the activity we'd need to change it in the workflow too, and if we are generating a unique value there for the child directory name it won't be possible to know it without returning it from the activity.

Diogenesoftoronto:
> relative path in the activity we'd need to change it in the workflow too, and if we are generating a unique value there for the child directory name it won't be possible to know it without returning it from the activity.

---
That is a good point.

Diogenesoftoronto (approved):
Looks good.

Issue #783, opened 2023-11-24
Problem: when using Archivematica, processing info is not displayed in Enduro


sallain:
**Is your feature request related to a problem? Please describe.**

When using Enduro with a3m, users can see some information about processing in the Enduro UI. However, when using Archivematica, this information is not available, meaning that users need to go to the Archivematica UI to see this information. We would like to keep users in the Enduro UI for simplicity.

**Describe the solution you'd like**

Add more details about the various processing steps and outcomes to the UI for each workflow run.

**Describe alternatives you've considered**

None.

**Additional context**

This is how the page looks when using a3m:

![image](https://github.com/artefactual-sdps/enduro/assets/4612276/15bd705b-aecb-425c-bf98-1a62165f6fdf)

Specifically, the tasks at the bottom of the page are what we would like to replicate when using Archivematica.

scollazo:
Related issue: https://github.com/archivematica/Issues/issues/1641

djjuhasz:
Fixed by #803 and #804.

PR #803, opened 2023-12-05
Save AM job data in poll transfer activity

djjuhasz:
Related issue #783.

- Create an `am.Jobs` package to query the AM jobs endpoint, and save the data to the package service as preservation tasks
- Integrate `am.Jobs` into PollTransferActivity
- Set the preservation task `startedAt` and `completedAt` to the current time when saving the data to the database becuase the Enduro dashboard won't show tasks that don't have a `startedAt` time

### DIFF
--- a/internal/am/poll_transfer.go
+++ b/internal/am/poll_transfer.go
@@ -7,42 +7,73 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/jonboulle/clockwork"
 	"go.artefactual.dev/amclient"
 	temporal_tools "go.artefactual.dev/tools/temporal"
 	temporalsdk_activity "go.temporal.io/sdk/activity"
+
+	"github.com/artefactual-sdps/enduro/internal/package_"
 )
 
 const PollTransferActivityName = "poll-transfer-activity"
 
 type PollTransferActivityParams struct {
-	TransferID string
+	PresActionID uint
+	TransferID   string
 }
 
 type PollTransferActivity struct {
 	logger logr.Logger
 	cfg    *Config
-	amts   amclient.TransferService
+	clock  clockwork.Clock
+	jobSvc amclient.JobsService
+	pkgSvc package_.Service
+	tfrSvc amclient.TransferService
 }
 
 type PollTransferActivityResult struct {
-	SIPID string
-	Path  string
+	SIPID         string
+	Path          string
+	PresTaskCount int
 }
 
-func NewPollTransferActivity(logger logr.Logger, cfg *Config, amts amclient.TransferService) *PollTransferActivity {
-	return &PollTransferActivity{logger: logger, cfg: cfg, amts: amts}
+func NewPollTransferActivity(
+	logger logr.Logger,
+	cfg *Config,
+	clock clockwork.Clock,
+	jobSvc amclient.JobsService,
+	pkgSvc package_.Service,
+	tfrSvc amclient.TransferService,
+) *PollTransferActivity {
+	return &PollTransferActivity{
+		logger: logger,
+		cfg:    cfg,
+		clock:  clock,
+		jobSvc: jobSvc,
+		pkgSvc: pkgSvc,
+		tfrSvc: tfrSvc,
+	}
 }
 
 // Execute polls Archivematica for the status of a transfer and returns when
 // the transfer is complete or returns an error status. Execute sends an
 // activity heartbeat after each poll.
 //
+// On each poll, Execute requests an updated list of AM jobs performed and saves
+// the job data to the package service as preservation tasks.
+//
 // A transfer status of "REJECTED", "FAILED", "USER_INPUT", or "BACKLOG" returns
 // a temporal.NonRetryableApplicationError to indicate that processing can not
 // continue.
 func (a *PollTransferActivity) Execute(ctx context.Context, params *PollTransferActivityParams) (*PollTransferActivityResult, error) {
-	a.logger.V(1).Info("Executing PollTransferActivity", "TransferID", params.TransferID)
+	var taskCount int
 
+	a.logger.V(1).Info("Executing PollTransferActivity",
+		"PresActionID", params.PresActionID,
+		"TransferID", params.TransferID,
+	)
+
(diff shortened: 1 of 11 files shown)


internal/am/jobs.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+var jobStatusToPreservationTaskStatus = map[amclient.JobStatus]package_.PreservationTaskStatus{
+	amclient.JobStatusUnknown:    package_.TaskStatusUnspecified,
+	amclient.JobStatusComplete:   package_.TaskStatusDone,
+	amclient.JobStatusProcessing: package_.TaskStatusInProgress,
+	amclient.JobStatusFailed:     package_.TaskStatusError,
+}
+
+type PreservationAction struct {
```
sevein: Delving into naming conventions once more! What do you think about renaming PreservationAction to JobTracker? This name might better reflect its role in maintaining a stateful tracking of jobs.

internal/am/jobs.go
```diff
@@ -0,0 +1,120 @@
...
+		clock:  clock,
+		jobSvc: jobSvc,
+		pkgSvc: pkgSvc,
+
+		id:       id,
+		savedIDs: make(map[string]struct{}),
+	}
+}
+
+// SaveTasks polls the Archivematica jobs list endpoint to get a list of
```
sevein: s/polls/fetches?

internal/am/jobs.go
```diff
@@ -0,0 +1,120 @@
...
+	clock  clockwork.Clock
+	jobSvc amclient.JobsService
+	pkgSvc package_.Service
+
+	// id is the Preservation Action ID.
+	id uint
+
+	// savedIDs caches the ID of jobs that have already been saved so we don't
+	// create duplicates.
+	savedIDs map[string]struct{}
```
sevein: This is perfect, but now wondering whether we should eventually adopt some kind of generic data structure, e.g. https://github.com/zyedidia/generic has a set type - see https://go.dev/play/p/hHNbIN2msXf. `struct{}{}` is a memory trick that may be as well hidden I guess.

internal/am/jobs.go
```diff
@@ -0,0 +1,120 @@
...
+	for _, job := range jobs {
+		// Wait until a job is complete (or failed) before saving it.
+		if job.Status == amclient.JobStatusProcessing {
+			continue
+		}
+
+		pt := ConvertJobToPreservationTask(job)
+		pt.PreservationActionID = pa.id
+		pt.StartedAt = sql.NullTime{Time: pa.clock.Now(), Valid: true}
+		pt.CompletedAt = sql.NullTime{Time: pa.clock.Now(), Valid: true}
```
sevein: We should probably call pa.clock.Now() just once.

sevein (approved):
Nicely done!

internal/am/poll_transfer.go
```diff
@@ -52,17 +82,33 @@ func (a *PollTransferActivity) Execute(ctx context.Context, params *PollTransfer
 			return nil, ctx.Err()
 		case <-ticker.C:
 			reqCtx, cancel := context.WithTimeout(ctx, a.cfg.PollInterval/2)
-			resp, httpResp, err := a.poll(reqCtx, params.TransferID, cancel)
-			if err == errWorkOngoing {
+			resp, _, err := a.poll(reqCtx, params.TransferID, cancel)
+			if err == ErrWorkOngoing {
+				// Update preservation task data.
+				count, err := pa.SaveTasks(ctx, params.TransferID)
```
sevein: Should we use the context with timeout previously defined to save the tasks? I'm thinking if things go wrong here we could be missing a heartbeat, but I don't know if that's a big deal.

internal/am/jobs.go
djjuhasz: I did s/polls/queries

internal/am/jobs.go
sevein: I meant in the long term really, not asking you to do it now.

internal/am/jobs_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"go.artefactual.dev/tools/mockutil"
+	temporal_tools "go.artefactual.dev/tools/temporal"
+	"go.uber.org/mock/gomock"
+	"gotest.tools/v3/assert"
+
+	"github.com/artefactual-sdps/enduro/internal/am"
+	fake_package "github.com/artefactual-sdps/enduro/internal/package_/fake"
+)
+
+func TestPreservationAction_UpdateTasks(t *testing.T) {
```
Diogenesoftoronto: We can have t.Parrellel here as well

internal/am/poll_transfer.go
Diogenesoftoronto: Considering that heartbeats are half the configurable time? seems like it could be. if someone decided polling once an hour or something crazy. Maybe just a note for setup that increasing polling time by a large amount may result in missing polling intervals for tasks that are saved.

internal/am/poll_transfer_test.go
```diff
@@ -17,48 +19,81 @@ import (
...
 	"github.com/artefactual-sdps/enduro/internal/am"
+	fake_package "github.com/artefactual-sdps/enduro/internal/package_/fake"
+)
+
+var (
+	http200Resp = http.Response{StatusCode: http.StatusOK, Status: "200 OK"}
+	http400Resp = http.Response{StatusCode: http.StatusBadRequest, Status: "400 Bad request"}
 )
 
 func TestPollTransferActivity(t *testing.T) {
```
Diogenesoftoronto, line 30: we can use t.Parallel here

Diogenesoftoronto (changes requested):
Looks good so far.

internal/am/poll_transfer.go
djjuhasz: I've updated this method call to use `reqCtx` but I refactored the code so I didn't have to pass the context cancel function to the `SaveTasks()` method too.  I've also increased the context timeout to the full `PollInterval`, because the Heartbeat Timeout is set to `2 * PollInterval` so we should be well under that deadline.

Diogenesoftoronto (approved):

internal/am/poll_transfer.go
sevein: Good call, the simplified loop is much easier to read now.

PR #804, opened 2023-12-06
Save AM ingest preservation tasks

djjuhasz:
Refs issue #783.

Update poll ingest activity to request an AM job list and save the job data as preservation tasks.


sevein (approved):
:+1:

Diogenesoftoronto (approved):

PR #808, opened 2023-12-07
Fix download activity parameters and returns

### LINKED ISSUE #807: Problem: the download activity doesn't follow best practice for input parameters or return values
**Describe the bug**

The download activity takes two string parameters (watcherName, key) and returns a string and an error.  The Temporal Go SDK [recommends passing a single struct to activities](https://docs.temporal.io/dev-guide/go/foundations#activity-parameters) to allow adding parameters in the future without breaking running workflow executions.  Likewise, returning a struct from the activity allows adding parameters in a future-safe way.

**To Reproduce**

This problem is only apparent when upgrading the workflow code when there are running workflow executions.

**Expected behavior**

The download activity should take a single struct parameter (in addition to ctx), and return a struct and an error.

**Additional context**

Issue #800 describes a similar problem, but limited to activities that on return an error.

djjuhasz:
- Use structs for the activity parameters and return value for future safety
- Change watcher server `Download()` method so it accepts a destination filepath (string) instead of a file writer to allow testing without needing to pass an open file writer
- Add a logger and log the download activity execution parameters
- Add unit tests
- Update workflow tests

### DIFF
--- a/cmd/enduro-a3m-worker/main.go
+++ b/cmd/enduro-a3m-worker/main.go
@@ -135,7 +135,7 @@ func main() {
 			os.Exit(1)
 		}
 
-		w.RegisterActivityWithOptions(activities.NewDownloadActivity(wsvc).Execute, temporalsdk_activity.RegisterOptions{Name: activities.DownloadActivityName})
+		w.RegisterActivityWithOptions(activities.NewDownloadActivity(logger, wsvc).Execute, temporalsdk_activity.RegisterOptions{Name: activities.DownloadActivityName})
 		w.RegisterActivityWithOptions(activities.NewBundleActivity(wsvc).Execute, temporalsdk_activity.RegisterOptions{Name: activities.BundleActivityName})
 		w.RegisterActivityWithOptions(a3m.NewCreateAIPActivity(logger, &cfg.A3m, pkgsvc).Execute, temporalsdk_activity.RegisterOptions{Name: a3m.CreateAIPActivityName})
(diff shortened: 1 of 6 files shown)


internal/watcher/watcher.go
```diff
@@ -74,7 +74,7 @@ type Service interface {
 	ByName(name string) (Watcher, error)
 
 	// Download blob given an event.
-	Download(ctx context.Context, w io.Writer, watcherName, key string) error
+	Download(ctx context.Context, dest, watcherName, key string) error
```
sevein: > Change watcher server Download() method so it accepts a destination filepath (string) instead of a file writer to allow testing without needing to pass an open file writer.

`io.Writer` is nice because it keeps the function decoupled from the file system - its sole responsibility is to write data, and it's up to the caller to do the rest. It's nice that these are all possible values we can pass: *os.File, *bytes.Buffer, *blob.Writer...

I suggest to maintain it. When it comes to TestDownloadActivity, this is how it coud look like: https://gist.github.com/sevein/e7529e067601dfa1a2a9c44adf113e4b. The trick used in `gomock.AssignableToTypeOf((*os.File)(nil))` wasn't obvious to me, I'm wondering if that was the original motivation for this change.

sevein (changes requested):
LGTM. I just have a tiny request: could we have our `io.Writer` back? :pray: :smile:

internal/watcher/watcher.go
djjuhasz: @sevein very nice! I was struggling with how to test that download actually writes data to the target file, so thank you for demonstrating how to do that. :)

sevein (approved):
LGTM

PR #809, opened 2023-12-13
Fix "can't find bundle activity" error

djjuhasz:
Refs #795.

This may or may not fix the periodic "can't find bundle activity" error I've been getting, but if it doesn't fix it the additional logging may help diagnose the problem.

- Use an intermediate variable to assign the bundle activity results to `tinfo.Bundle`
- Add logging to the bundle activity and log the execution params
- Update workflow tests
- Add bundle activity unit tests
- Generate a mock watcher for tests


djjuhasz:
I need to fix the directory and file permissions.

PR #810, opened 2023-12-15
Fix "can't find bundle activity" error

djjuhasz:
Refs #795.

This may or may not fix the periodic "can't find bundle activity" error I've been getting, but even if it doesn't fix the bug, the additional logging may be helpful in diagnosis.

- Use an intermediate variable to assign the bundle activity results to `tinfo.Bundle`
- Add logging to the bundle activity and log the execution params
- Update workflow tests
- Add bundle activity unit tests (39% -> 64% coverage in workflow activities :tada:)
- Generate a mock watcher for tests


internal/bundler/bundler.go
```diff
@@ -104,9 +104,7 @@ func (b *Bundler) Write(name string, r io.Reader) error {
 	if err != nil {
 		return fmt.Errorf("error writing file: %w", err)
 	}
-	defer func() {
-		_ = file.Close()
-	}()
+	defer file.Close()
```
sevein, line 107: I'm now realizing that the error returned by Close() shouldn't be ignored because we're writing.
We could try something like the following, but I haven't tested:

```go
func (b *Bundler) Write(name string, r io.Reader) error {
	file, err := b.Create(name)
	if err != nil {
		return fmt.Errorf("create file: %w", err)
	}

	var closeErr error
	defer func() {
		closeErr = fmt.Errorf("close file: %w", file.Close())
	}()

	_, copyErr := io.Copy(file, r)
	if copyErr != nil {
		copyErr = fmt.Errorf("copy file: %w", copyErr)
	}

	return errors.Join(copyErr, closeErr)
}
```

Note: `errors.Join` returns nil if both copyErr and closeErr are nil.

sevein (approved):
:+1:

internal/bundler/bundler.go
djjuhasz: @sevein it seems that the return values are evaluated *before* the deferred function runs, so `closeErr` will always be nil, e.g. https://go.dev/play/p/iBY6l14I9fx

Based on [understanding defer in go (Digital Ocean)](https://www.digitalocean.com/community/tutorials/understanding-defer-in-go) think it would work to explicitly call `file.Close()` like:

```
func (b *Bundler) Write(name string, r io.Reader) error {
	file, err := b.Create(name)
	if err != nil {
		return fmt.Errorf("create file: %w", err)
	}
	defer file.Close()

	_, copyErr := io.Copy(file, r)
	if copyErr != nil {
		return fmt.Errorf("copy file: %w", copyErr)
	}
        
        // Return any errors on file close
	return file.Close()
}
```

sevein: That makes sense, thank you!

Issue #811, opened 2023-12-18
Problem: frontend fails to create websocket conn after Dex auth forward


sevein:
**Describe the bug**

The dashboard seems to be unable to establish the WebSocket connection to /api/package/monitor after signing in successfully. It needs an aditional page refresh to connect successfully. This shows when submitting packages because the UI won't reflect the updates unless the connection is established beforehand.

**To Reproduce**

1. Open http://localhost:8080/
2. Log out
3. Open the Network tab and clear the network log
4. Sign in
5. Observe that the ws connection to /api/package/monitor is not established
6. Refresh page, connection is established.

**Expected behavior**

The dashboard should always be connected to the WebSocket updates channel.

**Additional context**

Perhaps the connection attempt is made but the authentication ticket is not accepted, needs some debugging.

PR #812, opened 2023-12-19
Establish WebSocket connection after login

jraddaoui:
Refs #811.


jraddaoui:
I initially tried to move the establishment to the store but that broke the tests and I could not find an example where we fake the API client.

sevein (approved):
Noice!

PR #816, opened 2024-01-04
Make sftp.dial() thread safe

djjuhasz:
Return a new connection on each call.

### DIFF
--- a/internal/sftp/goclient.go
+++ b/internal/sftp/goclient.go
@@ -21,9 +21,6 @@ import (
 type GoClient struct {
 	cfg    Config
 	logger logr.Logger
-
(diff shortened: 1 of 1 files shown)


sevein (approved):
Very nice! :+1: 

At some point we could consider adding a connection pool for performance or efficiency. [rclone's []*conn](https://github.com/rclone/rclone/blob/578b9df6eae50dd75283ba3190408542a0fb2f53/backend/sftp/sftp.go#L531-L532) could be a good reference, but I suspect the added complexity isn't going to pay off for now. I should set up OpenTelemetry so we can start measuring things like connection establishment overheads, etc. that way at least we can tell if we'd need the investment.

PR #817, opened 2024-01-04
Update Pulumi deps

sevein:
I've made a few upgrades of major versions. I scanned the changelogs and saw that the v4 notes of pulumi-docker include some notes with some additional steps required that you may need to run in existing deployments, see https://github.com/pulumi/pulumi-docker/releases/tag/v4.0.0.


djjuhasz (approved):
LGTM :+1:

jraddaoui (approved):

PR #818, opened 2024-01-04
Don't retry SSH connection when auth fails

djjuhasz:
Authentication failures can not be resolved by retrying the connection with the same credentials.

### DIFF
--- a/internal/sftp/client.go
+++ b/internal/sftp/client.go
@@ -2,23 +2,20 @@ package sftp
 
 import (
 	"context"
+	"fmt"
 	"io"
 )
 
 type AuthError struct {
-	err error
+	Message string
 }
 
 func (e *AuthError) Error() string {
-	return e.err.Error()
+	return fmt.Sprintf("auth: %s", e.Message)
 }
 
-func (e *AuthError) Unwrap() error {
-	return e.err
-}
-
-func NewAuthError(e error) *AuthError {
-	return &AuthError{err: e}
+func NewAuthError(e error) error {
+	return &AuthError{Message: e.Error()}
 }
 
 // A Client manages the transmission of data over SFTP.
--- a/internal/sftp/client.go
+++ b/internal/sftp/client.go
@@ -5,6 +5,22 @@ import (
 	"io"
 )
 
+type AuthError struct {
+	err error
+}
+
+func (e *AuthError) Error() string {
+	return e.err.Error()
+}
+
+func (e *AuthError) Unwrap() error {
+	return e.err
+}
+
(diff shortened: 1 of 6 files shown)


internal/sftp/client.go
```diff
@@ -5,6 +5,22 @@ import (
...
+	return e.err.Error()
+}
+
+func (e *AuthError) Unwrap() error {
+	return e.err
+}
+
+func NewAuthError(e error) *AuthError {
+	return &AuthError{err: e}
+}
```
sevein: A possible alternative:

```
type AuthError struct {
	Reason string
}

func (e *AuthError) Error() string {
	return fmt.Sprintf("auth: %s", e.Reason)
}

func NewAuthError(err error) error {
	return &AuthError{Reason: err.Error()}
}
```

This version:
* Doesn't use Unwrap because we don't need to give access to the underlying error.
* Uses the error type in the constructor signature, see https://go.dev/doc/faq#nil_error:
   > It's a good idea for functions that return errors always to use the error type in their signature rather than a concrete type such as *MyError, to help guarantee the error is created correctly. As an example, [os.Open](https://go.dev/pkg/os/#Open) returns an error even though, if not nil, it's always of concrete type [*os.PathError](https://go.dev/pkg/os/#PathError).

I think that we'd want the function to return a concrete type (*AuthError) if we were doing some performance-focused work, e.g. https://github.com/connectrpc/connect-go/blob/d88758dc0e89170db922ecd20f16cec57662ec23/error.go#L117-L128 could be a good example where the design was driven by performance objectives, like avoiding memory allocations. This is a guess though, not entirely sure.

sevein (approved):
LGTM, only one minor suggestion.

internal/sftp/client.go
djjuhasz: @sevein does `Reason` need to be exported?

djjuhasz: Exporting the `Reason` field does have the nice benefit of being able to set an arbitrary string without having to create an error first, e.g. 

```
e := &sftp.AuthError{Reason: "bad password"}
```
vs.

```
e := sftp.NewAuthError(errors.New("bad password"))
```

internal/sftp/ssh.go
```diff
@@ -17,32 +18,32 @@ import (
 // returns a client connection.
 //
 // Only private key authentication is currently supported, with or without a
-// passphrase.
+// passphrase.SSH: %v",
```
sevein, line 21: Looks like this was accidentally changed!

djjuhasz: Oops! :facepalm:

Issue #820, opened 2024-01-10
Problem: AIPs do not download via the Archivematica Storage Service


Diogenesoftoronto:
**Is your feature request related to a problem? Please describe.**

Currently when we download an AIP in the Dashboard's UI it comes from Minio, our clients need to be able to download it directly from the Archivematica Storage Service when using Enduro with Archivematica.

**Describe the solution you'd like**
Change the backend API such that if the preservation system identified in the configuration file is Archivematica, change the download endpoint behavior to use the storage service and not Minio.

**Describe alternatives you've considered**

We could change/add a button/option on the dashboard to have the option of downloading from the storage service instead of Minio. And we could bypass Enduro and talk directly to Archivematica, or add a new endpoint to Enduro for this.  

**Additional context**

#782

djjuhasz:
It looks like the legacy Enduro implementation of AIP download is here: https://github.com/artefactual-labs/enduro/blob/cced3fcdbd3eaf5bc1fff610a3da3b7899eba0a2/internal/collection/goa.go#L324

sevein:
I didn't mean to close the issue. Keith's PR implements a new storage location type based on the ssblob package, but I don't think we have a mechanism yet to enable it when deploying with Archivematica. I believe that locations are currently populated using raw SQL using k8s jobs.

PR #822, opened 2024-01-10
Add AIP download from AM Storage Service

Diogenesoftoronto:
- Draft, the actual download to the storage service is not implemented,
this is just a proof of concept for how it might work.

#820


sevein:
Replaced by https://github.com/artefactual-sdps/enduro/pull/831.

Issue #823, opened 2024-01-10
Problem: complex tooling and environment variables in development setup


Diogenesoftoronto:
**Is your feature request related to a problem? Please describe.**

It can take quite a long time to get different environments setup for each developer and all the tooling installed to fit their machine.
There is also the problem of version those tools when problems arise. 
**Describe the solution you'd like**

A shareable environment via devbox, just like how we containerise our applications, we should containerise or development.

**Describe alternatives you've considered**
Vagrant from hashicorp, devcontainers.

**Additional context**
...

joel-simpson:
another way to frame this is "our dev environment is complex and hard to debug"? a devbox container doesn't solve that exactly but you get more consistency in how all of the different variables are set

PR #824, opened 2024-01-10
Add devbox

Diogenesoftoronto:
This is a minimal devbox setup for our project. @latest for the versions
probably does not make sense past this initial concept as that will mean
different versions of the software when people start the project.

Adding installations instruction so that people can get started with devbox is another thing that can be added.

ref #823

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -1,11 +1,14 @@
 {
   "packages": [
-    "go_1_20@latest",
     "kubectl@latest",
     "tilt@latest",
     "k3d@latest",
     "nodejs_20@latest",
-    "git@latest"
+    "git@latest",
+    "go_1_21@latest",
+    "docker@latest",
+    "python3@latest",
+    "markdownlint-cli@latest"
   ],
   "shell": {
     "init_hook": [
--- a/devbox.json
+++ b/devbox.json
@@ -1,0 +1,20 @@
+{
+  "packages": [
+    "go_1_20@latest",
+    "kubectl@latest",
+    "tilt@latest",
+    "k3d@latest",
+    "nodejs_20@latest",
+    "git@latest"
+  ],
+  "shell": {
+    "init_hook": [
+      "echo 'Welcome to devbox!' > /dev/null"
+    ],
+    "scripts": {
+      "test": [
+        "echo \"Error: no test specified\" && exit 1"
+      ]
+    }
+  }
+}
--- a/docs/src/dev-manual/devel.md
+++ b/docs/src/dev-manual/devel.md
@@ -1,9 +1,20 @@
 # Local/Development environment
 
+## Requirements
+
+### Quickstart with Devbox
+
+Devbox is currently an experimental approach to managing and sharing our
+developer environments quickly and efficiently, it is not however _well
+tested_.
+
+First install [Devbox]. After installing Devbox, run `devbox shell`. Continue
+the instructions [here](#dex-host).
+
+### Standard installation
+
 Enduro uses Tilt to set up a local environment building the Docker images in a
 Kubernetes cluster. It has been tested with k3d, Minikube and Kind.
-
-## Requirements
 
 - [Docker] (v18.09+)
 - [kubectl]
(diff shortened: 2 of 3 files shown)


devbox.json
```diff
@@ -0,0 +1,20 @@
+{
+  "packages": [
+    "go_1_20@latest",
```
sevein: Shouldn't we be using "go_1_21@latest"?

devbox.json
```diff
@@ -0,0 +1,20 @@
...
+    "kubectl@latest",
+    "tilt@latest",
+    "k3d@latest",
```
sevein, line 5: Curious to know, what's the limit here? You list k3d, but why not docker or vscode? Genuinely asking!

docs/src/dev-manual/devel.md
```diff
@@ -1,10 +1,15 @@
 # Local/Development environment
 
-Enduro uses Tilt to set up a local environment building the Docker images in a
-Kubernetes cluster. It has been tested with k3d, Minikube and Kind.
+## Quickstart with Devbox
+
+First install [devbox]. After installing devbox, run `devbox shell`. Continue
+the instructions [here](#dex-host).
```
sevein: Can you make this a subsection of Requirements and briefly describe what devbox is and that it's experimental and not part of the standard dev workflow yet?

sevein (changes requested):
Thanks for doing this! I've always wanted to try devbox. It has potential to replace the hack/make stuff we use to build our tools.

I've noticed that you haven't included the lock file, is there any reason for not doing that? It feels like it's going to be important to have a shared lock file for reproducibility, just like our go.sum, etc.

devbox.json
Diogenesoftoronto: Yeah honestly I think I just forgot go1.21 released

devbox.json
Diogenesoftoronto: docker is a good addition, however not sure if we should package vscode, as not everyone uses the editor. I was trying get a minimal set of tools need for the project to work. Other developer tools like k9s, github-cli, glow, and fish or useful scripts that we like beyond make are also options.

Diogenesoftoronto:
> Thanks for doing this! I've always wanted to try devbox. It has potential to replace the hack/make stuff we use to build our tools.
> Yes it is very useful for that. There is a lot to explore.
> I've noticed that you haven't included the lock file, is there any reason for not doing that? It feels like it's going to be important to have a shared lock file for reproducibility, just like our go.sum, etc.

I considered adding the lockfile, and my initial belief about the lockfile was wrong. I am going to add it.

docs/src/dev-manual/devel.md
```diff
@@ -1,10 +1,21 @@
...
 
+## Requirements
+
+### Quickstart with Devbox
+
+Devbox is currently an experimental approach to managing and sharing our
+developer environments quickly and efficiently, it is not however _well
+tested_.
+
+First install [devbox]. After installing devbox, run `devbox shell`. Continue
```
sevein:
```suggestion
First install [Devbox]. After installing Devbox, run `devbox shell`. Continue
```

docs/src/dev-manual/devel.md
```diff
@@ -1,10 +1,21 @@
...
+the instructions [here](#dex-host).
+
+### Standard Installation
```
sevein: We're using sentence case capitalization:

```suggestion
### Standard installation
```

docs/src/dev-manual/devel.md
```diff
@@ -247,3 +258,4 @@ is sometimes not setup properly. To solve it, from the Tilt UI, restart the
 [gcc]: https://gcc.gnu.org/
 [visual studio code]: https://code.visualstudio.com/
 [working with archivematica]: archivematica.md
+[devbox]: https://www.jetpack.io/devbox/docs/quickstart/#install-devbox
```
sevein, line 261:
```suggestion
[Devbox]: https://www.jetpack.io/devbox/docs/quickstart/#install-devbox
```

sevein (approved):
LGTM, only a few minor changes requested.

docs/src/dev-manual/devel.md
Diogenesoftoronto: hmm this seems out of alighnment with how we have capitalised every other variable here.

sevein: Oh maybe it's case insensitive? That makes sense. I only meant that the proper name is Devbox as opposed to devbox.

PR #827, opened 2024-01-12
Send heartbeat with progress during SFTP upload

### LINKED ISSUE #815: Problem: Archivematica SFTP upload provides no progress updates
**Describe the bug**

When uploading a transfer to Archivematica for preservation processing, the SFTP upload process does not provide any progress updates on the upload until the upload is complete.  For large transfers this means the workflow could be waiting for minutes or hours without any indication of whether the upload is still progressing, or if it has failed or stalled.

**To Reproduce**

1. [Configure Enduro to use Archivematica for preservation processing](https://enduro.readthedocs.io/dev-manual/archivematica/)
2. Upload a large transfer to the Enduro MinIO "sips" bucket
3. After the "upload transfer" activity starts, the Temporal UI shows a Workflow "ActivityTaskScheduled" Event for "UploadTransferActivity" until the upload has completed (with a 2 hour timeout until the activity fails).

**Expected behavior**

Enduro should provide regular updates to the user that the upload is still running and making progress.

**Screenshot**

image

**Additional context**

None.

djjuhasz:
Fixes issue #815.


internal/am/upload_transfer.go
```diff
@@ -57,9 +76,30 @@ func (a *UploadTransferActivity) Execute(ctx context.Context, params *UploadTran
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-done:
+			return
+		case <-ticker.C:
+			temporalsdk_activity.RecordHeartbeat(ctx,
+				fmt.Sprintf("uploaded %d bytes of %d.", a.client.Progress(), fileSize),
```
sevein: An issue with Progress() is that in the scenario where multiple uploads are happening concurrently through the same client instance, it might be challenging to accurately report the progress of a specific upload. One option could be to have Upload() return a WriteCounter that is used only in the context of that upload. Since the count is probably updated and read from different goroutines I suspect you'll need to write and read atomically, e.g.: https://go.dev/play/p/PgOX4uZn9Jp. You could also use a mutex or a channel, but "sync/atomic" is pretty common, more efficient and lets you build cleaner APIs.

sevein (changes requested):
Looks like I forgot to commit the fakes generated by a new version of github.com/uber-go/mock? Perhaps that work could go into its own dedicated commit.

internal/am/upload_transfer.go
djjuhasz: Yeah, I realized last night that the current implementation isn't going to work with multiple connections. :cry: I think making the writeCounter specific to the connection should avoid simultaneous writes and the need for `sync/atomic` as we don't re-use connections right now.

sevein:
Looking good so far. This reminds me of http.Get which unblocks when the headers are received so you can read resp.Body (the underlying conn remains open until resp.Body.Close() is called).

djjuhasz:
> Looking good so far. This reminds me of http.Get which unblocks when the headers are received so you can read resp.Body (the underlying conn remains open until resp.Body.Close() is called).

Thanks @sevein! It's been a great opportunity to learn more about async processing with goroutines and channels.  My brain was smoking a bit yesterday trying to figure out the bugs, but I learned a lot. :)

And it turns out atomics came in handy after all! :laughing:

internal/am/upload_transfer.go
```diff
@@ -5,36 +5,54 @@ import (
...
 type UploadTransferActivityParams struct {
+	// Local path of the source file.
 	SourcePath string
 }
 
 type UploadTransferActivityResult struct {
-	BytesCopied int64
-	// Full path including `remoteDir` config path.
+	// Bytes copied to the remote file over the SFTP connection.
+	BytesCopied uint64
```
sevein: FYI: I default to signed types these days because that seems to be genera consensus, e.g. see [blob](https://github.com/google/go-cloud/blob/01064b751a87ca20722835090310a7034a9fc598/blob/driver/driver.go#L161-L162), [len](https://pkg.go.dev/builtin#len), [io.Writer](https://pkg.go.dev/io), etc...

I'm not 100% sure why that's the case. I think it's Alan Donovan who said to always use signed types unless you're planning to use bitwise operations. In [The Go Programming Language book (3.1 Integers)](https://learning.oreilly.com/library/view/the-go-programming/9780134190570/ebook_split_028.html), there's a section where he talks about preventing certain errors that may explain where this decision came from.

internal/sftp/client.go
```diff
@@ -24,9 +28,28 @@ func NewAuthError(e error) error {
...
-	// Upload transfers data from the provided source reader to a specified
-	// destination on the SFTP server.
-	Upload(ctx context.Context, src io.Reader, dest string) (bytes int64, remotePath string, err error)
-	// Delete removes data from the provided dest on the SFTP server.
-	Delete(ctx context.Context, dest string) (err error)
+	// Delete removes dest from the SFTP server.
+	Delete(ctx context.Context, dest string) error
+	// Upload asynchronously copies data from the src reader to the specified
+	// dest on the SFTP server.
+	Upload(ctx context.Context, src io.Reader, dest string) (remotePath string, upload AsyncUpload, err error)
```
sevein, line 35: I'm interested to see how our approach to this API evolves over time. Lately, I've noticed a trend in the community towards more synchronous APIs, and moving away from using channels due to their complexity (semantics) and potential for errors.

If we had to buidl a public API that prioritizes simplicity and ease of use, we could explore these two alternatives:

1. Provide a io.Writer, e.g. [gocloude.dev's bucket.NewWriter](https://gocloud.dev/howto/blob/#writing),
2. Provide a callback, e.g.:

```go
type UploadStatusCallback func(bytesWritten int64)
    
// Upload blocks.
Upload(ctx context.Context, src io.Reader, dest string, updater UploadStatusCallback) (remotePath string, err error)
```

Just to be clear, I'm not suggesting we need to change the current design.

sevein (approved):
LGTM

internal/am/upload_transfer.go
djjuhasz: Okay, I changed BytesCopied back to an `int64`.  Go int types are so weird. :confused:

internal/sftp/client.go
djjuhasz: @sevein my *original* plan was to add an `UploadChunk()` method that would take start and width parameters, and then just loop it in the upload activity with a heartbeat + bytes upload count each iteration.   I thought it would fairly easy to do chunked read/writes with the `io` interfaces, but my reading on it gave me the impression that reading/writing chunks like that is not very efficient.  The best answer I found for how to do an (up/down)load progress update was to use `io.TeeWriter()` like I've used here.  Unfortunately I don't see any way to use the TeeWriter for updates without goroutines and channels for communication. :shrug: 

I'm not sure how you get around the blocking problem with `io.Writer` or a callback to provide heartbeats during the upload, without using goroutines somehow. It seems like you would have to push the aysnc complexity to the caller instead of handling it in the Upload method.

sevein: I've read somewhere that the Go SDK throttles heartbeats. I suspect that's why in the code examples they just show `RecordHeartbeat` being used in loops. The SDK makes sure we don't flood the Temporal frontend service with unnecessary heartbeats. I'll try to find where I saw that.

sevein: Here: https://docs.temporal.io/activities#throttling.

sevein: So the activity code could look like:

```go
type progressReader struct {
    r        io.Reader
    callback func(int)
}

func (pr *progressReader) Read(p []byte) (int, error) {
    n, err := pr.r.Read(p)
    if err == nil {
        pr.callback(n)
    }
    return n, err
}

io.Copy(w, progressReader{myFileReader, myHeartbeater})
```

Not sure if that's any easier, maybe?

PR #831, opened 2024-01-16
Use location and PackageReader for Download

### LINKED ISSUE #820: Problem: AIPs do not download via the Archivematica Storage Service

Diogenesoftoronto:
This implementation does not use the preservation config for determining
the correct call to download, instead it leverages the package reader
function and the Location interface to manage the correct storage service
to use.

This is partially implemented. Not sure how to actually snag the url from
the location configuration. I really need guidance there. This also seems
rather complex compared to the other implementation.

Closes https://github.com/artefactual-sdps/enduro/issues/820.


sevein:
@Diogenesoftoronto I'm assuming you don't need this one anymore, I'll close it.

Replaced by https://github.com/artefactual-sdps/enduro/pull/847.

PR #838, opened 2024-01-22
Add path activation script

sevein:
This pull request includes a path activation script for the fish shell inspired by Python's virtual environments.

Usage:

    source hack/activate.fish

It adds the local binary cache directory of the project to the PATH environment string. Now you can use tools like ent, goa, golangci-lint, gomajor, gosec, gotestsum, mockgen, shfmt or tparse with the confidence of using the versions required by this project as opposed to any global version you have installed previously.

If you're using devbox, the equivalent would be to execute `devbox shell`.

### DIFF
--- a/hack/activate.fish
+++ b/hack/activate.fish
@@ -1,0 +1,46 @@
+# This file must be used with "source hack/activate.fish" *from fish*
+# (https://fishshell.com/). You cannot run it directly.
+
+function deactivate -d "Exit virtual environment and return to normal shell environment"
+    if test -n "$_OLD_VIRTUAL_PATH"
+        set -gx PATH $_OLD_VIRTUAL_PATH
+        set -e _OLD_VIRTUAL_PATH
+    end
+
+    if test -n "$_OLD_FISH_PROMPT_OVERRIDE"
+        set -e _OLD_FISH_PROMPT_OVERRIDE
+        # prevents error when using nested fish instances (Issue #93858)
+        if functions -q _old_fish_prompt
+            functions -e fish_prompt
+            functions -c _old_fish_prompt fish_prompt
+            functions -e _old_fish_prompt
+        end
+    end
+
+    set -e VIRTUAL_ENV
+    set -e VIRTUAL_ENV_PROMPT
+    if test "$argv[1]" != "nondestructive"
+        # Self-destruct!
+        functions -e deactivate
+    end
+end
+
+# Unset irrelevant variables.
+deactivate nondestructive
+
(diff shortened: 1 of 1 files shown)


hack/activate.fish
```diff
@@ -0,0 +1,46 @@
+# This file must be used with "source hack/activate.fish" *from fish*
+# (https://fishshell.com/). You cannot run it directly.
```
Diogenesoftoronto: Is there a more direct link to the information here?

Diogenesoftoronto (approved):

hack/activate.fish
sevein: What do you mean exactly?

Diogenesoftoronto: For example: https://fishshell.com/docs/current/cmds/source.html is the link to the man page of source

sevein: I see, will do, thanks.

djjuhasz:
@sevein would a similar script be helpful for Bash, or are the tool versions already handled in Bash by the Makefile?  I use Zsh and I don't want to introduce spurious changes because I have the wrong tool version.

Diogenesoftoronto:
> @sevein would a similar script be helpful for Bash, or are the tool versions already handled in Bash by the Makefile? I use Zsh and I don't want to introduce spurious changes because I have the wrong tool version.

zsh can run the normal source activate code that fish cannot. The reason is that fish is not posix complaint shell at all. for example in fish a variable is defined via the set command. This breaks almost all bash scripts when porting to fish. it is why i usually just execute bash scripts on my fish shell via bash -c "<command>".

sevein:
Virtual environments created with `python3 -m venv` populate four activate bin scripts for different shell types. I've uploaded an example from my a3m venv: https://gist.github.com/sevein/0f4f072143f6eb38954867282c56ec1c. 
Perhaps we should have: `hack/activate/bash` and `hack/activate/fish`? Sounds like that'd cover all our use cases.
Let me know and I'll do it.

djjuhasz:
@sevein I don't know if [hack/fish](https://github.com/artefactual-sdps/enduro/blob/main/hack/activate.fish) is necessary because the binary version control in the individual make scripts (e.g. [hack/make/dep_mockgen.mk](https://github.com/artefactual-sdps/enduro/blob/main/hack/make/dep_mockgen.mk)) doesn't work in fish, or if `hack/fish` has some other purpose.  If the fish activation script provides extra functionality, then yes, I'd like to have it for bash/zsh as well. :)

PR #842, opened 2024-01-25
Add timestamps to Preservation tasks

### LINKED ISSUE #832: Problem: lack of timestamps for task lifecycle in Archivematica microservices

We have the list of archivematica services in the enduro ui, but not their actual start/end times. 

Take advantage of the new "detailed" parameter in Archivematica API as described [here](https://github.com/artefactual/archivematica/pull/1879)

None

**Additional context**

Getting duration times in the enduro interface helps getting the system "real" performance .

Diogenesoftoronto:
This PR aims to add timestamps to Preservations Tasks and give additional detail to the user in the enduro dashboard about when archivematica jobs started and finished.

Closes #832

### DIFF
--- a/internal/am/job_tracker.go
+++ b/internal/am/job_tracker.go
@@ -62,7 +62,9 @@ func (jt *JobTracker) SavePreservationTasks(ctx context.Context, unitID string) 
 
 // list requests a job list for unitID from the Archivematica jobs endpoint.
 func (jt *JobTracker) list(ctx context.Context, unitID string) ([]amclient.Job, error) {
-	jobs, httpResp, err := jt.jobSvc.List(ctx, unitID, &amclient.JobsListRequest{})
+	jobs, httpResp, err := jt.jobSvc.List(ctx, unitID, &amclient.JobsListRequest{
+		Detailed: true,
+	})
 	if err != nil {
 		return nil, convertAMClientError(httpResp, err)
 	}
@@ -82,10 +84,6 @@ func (jt *JobTracker) savePreservationTasks(ctx context.Context, jobs []amclient
 
 		pt := ConvertJobToPreservationTask(job)
 		pt.PreservationActionID = jt.presActionID
-
-		now := sql.NullTime{Time: jt.clock.Now(), Valid: true}
-		pt.StartedAt = now
-		pt.CompletedAt = now
 
 		err := jt.pkgSvc.CreatePreservationTask(ctx, &pt)
 		if err != nil {
@@ -115,9 +113,38 @@ func filterSavedJobs(jobs []amclient.Job, saved map[string]struct{}) []amclient.
 // ConvertJobToPreservationTask converts an amclient.Job to a
 // package_.PreservationTask.
 func ConvertJobToPreservationTask(job amclient.Job) package_.PreservationTask {
+	st, co := jobTimeRange(job)
 	return package_.PreservationTask{
-		TaskID: job.ID,
-		Name:   job.Name,
-		Status: jobStatusToPreservationTaskStatus[job.Status],
+		TaskID:      job.ID,
+		Name:        job.Name,
+		Status:      jobStatusToPreservationTaskStatus[job.Status],
+		StartedAt:   st,
+		CompletedAt: co,
 	}
+}
+
+// jobTimeRange calculates the overall start and end times for a job.
+func jobTimeRange(job amclient.Job) (sql.NullTime, sql.NullTime) {
+	if len(job.Tasks) == 0 {
+		return sql.NullTime{}, sql.NullTime{}
+	}
+	st := job.Tasks[0].StartedAt.Time
+	ct := job.Tasks[0].CompletedAt.Time
+
+	for _, t := range job.Tasks[1:] {
+		// Update st to the earliest task start time.
+		if st.After(t.StartedAt.Time) {
+			st = t.StartedAt.Time
+		}
+		// Update ct to the latest task completion time.
+		if ct.Before(t.CompletedAt.Time) {
+			ct = t.CompletedAt.Time
+		}
+	}
+
+	// Emit NULLs if we see the zero value.
+	start := sql.NullTime{Time: st, Valid: !st.IsZero()}
+	end := sql.NullTime{Time: ct, Valid: !ct.IsZero()}
+
+	return start, end
 }
--- a/internal/am/job_tracker_test.go
+++ b/internal/am/job_tracker_test.go
@@ -2,10 +2,11 @@ package am_test
 
 import (
 	"context"
-	"database/sql"
 	"net/http"
 	"testing"
+	"time"
 
+	"entgo.io/ent/dialect/sql"
 	"github.com/google/uuid"
 	"github.com/jonboulle/clockwork"
 	"go.artefactual.dev/amclient"
@@ -16,6 +17,7 @@ import (
 	"gotest.tools/v3/assert"
 
 	"github.com/artefactual-sdps/enduro/internal/am"
+	"github.com/artefactual-sdps/enduro/internal/package_"
 	fake_package "github.com/artefactual-sdps/enduro/internal/package_/fake"
 )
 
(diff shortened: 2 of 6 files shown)


internal/am/job_tracker.go
```diff
@@ -115,9 +117,36 @@ func filterSavedJobs(jobs []amclient.Job, saved map[string]struct{}) []amclient.
...
+		Name:        job.Name,
+		Status:      jobStatusToPreservationTaskStatus[job.Status],
+		StartedAt:   st,
+		CompletedAt: co,
 	}
 }
+
+func getJobTimestamp(job amclient.Job) (sql.NullTime, sql.NullTime) {
+	st := job.Tasks[0].StartedAt.Time
+	ct := job.Tasks[0].CompletedAt.Time
```
sevein, line 132: You're assuming that the slice of tasks isn't empty but that could not always be true, e.g. a bug in the API server. We must check the length of the slice before to avoid a potential panic (index out of range).

internal/am/job_tracker.go
```diff
@@ -115,9 +117,36 @@ func filterSavedJobs(jobs []amclient.Job, saved map[string]struct{}) []amclient.
...
-		Status: jobStatusToPreservationTaskStatus[job.Status],
+		TaskID:      job.ID,
+		Name:        job.Name,
+		Status:      jobStatusToPreservationTaskStatus[job.Status],
+		StartedAt:   st,
+		CompletedAt: co,
 	}
 }
+
+func getJobTimestamp(job amclient.Job) (sql.NullTime, sql.NullTime) {
```
sevein: The name `getJobTimestamp` can be misleading because it implies that the function returns a single timestamp. A better name could be `jobTimeRange`. I've omitted the "get" prefix because of this recommendation: https://go.dev/doc/effective_go#Getters.

internal/am/job_tracker.go
```diff
@@ -115,9 +117,36 @@ func filterSavedJobs(jobs []amclient.Job, saved map[string]struct{}) []amclient.
...
+	st := job.Tasks[0].StartedAt.Time
+	ct := job.Tasks[0].CompletedAt.Time
+
+	for _, t := range job.Tasks {
```
sevein: Since [0] is already assigned, I think here you could use `range job.Tasks[1:]`, right?

sevein: Since [0] is already assigned, I think here you could use `range job.Tasks[1:]`, right?

internal/am/job_tracker.go
```diff
@@ -115,9 +117,36 @@ func filterSavedJobs(jobs []amclient.Job, saved map[string]struct{}) []amclient.
...
+		// if current st is after the task time set that as the startedat time.
```
sevein: Go doc comments use complete sentences, also meaning that they start with capital letter.

```suggestion
		// If current st is after the task time set that as the startedat time.
```

sevein (changes requested):
I've added a few comments. Have you confirmed that the timestamps are being properly recorded? In the `savePreservationTasks` method, I see the following:

```go
		pt := ConvertJobToPreservationTask(job)
		pt.PreservationActionID = jt.presActionID

		now := sql.NullTime{Time: jt.clock.Now(), Valid: true}
		pt.StartedAt = now
		pt.CompletedAt = now
```

With your changes, ConvertJobToPreservationTask is populating the timestamps but they're set to `now` right after. If you tested this PR and observed the timestamps recorded I suppose that'd become apparent.

Diogenesoftoronto:
> I've added a few comments. Have you confirmed that the timestamps are being properly recorded? In the `savePreservationTasks` method, I see the following:
> 
> ```go
> 		pt := ConvertJobToPreservationTask(job)
> 		pt.PreservationActionID = jt.presActionID
> 
> 		now := sql.NullTime{Time: jt.clock.Now(), Valid: true}
> 		pt.StartedAt = now
> 		pt.CompletedAt = now
> ```
> 
> With your changes, ConvertJobToPreservationTask is populating the timestamps but they're set to `now` right after. If you tested this PR and observed the timestamps recorded I suppose that'd become apparent.

Only thing I am worried about for this is that the tasks could very well be completed before this current time such that if any task that would have had a different completedAt time would have `now` as their time. I think this is a problem. Too fix this I could compare to check if the first value in the list is actually after the time completed. If it is then we know the value is both not null and the time that the task actually completed.

Diogenesoftoronto:
> > I've added a few comments. Have you confirmed that the timestamps are being properly recorded? In the `savePreservationTasks` method, I see the following:
> > ```go
> > 		pt := ConvertJobToPreservationTask(job)
> > 		pt.PreservationActionID = jt.presActionID
> > 
> > 		now := sql.NullTime{Time: jt.clock.Now(), Valid: true}
> > 		pt.StartedAt = now
> > 		pt.CompletedAt = now
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With your changes, ConvertJobToPreservationTask is populating the timestamps but they're set to `now` right after. If you tested this PR and observed the timestamps recorded I suppose that'd become apparent.
> 
> Only thing I am worried about for this is that the tasks could very well be completed before this current time such that if any task that would have had a different completedAt time would have `now` as their time. I think this is a problem. Too fix this I could compare to check if the first value in the list is actually after the time completed. If it is then we know the value is both not null and the time that the task actually completed.

I fixed this concern. I also added some of the tests. There is better test coverage and I individually tested both of the functions to make sure they return the correct times.

internal/am/job_tracker.go
```diff
@@ -81,11 +84,16 @@ func (jt *JobTracker) savePreservationTasks(ctx context.Context, jobs []amclient
...
 		pt := ConvertJobToPreservationTask(job)
-		pt.PreservationActionID = jt.presActionID
+		if !pt.StartedAt.Valid {
+			pt.StartedAt.Time = jt.clock.Now()
+			pt.StartedAt.Valid = true
+		}
+		if !pt.CompletedAt.Valid {
+			pt.CompletedAt.Time = jt.clock.Now()
+			pt.CompletedAt.Valid = true
+		}
```
sevein: Should we get rid of this block entirely? The goal of this snippet seems to be avoid NULLs and use the current time instead. I guess my point is that we don't need this anymore since now we can retrieve timestamps from Archivematica.

internal/am/job_tracker.go
```diff
@@ -115,9 +123,44 @@ func filterSavedJobs(jobs []amclient.Job, saved map[string]struct{}) []amclient.
...
+
+	var t time.Time
+	start := sql.NullTime{
+		Time:  st,
+		Valid: st != t,
+	}
+	end := sql.NullTime{
+		Time:  ct,
+		Valid: ct != t,
+	}
```
sevein: Perhaps this is a simpler alternative that should be functionally equivalent:

```go
	start := sql.NullTime{Time: st, Valid: !st.IsZero()}
	end := sql.NullTime{Time: ct, Valid: !ct.IsZero()}
```

internal/am/job_tracker_test.go
```diff
@@ -161,3 +173,91 @@ func TestJobTracker(t *testing.T) {
...
+					CompletedAt: amclient.TaskDateTime{Time: og},
+					Duration:    amclient.TaskDuration(time.Second / 2),
+				},
+			},
+		},
+	}
+	type test struct {
+		name string
+		job  amclient.Job
+		want []sql.NullTime
```
sevein: Because this test is named TestConvertJobToPreservationTask I think it should test its return value (`package_.PreservationTask`) as opposed to part of it (timestamp fields).

sevein (changes requested):
I think this is almost complete but I've included additional comments. I also submitted 9a77effb824afffc1f03b06e36ec440c54a46925 myself to save you the extra work. Please let me know your thoughts.

Diogenesoftoronto:
It looks great @sevein, I didn't realize you were suggesting removing the null check here, but I suppose that makes sense. Thanks for implementing the solutions to the tests as well. They look great. I have rebased and formatted the commit. It is ready to merge.

sevein (approved):
:+1:

PR #847, opened 2024-02-01
Implement ssblob package for downloading AIPs from Archivematica Storage Service

### LINKED ISSUE #820: Problem: AIPs do not download via the Archivematica Storage Service

Diogenesoftoronto:
- Allow AMSS to be used as a Blob
- fix config for ssblob type
- make timeout and base path configurable
- create a storage service config
- return error if multiple config values have been set

I considered not using reflections as the performance on them is not the best, I want to do a better test for this as well but I wonder what you peeps think of this so far. thoughts, ideas, beliefs and intentions???

Closes #820


internal/storage/types/ssblob/ssblob.go
```diff
@@ -0,0 +1,135 @@
+// Package ssblob provides a blob implementation for the Archivematica Storage
+// Service. Use OpenBucket to construct a *blob.Bucket.
+package ssblob
```
sevein: In Go, it is a common practice to favor flat package hierarchies over deeply nested ones for simplicity. A better location would be `internal/ssblob` or if you want something slightly more organized you could try `internal/storage/ssblob` but beyond that I think it's unnecessary.

internal/storage/types/ssblob/ssblob.go
```diff
@@ -0,0 +1,135 @@
...
+	"net/url"
+
+	"gocloud.dev/blob"
+	"gocloud.dev/blob/driver"
+	"gocloud.dev/gcerrors"
+)
+
+type NotImplemented error
+
+var errNotImpl NotImplemented
```
sevein: I think that you should go back to the solution I suggested in my [draft package](https://github.com/sevein/ssblob/blob/9a6d4024edbc25298ad61b1dcc2e7ad79efb14a8/ssblob.go#L18):

```go
var errNotImplemented = errors.New("not implemented")
```

We don't need to define a custom error type or export it, I believe that'd be unnecessary complexity. We're only expected to export gcerrors using the `ErrorCode` method that is part of the interface, like this:

```go
var (
       errNotFound       = errors.New("blob not found")
       errNotImplemented = errors.New("not implemented")
)

func (b *bucket) ErrorCode(err error) gcerrors.ErrorCode {
       switch err {
       case errNotFound:
               return gcerrors.NotFound
       case errNotImplemented:
               return gcerrors.Unimplemented
       default:
               return gcerrors.Unknown
       }
}
```

For example: https://github.com/google/go-cloud/blob/4e3dfbec38511935040c9805491cb8ac1341b799/blob/fileblob/fileblob.go#L340-L347.

internal/storage/types/location_config.go
```diff
@@ -53,14 +57,26 @@ func (c *LocationConfig) UnmarshalJSON(blob []byte) error {
 		return err
 	}
 
+	errMultipleConfig := errors.New("multiple config values have been assigned")
```
sevein: Please address this issue in a separate pull request at a later time.

internal/storage/types/location_config.go
```diff
@@ -143,6 +159,25 @@ func (c URLConfig) OpenBucket(ctx context.Context) (*blob.Bucket, error) {
 	if err != nil {
 		return nil, fmt.Errorf("open bucket by URL: %v", err)
 	}
+	return b, nil
+}
+
+type SSConfig struct {
+	SS string `json:"url"`
+}
```
sevein: `SS` isn't a great name for the URL attribute. Here's an alternative where I also break down the configuration into a few fields:

```go
type SSConfig struct {
	URL string `json:"url"`
	Username string `json:"username"`
	APIKey string `json:"api_key"`
}
```

For example:

```json
{
  "url": "http://127.0.0.1:12345",
  "username": "test",
  "api_key": "test"
}
```

I think in the past we've encoded the credentials in the URL, e.g. `http://test:test@127.0.0.1:12345` but unsure that's a good idea in this case given that we'll likely want to manage secrets externally at some point. That's why I think it's better to break it down like this.

internal/storage/types/location_config.go
```diff
@@ -143,6 +159,25 @@ func (c URLConfig) OpenBucket(ctx context.Context) (*blob.Bucket, error) {
...
 
+func (c SSConfig) Valid() bool {
+	return c.SS != ""
+}
+
+func (c SSConfig) OpenBucket(ctx context.Context) (*blob.Bucket, error) {
+	opts := ssblob.Options{
+		Context: ctx,
```
sevein: We don't need to share this context with ssblob.Options, just ignore it like we did in SFTPConfig. It's only useful when creating a bucket targeting blobstores like like Google Cloud Storage where the init process needs to deal with some authentication concerns that may require extra network connections.

internal/storage/types/location_config.go
```diff
@@ -143,6 +159,25 @@ func (c URLConfig) OpenBucket(ctx context.Context) (*blob.Bucket, error) {
 	if err != nil {
 		return nil, fmt.Errorf("open bucket by URL: %v", err)
 	}
+	return b, nil
+}
+
+type SSConfig struct {
```
sevein, line 156: Please remember to cover this from TestLocationConfig.

internal/storage/types/ssblob/ssblob.go
```diff
@@ -0,0 +1,135 @@
...
+
+type bucket struct {
+	Options
+}
+
+type Options struct {
+	Context context.Context
```
sevein: I mentioned it before but to make sure this is addressed: we dont' need a context in the struct.

internal/storage/types/ssblob/ssblob.go
```diff
@@ -0,0 +1,135 @@
...
+func (b *bucket) Attributes(ctx context.Context, key string) (*driver.Attributes, error) {
+	return nil, errNotImpl
+}
+
+func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driver.ListPage, error) {
+	return nil, errNotImpl
+}
+
+func (b *bucket) NewRangeReader(ctx context.Context, key string, offset, length int64, opts *driver.ReaderOptions) (driver.Reader, error) {
+	client := http.Client{}
```
sevein: I've suggested earlier to make the HTTP client a member of the bucket struct so here we would just be using `b.client` instead of having to create a new client each time the user wants to read. http.Client is thread-safe and best shared in a struct.

internal/storage/types/ssblob/ssblob.go
```diff
@@ -0,0 +1,135 @@
...
+	bu, err := url.Parse(b.URL)
+	if err != nil {
+		return nil, err
+	}
+	if b.Base == "" {
+		b.Base = "."
+	}
+	path, err := url.JoinPath(b.Base, key)
```
sevein: I didn't know about url.JoinPath, cool.

internal/storage/types/ssblob/ssblob.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
+	return "", errNotImpl
+}
+
+func (b *bucket) Close() error {
+	return nil
+}
+
+// Reader should be able to read an AIP object from Storage Service using *http.Client.
```
sevein: In the original package I wrote, my comment started with `reader` as opposed to `Reader` to adhere to the convention: start with the name of the item being described. We're describing `reader`, not `Reader`. In any case, perhaps this is a better description given that this is not really using a http.Client:

```suggestion
// reader reads an Archivematica Storage Service AIP. It implements driver.Reader.
```

internal/storage/types/ssblob/ssblob_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"io"
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	"github.com/artefactual-sdps/enduro/internal/storage/types/ssblob"
+)
+
+// fix tests to work for archivematica storage service. Add proper mocks etc for this I am not sure how exactly to go about it.
```
sevein: You can create a [httptest.Server](https://pkg.go.dev/net/http/httptest#NewServer) where `yourHandlerFunction` returns some data and validates the client request. E.g.:

```go
srv := httptest.NewServer(http.HandlerFunc(yourHandlerFunction))
```
    
Use `srv.URL` to configure the new bucket.

internal/storage/types/ssblob/ssblob.go
```diff
@@ -0,0 +1,135 @@
...
+
+var errNotImpl NotImplemented
+
+type bucket struct {
+	Options
+}
+
+type Options struct {
+	Context context.Context
+	URL     string `json:"string"`
```
sevein: I believe we won't need JSON tags here.

internal/storage/types/ssblob/ssblob.go
```diff
@@ -0,0 +1,135 @@
...
+	Base    string
+}
+
+func openBucket(opts *Options) (driver.Bucket, error) {
+	return &bucket{
+		*opts,
+	}, nil
+}
```
sevein: This is where we would be creating the `*http.Client` and store it in the `*bucket`.

Use preferably [cleanhttp.DefaultPooledClient](https://pkg.go.dev/github.com/hashicorp/go-cleanhttp?utm_source=godoc#DefaultPooledClient). We may have to adjust some settings down the road to allow for reliable downloads, not sure.

sevein (commented):
Looking good, I've added some comments!

internal/storage/ssblob/ssblob.go
```diff
@@ -0,0 +1,155 @@
...
+	URL      string
+	Key      string
+	Username string
+	Client   *http.Client
+}
+
+func openBucket(opts *Options) (driver.Bucket, error) {
+	// Will use the http client we pass with options if it is given.
+	var cl *http.Client
+	if reflect.ValueOf(opts.Client).IsZero() {
```
sevein: This is enough: opts.Client == nil

internal/storage/ssblob/ssblob_test.go
```diff
@@ -0,0 +1,46 @@
...
+		w.Write([]byte("hello"))
+	})
+	srv := httptest.NewServer(
+		handler,
+	)
+	cl := srv.Client()
+
+	opts := ssblob.Options{
+		URL:    srv.URL,
+		Client: cl,
```
sevein: Why do we need to pass the client? It should be enough with the URL afaik. It's ok to accept a client as an option but we should have a real use case, not just for testing purposes.

internal/storage/ssblob/ssblob_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/artefactual-sdps/enduro/internal/storage/ssblob"
+)
+
+func TestBucket(t *testing.T) {
+	t.Parallel()
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "image/jpeg")
+		w.Write([]byte("hello"))
+	})
```
sevein: This is where we can pretend we're the Storage Service. Here you can look up the request (r) and confirm the headers (e.g. Authorization header with the API key), URL or HTTP method. You can assert with gotest.tools/v3/assert here too.

internal/storage/ssblob/ssblob.go
Diogenesoftoronto: Forgot that I made it a pointer keep doing this haha.

sevein (approved):
Nice job, Keith. I've added an extra commit with some additional changes:
1. I decided not to use tables in `ssblob_test.go` since we're dealing with multiple scenarios that don't share much of a common structure, e.g. we can have a bucket open operation fail, or a bucket new reader fail, etc. I thought it'd look much simpler using simple subtests.
2. I've introduced a custom transport for the http.Client (ssblob/client.go) to inject HTTP headers consistently. Notice that the authentication header is formed as `Authorization: ApiKey username:key`. This is a bit different compared to how you were doing it. Check out the [docs](https://www.archivematica.org/en/docs/archivematica-1.15/dev-manual/api/api-reference-storage-service/).
3. I've updated the request path according to [API spec](https://www.archivematica.org/en/docs/archivematica-1.15/dev-manual/api/api-reference-storage-service/#download-package), e.g. `/api/v2/file/2db707f3-3cd2-44b7-9012-9b68eb10d207/download`,
4. I've made a few extra tweaks to the ssblob driver: implemented `As` and `ErrorAs` methods and started using a new APIError type to encode HTTP statuses so we can come up with better errors in the ErrorCode method,
5. In the new location config, I've made the username a required field.

I'm going to merge. It's likely not ready but I think that's ok. I guess we'll have to figure out how and when we're setting up the new location. Not sure if that's a concern that we want to address at deploy time for now just like we're doing in kube/base/mysql-create-locations-job.yaml for the s3 locations.

PR #848, opened 2024-02-01
Add activate for bash/zsh

sevein:
Please give it a test! It's copied again from Python's venv with a few simplifications.

Usage: 

```
$ source hack/activate/bash
$ which gotestsum
/home/jesus/.cache/enduro-sdps/Linux/x86_64/bin/gotestsum
```


djjuhasz (approved):
Yep, it works. :)

PR #849, opened 2024-02-01
Add make tools

sevein:
Add `make tools` to install all the dev tools that we manage with make.


Diogenesoftoronto:
This is awesome, I should take some of these scripts to devbox.

PR #854, opened 2024-02-02
Return error for multiple config options set

### LINKED ISSUE #853: Problem: Current storage configuration allows for mutually exclusive options to be chosen

Setting multiple configuration options for the storage service results in no errors. Yet these are mutually exclusive options. If the user sets up two or more only one of them will work whatever is the first in the switch statement.

Return an error or crash the program if multiple storage configuration options have been setup. This forces the user to set the expected configuration.

Set up a priority of configuration options. For example if the options for S3 have been setup but also URLs then URLs will be used instead. 

**Additional context**

Diogenesoftoronto:
An error is now returned when a more than one storage configuration has
been set. Previously, this would not return an error and simply use what
-ever the user had set that was first in the storage switch options.

Closes #853

### DIFF
--- a/internal/storage/types/location_config.go
+++ b/internal/storage/types/location_config.go
@@ -47,14 +47,22 @@ func (c LocationConfig) MarshalJSON() ([]byte, error) {
 }
 
 func (c *LocationConfig) UnmarshalJSON(blob []byte) error {
-	types := configTypes{}
+	// Raise an error if the doc describes multiple configs.
+	keys := map[string]json.RawMessage{}
+	err := json.Unmarshal(blob, &keys)
+	if err != nil {
+		return err
+	}
+	if len(keys) > 1 {
+		return errors.New("multiple config values have been assigned")
+	}
 
+	types := configTypes{}
 	if err := json.Unmarshal(blob, &types); err != nil {
 		return err
 	}
 
 	switch {
-	// TODO: return error if we have more than one config assigned (mutually exclusive)
 	case types.S3 != nil:
 		c.Value = types.S3
 	case types.SFTPConfig != nil:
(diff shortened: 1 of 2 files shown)


internal/storage/types/location_config.go
```diff
@@ -53,14 +54,21 @@ func (c *LocationConfig) UnmarshalJSON(blob []byte) error {
 		return err
 	}
 
+	errMultipleConfig := errors.New("multiple config values have been assigned")
 	switch {
-	// TODO: return error if we have more than one config assigned (mutually exclusive)
 	case types.S3 != nil:
 		c.Value = types.S3
 	case types.SFTPConfig != nil:
+		if reflect.ValueOf(c.Value).IsValid() {
```
sevein: I think that we can still use a simple nil check here, right? If you disagree please show me what's the value added because I can't see it. This could be a starting point for us to understand it: https://go.dev/play/p/y2XQtTZ1jnQ.

sevein (changes requested):
You forgot to add a test, I wrote one and it confirmed that your solution isn't working. See f9ba7a3a where I've added a different kind of check early in the function.

Diogenesoftoronto:
I expected this to work because I expected Cstyle switch statements. I this works if I write the fall through keywords then we don't need to unMarshall a struct and allocate any extra memory here.

sevein:
>  I this works if I write the fall through keywords then we don't need to unMarshall a struct and allocate any extra memory here.

I'd love to see what that looks like. Feel free to push a new commit with changes to UnmarshalJSON, ensuring the tests still pass.

sevein:
Going to merge, feel free to submit an additional PR showing your alternative to UnmarshalJSON if you feel like it. I personally feel like this is easy to read and the overhead is definitely not substantial.

PR #855, opened 2024-02-06
Rewrite location config tests using tables

sevein:
This pull request rewrites the location config tests using tables and subtests. It does not introduce functional changes.

It depends on https://github.com/artefactual-sdps/enduro/pull/854.


Diogenesoftoronto:
Looks good I was considering doing this once I saw what location configuration test was like in #842.

PR #857, opened 2024-02-06
Add package status tests

sevein:
Testing codecov comments.


sevein:
Not sure yet why codecov isn't reporting coverage as comments. Their PR report seems to reflect what the code is doing, see https://github.com/artefactual-sdps/enduro/pull/857 (1.3% increase), but seems to be stuck in "Processing" status.

image

Issue #858, opened 2024-02-06
Problem: inconsistent bucket configuration across components


sevein:
**Describe the bug**

Enduro bootstraps instances of `*blob.Bucket` in several places, e.g. `watcher`, `upload` or `location`. `*blob.Bucket` is an abstraction provided by gocloud.dev that we've adopted to interact with blobstore.

These are some issues observed with the current approach:
* Lack of a shared configuration struct embeddable in different contexts (appconfig, databases, etc),
* Lack of support for URL-based buckets (useful during testing, e.g. `mem://` or simpler config),
* Use of AWS SDKv1 ([end of support announced](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/)), should use v2 instead!

**Additional context**

I'm planning to suggest the introduction of a new `bucket` package providing a shared configuration struct and a simple function to open a `*blob.Bucket` given such configuration. Future iterations could focus on deeper integration with the new facilities, e.g. `watcher.MinioConfig` to make use of the new config struct to define the bucket configuration.

jraddaoui:
I think you solved this issue @sevein but, could this package be moved to the gotools repository? I think it will be useful for other projects.

sevein:
> Could this package be moved to the gotools repository? I think it will be useful for other projects.

Done! https://pkg.go.dev/go.artefactual.dev/tools@v0.9.0/bucket

PR #859, opened 2024-02-06
Add bucket package

sevein:
This commit introduces a new package `bucket` aiming to consolidate the
different ways we configure and create buckets across the application.

It enables future work such as embedding the shared config in different
contexts as well as using memory-based or fs-based buckets for testing purposes
when setting up some components of the application.

Relates to https://github.com/artefactual-sdps/enduro/issues/858.

### DIFF
--- a/internal/bucket/bucket.go
+++ b/internal/bucket/bucket.go
@@ -38,14 +38,29 @@ func Open(ctx context.Context, c *Config) (*blob.Bucket, error) {
 		return nil, errors.New("config is undefined")
 	}
 
+	var (
+		b   *blob.Bucket
+		err error
+	)
+
 	if c.URL != "" {
-		b, err := blob.OpenBucket(ctx, c.URL)
-		if err != nil {
-			return nil, fmt.Errorf("open bucket from URL %q: %v", c.URL, err)
-		}
-		return b, nil
+		b, err = openWithURL(ctx, c.URL)
+	} else {
+		b, err = openWithConfig(ctx, c)
 	}
 
+	return b, err
+}
+
+func openWithURL(ctx context.Context, url string) (*blob.Bucket, error) {
+	b, err := blob.OpenBucket(ctx, url)
+	if err != nil {
+		return nil, fmt.Errorf("open bucket from URL %q: %v", url, err)
+	}
+	return b, nil
+}
+
+func openWithConfig(ctx context.Context, c *Config) (*blob.Bucket, error) {
 	addr := c.Endpoint
 	if u, err := url.Parse(c.Endpoint); err == nil {
 		if !strings.HasPrefix(u.Scheme, "http") {
@@ -65,10 +80,7 @@ func Open(ctx context.Context, c *Config) (*blob.Bucket, error) {
 		config.WithEndpointResolverWithOptions(
 			aws.EndpointResolverWithOptionsFunc(
 				func(service, region string, options ...interface{}) (aws.Endpoint, error) {
-					if service == s3.ServiceID {
-						return aws.Endpoint{URL: addr}, nil
-					}
-					return aws.Endpoint{}, &aws.EndpointNotFoundError{}
+					return aws.Endpoint{URL: addr}, nil
 				},
 			),
 		),
--- a/internal/bucket/bucket.go
+++ b/internal/bucket/bucket.go
@@ -1,0 +1,89 @@
+// Package bucket provides a mechanism to open buckets using configuration.
+package bucket
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"gocloud.dev/blob"
+	"gocloud.dev/blob/s3blob"
+)
+
+type Config struct {
+	// Bucket URL as described in https://gocloud.dev/howto/blob/.
+	URL string
+
+	// Alternatively, a less flexible way to access S3-compatible buckets.
+	Endpoint  string
+	Bucket    string
+	AccessKey string
+	SecretKey string
+	Token     string
+	Profile   string
+	Region    string
+	PathStyle bool
+}
+
+// Open a bucket provided its configuration. Unless specified in the URL field,
+// it will open buckets with s3blob.OpenBucketV2.
+func Open(ctx context.Context, c *Config) (*blob.Bucket, error) {
+	if c == nil {
+		return nil, errors.New("config is undefined")
+	}
+
+	if c.URL != "" {
+		b, err := blob.OpenBucket(ctx, c.URL)
+		if err != nil {
+			return nil, fmt.Errorf("open bucket from URL %q: %v", c.URL, err)
+		}
+		return b, nil
+	}
+
+	addr := c.Endpoint
+	if u, err := url.Parse(c.Endpoint); err == nil {
+		if !strings.HasPrefix(u.Scheme, "http") {
+			addr = "http://" + addr
+		}
+	}
+
+	awscfg, err := config.LoadDefaultConfig(
+		ctx,
+		config.WithSharedConfigProfile(c.Profile),
+		config.WithRegion(c.Region),
+		config.WithCredentialsProvider(
+			credentials.NewStaticCredentialsProvider(
+				c.AccessKey, c.SecretKey, c.Token,
+			),
+		),
+		config.WithEndpointResolverWithOptions(
+			aws.EndpointResolverWithOptionsFunc(
+				func(service, region string, options ...interface{}) (aws.Endpoint, error) {
+					if service == s3.ServiceID {
+						return aws.Endpoint{URL: addr}, nil
+					}
+					return aws.Endpoint{}, &aws.EndpointNotFoundError{}
+				},
+			),
+		),
+	)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


internal/bucket/bucket_test.go
```diff
@@ -0,0 +1,90 @@
...
+				assert.Equal(t, opts.UsePathStyle, false)
+
+				_, err := client.ListBuckets(context.Background(), &s3v2.ListBucketsInput{})
+				assert.ErrorContains(t, err, "http://foobar:12345/?x-id=ListBuckets")
+			},
+		},
+	}
+
+	for name, tc := range tests {
+		tc := tc
```
Diogenesoftoronto: 1.22 makes this unnecessary.

internal/bucket/bucket_test.go
```diff
@@ -0,0 +1,90 @@
...
+		},
+		"Appends http if scheme is undefined": {
+			config: &bucket.Config{
+				Endpoint:  "foobar:12345",
+				Bucket:    "name",
+				Region:    "region",
+				AccessKey: "access",
+				SecretKey: "secret",
+			},
+			require: func(b *blob.Bucket) {
```
Diogenesoftoronto, line 59: If these functions remain the same, why not just have a default here. Extract the function as a variable.

Diogenesoftoronto (changes requested):
Hey looks good so far, but are we covering this in location config test?

internal/bucket/bucket.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if c.URL != "" {
+		b, err := blob.OpenBucket(ctx, c.URL)
+		if err != nil {
+			return nil, fmt.Errorf("open bucket from URL %q: %v", c.URL, err)
+		}
+		return b, nil
+	}
+
+	addr := c.Endpoint
```
djjuhasz, line 64: I think the code would be easier to read and understand if the `s3blob.OpenBucket()` portion was refactored into a separate, private method.

internal/upload/service.go
```diff
@@ -48,49 +47,37 @@ func NewService(logger logr.Logger, config Config, uploadMaxSize int, tokenVerif
 		tokenVerifier: tokenVerifier,
 	}
 
-	err = s.openBucket(config)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+	defer cancel()
+
+	err = s.openBucket(ctx, config)
```
djjuhasz, line 53: I think it's a mistake to open the bucket at this point - it's possible it's never accessed, and it hides the need to call `bucket.Close()` when the resource is no longer needed.  I think it's better to open the bucket in the service methods that need to access the contents.

djjuhasz: In general I think we are sloppy about closing buckets in Enduro.

djjuhasz (commented):
Overall, I think this a much needed change, and it's looking good. I do have some concerns about the lifespan of open buckets in Enduro.

internal/upload/service.go
sevein: I believe that it's meant to be used like *http.Client or *sql.DB, i.e. they're meant to be shared, can be used by multiple goroutines, etc. The service Close method closes the bucket. But I will look it up again. I think we were doing this in sdps_preprocessing if I remember correctly.

djjuhasz: The example at https://gocloud.dev/howto/blob/#opening includes an immediate `defer bucket.Close()` which makes me think there is a cost to keeping buckets open.

sevein:
Thank you both, will address your feedback soon.

internal/upload/service.go
Diogenesoftoronto: should we add a lint check for bucket closing and other structs with a close method?

Diogenesoftoronto:
I added some of the code changes as suggestions and trying it out I noticed that defer on the close throws a lint error which should be ignored. I do not think we should be checking every close error here especially not in a test.

internal/upload/service.go
sevein: I've looked this up more closely and this is what I've found:
* A quick search on GitHub ([link](https://github.com/search?q=*blob.Bucket+language%3AGo&type=code)) reveals that it's a common practice to have a shared bucket as a struct field. Some examples: [goreleaser](https://github.com/goreleaser/goreleaser/blob/c9cc1065b473aa5ae96c4a0074b360a9f37e6f2b/internal/pipe/blob/upload.go#L233), [open-saves](https://github.com/googleforgames/open-saves/blob/536314ac9be11e64a3e642e213e9c8fa354b6d86/internal/pkg/blob/gcs.go#L31), [go-iiif](https://github.com/go-iiif/go-iiif/blob/6aba66ab5764973c0b1ef563d3c6ec8ec54d904a/source/memory.go#L13), [vroom](https://github.com/getsentry/vroom/blob/6ac471577674e6628e3aa2575e0f9be7f6a7a5c4/cmd/vroom/main.go#L41)... 
* `Close()` seems to be a no-op for most driver implementations, e.g.: [memblob](https://github.com/google/go-cloud/blob/5ec203b7dc558ebb940d0847f8dae12699f1d873/blob/memblob/memblob.go#L102-L104), [s3blob](https://github.com/google/go-cloud/blob/5ec203b7dc558ebb940d0847f8dae12699f1d873/blob/s3blob/s3blob.go#L370-L372), [gcsblob](https://github.com/google/go-cloud/blob/5ec203b7dc558ebb940d0847f8dae12699f1d873/blob/gcsblob/gcsblob.go#L400-L402) or [fileblob](https://github.com/google/go-cloud/blob/5ec203b7dc558ebb940d0847f8dae12699f1d873/blob/fileblob/fileblob.go#L293-L295). [*blob.Bucket](https://github.com/google/go-cloud/blob/5ec203b7dc558ebb940d0847f8dae12699f1d873/blob/blob.go#L1276-L1286) (which performs as the wrapper in the driver pattern) is only wrapping the call to the underlying driver and seems like it's only preventing concurrent access (with a shared lock), but it does not seem to be performing other clean-up operations.
* Underlying drivers, e.g. [s3blob](https://github.com/google/go-cloud/blob/5ec203b7dc558ebb940d0847f8dae12699f1d873/blob/s3blob/s3blob.go#L362-L368) or [gcsblob](https://github.com/google/go-cloud/blob/5ec203b7dc558ebb940d0847f8dae12699f1d873/blob/gcsblob/gcsblob.go#L344), use their bucket struct to store the client objects (provided by the SDK) which seem to be expensive to create, have thread-safety capabilities and generally seem to be designed to be long-lived. Opening a s3 or gcs bucket seems to be particularly expensive because the underlying SDKs need to deal with concerns like authentication or metadata, e.g. opening a gcs bucket results in network operations, i.e. see [Creating a client](https://pkg.go.dev/cloud.google.com/go/storage#hdr-Creating_a_Client) which reads: clients should be reused instead of created as needed. 
* I wrote a quick [benchmark](https://gist.github.com/sevein/856fc782803eebeb222a8be76193388d). If I'm intepreting it correctly I think it's showing that `BenchmarkWithOpen` allocates approximately 752 more bytes and performs 15 more allocations than `BenchmarkWithoutOpen`, and took 1038 ns more to run (lol), but this exacerbated considerably when using s3blob.

My understanding is that the cost of keeping a bucket open is negligible compared to the cost of opening the bucket each time it's needed. The bootstrap process appears to be more resource-intensive, so it's better to do it just once. Some memory is used for caching purposes, but this will save computation time later on.

internal/bucket/bucket_test.go
sevein: I made some changes and ended up not having the same require twice, lmk if you think it's still needed somehow.

internal/upload/service.go
sevein: > should we add a lint check for bucket closing and other structs with a close method?

There is bodyclose or sqlclosecheck in golangci-lint, but I don't think there is a general-purpose linter probably because it's an application-specific concern, not a primitive in the language, if that makes sense.

internal/bucket/bucket_test.go
```diff
@@ -69,14 +69,16 @@ func TestOpen(t *testing.T) {
...
 
 	for name, tc := range tests {
-		tc := tc
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 
 			b, err := bucket.Open(context.Background(), tc.config)
+			// We should definitely close buckets when we are testing!
+			// This throws a lint error which should be ignored!
+			defer b.Close()
```
sevein: Thank you. I noticed that this panic when b is nil, which is possible in another test I wrote where Open is supposed to return an error instead. I've ended up doing this:

```go
if b != nil {
	b.Close()
}
```

gocloud seems to be doing the same, see [here](https://github.com/google/go-cloud/blob/5ec203b7dc558ebb940d0847f8dae12699f1d873/blob/s3blob/s3blob_test.go#L477-L480).

internal/bucket/bucket_test.go
```diff
@@ -69,14 +69,16 @@ func TestOpen(t *testing.T) {
...
 
 			if tc.errMsg != "" {
-				assert.Assert(t, b == nil)
+				assert.Equal(t, b, nil)
```
sevein: Thank you. However I think this is not an improvement compared to the original version because `nil` is untyped, i.e. not the same as `(*blob.Bucket)(nil)`. More details here: https://go101.org/article/nil.html. You could see errors like the following:

    assertion failed: <nil> (b *blob.Bucket) != <nil> (nil <nil>)

You'd need to write:

    assert.Equal(t, b, (*blob.Bucket)(nil))
    
At that point I much prefer the original:

    assert.Assert(t, b == nil)

djjuhasz (approved):
LGTM :)

sevein:
Thanks, I forgot I had submitted this PR! I'll merge.

PR #861, opened 2024-02-08
Add preprocessing child workflow

jraddaoui:
Allows to configure and trigger a custom preprocessing child workflow
to be handled by a different worker. This initial implementation requires
both workers to have access to the same filesystem to share the package.

Refs #886.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -63,8 +63,35 @@ KUBE_OVERLAY = 'hack/kube/overlays/dev-a3m'
 if PRES_SYS == 'am':
   KUBE_OVERLAY = 'hack/kube/overlays/dev-am'
 
+# Load Kustomize YAML
+yaml = kustomize(KUBE_OVERLAY)
+
+# Preprocessing
+PREPROCESSING_PATH = os.environ.get("PREPROCESSING_PATH", "")
+if PREPROCESSING_PATH != "":
+  # Load preprocessing Tiltfile for Enduro
+  load_dynamic(PREPROCESSING_PATH + "/Tiltfile.enduro")
+  # Get Enduro a3m/am worker k8s manifest
+  if PRES_SYS == "a3m":
+    pres_yaml, yaml = filter_yaml(yaml, name="^enduro-a3m$", kind="StatefulSet")
+  else:
+    pres_yaml, yaml = filter_yaml(yaml, name="^enduro-am$", kind="Deployment")
+  # Append preprocessing volume and volume mount to worker container,
+  # this will only work in single node k8s cluster deployments
+  volume = {"name": "shared-dir", "persistentVolumeClaim": {"claimName": "preprocessing-pvc"}}
+  volume_mount = {"name": "shared-dir", "mountPath": "/home/enduro/preprocessing"}
+  pres_obj = decode_yaml(pres_yaml)
+  if "volumes" not in pres_obj["spec"]["template"]["spec"]:
+    pres_obj["spec"]["template"]["spec"]["volumes"] = []
+  pres_obj["spec"]["template"]["spec"]["volumes"].append(volume)
+  for container in pres_obj["spec"]["template"]["spec"]["containers"]:
+    if container["name"] in ["enduro-a3m-worker", "enduro-am-worker"]:
+      container["volumeMounts"].append(volume_mount)
+  pres_yaml = encode_yaml(pres_obj)
+  yaml = [yaml, pres_yaml]
+
 # Load Kubernetes resources
-k8s_yaml(kustomize(KUBE_OVERLAY))
+k8s_yaml(yaml)
 
 # Configure trigger mode
 trigger_mode = TRIGGER_MODE_MANUAL
--- a/docs/src/dev-manual/devel.md
+++ b/docs/src/dev-manual/devel.md
@@ -192,6 +192,14 @@ are planning to use Archivematica as preservation system.
 Build and use a local version of a3m. Requires to have the `a3m` repository
 cloned as a sibling of this repository folder.
 
+### PREPROCESSING_PATH
+
+Relative path to a preprocessing child workflow repository. It loads a Tiltfile
+called `Tiltfile.enduro` from that repository and mounts a presistent volume
+claim (PVC) in the preservation system pod. That PVC must be defined in the
+preprocessing and be called `preprocessing-pvc`. Check the [Preprocessing child
+workflow] docs to configure the child workflow execution.
+
 ## Tilt UI helpers
 
 ### Upload to Minio
@@ -259,3 +267,4 @@ is sometimes not setup properly. To solve it, from the Tilt UI, restart the
 [visual studio code]: https://code.visualstudio.com/
 [working with archivematica]: archivematica.md
 [devbox]: https://www.jetpack.io/devbox/docs/quickstart/#install-devbox
+[preprocessing child workflow]: preprocessing.md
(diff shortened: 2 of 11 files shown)


Tiltfile
```diff
@@ -63,8 +63,35 @@ KUBE_OVERLAY = 'hack/kube/overlays/dev-a3m'
...
+  load_dynamic(PREPROCESSING_PATH + "/Tiltfile.enduro")
+  # Get Enduro a3m/am worker k8s manifest
+  if PRES_SYS == "a3m":
+    pres_yaml, yaml = filter_yaml(yaml, name="^enduro-a3m$", kind="StatefulSet")
+  else:
+    pres_yaml, yaml = filter_yaml(yaml, name="^enduro-am$", kind="Deployment")
+  # Append preprocessing volume and volume mount to worker container,
+  # this will only work in single node k8s cluster deployments
+  volume = {"name": "shared-dir", "persistentVolumeClaim": {"claimName": "preprocessing-pvc"}}
+  volume_mount = {"name": "shared-dir", "mountPath": "/home/enduro/preprocessing"}
```
jraddaoui, line 82: I was planning to make the `claimName` and `mountPath` configurable through the environment, but I don't like this solution in general. I'll try to find something better while working on the remote option, maybe just working with the k8s manifests, ideas are welcome.

docs/src/dev-manual/devel.md
```diff
@@ -192,6 +192,14 @@ are planning to use Archivematica as preservation system.
...
 cloned as a sibling of this repository folder.
 
+### PREPROCESSING_PATH
+
+Relative path to a preprocessing child workflow repository. It loads a Tiltfile
+called `Tiltfile.enduro` from that repository and mounts a presistent volume
+claim (PVC) in the preservation system pod. That PVC must be defined in the
+preprocessing and be called `preprocessing-pvc`. Check the [Preprocessing child
+workflow] docs to configure the child workflow execution.
+
```
jraddaoui, line 202: This will be better documented (if it's still needed) in a `preprocessing-base` repository.

internal/workflow/activities/download.go
```diff
@@ -22,8 +22,9 @@ type DownloadActivity struct {
 }
 
 type DownloadActivityParams struct {
-	Key         string
-	WatcherName string
+	Key             string
+	WatcherName     string
+	DestinationPath string
```
jraddaoui, line 27: Check https://github.com/artefactual-sdps/enduro/issues/926 for more info.

jraddaoui (commented):
@djjuhasz @sevein this is ready for a first review.

I was planning to include a remote option in this PR, but I think it's better to do that incrementally in another one, after we talk a little about https://github.com/artefactual-sdps/enduro/issues/927.

docs/src/dev-manual/preprocessing.md
```diff
@@ -0,0 +1,34 @@
...
+child workflow.
+
+## Configuration
+
+### `.tilt.env`
+
+Check the [Tilt environment configuration].
+
+### `enduro.toml`
+
```
djjuhasz, line 13: I think it would be better to reference (or link to) the enduro.toml preprocessing section, rather than duplicating all of the options and comments here and in enduro.toml.  Mostly my concern is the maintenance of syncing the toml here and in  enduro.toml.

internal/config/config.go
```diff
@@ -30,17 +31,18 @@ type Configuration struct {
...
+	API           api.Config
+	Database      db.Config
+	Event         event.Config
+	Preservation  pres.Config
+	Storage       storage.Config
+	Temporal      temporal.Config
+	Upload        upload.Config
+	Watcher       watcher.Config
+	Telemetry     telemetry.Config
+	Preprocessing preprocessing.Config
```
djjuhasz: I'd like to keep the config sections alphabetized.

internal/config/config.go
```diff
@@ -57,6 +59,12 @@ func (c Configuration) Validate() error {
 			return err
 		}
 	}
+	if config, ok := interface{}(c.Preprocessing).(ConfigurationValidator); ok {
```
djjuhasz: I don't understand why this is necessary? Won't `c.Preprocessing` always implement the `ConfigurationValidator` interface?

internal/workflow/processing.go
```diff
@@ -799,3 +811,36 @@ func (w *ProcessingWorkflow) transferAM(sessCtx temporalsdk_workflow.Context, ti
...
 	return nil
 }
+
+func (w *ProcessingWorkflow) preprocessing(ctx temporalsdk_workflow.Context, tinfo *TransferInfo) error {
+	if !w.cfg.Preprocessing.Enabled {
+		return nil
+	}
+
+	// TODO: move package if tinfo.TempPath is not inside w.prepConfig.SharedPath.
+	realPath, err := filepath.Rel(w.cfg.Preprocessing.SharedPath, tinfo.TempPath)
```
djjuhasz: I think `realPath` should be `relPath`?

internal/workflow/processing_test.go
```diff
@@ -68,8 +78,9 @@ func (s *ProcessingWorkflowTestSuite) SetupWorkflowTest(taskQueue string) {
 	ctrl := gomock.NewController(s.T())
 	logger := logr.Discard()
 	cfg := config.Configuration{
-		Preservation: pres.Config{TaskQueue: taskQueue},
-		A3m:          a3m.Config{ShareDir: s.transferDir},
+		Preservation:  pres.Config{TaskQueue: taskQueue},
+		A3m:           a3m.Config{ShareDir: s.transferDir},
+		Preprocessing: prepConfig,
```
djjuhasz: Nitpick: this is really minor, but "prep" is usually a short form for "prepare" so I read "prepConfig" as "prepare config".  Maybe "ppConfig" or "preprocessingConfig" instead?

djjuhasz (changes requested):
Looks really good to me @jraddaoui.  I had a few notes, but it's all pretty minor stuff.

docs/src/dev-manual/preprocessing.md
jraddaoui: I had my conflicts about this too, linking to that section will also require maintenance if the lines change, which could happen more often than changing the actual preprocessing config. Also, it'll be easier to forget when you are changing other sections in that file.

Maybe we just link to the entire file and indicate the section without setting the lines, but I don't know how is that link going to work in the built website.

internal/config/config.go
jraddaoui: Ups, didn't notice they were sorted.

internal/config/config.go
jraddaoui: Yeah, me neither, maybe this was part of a loop or something in another implementation and the check was needed because some configs didn't implement `ConfigurationValidator`.

docs/src/dev-manual/preprocessing.md
jraddaoui: Not sure what to do with this, I'd leave it as it is for now. Other thoughts @djjuhasz?

internal/config/config.go
jraddaoui: I changed this so it also returns all errors and not only the first one found.

docs/src/dev-manual/preprocessing.md
djjuhasz: I'm fine with leaving it as-is. :+1:

djjuhasz (approved):
:+1:

jraddaoui:
Thanks @djjuhasz!

Issue #864, opened 2024-02-21
Problem: Changing the configured a3m shared directory breaks the a3m workflow


djjuhasz:
**Describe the bug**

Changing the value of the a3m "shareDir" configuration setting in the "enduro.toml" configuration file causes the a3m processing workflow to fail with a fatal error.

**To Reproduce**

1. Configure Enduro to use a3m for preservation
2. In the "enduro.toml" configuration file, change the default value of the "a3m.shareDir" setting to a different value
3. Restart the enduro and a3m workers to load the new configuration
4. Attempt to process a transfer by uploading it to the MinIO "sips" bucket
5. The processing workflow fails with the error:

```
 "message": "error creating temporary directory: stat /home/a3m/.local/share/a3m/share: no such file or directory",
  "source": "GoSDK",
```

**Expected behavior**

Changing the configured a3m shared directory location to a valid filesystem path should not break the a3m workflow.

Screenshot:
![image](https://github.com/artefactual-sdps/enduro/assets/626135/2f1f9424-151e-4d7b-b5d1-aa355ae62c56)

Failure message (JSON):
```
{
  "message": "activity error",
  "source": "GoSDK",
  "stackTrace": "",
  "encodedAttributes": null,
  "cause": {
    "message": "error creating temporary directory: stat /home/a3m/.local/share/a3m/share: no such file or directory",
    "source": "GoSDK",
    "stackTrace": "",
    "encodedAttributes": null,
    "cause": null,
    "applicationFailureInfo": {
      "type": "",
      "nonRetryable": true,
      "details": {
        "payloads": [
          null
        ]
      }
    }
  },
  "activityFailureInfo": {
    "scheduledEventId": "20",
    "startedEventId": "21",
    "identity": "1@enduro-a3m-0@",
    "activityType": {
      "name": "bundle-activity"
    },
    "activityId": "20",
    "retryState": "NonRetryableFailure"
  }
}
```

**Additional context**

The a3m share directory path is hardcoded at https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/processing.go#L321

sevein:
Can this issue be closed?

djjuhasz:
@sevein that's a tough question.  Changing the shareDir will still break the a3m integration in dev because a3m is still trying to send the AIP back to `/home/a3m/.local/share/a3m/share`.  I couldn't find any way to configure a3m (in the dev env) to send the AIP to a different directory. :(

[...]

PR #865, opened 2024-02-21
Use a3m shareDir config value in processing

djjuhasz:
Refs #864

- Pass the full enduro config to processing workflow, and remove the separate taskQueue parameter
- Use config value for A3m shareDir in processing.go
- Use a temporary directory as a shareDir in processing_test.go

### DIFF
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/artefactual-sdps/enduro/internal/a3m"
 	"github.com/artefactual-sdps/enduro/internal/am"
+	"github.com/artefactual-sdps/enduro/internal/config"
 	"github.com/artefactual-sdps/enduro/internal/fsutil"
 	"github.com/artefactual-sdps/enduro/internal/package_"
 	"github.com/artefactual-sdps/enduro/internal/temporal"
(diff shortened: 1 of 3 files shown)


djjuhasz:
Commit eca89b656e9e4b825acae07b50230aa73cbe7732 fixes a hardcoded a3m shared directory path in Enduro, but the processing workflow is still failing if the path is changed from the default.  A3m finds the deposited SIP in the new path, and succesfully creates and AIP, but a3m is still saving the final AIP to the default "/home/a3m/.local/share/a3m/share/" directory instead of the new path.  A3m doesn't pass the stored AIP path back to Enduro directly so Enduro is assuming the path of the AIP is "/home/a3m/.local/share/a3m/new_share/completed", which is not correct.

Fixing this completely will require changes to a3m.

internal/workflow/processing_test.go
```diff
@@ -32,6 +34,9 @@ type ProcessingWorkflowTestSuite struct {
 
 	env *temporalsdk_testsuite.TestWorkflowEnvironment
 
+	// Each test creates it's own temporary transfer directory.
```
sevein:
```suggestion
	// Each test creates its own temporary transfer directory.
```

internal/workflow/processing.go
```diff
@@ -26,18 +27,23 @@ import (
 )
 
 type ProcessingWorkflow struct {
-	logger    logr.Logger
-	pkgsvc    package_.Service
-	wsvc      watcher.Service
-	taskQueue string
+	logger logr.Logger
+	cfg    config.Configuration
```
sevein, line 31: This could result in a lot of sharing, but I guess that's probably okay?

sevein (approved):
:+1:

Issue #867, opened 2024-02-22
Problem: processing workflow has separate code paths for filesystem and minio watchers


djjuhasz:
**Describe the problem**

The processing workflow and bundle activity use conditional branches to manage differences between the filesystem watcher and minio watcher implementations. The branching logic is hard to understand, requires some code duplication and makes the assumption that only a filesystem watcher can provide a "blob" that is a directory.

**Possible solutions**

The differences between the two watchers could be managed by moving the difference in implementation for copying the blob data to a local "processing" directory into their individual implementations instead of requiring branching logic (see "Avoiding conditionals by obeying the Liskov Substitution Principle" - https://sandimetz.com/99bottles).

**Additional context**

- https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/processing.go#L295 branches based on whether `tinfo.req.IsDir` is true (which is only possible for a filesystem watcher) or false (possible for both minio and filesystem watchers)
- https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/activities/bundle.go#L89 branches based on whether  `param.IsDir` (a local copy of `tinfo.req.IsDir`) is true. This branch requires duplicate code for creating a temporary dir in the transfer dir, and stripping a top-level directory
- the current implementation assumes that a SIP uploaded to MinIO can not be a directory of files, but we may want to change this in the future (see #845)

djjuhasz:
This is complicated by the fact that the current `watcher.Service.Download()` method takes an `io.Writer` parameter as the destination to write the byte stream.  An `io.Writer` stream works for a single file, as we expect from a minioWatcher, but doesn't work for a filesystem directory as we might receive from a filesystemWatcher.

PR #868, opened 2024-02-23
Standardize watcher Download() interface

djjuhasz:
Refs #867

- Update existing watcher Download() method and interface signatures to take a destination filepath (string) instead of an io.Writer stream
- Add a `filewatcher.Download()` method
- Create a `minio.Download()` method and move implementation specific code from `serviceImpl.Download()` to it
- Update mocks

### DIFF
--- a/internal/filenotify/filenotify.go
+++ b/internal/filenotify/filenotify.go
@@ -4,11 +4,17 @@
 package filenotify
 
 import (
+	"time"
+
 	"github.com/fsnotify/fsnotify"
 	"github.com/radovskyb/watcher"
 )
 
-// FileWatcher is an interface for implementing file notification watchers
+type Config struct {
+	PollInterval time.Duration
+}
+
+// FileWatcher is an interface for implementing file notification watchers.
 type FileWatcher interface {
 	Events() <-chan fsnotify.Event
 	Errors() <-chan error
@@ -18,11 +24,11 @@ type FileWatcher interface {
 }
 
 // New tries to use an fs-event watcher, and falls back to the poller if there is an error
-func New() (FileWatcher, error) {
+func New(cfg Config) (FileWatcher, error) {
 	if watcher, err := NewEventWatcher(); err == nil {
 		return watcher, nil
 	}
-	return NewPollingWatcher()
+	return NewPollingWatcher(cfg)
 }
 
 // NewEventWatcher returns an fs-event based file watcher
@@ -37,7 +43,7 @@ func NewEventWatcher() (FileWatcher, error) {
 }
 
 // NewPollingWatcher returns a poll-based file watcher
-func NewPollingWatcher() (FileWatcher, error) {
+func NewPollingWatcher(cfg Config) (FileWatcher, error) {
 	poller := &filePoller{
 		wr:     watcher.New(),
 		events: make(chan fsnotify.Event),
@@ -49,7 +55,7 @@ func NewPollingWatcher() (FileWatcher, error) {
 	done := make(chan error)
 	{
 		go func() {
-			err := poller.wr.Start(watchWaitTime)
+			err := poller.wr.Start(cfg.PollInterval)
 			if err != nil {
 				done <- err
 			}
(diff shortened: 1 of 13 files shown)


sevein:
@djjuhasz I think that you're environment is somehow still using an old version of mockgen.

Using bash:

```
$ source hack/activate/bash 
$ mockgen -version
v0.4.0
$ ls ~/.cache/enduro-sdps/Linux/x86_64/versions/mockgen/
0.4.0
```

Not sure what's going on. Maybe you're using a version of mockgen installed globally, but that shouldn't happen afaik. You could try this:

```
$ rm ~/.cache/enduro-sdps/Linux/x86_64/versions/mockgen/*
```

That would force mockgen to be downloaded again when running: `make gen-mock`.

djjuhasz:
@sevein I forgot to do the `source hack/activate/bash`, so I was generating mocks with mockgen v0.3.0. :(  I've just added a commit with the mockgen v0.4.0 output.

internal/watcher/config.go
```diff
@@ -29,14 +31,24 @@ func (c Config) CompletedDirs() []string {
...
 	RetentionPeriod  *time.Duration
-	CompletedDir     string
 	StripTopLevelDir bool
+
+	// PollInterval sets the length of time between filesystem polls (default:
+	// 200ms). If Inotify is true then PollInterval is ignored.
+	PollInterval time.Duration
+}
+
+func (cfg *FilesystemConfig) SetDefaults() {
```
sevein: Should this be unexported?

internal/filenotify/filenotify.go
```diff
@@ -4,11 +4,17 @@
...
+	"time"
+
 	"github.com/fsnotify/fsnotify"
 	"github.com/radovskyb/watcher"
 )
 
-// FileWatcher is an interface for implementing file notification watchers
+type Config struct {
+	PollInterval time.Duration
+}
```
sevein, line 15: Since this isn't directly related, I recommend submitting multiple commits next time. This approach is generally more convenient for the reviewer and improves the functionality of git blame. Archivematica provides a detailed explanation on this topic: https://github.com/artefactual/archivematica/blob/qa/1.x/CONTRIBUTING.md#commits-should-be-specific-and-atomic.

internal/watcher/watcher.go
```diff
@@ -31,8 +30,12 @@ type Watcher interface {
 	// Implementors must not return a nil function.
 	Watch(ctx context.Context) (*BlobEvent, Cleanup, error)
 
+	// Download copies the file or directory identified by key to dest.
+	Download(ctx context.Context, dest, key string) error
```
sevein, line 34: I wish that a `Watcher` didn't have to know how to download, or maybe that's ok? It sounds like something that we should be able to do using the `*blob.Bucket` returned by `OpenBucket` in the service layer, but I don't know if that's possible. `*blob.Bucket` was recently updated to implement the standard io/fs.FS and io/fs.SubFS interfaces (https://github.com/google/go-cloud/pull/3272) so you should be able to walk a bucket with [io/fs.WalkDir](https://pkg.go.dev/io/fs#WalkDir), but I don't know if that's enough. Maybe BlobEvent.IsDir() is also relevant. There's probably some room for improvement, but I'm happy with what you've done here. The internal/watcher package has always been at the back of my mind as something not very well-conceived, rushed and disorganized.

sevein (approved):
:+1:

internal/filenotify/filenotify.go
djjuhasz: Yeah, fair point.  I wanted a configurable poll interval for the tests, but I don't think using a shorter (5ms) poll interval actually sped up the tests much anyway. :shrug:

internal/watcher/watcher.go
djjuhasz: I agree that it would be better if the Watcher service was solely concerned with watching for transfers and emitting change events, with downloading the blob the responsibility of a separate module - but I think this would be a big change. I'll file an issue about separating the watcher and download functionality and maybe we can tackle it later.

internal/watcher/config.go
djjuhasz: Yes, good catch. :+1:

PR #871, opened 2024-02-28
Remove Watcher Service from Bundle activity

### LINKED ISSUE #870: Problem: The Bundle Activity is dependent on the Watcher Service
**Describe the problem**

The Bundle Activity has a dependency on the Watcher Service, which is only ever used to determine the path of a transfer directory inside a watched filesystem directory.  

**Solutions considered**

The full path of the transfer could easily be passed as string as a simpler and lighter alternative to injecting the Watcher Service.

djjuhasz:
Fixes #870


sevein (approved):
Nice catch.

PR #872, opened 2024-03-02
Add an unarchive activity

### LINKED ISSUE #869: Problem: The Bundle Activity extracts the contents of an archive file package
**Describe the problem**

The Bundle Activity is currently responsible for extracting an archive (e.g. zip, tgz) package, then "bundling" (restructuring) the package contents as an Archivematica standard transfer.  Extracting the archive contents in the bundle activity means that no other activities can operate on the extracted contents before bundling happens.  The mixing of responsibilities also makes the Bundle activity more complex and harder to understand. 

**Solutions considered**

1. Create a separate activity that is solely responsible for extracting an archive package's contents to a working directory, and remove the archive extraction code from the bundle activity.

djjuhasz:
Fixes #869

- Move archive extraction code from bundle activity to unarchive activity
- Add an `fsutil.SetFileModes()` method to recursively set file modes on a dir
- Add an unzipped bag testdata transfer


internal/workflow/processing.go
```diff
@@ -313,11 +311,30 @@ func (w *ProcessingWorkflow) SessionHandler(sessCtx temporalsdk_workflow.Context
 				if err != nil {
 					return err
 				}
-				tinfo.TempFile = downloadResult.Path
+				tinfo.TempPath = downloadResult.Path
 			}
 		}
 	}
 
+	// Unarchive.
```
sevein: Can we short-circuit if `tinfo.req.IsDir` is true? I'm not certain if that's the case, but if there's a chance to avoid running the activity, I think we should take it.

internal/workflow/activities/unarchive.go
```diff
@@ -0,0 +1,178 @@
...
+type UnarchiveParams struct {
+	// SourcePath is the path of the transfer to be unarchived.
+	SourcePath string
+
+	// StripTopLevelDir indicates whether the top-level "container" directory of
+	// the archive should be excluded from the destination directory.
+	StripTopLevelDir bool
+}
+
+type UnarchiveResult struct {
```
Diogenesoftoronto: It is standard to write NameActivityResult, NameActivityParams. This is done in all other activities as far as I am aware.

internal/workflow/activities/unarchive.go
```diff
@@ -0,0 +1,178 @@
...
+			return nil, fmt.Errorf("unarchive: strip top-level dir: %v", err)
+		}
+	}
+
+	if err := fsutil.SetFileModes(dest, ModeDir, ModeFile); err != nil {
+		return nil, fmt.Errorf("unarchive: %v", err)
+	}
+
+	if err := os.Remove(params.SourcePath); err != nil {
+		a.logger.V(1).Info("Unarchive: couldn't delete source archive: %v", err)
```
Diogenesoftoronto: should this not be an error log? if the source archive could not be deleted that might require some action because limited storage. Might make sense to alert the user here above info.

internal/workflow/activities/unarchive.go
```diff
@@ -0,0 +1,178 @@
...
+// stripDirContainer strips the top-level directory of a transfer.
+func stripDirContainer(dir string) error {
+	tld, err := topLevelDir(dir)
+	if err != nil {
+		return fmt.Errorf("get top-level dir: %v", err)
+	}
+
+	tmpDir, err := os.MkdirTemp("", "enduro-")
+	if err != nil {
+		return fmt.Errorf("make temp dir: %v", err)
```
Diogenesoftoronto, line 116: not sure the error "make temp dir" adds information to the error here. might as well just return the error.

internal/workflow/activities/unarchive.go
```diff
@@ -0,0 +1,178 @@
...
+	return nil
+}
+
+// topLevelDir reads the directory at path and returns the name of it's
+// immediate sub-directory. If path contains anything other than a single
+// sub-directory then topLevelDirectory will return an error.
+func topLevelDir(path string) (string, error) {
+	r, err := os.Open(filepath.Clean(path))
+	if err != nil {
+		return "", fmt.Errorf("cannot open path: %v", err)
```
Diogenesoftoronto: os open will already tell us that we could not open a path. This is another case I think it might make sense to give more useful error messages or just return the base error.

internal/workflow/activities/unarchive_test.go
```diff
@@ -0,0 +1,189 @@
...
+			assert.NilError(t, err)
+
+			var res activities.UnarchiveResult
+			_ = enc.Get(&res)
+
+			if tt.want != nil {
+				tt.want.DestPath = sourceDir.Join(tt.want.DestPath)
+				assert.DeepEqual(t, &res, tt.want)
+			}
+			if tt.wantFs != (fs.Manifest{}) {
```
Diogenesoftoronto, line 174: just to be clear why the paranthesis here? what do they add?

Diogenesoftoronto (changes requested):
Great job on this! I think I have some minor feedback but it looks pretty good overall.

internal/workflow/activities/unarchive.go
djjuhasz: Huh, I think at some point @sevein discouraged me from using `a.logger.Error()`, but I can't remember the details now. :thinking:

internal/workflow/activities/unarchive.go
djjuhasz: I don't actually know what the error output from `os.MkdirTemp()` looks like.  My current philosophy on Go error messages is to add some context to make it easy to find the caller (e.g. Oh, this error is in "unarchive: stripDirContainer: make temp dir: [error message]").   I notice I haven't included the "stipDirContainer" context in this case though, which is probably helpful to add.

internal/workflow/activities/unarchive.go
djjuhasz: This error message was copied when I moved this code from `activities/bundle.go` - the only change I made was to remove the "errPrefix" string.

internal/workflow/activities/unarchive_test.go
djjuhasz: The code doesn't compile without them. :shrug:

internal/workflow/activities/unarchive.go
sevein: I think that we'd want to log an error if we want to alert the system operator because there's an actionable issue that needs to be addressed. Running out of disk space is a good example but we can also assume that system monitoring should already be in place to monitor that kind of problem earlier in production. I'm ambivalent but I generally lean towards not logging errors unless strictly needed - one risk of logging too many errors is that they may ultimately be overlooked.

internal/workflow/activities/unarchive.go
sevein: https://github.com/uber-go/guide/blob/master/style.md#error-wrapping reads that the underlying error can be returned as-is if there is no additional context that can be added, but I think I tend to always add some context just like you did because I can't never remember which functions in the standard library come with enough context (they're definitely not consistent). I think I'd be ok with both approaches, the worst that could happen is to have a little repetition, which is better than no context.

> I notice I haven't included the "stipDirContainer" context in this case though, which is probably helpful to add.

Yeah good point, I feel like we haven't done this enough. I could be a nice practice to adopt.


On this topic, while browsing [pkgsite](https://github.com/golang/pkgsite) the other day I saw a pretty nifty [pattern using named return values and defer](https://github.com/golang/pkgsite/blob/ea43129276ed1c7e85557321bd2af0504fbc4a7f/internal/auth/auth.go#L34-L55) to wrap the error with the name of the function.

```go
func Header(ctx context.Context, jsonCreds []byte, useExp bool) (_ string, err error) {
	defer derrors.Wrap(&err, "auth.Header(jsonCreds, %t)", useExp)
	// [...]
}
```

Sadly it only works in combination with named return parameters (try [playground](https://go.dev/play/p/ME004ReW6J2)) which are avoided by a lot of people, e.g. see [nonamedreturns](https://github.com/firefart/nonamedreturns) supported by golangci-lint. It's in the [spec](https://go.dev/ref/spec#Defer_statements) but I guess they can be misleading? Idk.

jraddaoui: That's nice @sevein, thanks!

Make sure you use and **return** `err` inside `// [...]`. I think the example you shared will miss the extra wrapping done in lines [46](https://github.com/golang/pkgsite/blob/ea43129276ed1c7e85557321bd2af0504fbc4a7f/internal/auth/auth.go#L46) and [51](https://github.com/golang/pkgsite/blob/ea43129276ed1c7e85557321bd2af0504fbc4a7f/internal/auth/auth.go#L51) since that's not returning the actual `err` like in line [42](https://github.com/golang/pkgsite/blob/ea43129276ed1c7e85557321bd2af0504fbc4a7f/internal/auth/auth.go#L42).

I use this pattern to be able to join errors from defer functions and return them with the original one.

internal/workflow/activities/bundle.go
```diff
@@ -101,51 +90,28 @@ func (a *BundleActivity) Execute(ctx context.Context, params *BundleActivityPara
...
-		return nil, temporal_tools.NewNonRetryableError(fmt.Errorf(
-			"error calculating relative path to transfer (base=%q, target=%q): %v",
-			params.TransferDir, res.FullPath, err,
-		))
+		return nil, temporal_tools.NewNonRetryableError(
+			fmt.Errorf("bundle: unbag: %v", err),
+		)
 	}
 
 	if err = setPermissions(res.FullPath); err != nil {
```
jraddaoui: Should this use the new `fsutil.SetFileModes` function?

internal/workflow/activities/bundle.go
```diff
@@ -51,9 +50,7 @@ type BundleActivityParams struct {
 }
 
 type BundleActivityResult struct {
-	RelPath             string // Path of the transfer relative to the transfer directory.
-	FullPath            string // Full path to the transfer in the worker running the session.
-	FullPathBeforeStrip string // Same as FullPath but includes the top-level dir even when stripped.
```
jraddaoui, line 56: I think you could remove the `StripTopLevelDir` parameter after this changes too.

internal/workflow/activities/unarchive.go
```diff
@@ -0,0 +1,178 @@
...
+
+type UnarchiveResult struct {
+	// DestPath is the path to the extracted archive contents.
+	DestPath string
+
+	// IsDir is true if DestPath is a directory.
+	IsDir bool
+}
+
+func NewUnarchiveActivity(logger logr.Logger) *UnarchiveActivity {
```
jraddaoui, line 40: So many versions of this activity already, it would be nice to be able to share it with other repositories.

internal/workflow/processing_test.go
```diff
@@ -162,16 +163,20 @@ func (s *ProcessingWorkflowTestSuite) TestPackageConfirmation() {
 		&activities.DownloadActivityResult{Path: "/tmp/enduro123456/" + key}, nil,
 	)
 
+	s.env.OnActivity(activities.UnarchiveActivityName, sessionCtx,
+		&activities.UnarchiveParams{SourcePath: "/tmp/enduro123456/" + key},
+	).Return(
+		&activities.UnarchiveResult{DestPath: "/tmp/enduro123456/extract"}, nil,
```
jraddaoui: It would be nice to have these two paths as variables and reuse them. That would have helped with some modifications I made in a fork where the path changed in a preprocessing step and I had to modify its usage in all calls later.

internal/workflow/activities/unarchive.go
jraddaoui: E.g.: https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/internal/workflow/preprocessing.go#L49

internal/workflow/processing.go
djjuhasz: @sevein it's cool that we have data now on the latency for scheduling and running Temporal activities - I had no idea. :open_mouth:

internal/workflow/activities/bundle.go
djjuhasz: I thought I had! :facepalm: :laughing:

internal/workflow/activities/bundle.go
djjuhasz: Yes, totally :+1:

internal/workflow/activities/unarchive.go
djjuhasz: @jraddaoui are you thinking that we would add an unarchive package to https://github.com/artefactual-labs/gotools/, or create a separate repo, or something else?

internal/workflow/activities/unarchive.go
djjuhasz: I don't know if `Info()` or `Error()` is better here, but I lean slightly to `Info()` because `SourcePath()` is a temporary directory.  I think the worst case scenario is the system has no monitoring, `Remove()` fails for some reason and the "/tmp" partition fills up.  In this case a system restart will clear out any lingering transfers in "/tmp".

internal/workflow/activities/unarchive.go
djjuhasz: I did actually include the error message "stripDir:" context at https://github.com/artefactual-sdps/enduro/blob/dev/issue-869-unarchive-activity/internal/workflow/activities/unarchive.go#L77

jraddaoui: Thanks @sevein for the background chat. It actually seems to work without returning the named return: https://go.dev/play/p/m8j5Gr0ZHDZ

internal/workflow/activities/unarchive.go
jraddaoui: I'd create a new repository in labs or sdps, `temporal-activities` or something like that and add this activity there alongside the one from https://github.com/artefactual-sdps/remove-files-activity. Not sure the best way to organize them there, if we should use different packages for each activity o have them all together in a single package.

internal/workflow/activities/unarchive.go
Diogenesoftoronto: > I don't know if `Info()` or `Error()` is better here, but I lean slightly to `Info()` because `SourcePath()` is a temporary directory. I think the worst case scenario is the system has no monitoring, `Remove()` fails for some reason and the "/tmp" partition fills up. In this case a system restart will clear out any lingering transfers in "/tmp".

The worst case is actually the system continues running and the tmp directories eat up all the memory and the program panics due to Out of Memory error during an activity. I would expect full system restarts to not even be that common. 

Is it possible to do something raise a 'Warn()' or something?

djjuhasz: https://github.com/go-logr/logr only has `Info()` (with a verbosity level) and `Error()` methods because it's philosophy is to minimize the number of log levels. I'll switch to `Error()` here because you seem to feel strongly about it @Diogenesoftoronto.  In practice I don't expect `os.Remove()` to return an error very often, so I'm not worried about it cluttering up the error logs.

internal/workflow/activities/unarchive.go
djjuhasz: I've updated this error message and the others in `topLevelDir()`.

internal/workflow/activities/unarchive.go
djjuhasz: @jraddaoui I think that's a good plan for the future, but I'm going to leave the unarchive activity in this project for this PR; Moving the activity to a new project is too big a task for me to tackle it here.

internal/workflow/activities/unarchive.go
Diogenesoftoronto: > https://github.com/uber-go/guide/blob/master/style.md#error-wrapping reads that the underlying error can be returned as-is if there is no additional context that can be added, but I think I tend to always add some context just like you did because I can't never remember which functions in the standard library come with enough context (they're definitely not consistent). I think I'd be ok with both approaches, the worst that could happen is to have a little repetition, which is better than no context.
> 
> > I notice I haven't included the "stipDirContainer" context in this case though, which is probably helpful to add.
> 
> Yeah good point, I feel like we haven't done this enough. I could be a nice practice to adopt.
> 
> On this topic, while browsing [pkgsite](https://github.com/golang/pkgsite) the other day I saw a pretty nifty [pattern using named return values and defer](https://github.com/golang/pkgsite/blob/ea43129276ed1c7e85557321bd2af0504fbc4a7f/internal/auth/auth.go#L34-L55) to wrap the error with the name of the function.
> 
> ```go
> func Header(ctx context.Context, jsonCreds []byte, useExp bool) (_ string, err error) {
> 	defer derrors.Wrap(&err, "auth.Header(jsonCreds, %t)", useExp)
> 	// [...]
> }
> ```
> 
> Sadly it only works in combination with named return parameters (try [playground](https://go.dev/play/p/ME004ReW6J2)) which are avoided by a lot of people, e.g. see [nonamedreturns](https://github.com/firefart/nonamedreturns) supported by golangci-lint. It's in the [spec](https://go.dev/ref/spec#Defer_statements) but I guess they can be misleading? Idk.

I believe the major reason named returns are not encouraged is because they allow naked returns with those values which can get hairy pretty quickly. For example if you have a named return here you can do something like 

```go
func add(a, b int) (sum int) {
    sum = a + b
    return
} 
// this return the sum of a and b
```

https://go.dev/play/p/22eUUbi0vsI for the example with @jraddaoui will return two EOF errors to the console with a new line.

internal/workflow/activities/unarchive.go
Diogenesoftoronto: I mean it is not just me this is from the creators of go-logr on when to use the Error():

`Info logs are things you want to tell the user which are not errors. Error logs are, well, errors. If your code receives an error from a subordinate function call and is logging that error and not returning it, use error logs.`

Diogenesoftoronto (approved):

PR #878, opened 2024-03-07
Update Codecov ignore list

djjuhasz:


djjuhasz:
This change didn't impact the test coverage report at all, but I think it's a more accurate list of directories that should be ignored by Codecov.

sevein (approved):

PR #879, opened 2024-03-08
Block in filePoller.Close until loop is finished

sevein:
Regarding the indirect changes often reported by Codecov, I've observed on https://app.codecov.io/gh/artefactual-sdps/enduro/pull/877/indirect-changes that the close case in the filePoller pool's select statement is not consistently executed. This is probably happening because the test occasionally finishes before that section of code is executed. This behavior might be attributed to the inherent non-determinism of Go's select statement. To address it, I've updated `filePoller.Close` to ensure that it waits until the pool has completely finished, ensuring that the loop has always an opportunity to be released.

I've also modified the job step "Upload coverage to Codecov" in our CI to enable skipping this step if the commit message includes the string `[skip codecov]`. I needed this because it is not obvious how we can test the timeout case that I've introduced in the Close method. I suspect that we'll find other situations where we can't provide enough coverage but we still need successful CI builds for deployments, merge checks, etc.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,8 +22,22 @@ jobs:
         cache: true
     - name: Test
       run: make test-ci
+    - name: Determine skip-codecov
+      id: skip-codecov
+      with:
+        script: |
+          const { repo, owner } = context.repo;
+          const { data: commit } = await github.rest.repos.getCommit({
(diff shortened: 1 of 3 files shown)


djjuhasz (approved):
LGTM! :+1:

PR #881, opened 2024-03-08
Add result struct to activity function signature

### LINKED ISSUE #800: Problem: some activities only return `error`
**Describe the bug**

A number of activities in this project are not returning a struct. This can lead to backward-compatibility issues if we add one in the future.  If we use a struct, the activity remains compatible even if the fields change. This means we can add, remove, or change fields in our struct without breaking the function signature.

Some examples:

* CopyToPermanentLocationActivity,
* CleanUpActivity,
* DeleteOriginalActivity,
* DisposeOriginalActivity,
* MoveToPermanentStorageActivity,
* PollMoveToPermanentStorageActivity,
* RejectPackageActivity,
* UploadActivity, or
* [DeleteTransferActivity].

**Expected behavior**

Always return a struct even if empty. Refer to the code example in https://docs.temporal.io/dev-guide/go/foundations#activity-return-values:

```go
// YourActivityResultObject is the struct returned from your Activity.
// Use a struct so that you can return multiple values of different types.
// Additionally, your function signature remains compatible if the fields change.
type YourActivityResultObject struct {
    ResultFieldX string
    ResultFieldY int
}
```

**Additional context**

I believe this is a future-proofing effort that will become more important when we hit production and as upgrades are rolled out progressively the system will have both old and new versions of workers running concurrently.

Related: https://github.com/artefactual-sdps/enduro/pull/787#discussion_r1409683586.

[DeleteTransferActivity]: https://github.com/artefactual-sdps/enduro/blob/174f7f057ac5fc7384292d34e9820b4ac4962f0e/internal/am/delete_transfer.go#L27

Diogenesoftoronto:
closes #800

### DIFF
--- a/internal/am/delete_transfer.go
+++ b/internal/am/delete_transfer.go
@@ -20,19 +20,21 @@ type DeleteTransferActivity struct {
 	logger logr.Logger
 }
 
+type DeleteTransferActivityResult struct{}
+
 func NewDeleteTransferActivity(logger logr.Logger, client sftp.Client) *DeleteTransferActivity {
 	return &DeleteTransferActivity{client: client, logger: logger}
 }
 
-func (a *DeleteTransferActivity) Execute(ctx context.Context, params *DeleteTransferActivityParams) error {
+func (a *DeleteTransferActivity) Execute(ctx context.Context, params *DeleteTransferActivityParams) (*DeleteTransferActivityResult, error) {
 	a.logger.V(1).Info("Execute DeleteTransferActivity",
 		"destination", params.Destination,
 	)
 
 	err := a.client.Delete(ctx, params.Destination)
 	if err != nil {
-		return fmt.Errorf("delete transfer: path: %q: %v", params.Destination, err)
+		return &DeleteTransferActivityResult{}, fmt.Errorf("delete transfer: path: %q: %v", params.Destination, err)
 	}
 
-	return nil
+	return &DeleteTransferActivityResult{}, nil
 }
--- a/internal/workflow/activities/delete_original.go
+++ b/internal/workflow/activities/delete_original.go
@@ -10,10 +10,12 @@ type DeleteOriginalActivity struct {
 	wsvc watcher.Service
 }
 
+type DeleteOriginalActivityResult struct{}
+
 func NewDeleteOriginalActivity(wsvc watcher.Service) *DeleteOriginalActivity {
 	return &DeleteOriginalActivity{wsvc: wsvc}
 }
 
-func (a *DeleteOriginalActivity) Execute(ctx context.Context, watcherName, key string) error {
-	return a.wsvc.Delete(ctx, watcherName, key)
+func (a *DeleteOriginalActivity) Execute(ctx context.Context, watcherName, key string) (*DeleteOriginalActivityResult, error) {
+	return &DeleteOriginalActivityResult{}, a.wsvc.Delete(ctx, watcherName, key)
 }
--- a/internal/workflow/activities/dispose_original.go
+++ b/internal/workflow/activities/dispose_original.go
@@ -10,10 +10,12 @@ type DisposeOriginalActivity struct {
 	wsvc watcher.Service
 }
 
+type DisposeOriginalActivityResult struct{}
+
 func NewDisposeOriginalActivity(wsvc watcher.Service) *DisposeOriginalActivity {
 	return &DisposeOriginalActivity{wsvc: wsvc}
 }
 
-func (a *DisposeOriginalActivity) Execute(ctx context.Context, watcherName, completedDir, key string) error {
-	return a.wsvc.Dispose(ctx, watcherName, key)
+func (a *DisposeOriginalActivity) Execute(ctx context.Context, watcherName, completedDir, key string) (*DisposeOriginalActivityResult, error) {
+	return &DisposeOriginalActivityResult{}, a.wsvc.Dispose(ctx, watcherName, key)
 }
--- a/internal/storage/activities/copy.go
+++ b/internal/storage/activities/copy.go
@@ -11,47 +11,49 @@ type CopyToPermanentLocationActivity struct {
 	storagesvc storage.Service
 }
 
+type CopyToPermanentLocationActivityResult struct{}
+
 func NewCopyToPermanentLocationActivity(storagesvc storage.Service) *CopyToPermanentLocationActivity {
 	return &CopyToPermanentLocationActivity{storagesvc: storagesvc}
 }
 
-func (a *CopyToPermanentLocationActivity) Execute(ctx context.Context, params *storage.CopyToPermanentLocationActivityParams) error {
+func (a *CopyToPermanentLocationActivity) Execute(ctx context.Context, params *storage.CopyToPermanentLocationActivityParams) (*CopyToPermanentLocationActivityResult, error) {
 	p, err := a.storagesvc.ReadPackage(ctx, params.AIPID)
 	if err != nil {
-		return err
+		return &CopyToPermanentLocationActivityResult{}, err
 	}
 
 	reader, err := a.storagesvc.PackageReader(ctx, p)
 	if err != nil {
-		return err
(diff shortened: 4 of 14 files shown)


djjuhasz (approved):
LGTM! :+1:

internal/am/delete_transfer.go
```diff
@@ -20,19 +20,21 @@ type DeleteTransferActivity struct {
...
 	)
 
 	err := a.client.Delete(ctx, params.Destination)
 	if err != nil {
-		return fmt.Errorf("delete transfer: path: %q: %v", params.Destination, err)
+		return nil, fmt.Errorf("delete transfer: path: %q: %v", params.Destination, err)
 	}
 
-	return nil
+	return nil, nil
```
sevein: This should read:

    return &DeleteTransferActivityResult{}, nil

internal/workflow/activities/delete_original.go
```diff
@@ -10,10 +10,12 @@ type DeleteOriginalActivity struct {
...
+type DeleteOriginalActivityResult struct{}
+
 func NewDeleteOriginalActivity(wsvc watcher.Service) *DeleteOriginalActivity {
 	return &DeleteOriginalActivity{wsvc: wsvc}
 }
 
-func (a *DeleteOriginalActivity) Execute(ctx context.Context, watcherName, key string) error {
-	return a.wsvc.Delete(ctx, watcherName, key)
+func (a *DeleteOriginalActivity) Execute(ctx context.Context, watcherName, key string) (*DeleteOriginalActivityResult, error) {
+	return nil, a.wsvc.Delete(ctx, watcherName, key)
```
sevein: This is also a problem, you're returnil a nil result regardless the result of Delete.

sevein (changes requested):
Your approach is nearly correct; however, I've noticed an oversight. You're returning a nil pointer instead of a pointer to the result struct when there's no error. I'll highlight a few examples, although I've observed this issue in several places.

internal/workflow/activities/delete_original.go
sevein: ~~returnil~~ lol, I meant "returning"

internal/am/delete_transfer.go
Diogenesoftoronto: I think is fine because we are not using these pointers. When we do actually use the pointer this will no longer be an issue because we would have changed activity result fields  anyway, which means turning this result into a struct. What a nil pointer return signals here is a currently unused value. if that value is used it should panic. I think this is just easier to read as well.

internal/workflow/activities/delete_original.go
Diogenesoftoronto: see https://github.com/artefactual-sdps/enduro/pull/881#discussion_r1520143356 for why.

internal/am/delete_transfer.go
sevein: What you're saying doesn't seem correct to me. We want to return the struct even if empty because the JSON data converter used by the SDK will encode it as an object (`{}`), which can later have new fields added while maintaining backward compatibility. Returning a nil will result in something like an empty string or similar.

sevein: I've just tested it, the converter encodes `(*T)(nil)` as an empty string (`""`) whereas `&T{}` becomes: `"{}"`. See how the background-checks app avoids returning `nil`, e.g. see [`SendDeclineEmail`](https://github.com/temporalio/background-checks/blob/6a23013be4f2cf03ae8463dcbb64536abc98db93/activities/activities.go#L169-L174). Let me know if this makes sense. It's hard to imagine scenarios where this can be relevant (backward and forward compatibility in a distributed environment, rolling upgrades while having running workflow instances, etc), but it seems to be a good idea as a precaution?

```go
package main

import (
	"fmt"
	"log"

	"go.temporal.io/sdk/converter"
)

type Result struct{}

func main() {
	conv := converter.GetDefaultDataConverter()

	r1 := (*Result)(nil)
	if payload, err := conv.ToPayload(r1); err != nil {
		log.Fatal(err)
	} else {
		fmt.Printf("r1: JSON %+q\n", payload.Data)
		// r1: JSON ""
	}

	r2 := &Result{}
	if payload, err := conv.ToPayload(r2); err != nil {
		log.Fatal(err)
	} else {
		fmt.Printf("r2: JSON %+q\n", payload.Data)
		// r2: JSON "{}"
	}
}
```

sevein: Here's a hypothetical scenario: we deploy an updated version of a workflow function. In our production environment, there are active workflow instances that, upon replay by the workflow worker to restore the state, encounter a situation where they need to deserialize data previously encoded as an empty string (resulting from returning `(*T)(nil)` in older versions of an activity) into a `*T` struct. This is where we would see the discrepancy between the expected data formats: the data converter, tasked with unmarshalling the empty string into a `*T` would fail (panic?).

Diogenesoftoronto: Just an update on this, a few things that I noticed for future prs.

1. We are not using structs in three places, (they are only in local activities so this might be fine anyway) we are using uints
2. This might not be a good solution but alternative I explored is changing the default data converter to a custom data converter that deserializes structs and empty pointers in the same way. It might be useful to explore doing this for a couple reasons and bad for others: 
     - good if we need to switch away from using json at some point (a common use case demonstrated in the temporal docs) or for some other reason want to change the data converter.
     - bad because it might result in unexpected behavior
 Here are the docs on data conversion: https://docs.temporal.io/dataconversion
 
 Temporal's ToPayload is kind of simple in that it checks if something is an empty interface and if it is it just creates a new payload with no data. We could override this so that we send the zero value of the referent instead just like temporal does with &SomeEmptyActivityResult{}

Diogenesoftoronto: 3. Temporal recommends pointers to struct for the scenario's @sevein outlines, that is it means you do not have breaking changes and we can deploy upgraded versions while the services is up. If that is not a use case that we have (which we do not right now) then I am not sure how helpful this PR is. 
from temporal docs -> 
```go
// Use a struct so that you can return multiple values of different types.
// Additionally, your function signature remains compatible if the fields change.
```

Diogenesoftoronto:
I made the recommend changes in the new commits. I also added some comments with my thoughts as a whole.

internal/am/delete_transfer.go
Diogenesoftoronto: 4. If this is an issue, perhaps we should also change workflow definitions to pass empty structs as they currently only return an error. May also be the case for child workflows we design in the future as well.

internal/workflow/activities/dispose_original.go
```diff
@@ -10,10 +10,12 @@ type DisposeOriginalActivity struct {
...
+type DisposeOriginalActivityResult struct{}
+
 func NewDisposeOriginalActivity(wsvc watcher.Service) *DisposeOriginalActivity {
 	return &DisposeOriginalActivity{wsvc: wsvc}
 }
 
-func (a *DisposeOriginalActivity) Execute(ctx context.Context, watcherName, completedDir, key string) error {
-	return a.wsvc.Dispose(ctx, watcherName, key)
+func (a *DisposeOriginalActivity) Execute(ctx context.Context, watcherName, completedDir, key string) (*DisposeOriginalActivityResult, error) {
+	return nil, a.wsvc.Dispose(ctx, watcherName, key)
```
sevein: This is still returning `(*DisposeOriginalActivityResult)(nil)`.

sevein (changes requested):

internal/am/delete_transfer.go
sevein: Make sure you bring this up to the team! I checked `preprocessing-sfa` and its child workflow seems to be returning a struct.

internal/workflow/activities/dispose_original.go
Diogenesoftoronto: Thanks for finding this, I found one other one while checking as well.

sevein:
Can you rebase? :pray:

sevein (approved):

PR #884, opened 2024-03-09
Fix skip-codecov check in test workflow

sevein:
I added an extra step in test.yml to optionally disable Codecov in commit e5676993. I believed it was working, meaning that Codecov wasn't running, but that was a false positive; the entire job was actually broken. This commit fixes it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,21 +23,22 @@ jobs:
     - name: Test
       run: make test-ci
     - name: Determine skip-codecov
+      uses: actions/github-script@v7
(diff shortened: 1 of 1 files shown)


sevein:
This seems to be working now - tested with https://github.com/artefactual-sdps/enduro/pull/884/commits/8c53e00a5cae1576d882b1c847bdb14d2dd3c941 (now removed from this branch).
The [run logs](https://github.com/artefactual-sdps/enduro/actions/runs/8213099914/job/22464093650) show that the coverage report step is successfully skipped.

PR #887, opened 2024-03-13
Restore unparam linter

sevein:
Restore the [`unparam`](https://github.com/mvdan/unparam) linter in golangci-lint since it's working again with recent releases of Go and `golang.org/x/tools`.


djjuhasz (approved):
LGTM :+1:

PR #889, opened 2024-03-13
Add the persistence service to the package service

### LINKED ISSUE #888: Problem: the package service can only use a MySQL datastore
**Describe the problem**

The package service assumes that the underlying data storage is a MySQL database, directly querying the database using SQL statements.  This tight coupling of the package service layer with the persistence layer makes it very difficult to switch to a different datastore in the future, or to test against a different datastore (e.g. SQLite).

**Solutions explored**

1. Using https://entgo.io/. The storage service has a separate persistence service which uses an https://entgo.io/ to provide an ORM that abstracts the data modelling and queries from the underlying datastore.  Using the same pattern of a persistence layer + https://entgo.io/ would harmonize the package and storage service code, reducing the cognitive burden of switching between different systems in the two different packages, while reducing coupling to MySQL.

**Additional context**

I did some initial work in Sept 2023 to add a package persistence layer and an entgo model of the package database:

1. https://github.com/artefactual-sdps/enduro/commit/52318a91e6e68d6526433b2ddd11ffc03ea5f09e
2. https://github.com/artefactual-sdps/enduro/commit/a5bf7465fb9568033ca985c75d25c37df16fb10d
3. https://github.com/artefactual-sdps/enduro/commit/5f5e04b400fa84a397d1ff2d2411732b6d71dedf

But other work came up that prevented further progress on this path.

djjuhasz:
Refs #888

By adding the persistence service to the package service alongside the existing `db` client, I am introducing a way to incrementally switch the service implementation from using raw SQL queries to relying on the persistence service for data storage and retrieval.

- Inject the persistence service into the package service
- Move the `package_.Package` struct to `datatype.Package` to allow importing it in the persistence and package_ packages without circular import problems
- Move the `package_.Status` enum to `enums.PackageStatus` to avoid circular import problems
- Rename the `enums.PackageStatus` constants to avoid ambiguous names


sevein (approved):
I appreciate the effort you've put into submitting multiple commits, it makes things much easier to follow! :+1:

sevein (approved):
I've been testing this today, works great!

PR #890, opened 2024-03-14
Update go in devbox

Diogenesoftoronto:


devbox.json
```diff
@@ -9,7 +9,8 @@
     "docker@latest",
     "python3@latest",
     "markdownlint-cli@latest",
-    "python311Packages.pip-tools@latest"
+    "python311Packages.pip-tools@latest",
+    "go_1_22@latest"
```
sevein, line 12: Do we have to keep go_1_21?

sevein: Have you noticed that it's installing 1.22.0 instead of 1.22.1? Can that be changed?

Diogenesoftoronto: Oh absolutely. Seems that stuck around during some work I was doing in the repo.

devbox.lock
```diff
@@ -41,23 +41,23 @@
...
-          "store_path": "/nix/store/id3ygqxp1hvbh7dyjlwngx53fvcpqg84-go-1.21.1"
+          "store_path": "/nix/store/7wxzkvjv8qc2awhagpz0r8q9ay38q3wj-go-1.22.0"
         },
         "x86_64-darwin": {
-          "store_path": "/nix/store/yn0g820ayb963s0i4nnw9pf72l49kkg0-go-1.21.1"
+          "store_path": "/nix/store/fgkl3qk8p5hnd07b0dhzfky3ys5gxjmq-go-1.22.0"
         },
         "x86_64-linux": {
-          "store_path": "/nix/store/pvvv2lmx9m8b2n0447mljrl5xmla8r3h-go-1.21.1"
+          "store_path": "/nix/store/88y9r33p3j8f7bc8sqiy9jdlk7yqfrlg-go-1.22.0"
```
sevein: Do you know why it chooses to install 1.22.0 instead of 1.22.1?

Diogenesoftoronto: huh must have been before the nix package repo had been updated, when i load the latest it works with 1.22.1 now

sevein (changes requested):
Weird, I still see "1.22.0".

Diogenesoftoronto:
> Weird, I still see "1.22.0".

Oh my gosh, sorry for the long update, some of my work got lost. I am learning to use gitbutler and some of my changes were not recovered because of the way that it works.

sevein (approved):

PR #893, opened 2024-03-15
Enable tracing in Redis clients

sevein:
This pull request enables tracing instrumentation in our Redis clients using the `redisotel` package.

### Find watcher triggers

This is likely the most interesting outcome of this work. Now watchers start a root span (`watcher.poll`) representing the interaction with Redis using the `blmove` command to wait until the notification is delivered by MinIO. In Grafana, search for:

    {name=~"watcher.poll"}

image

The gantt chart shows that we remove the notification from the queue using `lrem` command, but only once we've successfully started the workflow instance. The workflow continues running and recording spans as part of the same trace, including the entire workflow execution, activities, etc...

### Find creation of WebSocket auth tickets

We can see how authentication tickets are stored in Redis:

    {name="api" && .http.route="/package/monitor"}
    
image

### Find application publishing events

We can also find spans related to the operation of publishing events into the event service, e.g.:

    {name="publish"}

#### Find Redis traces

A more general approach:

    {.db.system="redis"}
    
image

### DIFF
--- a/cmd/enduro-a3m-worker/main.go
+++ b/cmd/enduro-a3m-worker/main.go
@@ -115,7 +115,7 @@ func main() {
 	}
 
 	// Set up the event service.
-	evsvc, err := event.NewEventServiceRedis(logger.WithName("events"), &cfg.Event)
+	evsvc, err := event.NewEventServiceRedis(logger.WithName("events"), tp, &cfg.Event)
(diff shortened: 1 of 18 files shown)


djjuhasz (approved):
Cool! :sunglasses:

Issue #892, opened 2024-03-14
Problem: unexpected coverage changes


sevein:
**Describe the bug**

Some pull requests are not building successfully because codecov is reporting negative coverage, i.e. unexpected coverage changes. It's easy to find "indirect changes" via the Codecov web interface but it's not always easy to fix them.

This issue aims to be used track areas of the code where this is happening.

- [x] `filenotify` (fixed in https://github.com/artefactual-sdps/enduro/pull/879)
- [ ] `sftp` (follow [link](https://app.codecov.io/gh/artefactual-sdps/enduro/pull/891/indirect-changes) for more details)
- [ ] `watcher` (follow [link](https://app.codecov.io/gh/artefactual-sdps/enduro/pull/891/indirect-changes) for more details)

Also:

- [x] We can [configure codecov status checks](https://docs.codecov.com/docs/commit-status) with a threshold allowing a certain number of lines to transition from covered to uncovered without reporting a negative patch coverage result.

**To Reproduce**

* Submit a pull request
* Watch codecov/project fail
* Open pull request on Codecov, e.g. https://app.codecov.io/gh/artefactual-sdps/enduro/pull/891/indirect-changes

**Additional context**

* Use `[skip codecov]` to skip codecov.
* https://docs.codecov.com/docs/unexpected-coverage-changes

PR #894, opened 2024-03-15
Introduce coverage thresholds

sevein:
A tentative solution for https://github.com/artefactual-sdps/enduro/issues/892 that updates the codecov config to allow a certain number of lines to transition from covered to uncovered without reporting a negative patch coverage result.


djjuhasz (approved):
LGTM :+1:

Issue #875, opened 2024-03-06
Problem: AM processing poll transfer activity nil pointer error


djjuhasz:
**Describe the bug**

When processing a transfer with AM the workflow fails at "poll-transfer-activity" with a nil pointer error.

**To Reproduce**

1. Configure Enduro for AM preservation
2. Start processing by uploading a valid transfer to Enduro (via MinIO or Tilt upload script)
3. Processing fails with the error:

```
{
  "message": "activity error",
  "source": "GoSDK",
  "stackTrace": "",
  "encodedAttributes": null,
  "cause": {
    "message": "runtime error: invalid memory address or nil pointer dereference",
    "source": "GoSDK",
[...]

PR #895, opened 2024-03-19
Avoid panic reading preservation actions/tasks

jraddaoui:
Transform NULL `started_at` dates to zero value dates on their Goa representations.

Refs https://github.com/artefactual-sdps/enduro/issues/875.


Diogenesoftoronto (approved):
LGTM. Remember to put [skip codecov] in your commit so that it doesn't break ci

PR #896, opened 2024-03-25
Enable tracing in http.Client

sevein:
Register a propagator with OpenTelemetry to enable propagation of trace context over the wire [^1]. Update HTTP clients with `otelhttp.Transport` to wrap outbound requests with a span and enrich it with metrics.

This change enable us to relate spans occuring beyond an individual service, e.g. the activity "move-to-permanent-storage" executed by enduro-a3m-worker uses a HTTP client to request to the Storage Service the long-term storage of the AIP. With this commit we'll start seeing the `HTTP POST` operation performed by the client but also the rest of the details happening on the server side, i.e. the API handler, SQL queries, workflow, activities, etc.

image

[^1]: See the [Trace Context spec](https://www.w3.org/TR/trace-context/) (W3C Recommendation) for more details. It defines HTTP headers and a value format to standardize how context information is propagated over the wire.


internal/persistence/persistence.go
```diff
@@ -29,3 +34,50 @@ type Service interface {
 	CreatePackage(context.Context, *datatypes.Package) error
 	UpdatePackage(context.Context, uint, PackageUpdater) (*datatypes.Package, error)
 }
+
+type wrapper struct {
```
djjuhasz: I'd prefer to see this in a separate file (maybe `persistence/telemetry.go`?) to keep the service definition separate from the telemetry implementation of the service.  When we add a new method to the service interface, it will need to be added to the entclient and the instrumentation wrapper and I think having the wrapper here kind of muddies that requirement.

djjuhasz (approved):
I have one minor change suggestion, but it's not a blocker. :+1:

internal/persistence/persistence.go
sevein: I can do that!

PR #897, opened 2024-03-26
Fix long lines with golines

sevein:
Here's another attempt to fix long lines, this time using [golines]. This tool does not integrate with golangci-lint but it's also not incompatible either, i.e. golangci-lint is not raising issues with code formatted by golines. The result isn't perfect, but seems good enough. While not integrated with the linter/formatter, I'm considering keeping it installed so we can run it manually for now, or perhaps in CI in the future. Thoughts?

Alternatives:
* [golangci-lint/lll](https://golangci-lint.run/usage/linters/#lll) - part of golangci-lint but it doesn't have autofix.
* [gofumpt](https://github.com/mvdan/gofumpt) - the line splitter is an experimental feature not compatible with golangci-lint and it doesn't seem to identify issues in our project.

[golines]: https://github.com/segmentio/golines

### DIFF
--- a/internal/telemetry/traces.go
+++ b/internal/telemetry/traces.go
@@ -17,9 +17,15 @@ import (
 
 // TracerProvider provides Tracers that are used by instrumentation code to
 // trace computational workflows.
-func TracerProvider(ctx context.Context, logger logr.Logger, cfg Config, appName, appVersion string) (trace.TracerProvider, ShutdownProvider, error) {
+func TracerProvider(
+	ctx context.Context,
+	logger logr.Logger,
+	cfg Config,
+	appName, appVersion string,
+) (trace.TracerProvider, ShutdownProvider, error) {
 	if !cfg.Traces.Enabled || cfg.Traces.Address == "" {
-		logger.V(1).Info("Tracing system is disabled.", "enabled", cfg.Traces.Enabled, "addr", cfg.Traces.Address, "sampling-ration", cfg.Traces.SamplingRatio)
+		logger.V(1).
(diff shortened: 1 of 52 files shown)


internal/telemetry/traces.go
```diff
@@ -17,9 +17,15 @@ import (
...
+func TracerProvider(
+	ctx context.Context,
+	logger logr.Logger,
+	cfg Config,
+	appName, appVersion string,
+) (trace.TracerProvider, ShutdownProvider, error) {
 	if !cfg.Traces.Enabled || cfg.Traces.Address == "" {
-		logger.V(1).Info("Tracing system is disabled.", "enabled", cfg.Traces.Enabled, "addr", cfg.Traces.Address, "sampling-ration", cfg.Traces.SamplingRatio)
+		logger.V(1).
+			Info("Tracing system is disabled.", "enabled", cfg.Traces.Enabled, "addr", cfg.Traces.Address, "sampling-ration", cfg.Traces.SamplingRatio)
```
sevein, line 28: This isn't great, but the alternative is to use `golines --nochain-split-dots` which will break this simple logging entry into something like the following:

```go
logger.V(1).
	Info(
		"Starting new workflow",
		"watcher",
		event.WatcherName,
		"bucket",
		event.Bucket,
		"key",
		event.Key,
		"dir",
		event.IsDir,
	)
```

Horrible! I'd rather fix this specific case manually.

djjuhasz: Hmm, in general I like the `--no-chain-split-dots` style better then the default (See: https://github.com/segmentio/golines?tab=readme-ov-file#chained-method-splitting), but I agree that it's bad with these key / value pairs.

djjuhasz (approved):
Overall I'm happy with the golines formatting.  The chained method splitting is unfortunate (as you commented), but I think this is better than no line length formatter. :+1:

PR #898, opened 2024-03-26
Fix user hydration when authenticated

sevein:
After updating to Vue 3.4, the global navigation guard in the router stopped
being able to validate authenticated users during the initial loading of the
application, e.g. after a page refresh. The root issue lies in the `getUser`
success callback in `main.ts` executing too late, i.e. not having run by the
time when the navigation guard was evaluating `isUserValid`. I'm not sure what
is the exact change in Vue that caused this issue, but I suspect it's related
to the updates in the reactivity system.

To reproduce the issue:
1. Open http://localhost:8080 and sign in.
2. Open http://localhost:8080 again from a new tab,
3. You'll be redirected (unexpectedly) to the sign-in page.

The fix proposed consists in loading the user from the layout store using a new
action that is triggered by the navigation guard.

The `App` component now uses `watch` to observe `layoutStore.isUserValid` and
react by establishing the package monitor connection when it's set to true,
effectively decoupling the connection process from the authentication workflow.

### DIFF
--- a/dashboard/src/App.vue
+++ b/dashboard/src/App.vue
@@ -1,10 +1,24 @@
 <script setup lang="ts">
+import { watch } from "vue";
+import { client } from "@/client";
 import { useLayoutStore } from "./stores/layout";
 import Header from "@/components/Header.vue";
 import Sidebar from "@/components/Sidebar.vue";
 import { DialogWrapper } from "vue3-promise-dialog";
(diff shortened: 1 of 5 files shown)


jraddaoui (approved):
Thanks @sevein! I was facing this issue the other day.

PR #904, opened 2024-03-28
Add upload-sample-transfer helper

sevein:
Would this be useful? I see myself all the time running that command locally to test new transfers. I know that the Tilt button exists but often I'm closer to the terminal than to the Tilt dashboard.


djjuhasz (approved):
LGTM :+1:

PR #905, opened 2024-03-28
Enable tracing in the a3m gRPC client

sevein:
Enable OpenTelemetry tracing in the gRPC client and create it from main following recommended practices. I've also added a small test for the a3m activity, but it's improvable.

After checking out the traces in Grafana I realized that we were sending hundreds of Read requests to a3m so I've also introduced a small wait between reads to reduce the frequency.

image


internal/a3m/client.go
```diff
@@ -4,36 +4,34 @@ import (
...
 	conn, err := grpc.DialContext(
 		ctx,
 		addr,
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
-		grpc.WithBlock(),
+		grpc.WithStatsHandler(
+			otelgrpc.NewClientHandler(
+				otelgrpc.WithTracerProvider(tp),
+			),
+		),
```
sevein, line 29: This is the tracing enabler in case that it's unclear.

internal/a3m/a3m.go
```diff
@@ -190,6 +190,7 @@ func savePreservationTasks(
 			PreservationActionID: paID,
 		}
 		pt.StartedAt.Time = job.StartTime.AsTime()
```
djjuhasz: I think this would be easier to read with `StartedAt` assigned in the struct literal:

```go
pt := datatypes.PreservationTask{
	TaskID: job.Id,
	Name:   job.Name,
	Status: jobStatusToPreservationTaskStatus[job.Status],
	StartedAt: sql.NullTime{
		Time:  job.StartTime.AsTime(),
		Valid: true,
	},
	PreservationActionID: paID,
}
```

djjuhasz (approved):
One nitpick, but it's not a blocker. :+1:

PR #921, opened 2024-04-10
Upgrade mholt/archiver from v3 to v4

djjuhasz:
Fixes security vulnerability:
https://github.com/advisories/GHSA-rhh4-rh7c-7r5v

### DIFF
--- a/internal/workflow/activities/unarchive.go
+++ b/internal/workflow/activities/unarchive.go
@@ -2,15 +2,16 @@ package activities
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 
 	"github.com/go-logr/logr"
-	"github.com/mholt/archiver/v3"
-
-	"github.com/artefactual-sdps/enduro/internal/fsutil"
+	"github.com/google/safeopen"
+	"github.com/mholt/archiver/v4"
 )
 
 const UnarchiveActivityName = "unarchive-activity"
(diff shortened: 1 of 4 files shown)


internal/workflow/activities/unarchive.go
```diff
@@ -2,15 +2,16 @@ package activities
...
 	"io/fs"
 	"os"
 	"path/filepath"
 
+	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/go-logr/logr"
-	"github.com/mholt/archiver/v3"
-
-	"github.com/artefactual-sdps/enduro/internal/fsutil"
+	archiver "github.com/mholt/archiver/v4"
```
sevein: I don't think you need the import alias, i.e. v4 will be ignored.

internal/workflow/activities/unarchive.go
```diff
@@ -53,20 +54,35 @@ func (a *UnarchiveActivity) Execute(
...
+		return nil, fmt.Errorf("unarchive: open: %v", err)
 	}
-	if u == nil {
-		// Couldn't find an unarchiver, so this is probably a regular file.
-		// Return the source path unaltered, and IsDir as false.
-		a.logger.V(2).Info("Unarchive: not an archive, skipping.")
-		return &UnarchiveActivityResult{DestPath: params.SourcePath}, nil
+	defer f.Close()
+
+	r := io.Reader(f)
```
sevein: I suspect this line is not needed, i.e. `*os.File` implements `io.Reader` and Go should be doing the conversion implicitly.

sevein (approved):
LGTM. CI has a couple of minor complains.

internal/workflow/activities/unarchive.go
djjuhasz: You're right.  `archiver.Identify` returns an `io.Reader`, so I can't reassign the output to `f` like:
```go
format, f, err := archiver.Identify(params.SourcePath, f)
```

But I can just assign the output reader to a new var `r`:
```go
format, r, err := archiver.Identify(params.SourcePath, f)
```

PR #925, opened 2024-04-15
Use tparallel linter

sevein:
This PR configures the `tparallel` linter which detects a couple of common mistakes when using `t.Parallel()`.

Context:
- https://github.com/moricho/tparallel
- https://brandur.org/fragments/go-prefer-t-cleanup-with-parallel-subtests


djjuhasz (approved):
LGTM :+1:

PR #930, opened 2024-04-24
Add failed SIPs/PIPs buckets

jraddaoui:
If there is an error before the SIP is sent to preservation, copy the
downloaded SIP file to a failed SIPs bucket. If there is an error on
preservation, send a copy of the PIP to a failed PIPs bucket, with all
the transformations made before it was sent to preservation.

- Add failed buckets configuration.
- Use temporal-activities/bucketupload registered with two different
  names for each failed bucket.
- Modify Kubernetes jobs to set up and re-create buckets.

Refs #929.

### DIFF
--- a/cmd/enduro-am-worker/main.go
+++ b/cmd/enduro-am-worker/main.go
@@ -18,6 +18,7 @@ import (
 	"github.com/artefactual-sdps/temporal-activities/archivezip"
 	"github.com/artefactual-sdps/temporal-activities/bagcreate"
 	"github.com/artefactual-sdps/temporal-activities/bagvalidate"
+	"github.com/artefactual-sdps/temporal-activities/bucketupload"
 	"github.com/artefactual-sdps/temporal-activities/removepaths"
 	"github.com/hashicorp/go-cleanhttp"
 	"github.com/jonboulle/clockwork"
@@ -25,6 +26,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/pflag"
 	"go.artefactual.dev/amclient"
+	"go.artefactual.dev/tools/bucket"
 	"go.artefactual.dev/tools/log"
 	temporal_tools "go.artefactual.dev/tools/temporal"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace"
@@ -200,6 +202,22 @@ func main() {
 		}
 	}()
 
+	// Set-up failed SIPs bucket.
+	failedSIPs, err := bucket.NewWithConfig(ctx, &cfg.FailedSIPs)
+	if err != nil {
+		logger.Error(err, "Error setting up failed SIPs bucket.")
+		os.Exit(1)
+	}
+	defer failedSIPs.Close()
+
+	// Set-up failed PIPs bucket.
+	failedPIPs, err := bucket.NewWithConfig(ctx, &cfg.FailedPIPs)
+	if err != nil {
+		logger.Error(err, "Error setting up failed PIPs bucket.")
+		os.Exit(1)
+	}
+	defer failedPIPs.Close()
+
 	var g run.Group
 
 	// Activity worker.
@@ -302,6 +320,14 @@ func main() {
 		w.RegisterActivityWithOptions(
 			removepaths.New().Execute,
 			temporalsdk_activity.RegisterOptions{Name: removepaths.Name},
+		)
+		w.RegisterActivityWithOptions(
+			bucketupload.New(failedSIPs).Execute,
+			temporalsdk_activity.RegisterOptions{Name: activities.SendToFailedSIPsName},
+		)
+		w.RegisterActivityWithOptions(
+			bucketupload.New(failedPIPs).Execute,
+			temporalsdk_activity.RegisterOptions{Name: activities.SendToFailedPIPsName},
 		)
 
 		g.Add(
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -18,6 +18,7 @@ import (
 	"github.com/artefactual-sdps/temporal-activities/archivezip"
 	"github.com/artefactual-sdps/temporal-activities/bagcreate"
 	"github.com/artefactual-sdps/temporal-activities/bagvalidate"
+	"github.com/artefactual-sdps/temporal-activities/bucketupload"
 	"github.com/artefactual-sdps/temporal-activities/removepaths"
 	"github.com/google/uuid"
 	"go.artefactual.dev/tools/ref"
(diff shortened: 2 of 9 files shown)


cmd/enduro-am-worker/main.go
```diff
@@ -172,6 +174,36 @@ func main() {
...
+		fl, err := storage.NewInternalLocation(&cfg.FailedTransfers)
+		if err != nil {
+			logger.Error(err, "Error setting up failed transfers location.")
+			os.Exit(1)
+		}
+		ft, err = fl.OpenBucket(ctx)
+		if err != nil {
+			logger.Error(err, "Error getting failed transfers bucket.")
+			os.Exit(1)
+		}
```
sevein, line 189: Remember to close them if err == nil!

internal/workflow/processing_test.go
```diff
@@ -1061,3 +1072,400 @@ func (s *ProcessingWorkflowTestSuite) TestPreprocessingChildWorkflow() {
 	s.True(s.env.IsWorkflowCompleted())
 	s.NoError(s.env.GetWorkflowResult(nil))
 }
+
```
jraddaoui, line 1075: I tried to parameterize these tests, just using a for loop and calling `s.SetupWorkflowTest(cfg)` on each loop. I think that may be enough, but it makes it hard to debug the tests individually.

Since the duplication in these workflow tests is already an issue, it may be better that we try to improve it all together in a different PR.

internal/workflow/processing.go
```diff
@@ -110,6 +111,13 @@ type TransferInfo struct {
 	// It is populated by the workflow request.
 	GlobalTaskQueue       string
 	PreservationTaskQueue string
+
+	// Send to failed variables used to keep track of the SIP
+	// location, if it requires zipping (a3m PIP) and what activity
+	// needs to be called to be uploaded to the expected bucket.
+	SendToFailedPath         string
```
djjuhasz: I think it would be better to try and make this field more general to try and avoid variable proliferation. Maybe something like "WorkingDir" to indicate where the current working version of the package is located?

internal/workflow/processing.go
```diff
@@ -110,6 +111,13 @@ type TransferInfo struct {
...
+	SendToFailedActivityName string
```
djjuhasz: It's not necessary in this PR, but I think in the future we'd benefit from breaking `tinfo` into a couple of smaller structs: `sip` for metadata about the original package submitted to Enduro, and `pip` for the modified package metadata.  The pip should be linked back to the sip somehow (by UUID?).  I think `tinfo` is doing too much right now and doesn't distinguish between fields we need early (e.g. the bucket and key of the original SIP) from those we need later (e.g. the current location of the PIP).

internal/workflow/processing.go
```diff
@@ -110,6 +111,13 @@ type TransferInfo struct {
...
+	SendToFailedNeedsZipping bool
```
djjuhasz: Again, I think it would be better to make this field more general, like "IsZipped".  The cleanup code can then decide if the SIP/PIP needs to be zipped based on the "IsZipped" field.

internal/workflow/processing.go
```diff
@@ -327,9 +335,43 @@ func (w *ProcessingWorkflow) SessionHandler(
 	sessCtx temporalsdk_workflow.Context,
 	attempt int,
 	tinfo *TransferInfo,
-) error {
+) (e error) {
 	var cleanup cleanupRegistry
-	defer w.sessionCleanup(sessCtx, &cleanup)
+	defer func() {
+		if e != nil && tinfo.SendToFailedActivityName != "" && tinfo.SendToFailedPath != "" {
```
djjuhasz: I'd like to see this all moved into two named methods to avoid the cognitive load of reading all this logic in-line in the workflow. E.g. "zipPackage()" (would be no-op if it's already zipped) and "moveFailedPackage()".

internal/workflow/processing.go
```diff
@@ -327,9 +335,43 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+					Key:        fmt.Sprintf("Failed_%s", tinfo.req.Key),
+					BufferSize: 100_000_000,
+				},
+			).Get(activityOpts, &sendToFailedResult)
+			if err != nil {
+				e = errors.Join(e, err)
+			}
+		}
+
+		w.sessionCleanup(sessCtx, &cleanup)
```
djjuhasz, line 350: Should we be deleting a failed SIP from the "sip" bucket after uploading a copy to the "failed SIPs" bucket?  Should we be moving the SIP *and* PIP to failed buckets if the workflow fails after PIP creation?

djjuhasz (approved):
Thanks @jraddaoui, this looks good to me and I don't see any blockers.  I did make a few comments about how we might want to manage `tinfo` going forward.  I also had a couple of questions about how we manage SIPs in the "sips" bucket.

internal/workflow/processing.go
djjuhasz: Should we be removing SIPs from the "sips" bucket when the workflow succeeds too?  I can see "sips" getting very full over time. :thinking:

internal/workflow/processing.go
jraddaoui: But that will change over the workflow and I want to upload the original downloaded SIP (for the failed SIP).

internal/workflow/processing.go
jraddaoui: I agree, I thought about moving the three vars to a new struct, but didn't see a lot of benefits without a bigger restructure.

internal/workflow/processing.go
jraddaoui: We have an `isDir` variable that I initially re-used for this. The issue was in preprocessing, where `isDir` is set to true even if it "fails". When preprocessing fails, we still want to upload the original so I ended with an ugly condition to avoid zipping in that case.

It's important to differentiate the current status of the package vs. the moment we want to use to upload to the failed bucket, that's why I added the new variables.

internal/workflow/processing.go
jraddaoui: Okay, would a single method be better? That would be even less cognitive load, no need to know if it requires zipping or not.

internal/workflow/processing.go
jraddaoui: Good questions about the failed scenarios, I'll bring that to the team as follow up work to discuss. On success, we have a delete originals activity that deletes the original SIP. It's based on the retention period configuration from the watcher, I think we just don't set that config in the dev. env. so they never get deleted.

internal/workflow/processing.go
jraddaoui: In the end I did move them to their own struct, something that I could pass directly to the new method.

internal/workflow/processing.go
jraddaoui: Addressed in [668d564](https://github.com/artefactual-sdps/enduro/pull/930/commits/668d5642becfe00656bf02171f9fee91c3a864d4).

internal/workflow/processing_test.go
djjuhasz: Yeah, the processing workflow tests are long and repetitive.

djjuhasz (approved):
Thanks @jraddaoui, LGTM. :+1:

internal/workflow/processing.go
djjuhasz: Nice! :)

internal/workflow/processing.go
djjuhasz: Thank you! :grin:

PR #932, opened 2024-04-30
Use temporal-activities archive extractor

djjuhasz:
Use github.com/artefactual-sdps/temporal-activities/archive to extract archive contents instead of custom unarchive activity.


internal/config/config.go
```diff
@@ -35,6 +36,7 @@ type Configuration struct {
 	A3m           a3m.Config
 	AM            am.Config
 	API           api.Config
+	Archive       archive.Config
```
jraddaoui: Can we add a configuration example (even if it is with the defaults) to `enduro.toml`? I think a section called `[archive]` may not be clear enough, so we may want to change this field to `ExtractArchiveActivity` or something like that.

internal/workflow/processing.go
```diff
@@ -328,22 +335,26 @@ func (w *ProcessingWorkflow) SessionHandler(
...
-				StripTopLevelDir: tinfo.req.StripTopLevelDir,
-			},
+			archive.ExtractActivityName,
+			&archive.ExtractActivityParams{SourcePath: tinfo.TempPath},
 		).Get(activityOpts, &result)
 		if err != nil {
-			return err
+			switch err {
+			case archive.ErrInvalidArchive:
+				break // Not an archive file, bundle the source file as-is.
```
jraddaoui: The `break` is implied in Go.

internal/workflow/processing.go
```diff
@@ -328,22 +335,26 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+			default:
+				return temporal_tools.NewNonRetryableError(err)
+			}
+		}
+		if err == nil {
```
jraddaoui: `else`.

jraddaoui (changes requested):
Looking great @djjuhasz! Just a minor concern about the configuration.

jraddaoui (approved):
Thanks @djjuhasz!

djjuhasz:
> Thanks @djjuhasz!

Oops, didn't see your approval. Thanks @jraddaoui! :grin:

PR #939, opened 2024-05-08
Enable Archivematica Storage Service AIP downloads 

### LINKED ISSUE #820: Problem: AIPs do not download via the Archivematica Storage Service

djjuhasz:
Fixes #820.

Enables the Enduro Dashboard "download" button to download AIPs from the Archivematica Storage Service.


djjuhasz:
Testing this requires adding a few new fields to your Archivematica secrets file:

```
amss_url=https://enduro-temporal.ss.analyst.archivematica.net:8000
amss_user=analyst
amss_api_key=[your_key_here]
```

sevein (approved):
I haven't tested yet but it looks good to me. There are many layers involved but you have figured it all out. It's good to see ssblob finally put to use.

internal/storage/persistence/ent/client/client_test.go
```diff
@@ -61,25 +61,90 @@ func setUpClient(t *testing.T) (*db.Client, *client.Client) {
...
+		wantErr string
+	}
+
+	for _, tt := range []test{
+		{
+			name: "Creates a package with minimal data",
+			params: &goastorage.Package{
+				Name:      "test_package",
+				AipID:     uuid.MustParse("488c64cc-d89b-4916-9131-c94152dfb12e"),
+				ObjectKey: uuid.MustParse("e2630293-a714-4787-ab6d-e68254a6fb6a"),
```
jraddaoui: Could parse these UUIDs only once.

internal/storage/persistence/migrations/20240507234842_location_source_amss.down.sql
```diff
@@ -0,0 +1,2 @@
+-- reverse: modify "location" table
+ALTER TABLE `location` MODIFY COLUMN `source` enum('unspecified','minio','sftp') NOT NULL;
```
jraddaoui, line 2: Didn't recall we were adding down migrations here. It's the same approach as the previous migrations but they could easily break downgrading existing DBs.

internal/storage/service_test.go
```diff
@@ -353,6 +355,96 @@ func TestServiceSubmit(t *testing.T) {
...
+
+		attrs.persistenceMock.
+			EXPECT().
+			CreatePackage(
+				mockutil.Context(),
+				&goastorage.Package{
+					Name:       name,
+					AipID:      uuid.MustParse(aipID),
+					Status:     status,
+					ObjectKey:  uuid.MustParse(objKey),
```
jraddaoui: Same, we could parse only once.

internal/storage/types/location_config.go
```diff
@@ -27,7 +27,7 @@ type configTypes struct {
 	S3         *S3Config   `json:"s3,omitempty"`
 	SFTPConfig *SFTPConfig `json:"sftp,omitempty"`
 	URLConfig  *URLConfig  `json:"url,omitempty"`
-	SSConfig   *SSConfig   `json:"ss,omitempty"`
+	SSConfig   *AMSSConfig `json:"amss,omitempty"`
```
jraddaoui: Do we want to change the key too?

internal/workflow/activities/storage.go
```diff
@@ -2,15 +2,86 @@ package activities
...
+type CreateStoragePackageActivity struct {
+	client storage.Client
+}
+
+type CreateStoragePackageActivityParams struct {
+	Name       string
+	AIPID      string
+	ObjectKey  string
+	Status     string
+	LocationID string
```
jraddaoui: Should we reuse the `goastorage.CreatePayload` struct and do the transformation from strings outside?

internal/workflow/activities/storage_test.go
```diff
@@ -0,0 +1,159 @@
...
+			},
+			mockCalls: func(m *storage_fake.MockClientMockRecorder) {
+				m.Create(
+					mockutil.Context(),
+					&goastorage.CreatePayload{
+						Name:       "Package 1",
+						AipID:      "88e5a5fc-5c74-40e8-aa50-d6542b45f251",
+						ObjectKey:  "3776a8f2-5ad5-4b4e-80c7-a888d9229cc1",
+						Status:     "stored",
+						LocationID: ref.New(uuid.MustParse("92c3934d-8911-4ce4-8a44-ffa72a0b5720")),
```
jraddaoui: Same. I doubt it's noticeable, but it could be at some point with so many tests parsing UUIDs all the time.

internal/workflow/processing.go
```diff
@@ -809,6 +809,40 @@ func (w *ProcessingWorkflow) transferAM(sessCtx temporalsdk_workflow.Context, ti
...
 
+	// Set package location
+	{
+		ctx := withLocalActivityOpts(sessCtx)
+		err := temporalsdk_workflow.ExecuteLocalActivity(
+			ctx,
+			setLocationIDLocalActivity,
+			w.pkgsvc,
+			tinfo.req.PackageID,
+			uuid.MustParse(w.cfg.AM.AMSSLocationID),
```
jraddaoui: Brave.

jraddaoui (changes requested):
As @sevein said, nice work figuring all this @djjuhasz!

Some minor comments below, feel free to ignore. My only request would be to include the comment you added here in the "Working with Archivematica" readme.

internal/workflow/activities/storage.go
jraddaoui: It's probably better as it is right now, ignore me.

internal/workflow/processing.go
djjuhasz: Yeah, that's a bad panic waiting to happen. :sweat:

sevein: What would Viper do if you define AMSSLocationID as a uuid.UUID? If undefined, you can always check for `uuid.Nil`.

sevein: Oh forget it, you'd need to pass a custom decoder func to viper, see https://pkg.go.dev/github.com/spf13/viper#DecoderConfigOption.

https://github.com/spf13/viper/blob/ea35b92596da4253bd5e5514a7ce363de374bd63/viper.go#L1159-L1162

jraddaoui: I thought you were being brave because that config already passed some validation. If not, we shouldn't be that brave ;)

sevein: Look, found one!

https://github.com/go-saas/kit/blob/8e55a6f58fa1e5f3ae8d7aeff025b38f8fed8a93/pkg/mapstructure/mapstructure.go#L10-L21

internal/storage/persistence/migrations/20240507234842_location_source_amss.down.sql
djjuhasz: I think the db migrations would benefit from a second look and maybe some changes.  I've never fully understood how the ent/atlas db migration system works, but it looks to me like https://github.com/artefactual-sdps/enduro/blob/main/cmd/migrate/main.go is pretty close to the example provide by ent at https://entgo.io/docs/versioned-migrations.

I've been reading the ent and Atlas docs, but I can't figure out if Atlas generates "down" migration files by default, or if this is something that ent migrations do by default.  Looking at the code I don't see any flags that we are using that would enable generation of down migration files. :confused:

djjuhasz: https://atlasgo.io/blog/2024/04/01/migrate-down is an interesting read.  My main takeaway is that the atlas down migrations are mainly intended for use in development due to their destructive nature.

jraddaoui: I see a few differences between this migration script and the ones we have in other projects (where the down migrations are not created). Most important ones, the usage of `sqltool.NewGolangMigrateDir(path)` instead of `atlas.NewLocalDir(migrationsDir)` and the addition of `schema.WithFormatter(atlas.DefaultFormatter)` to the migrate options, maybe those have something to do. Not a big deal in any case.

internal/workflow/processing.go
djjuhasz: Okay, I implemented https://github.com/go-saas/kit/blob/8e55a6f58fa1e5f3ae8d7aeff025b38f8fed8a93/pkg/mapstructure/mapstructure.go#L10-L21 with viper to allow decoding AMSSLocationID to a `uuid.UUID`.  It definitely simplifies the type checking inside the app. :)

It was also a good reminder that viper decodes comma-separated strings to a slice by default. It may be better if I  update https://github.com/artefactual-sdps/temporal-activities/blob/main/removefiles/activity.go#L20 to a slice instead of string. :thinking:

jraddaoui:
Thinking about this and the deployment implications @djjuhasz. Should we try to make everything configurable? Maybe we could add the AMSS location information to a new [am.ss] config section and check its existence on startup, creating it if it doesn't exists. Otherwise, the creation of that location directly in the DB has to be considered on the deployment, while all the other AM options are part of the configuration.

djjuhasz:
> Thinking about this and the deployment implications @djjuhasz. Should we try to make everything configurable? Maybe we could add the AMSS location information to a new [am.ss] config section and check its existence on startup, creating it if it doesn't exists. Otherwise, the creation of that location directly in the DB has to be considered on the deployment, while all the other AM options are part of the configuration.

@jraddaoui I think the downside of putting the full AMSS location info in the config file is then we have the data in two places: in the config, and in the enduro_storage database.  Once the AMSS location data is added to the database the data in the config file  becomes superfluous. 

We need to have the AMSS location data in the database for the Enduro Dashboard "Download AIP" link to work.  I think for now it's better to stick with adding the AMSS location data to the enduro_storage database in deployment, like we are doing in the dev environment.  If it's a hassle to add the data directly with SQL, we could create a go cmd script to add locations to the database.

I don't think it's ideal to have the AMSSLocationID in the config file really - we may want to consider adding it (along with other config information) to the persistence layer in the future.

jraddaoui:
> @jraddaoui I think the downside of putting the full AMSS location info in the config file is then we have the data in two places: in the config, and in the enduro_storage database. Once the AMSS location data is added to the database the data in the config file becomes superfluous.

I don't think that's an issue, but I can see them updating the configuration at some point. It's not that i like this solution either, just having a lot of conflicts about how tied are those config options in this case. And also about how the entire storage service (from Enduro) works.

> We need to have the AMSS location data in the database for the Enduro Dashboard "Download AIP" link to work. I think for now it's better to stick with adding the AMSS location data to the enduro_storage database in deployment, like we are doing in the dev environment. If it's a hassle to add the data directly with SQL, we could create a go cmd script to add locations to the database.

I think it could also be done through the API, but I may be wrong.

> I don't think it's ideal to have the AMSSLocationID in the config file really - we may want to consider adding it (along with other config information) to the persistence layer in the future.

I think they are too tied together but then our services are supposed to work independently, that's why I'm having these conflicts. Lets go with what you have done for now, hopefully we can make it clearer/easier in the future.

hack/kube/overlays/dev-am/mysql-create-amss-location-job.yaml
```diff
@@ -0,0 +1,49 @@
...
+                  name: enduro-am-secret
+                  key: amss_api_key
+          command:
+            - sh
+            - -c
+            - >
+              LOC_AMSS='{"amss": {"url": "'$AMSS_URL'", "api_key": "'$AMSS_API_KEY'", "username": "'$AMSS_USER'"}}' &&
+              mysql -h mysql.enduro-sdps -u $MYSQL_USER -p$MYSQL_PASSWORD --execute "
+                INSERT IGNORE INTO enduro_storage.location (name, description, source, purpose, uuid, created_at, config)
+                VALUES ('amss', '', 'amss', 'aip_store', 'e0ed8b2a-8ae2-4546-b5d8-f0090919df04', NOW(), '$LOC_AMSS');
```
jraddaoui: Can we get the location UUID from the secret too? To make it match the one from the config directly in this environment.

internal/config/config_test.go
```diff
@@ -18,10 +19,17 @@ debugListen = "127.0.0.1:9001"
 
 [temporal]
 address = "host:port"
+
+[am]
+AMSSLocationID = "2d3aba70-035d-43f5-8d73-c97f35237006"
```
jraddaoui: Nitpick: using `amssLocationId` in the config file.

docs/src/dev-manual/archivematica.md
```diff
@@ -76,6 +80,9 @@ AM API and SFTP configuration:
     sftp_user=archivematica
     sftp_remote_dir=/enduro_transfers
     sftp_private_key_passphrase=
+    amss_url=http://host.k3d.internal:62081
+    amss_user=test
+    amss_api_key=secret
```
jraddaoui, line 86: If we add the `amss_location_id` to the secret and use it in the environment for the enduro services and the create location job, we could be sure they match in this environment.

docs/src/dev-manual/archivematica.md
```diff
@@ -108,5 +115,16 @@ Preservation system value needed for workflow to be Archvimatica specific:
...
+
+    [am]
+    amssLocationId = "e0ed8b2a-8ae2-4546-b5d8-f0090919df04"
+
+!!! note
+
+    The amssLocationId UUID must match the AMSS location ID in the
+    *enduro_storage* database — it does *not* need to match the actual location
+    UUID in the AM Storage Service.
+
```
jraddaoui: Not sure if we would want to remove this note after using the secret to configure the `amss_location_id`.

jraddaoui (changes requested):
Thanks @djjuhasz!

hack/kube/overlays/dev-am/mysql-create-amss-location-job.yaml
djjuhasz: Yeah, good idea :+1:

djjuhasz: @jraddaoui I decided to make the default storage location configurable instead of having a separate AMSS location config variable.  For the AM dev environment I just override the config variable to use the AMSS location. Check out b13439c and see what you think.

I'll update the docs to match the changes in b13439c if we decide to go with it.

hack/kube/overlays/dev-am/enduro-am.yaml
```diff
@@ -55,6 +55,11 @@ spec:
               value: $(MYSQL_USER):$(MYSQL_PASSWORD)@tcp(mysql.enduro-sdps:3306)/enduro
             - name: ENDURO_STORAGE_DATABASE_DSN
               value: $(MYSQL_USER):$(MYSQL_PASSWORD)@tcp(mysql.enduro-sdps:3306)/enduro_storage
+            - name: ENDURO_STORAGE_DEFAULTPERMANENTLOCATIONID
```
jraddaoui, line 58: Nitpick: I like `ENDURO_STORAGE_DEFAULT_PERMANENT_LOCATION_ID` better.

internal/workflow/processing.go
```diff
@@ -809,6 +809,40 @@ func (w *ProcessingWorkflow) transferAM(sessCtx temporalsdk_workflow.Context, ti
...
 
+	// Set package location
+	{
+		ctx := withLocalActivityOpts(sessCtx)
+		err := temporalsdk_workflow.ExecuteLocalActivity(
+			ctx,
+			setLocationIDLocalActivity,
+			w.pkgsvc,
+			tinfo.req.PackageID,
+			*tinfo.req.DefaultPermanentLocationID,
```
jraddaoui: We should make sure it's not a nil pointer. I'd consider not having it as a pointer in the ProcessingWorkflowRequest but that may be a backwards incompatible change. Also, it's a shame this activity is not expecting a params struct.

internal/workflow/processing.go
```diff
@@ -809,6 +809,40 @@ func (w *ProcessingWorkflow) transferAM(sessCtx temporalsdk_workflow.Context, ti
...
+	{
+		activityOpts := withLocalActivityOpts(sessCtx)
+		err := temporalsdk_workflow.ExecuteActivity(
+			activityOpts,
+			activities.CreateStoragePackageActivityName,
+			&activities.CreateStoragePackageActivityParams{
+				Name:       tinfo.req.Key,
+				AIPID:      tinfo.SIPID,
+				ObjectKey:  tinfo.SIPID,
+				LocationID: *tinfo.req.DefaultPermanentLocationID,
```
jraddaoui: Similar here.

hack/kube/overlays/dev-am/mysql-create-amss-location-job.yaml
jraddaoui: Neat!

jraddaoui (commented):
Thanks @djjuhasz!

I like that approach for now until we make the locations really configurable.

hack/kube/overlays/dev-am/enduro-am.yaml
djjuhasz: @jraddaoui me too, but viper won't accept it. :(

jraddaoui (approved):
Thanks @djjuhasz, great work!

PR #942, opened 2024-05-14
Revert "Run Dashboard container as non-root user"

sbreker:
This reverts commit a9d53698493df45c1f5b0daf4050484c32d05d65.

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -1,0 +1,2 @@
+# Need to revisit this issue in enduro-dashboard.
+AVD-DS-0002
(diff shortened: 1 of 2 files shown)


djjuhasz:
@sbreker there is a nginx "unprivileged" docker image we could use instead of the default "nginx:alpine" image: https://hub.docker.com/r/nginxinc/nginx-unprivileged

Alternatively the default nginx docker image can be run as non-root with some config changes: https://hub.docker.com/_/nginx "Running nginx as a non-root user"

djjuhasz (approved):
LGTM :+1:

PR #943, opened 2024-05-17
WIP: Add cluster control to makefile

mcantelon:
Just an idea... I roughed this up, based on the docs and internal recommendations, in case folks think it'd be useful to add.

Adding a RAM recommendation to it (I don't see one in the docs) could be good too so devs/potential devs don't end up running it with too little RAM (I've got 32GB and it seems to be working fine).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,3 +237,31 @@ upload-sample-transfer:
 workflowcheck: # @HELP Detect non-determinism in workflow functions.
 workflowcheck: $(WORKFLOWCHECK)
 	workflowcheck ./...
+
+cluster-start: # @HELP Start cluster.
+	k3d cluster start sdps-local
+	tilt up
+
+cluster-stop: # @HELP Stop cluster.
+	k3d cluster stop sdps-local
+	tilt down
+
+cluster-rebuild: # @HELP Rebuild cluster.
+	k3d cluster stop sdps-local
+	tilt down
+	kubectl config set-context --current --namespace sdps-local
+	echo
+	echo "*******************************************************************************"
+	echo "After deleting the volumes/cluster you'll see the status of the core pods being"
+	echo "created."
+	echo
+	echo "After all pods are running or completed, press CTRL-C and start the cluster."
+	echo "****************************************************************************************"
+	echo
+	echo "Press [enter] to continue..."
+	read
+	kubectl delete pvc,pv --all
+	k3d cluster delete sdps-local
+	docker system prune --all
+	k3d cluster create sdps-local --registry-create sdps-registry
+	watch kubectl get pod -A


Makefile
```diff
@@ -237,3 +237,31 @@ upload-sample-transfer:
...
+	echo "created."
+	echo
+	echo "After all pods are running or completed, press CTRL-C and start the cluster."
+	echo "****************************************************************************************"
+	echo
+	echo "Press [enter] to continue..."
+	read
+	kubectl delete pvc,pv --all
+	k3d cluster delete sdps-local
+	docker system prune --all
```
sevein, line 265: > docker system prune --all

What do we gain from this in this context? It looks like this is something that could impact other dev environments on my machine.

sevein (commented):
This seems like it could be useful! Perhaps the user experience could be improved by introducing the notion of idempotency? For example:

- `cluster` - Ensures that the cluster is created and running. This is the first thing you would do when bringing up the environment for the first time or after being already created. If you run it twice, the second time should be a no-op.
- `cluster-destroy`- Ensures that the cluster is destroyed (without failing if it doesn't exist).
- `cluster-stop` - Ensures that the cluster is stopped (without failing if it's already stopped or doesn't exist).

On top of that:

```
cluster-rebuild: cluster-destroy cluster-ensure # @HELP Rebuild the cluster.
	@echo "Rebuilding the cluster complete."
```

Not sure if that's a lot to ask or whether it'd be more sensible to implement it in `hack/cluster.sh` as a bash script and have these targets become simple wrappers.
In any case, this is probably a good start, we can improve it later.

Makefile
```diff
@@ -237,3 +237,31 @@ upload-sample-transfer:
...
 workflowcheck: $(WORKFLOWCHECK)
 	workflowcheck ./...
+
+cluster-start: # @HELP Start cluster.
+	k3d cluster start sdps-local
+	tilt up
+
+cluster-stop: # @HELP Stop cluster.
+	k3d cluster stop sdps-local
+	tilt down
```
jraddaoui, line 247: `tilt down` is not going to work with the cluster stopped. I also like to keep the cluster as it was between restarts, so I rarely run `tilt down` in general.

Makefile
```diff
@@ -237,3 +237,31 @@ upload-sample-transfer:
...
+
+cluster-rebuild: # @HELP Rebuild cluster.
+	k3d cluster stop sdps-local
+	tilt down
```
jraddaoui, line 251: Same here, it's also not needed if you are going to delete the cluster after.

Makefile
```diff
@@ -237,3 +237,31 @@ upload-sample-transfer:
...
+	echo "*******************************************************************************"
+	echo "After deleting the volumes/cluster you'll see the status of the core pods being"
+	echo "created."
+	echo
+	echo "After all pods are running or completed, press CTRL-C and start the cluster."
+	echo "****************************************************************************************"
+	echo
+	echo "Press [enter] to continue..."
+	read
+	kubectl delete pvc,pv --all
```
jraddaoui, line 263: No need to to this if you are going to delete the entire cluster.

Makefile
```diff
@@ -237,3 +237,31 @@ upload-sample-transfer:
...
+	k3d cluster delete sdps-local
+	docker system prune --all
+	k3d cluster create sdps-local --registry-create sdps-registry
```
jraddaoui, line 266: I like to use the same registry for all the clusters as many projects share the same images/layers, you could use `--registry-use` instead. Because it's a rebuild it should exist, but we are missing `cluster-create` and `cluster-delete` target rules to make sure the names match. It can get tricky if you think in all the possible states of the cluster and registry.

jraddaoui (changes requested):
Hi @mcantelon!

I'm not sure this adds a lot of value, my cluster and registry are called differently, I could easily change that to match this and the docs, but see my comments below for other reasons. I'm also happy to add it it if you think it's useful, but working with `k3d` and `kubectl` from the terminal is going to happen eventually.

Issue #944, opened 2024-05-21
Problem: the API logs report recovered panics


jraddaoui:
**Describe the bug**

Reported by @scollazo in Slack:

```

[...]

PR #945, opened 2024-05-24
Fix API monitor endpoint with auth disabled

jraddaoui:
Refs https://github.com/artefactual-sdps/enduro/issues/944.


sevein (approved):
LGTM

PR #946, opened 2024-05-30
Remove working dirs when worker session completes

djjuhasz:


jraddaoui (approved):
Nice! Thanks @djjuhasz!

PR #948, opened 2024-05-31
Move to github.com/go-jose/go-jose/v3

jraddaoui:
Fixes a vulnerability reported in the unmantained version we were using.

Refs https://github.com/artefactual-sdps/enduro/security/dependabot/52.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,16 +7,17 @@ require (
 	ariga.io/sqlcomment v0.1.0
 	buf.build/gen/go/artefactual/a3m/grpc/go v1.3.0-20230508184533-2e9432075630.2
 	buf.build/gen/go/artefactual/a3m/protocolbuffers/go v1.31.0-20230508184533-2e9432075630.2
-	chainguard.dev/go-oidctest v0.2.0
+	chainguard.dev/go-oidctest v0.3.1
 	entgo.io/ent v0.13.1
 	github.com/XSAM/otelsql v0.29.0
 	github.com/alicebob/miniredis/v2 v2.32.1
 	github.com/artefactual-sdps/temporal-activities v0.0.0-20240530153613-e610eaf238ed
-	github.com/coreos/go-oidc/v3 v3.9.0
+	github.com/coreos/go-oidc/v3 v3.10.0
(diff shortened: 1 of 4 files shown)


djjuhasz (commented):
@jraddaoui I don't understand the why the following modules were updated:

chainguard.dev/go-oidctest v0.2.0 -> v0.3.1
github.com/coreos/go-oidc/v3 v3.9.0 -> v3.10.0

Were they using an old version of github.com/go-jose/go-jose too?  I would expect this to be reflected in the indirect dependencies, but that updates to v4:

github.com/go-jose/go-jose/v3 v3.0.3 -> github.com/go-jose/go-jose/v4 v4.0.2 // indirect

:confused:

jraddaoui:
I upgraded chainguard.dev/go-oidctest manually to be able to use github.com/go-jose/go-jose/v3 in the test, that probably upgraded the indirect dependency. I tried with github.com/go-jose/go-jose/v4, but I could not make it work easily. I'll give it another try ;)

jraddaoui:
The signer created by `oidctest.NewIssuer(t)` is a v3 version, which can't be used with v4:

```
cannot use signer (variable of type "github.com/go-jose/go-jose/v3".Signer) as "github.com/go-jose/go-jose/v4".Signer value in argument to token: "github.com/go-jose/go-jose/v3".Signer does not implement "github.com/go-jose/go-jose/v4".Signer (wrong type for method Options)
		have Options() "github.com/go-jose/go-jose/v3".SignerOptions
		want Options() "github.com/go-jose/go-jose/v4".SignerOptionscompiler[InvalidIfaceAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#InvalidIfaceAssign)
```

`oidctest` seems to be working with v3, but it also imports github.com/coreos/go-oidc/v3, which is the one indirectly importing v4: https://github.com/chainguard-dev/go-oidctest/blob/v0.3.1/go.mod.

I'd not worry about it @djjuhasz.

djjuhasz (approved):
Okay, good enough for me. :)

Issue #949, opened 2024-05-31
Problem: child workflow tasks aren't visible in the UI


sallain:
**Is your feature request related to a problem? Please describe.**

Users can develop a child workflow to run a custom workflow on their SIPs. For example, a client may want to run a validation check to ensure that the SIP complies with their local SIP specification, or ensure that all of the file formats in the SIP are permitted. At present, none of this information appears in the Enduro user interface - meaning that the user has no visibility into whether the workflow was actually run. If a SIP fails one of the checks, it is not easy for operators to determine why the workflow halts, unless they go digging in the Temporal logs for more information.

**Describe the solution you'd like**

The goal is to make custom task information and outcomes visible in the Enduro UI. We have a client with this need (SFA) so we can use their child workflow as an example implementation, but this will be a constant need for any Enduro client/user with custom child workflow activities - meaning that ideally, we define some reusable abstractions in how we do this so it is easy to replicate for future custom tasks.

For SFA, a breakdown of the suggested task names, success messages (to be added to the Notes col), and failure outputs, can be found [here in the SFA Miro board](https://miro.com/app/board/uXjVMlYiVgs=/?moveToWidget=3458764589906174754&cot=14)

NOTE: there are two additional proposed pre-ingest tasks there (in the blue rows) that do not currently exist in PoC#1, and are not part of SFA pre-ingest validation. These will be described further in another card, but this card should be used for any work to add them to the Enduro UI.

There is also one task in yellow, that actually happens at the end of ingest / during post-ingest - the creation and delivery of the metadata bundle for delivery to the AIS. The linked table above also includes suggested UI info for that custom activity as well.

Finally, since we will want to reuse this pattern, we should add some developer notes - perhaps to the pre-processing base repo's README, or even an internal doc for now - on how other developers can reuse the pattern for future custom workflow activities.

**Describe alternatives you've considered**

None

**Additional context**

We have a related open issue, #934, about the Notes column, and adding more feedback and information to the Enduro UI. There is still some discussion about how to implement these in a way that is user-friendly and accessible.

djjuhasz:
I'm implementing this by adding Enduro API endpoints to add preservation actions and preservation tasks, then I'll POST to those endpoints from pre-processing to add the pre-processing tasks to the Enduro database.

I'll do this work as a set of smaller pull requests:

- [ ] Add a "Create preservation action" API endpoint to Enduro (#952)
- [ ] Add a "Create preservation task" API endpoint to Enduro
[...]

PR #952, opened 2024-06-01
Add a "create preservation action" API endpoint

djjuhasz:


djjuhasz:
Adding the API endpoint is no longer required to address issue #949.

Issue #957, opened 2024-06-05
Problem: API lacks an authorization implementation


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Enduro's API has optional authentication using OpenID Connect (OIDC), but there is no mechanism in place to control access to resources based on user roles or permissions.

**Describe the solution you'd like**

I looked at different options to provide this functionally:

- Access Control Lists (ACLs)
- Role-Based Access Control (RBAC)
- Attribute-Based Access Control (ABAC)
- Policy-Based Access Control (PBAC)

ACLs allows granular control over each resource, but it has scalability issues as the application grows. Like RBAC, they won't provide a flexible and fine-grained access control that will allow us to manage complex access needs dynamically as the application grows.

Both ABAC and PBAC grant or deny access based on attributes like user roles, resource types, operations, etc., evaluating them through defined policies. This attributes can have a hierarchical organization, policy inheritance and delegated administration (as the system and organization evolves). While PBAC provides comprehensive policy management, it can be implemented following ABAC when more defined policies and integration with external decision points are needed.

As an initial implementation or PoC, I'd like to see how Enduro is capable of getting a list of attributes from a configurable claim in the OIDC access token and limit user actions based on them. For example, considering the following actions (API-endpoints/UI-views):

- List packages
- Read package
- Download package

And the following attributes:

- `package:*`
- `package:list`
- `package:read`
- `package:download`

```json
"configurable_claim": [ "package:*" ]
```
Should be able to perform all actions.

```json
"configurable_claim": [ "package:list", "package:read" ]
```
Should only be able to list and read packages, but not download them.

**Describe alternatives you've considered**

Mentioned above.

**Additional context**

- Similar to the API authentication, we should try to make it optional and configurable.
- Actions and attributes above are just an example for an initial implementation.
- Should we consider attribute negation (`not:package:download`, `!package:download`, `package:!download`)?

aseles13:
This is a good incremental approach to validating and building out our identity management capabilities in Enduro.

fiver-watson:
@jraddaoui I believe this is the issue related to our discussion about adding a prototype landing page so we can test users w no permissions.  If that's correct, below is a super simple, rough first pass: 

![Image](https://github.com/artefactual-sdps/enduro/assets/5068277/376ec25d-bd02-4d14-8692-e6d4c4ef087d)

[...]

PR #962, opened 2024-06-11
Simplify OIDC implementation

jraddaoui:
- Remove client secret config. It's not used to verify the tokens. It will be added back if we implement a BFF flow with client credentials.
- Expect a single client in the OIDC provider. Since the secret is not needed, we don't need another client nor cross-client audience config.

Refs #957.


djjuhasz (commented):
@jraddaoui how can OIDC authenticate the enduro client without a secret?

jraddaoui:
The current implementation uses the [the authorization code flow (with PKCE)](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-browser-based-apps#name-the-authorization-code-grant-2) from the UI to obtain the token and validates it against the provider from the API on each request. These changes just remove unnecessary pieces from that implementation, as the secret is not used to verify the tokens either, we don't need the cross-client support.

We'll probably change this at some point to use a more secure [Backend For Frontend (BFF) flow](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-browser-based-apps#name-backend-for-frontend-bff), where we'll get the token from the backend using client credentials.

djjuhasz (approved):
LGTM :+1:

PR #963, opened 2024-06-11
Fix security schema in API design

jraddaoui:
Use Goa's JWTSecurity, the API is only verifying the token against the OIDC provider, it's not actually doing the authentication flow.

Refs #957.


dashboard/src/openapi-generator/models/Location.ts
djjuhasz, line 1: The changes to this file are all adding unnecessary whitespace. :(  Maybe it's from using a different version of the generator?

djjuhasz (approved):
Mostly LGTM - I just noticed one of the re-generated files added spurious whitespace. It's not a blocker, but it would be nice to remove the whitespace.

dashboard/src/openapi-generator/models/Location.ts
jraddaoui: I used `make gen-dashboard-client`, which has `openapitools/openapi-generator-cli:v7.1.0` hardcoded in it, and each time I run it is the same generated code.

PR #964, opened 2024-06-14
Save preprocesing tasks to Enduro DB

### LINKED ISSUE #949: Problem: child workflow tasks aren't visible in the UI

djjuhasz:
Fixes #949.

- Add a local SavePreprocessingTask activity
- Save tasks returned by the preprocessing child workflow to the Enduro DB as PreservationTasks so they are visible in the Dashboard UI


internal/ingest/preprocessing_task.go
```diff
@@ -0,0 +1,24 @@
...
+
+	// Outcome indicates the completion state of the task.
+	Outcome enums.IngestTaskOutcome
+
+	// StartedAt is the timestamp of the task initiation.
+	StartedAt time.Time
+
+	// StartedAt is the timestamp of the task completion.
+	CompletedAt time.Time
+}
```
jraddaoui: I think it's better to have this in the preprocessing package,

internal/preprocessing/preprocessing.go
```diff
@@ -24,9 +28,25 @@ type WorkflowParams struct {
...
+	// Outcome is an integer indicating if the workflow completed successfully,
+	// or with errors.
+	Outcome Outcome
+
 	// Relative path to the shared path.
 	RelativePath string
+
+	// IngestTasks is a log of the preservation tasks performed by
+	// preprocessing.
+	IngestTasks []ingest.Task
```
jraddaoui: How is this mapped to https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/internal/workflow/preprocessing.go#L35?

internal/workflow/localact/save_ingest_tasks.go
```diff
@@ -0,0 +1,81 @@
...
+		enums.IngestTaskOutcomeValidationFailure: enums.PreservationTaskStatusError,
+	}
+
+	status, found := taskOutcomeToPresTaskStatus[t.Outcome]
+	if !found {
+		status = enums.PreservationTaskStatusUnspecified
+	}
+
+	return datatypes.PreservationTask{
+		TaskID:      uuid.New().String(),
```
jraddaoui: No need to create one if you are going to overwrite it, just use "".

internal/workflow/localact/save_ingest_tasks.go
```diff
@@ -0,0 +1,81 @@
...
+		Name:        t.Name,
+		Status:      status,
+		StartedAt:   sql.NullTime{Time: t.StartedAt, Valid: true},
+		CompletedAt: sql.NullTime{Time: t.CompletedAt, Valid: true},
```
jraddaoui: Should we check they are not zero dates?

internal/workflow/processing.go
```diff
@@ -911,18 +916,41 @@ func (w *ProcessingWorkflow) preprocessing(ctx temporalsdk_workflow.Context, tin
...
+	if err != nil {
+		return err
+	}
+
+	switch ppResult.Outcome {
+	case preprocessing.OutcomeSystemError:
+		return errors.New("preprocessing workflow: system error")
+	case preprocessing.OutcomeContentError:
+		return errors.New("preprocessing workflow: validation failed")
+	default:
```
jraddaoui: Should we check for unexpected values or is the enum enough to verify that?

jraddaoui (changes requested):
Looking great @djjuhasz!

internal/ingest/preprocessing_task.go
djjuhasz: This was my poorly thought out, last minute attempt to shorten "preprocessing" (14 chars) to "ingest" (7 chars) and shorten awfully long names like "PreprocessingTaskOutcomeValidationFailure". :grimacing:  I went looking at some of Dan's newer architecture diagrams for SFA, and I thought "Ingest" was the new "official" name for "Preprocessing".  I talked with Dan & Sara today though, and  "Ingest" refers to everything before Preservation, which would include preprocessing and processing. :(

I'll switch back from ingest -> preprocessing.

internal/preprocessing/preprocessing.go
djjuhasz: Hmm, I don't know how that would map.  Maybe I broke the integration when I changed the name? I'll switch back to "PreservationTasks".

internal/workflow/localact/save_ingest_tasks.go
djjuhasz: Good catch, I forgot to delete this line after creating the TaskID in `SavePreprocessingTasksActivity`.

internal/workflow/localact/save_ingest_tasks.go
djjuhasz: Yes, good idea.  I've added a helper function to convert `time.Time` to `db.NullTime` properly, including zero times.

We've implemented a complementary conversion `db.FormatOptionalTime` in the database package, which makes me realize we shouldn't be using an implementation specific type (`db.NullTime`) outside of the DB implementation layer.  I will change the types of the `StartedAt` and `CompletedAt` fields in a future PR.

internal/workflow/processing.go
djjuhasz: Yes, good idea.

internal/workflow/processing.go
```diff
@@ -911,18 +916,43 @@ func (w *ProcessingWorkflow) preprocessing(ctx temporalsdk_workflow.Context, tin
...
+		localact.SavePreprocessingTasksActivityParams{
+			PkgSvc:               w.pkgsvc,
+			RNG:                  w.rng,
+			PreservationActionID: tinfo.PreservationActionID,
+			Tasks:                ppResult.PreservationTasks,
+		},
+	).Get(opts, &savePPTasksResult)
+	if err != nil {
+		return err
+	}
```
jraddaoui: We could check if there are PreservationTasks in the preprocessing result and only execute the activity if needed.

jraddaoui (approved):
Thanks @djjuhasz! One final thought, but looks great me.

internal/workflow/processing.go
djjuhasz: Yeah, good idea. :+1:

internal/workflow/processing.go
```diff
@@ -911,18 +916,48 @@ func (w *ProcessingWorkflow) preprocessing(ctx temporalsdk_workflow.Context, tin
...
+		switch ppResult.Outcome {
+		case preprocessing.OutcomeSuccess:
+			return nil
+		case preprocessing.OutcomeSystemError:
+			return errors.New("preprocessing workflow: system error")
+		case preprocessing.OutcomeContentError:
+			return errors.New("preprocessing workflow: validation failed")
+		default:
+			return fmt.Errorf("preprocessing workflow: unknown outcome %d", ppResult.Outcome)
+		}
```
jraddaoui, line 959: @djjuhasz, just noticed you moved the outcome check inside the preservation tasks condition too, was that intentional?

djjuhasz: It was intentional, but I didn't think about it for more then 10 seconds.  I included the `Outcome` in the conditional because I can't think of a real-world scenario where preprocessing would return an empty `PreservationTasks` slice and meaningful `Outcome`.  I guess a no-op preprocessing workflow could return an `Outcome` with no `PreservationTasks`, but I can't see this actually happening in a real deployment.

Anyway, I'm happy to move the `ppResult.Outcome` switch outside the conditional - I think it reads cleaner that way.

jraddaoui: Yeah, I agree. But I think it makes more sense outside, even if the result doesn't include an outcome, they way you implemented it, it should be success by default, right?

djjuhasz: Yes, "success" is the default. :)

djjuhasz: See https://github.com/artefactual-sdps/enduro/pull/966

PR #965, opened 2024-06-18
Add optional access control to API

jraddaoui:
Refs #957.

## Use Keycloak instead of Dex in dev env

Dex allowed us to implement OIDC authentication in front of an
OpenLDAP instance in the development environment. However, it
does not provide the IAM tools to manage access control in Enduro.

- Remove Dex and OpenLDAP from the Kubernetes environment.
- Add Keycloak to the environment:
  - Use development server.
  - Listen in `7470` to avoid collisions.
  - Import a custom `enduro` realm with:
    - Clients for Enduro, Temporal and MinIO.
    - Minimal OIDC built-in scopes.
    - Custom client scopes for Enduro and Minio.
    - Custom user attributes and claims mapped to those scopes.
    - Add three default users with different sets of attributes.
  - Without persistence to recreate the custom realm on each start.
  - Update docs to use `keycloack` instead of `dex` in `/etc/hosts`.
  - Update access section to indicate new credentials and service.
- Set up MinIO and Temporal to SSO with Keycloack.
- Update default Enduro configuration to work with Keycloack.
- Remove `offline_access` from the requested scopes (a following
  commit/PR will make scopes configurable).
- Update Pulumi project and related Github action.
- Add TODO note in dashboard to end sessions in the provider.

## Add optional access control to API

Optionally enable Attribute Based Access Control (ABAC) in the API. When
enabled, the API will look for a custom claim in the access token and
extract the attributes relevant to Enduro. If a custom scope needs to be
requested to get that claim, it has to be configured in the dashboard.

- Add OIDC ABAC configuration. Allows to enable or disable access
  control entirely and set a claim to get the attributes. That claim
  could be nested and include values unrelated to Enduro.
- Extend claims and token verifier to parse ABAC attributes from the
  access token based on configuration.
- Allow custom scopes in dashboard and configure it to work with the
  Keycloak instance for the dev. env. by default.
- Add scopes, forbidden error and security to all endpoints in Goa's
  API design.
- Check scopes against attributes in API requests. As a first access
  control check, the scopes configured on the API design need to be
  included in the user attributes.
  - This check allows attributes with wildcards, so:
    - `package:read` will only provide access to that action(s).
    - `package:*` will provide access to all package related actions.
    - `*` will provide full access to all actions.
- Add claims with the extracted attributes to the context on each
  request. After that initial check is passed, the claims are included
  in the context to make them available for other possible access
  control checks on the endpoints implementation.
- Re-order API design methods in storage service:
  - Keep locations and packages methods together.
  - Move `submit` to indicate `update` is related to that method
    and not to `create`.

## TODO

Follow-up work that may or may not happen in this pull request.

- Determine the attributes used for each endpoint and user action [done].
- Consider access control in the dashboard:
  - Add similar configuration and attribute extraction.
  - Add similar checks to hide/show elements/pages.
  - Consider websocket connection and access control (disable it for
    now when access control is enabled) [done].
  - Add landing page for users without the required attributes. Right
    now the landing page is the packages list.
- Add administration section to the documentation:
  - Add notes to configure OIDC ABAC in the API and the dashboard,
    explaining the current behavior.
  - Add notes to build and serve the dashboard, considering that
    configuration.
  - Document API, including the attributes/scopes needed.

**Note:** moving to a BFF auth. flow could simplify some of these changes
and documentation, but that won't happen in this PR.


jraddaoui:
@djjuhasz @mcantelon @sevein @sbreker 

While this is still a work in progress and I'm working in a document explaining the existing OIDC implementation, I think there is enough information in this PR description and in code for you to see where this is going and provide some initial feedback.

I'd appreciate if you can take a look when you have some time. It will be better if you review each commit individually, they have a detailed message (the same as in the PR description), and also check the TODO section. Thanks!

.github/workflows/pulumi.yml
```diff
@@ -41,5 +41,4 @@ jobs:
           kubectl rollout restart deployment temporal
           kubectl rollout restart deployment enduro
           kubectl rollout restart statefulset enduro-a3m
-          kubectl rollout restart deployment dex
```
djjuhasz, line 44: Does keycloak not need to be restarted here?

Tiltfile
```diff
@@ -170,7 +168,6 @@ cmd_button(
     kubectl rollout restart deployment temporal; \
     kubectl rollout restart deployment enduro; \
     kubectl rollout restart statefulset enduro-{pres_sys}; \
-    kubectl rollout restart deployment dex; \
```
djjuhasz, line 173: No Keycloak restart required?

enduro.toml
```diff
@@ -15,12 +15,40 @@ debug = false
 corsOrigin = "http://localhost"
 
 [api.auth]
+# Enable API authentication, only OIDC auth. is available at the moment.
```
djjuhasz: Thanks for the comments! :grin:  

I have a some suggested edits for the text, take them or leave them:

```
# Enable API authentication. OIDC is the only authentication protocol supported 
# at the moment. When enabled the API verifies the access token submitted with 
# each request. The Dashboard is responsible for obtaining and configuring an 
# access token.
```

enduro.toml
```diff
@@ -15,12 +15,40 @@ debug = false
...
+# The API only verifies the access token submitted on each request when
+# enabled. Obtaining the token is done and configured in the dashboard.
 enabled = true
 
 [api.auth.oidc]
+# OIDC provider URL (required with auth. is enabled).
```
djjuhasz: Suggested: 

```
# OIDC provider URL. Required when auth. is enabled. 
```

enduro.toml
```diff
@@ -15,12 +15,40 @@ debug = false
...
 [api.auth]
+# Enable API authentication, only OIDC auth. is available at the moment.
+# The API only verifies the access token submitted on each request when
+# enabled. Obtaining the token is done and configured in the dashboard.
 enabled = true
 
 [api.auth.oidc]
+# OIDC provider URL (required with auth. is enabled). 
 providerURL = "http://keycloak:7470/realms/enduro"
+# OIDC client ID that must be included in the `aud` claim from the access
```
djjuhasz: Suggested:

```
# OIDC client ID. The client ID must be included in the `aud` claim of the access
# token. Required when auth. is enabled. 
```

enduro.toml
```diff
@@ -15,12 +15,40 @@ debug = false
...
+# token (required with auth. is enabled). 
 clientID = "enduro"
 
+[api.auth.oidc.abac]
+# Enable Attribute Based Access Control (ABAC). If enabled, the API will
+# check a configurable multivalue claim against required attributes based
+# on each endpoint configuration. If extra scopes are needed to obtain
+# such claim they can be configured in the dashboard.
```
djjuhasz: Suggested: ~~such~~ a claim they can be configured in the dashboard.

enduro.toml
```diff
@@ -15,12 +15,40 @@ debug = false
...
+enabled = false
+# Claim name/path in the access token where the Enduro attributes should
+# be found. If the claim is nested include all fields with a separator.
+# For example "attributes.enduro", using "." as separator below. Required
+# when ABAC is enabled.
```
djjuhasz: Suggested:

```
# Claim path of the Enduro attributes within the access token. If the claim
# path is nested then include all fields separated by `claimPathSeparator` (see 
# below). E.g. "attributes.enduro" with `claimPathSeparator = "."`. 
#
# Required when ABAC is enabled.
```

enduro.toml
```diff
@@ -15,12 +15,40 @@ debug = false
...
+claimPath = "enduro"
+# Separator used to split the claim path fields. Defaults to "", it will
+# try to find the claim path as it is in the top-level fields.
```
djjuhasz: I'm not sure what you mean by "it will try to find the claim path as it is in the top-level fields."

enduro.toml
```diff
@@ -15,12 +15,40 @@ debug = false
...
+claimPathSeparator = ""
+# Set a preffix to filter values from the configured claim. If the claim
+# contains values unrelated to Enduro's ABAC, the ones that should be used
+# in Enduro can be prefixed so they are the only ones considered. For
+# example, a claim with the following values ["enduro:*", "unrelated"] will
+# be filtered to ["*"] if "enduro:" is used as prefix. Defaults to "", all
+# claim values will be considered.
```
djjuhasz: Suggested:

```
# Add a prefix to filter the values of the configured claim. If the claim
# contains values unrelated to Enduro's ABAC, the values relevant to Enduro
# should be prefixed so they are the only values used for authorization. For 
# example, a claim with values ["enduro:*", "unrelated"] will be filtered to a 
# value of ["*"] when `claimValuePrefix = "enduro:"`. 
# 
# The default claimValuePrefix = "" will pass all access token claim values to 
# the Enduro API for authorization.
```

internal/api/auth/claims.go
```diff
@@ -0,0 +1,67 @@
...
+
+// CheckAttributes verifies all required attributes are present in the claim
+// attributes. It always verifies if the claim is nil (authentication disabled)
+// or the attributes are nil (access control disabled). Attributes are verified
+// by exact match or by having an ancestor with wildcard. For example, a claim
+// with "*" or "package:*" as one of it's attributes will verify all package
+// actions, like "package:list", "package:read", etc.
+func (c *Claims) CheckAttributes(required []string) bool {
+	// Authentication disabled, access control disabled or all wildcard in claims.
+	if c == (*Claims)(nil) || c.Attributes == nil || slices.Contains(c.Attributes, "*") {
```
djjuhasz: I think `if c == nil` is equivalent to `if c == (*Claims)(nil)` here, and is preferable.

internal/api/auth/claims.go
```diff
@@ -0,0 +1,67 @@
...
+		return true
+	}
+
+	// Check for all required attributes considering wildcards.
+	for _, attr := range required {
+		for {
+			if slices.Contains(c.Attributes, attr) {
+				break
+			}
+			attr, _ = strings.CutSuffix(attr, ":*")
```
djjuhasz, line 33: It had to play with the tests for a bit to understand what this line was doing, but I get it now.  Thank $DEITY for tests! :raised_hands:

internal/api/auth/claims.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return true
+}
+
+type contextUserClaimsType struct{}
+
+var contextUserClaimsKey = &contextUserClaimsType{}
+
+// WithUserClaims puts the user claims into the current context.
+func WithUserClaims(ctx context.Context, claims *Claims) context.Context {
```
djjuhasz, line 51: Nice!  Is there anything `ctx` can't do? :stuck_out_tongue_winking_eye: :

internal/api/auth/claims_test.go
```diff
@@ -0,0 +1,122 @@
...
+		ctx := context.Background()
+		ctx = auth.WithUserClaims(ctx, &claims)
+		assert.Equal(t, auth.UserClaimsFromContext(ctx), &claims)
+	})
+
+	t.Run("Returns nil when not found", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := context.Background()
+		assert.Equal(t, auth.UserClaimsFromContext(ctx), (*auth.Claims)(nil))
```
djjuhasz: You could also do:
```go
assert.Assert(t, auth.UserClaimsFromContext(ctx) == nil)
```

But I think it's fine either way.

internal/api/auth/claims_test.go
```diff
@@ -0,0 +1,122 @@
...
+			name: "Checks attribute is missing",
+			claims: &auth.Claims{
+				Attributes: []string{},
+			},
+			attributes: []string{"package:download"},
+			want:       false,
+		},
+		{
+			name:       "Checks attributes on nil claim (auth disabled)",
+			claims:     (*auth.Claims)(nil),
```
djjuhasz: You can just remove this line altogether - it's equivalent.

internal/api/auth/claims_test.go
```diff
@@ -0,0 +1,122 @@
...
+			want:       true,
+		},
+		{
+			name: "Checks multiple attributes exist",
+			claims: &auth.Claims{
+				Attributes: []string{"package:list", "package:read"},
+			},
+			attributes: []string{"package:list", "package:read"},
+			want:       true,
+		},
```
djjuhasz, line 71: I found it interesting that this passes:

```go
{
			name: "Checks multiple attributes exist",
			claims: &auth.Claims{
				Attributes: []string{"package:list", "package:read"},
			},
			attributes: []string{"package:list"},
			want:       true,
		},
```

But this doesn't:

```go
{
			name: "Checks multiple attributes exist",
			claims: &auth.Claims{
				Attributes: []string{"package:list"},
			},
			attributes: []string{"package:list", "package:read"},
			want:       true,
		},
```

Again it's great to have the tests to help learn what the code is supposed to do. :grin:

internal/api/auth/claims_test.go
```diff
@@ -0,0 +1,122 @@
...
+		{
+			name: "Checks missing attributes with wildcard",
+			claims: &auth.Claims{
+				Attributes: []string{"package:*"},
+			},
+			attributes: []string{"package:list", "storage:download"},
+			want:       false,
+		},
```
djjuhasz, line 114: I think this is expected behaviour, but I just wanted to check:

```go
{
	name: "Checks a more specific attribute doesn't match a general one",
	claims: &auth.Claims{
		Attributes: []string{"package:list"},
	},
	attributes: []string{"package:list:something"},
	want:       false,
},
```

I guess the expectation is the user would need an explicit wildcard grant (e.g. `package:list:*`) to match the more specific requirement of `package:list:something`?

internal/api/auth/config.go
```diff
@@ -13,6 +13,14 @@ type Config struct {
 type OIDCConfig struct {
 	ProviderURL string
 	ClientID    string
+	ABAC        OIDCABACConfig
```
djjuhasz: Nice to have: A comment that expands ABAC for readers unfamiliar with the acronym.

internal/api/auth/config.go
```diff
@@ -26,8 +34,14 @@ type RedisConfig struct {
 
 // Validate implements config.ConfigurationValidator.
 func (c Config) Validate() error {
-	if c.Enabled && c.OIDC == nil {
+	if !c.Enabled {
+		return nil
+	}
+	if c.OIDC == nil || c.OIDC.ProviderURL == "" || c.OIDC.ClientID == "" {
```
djjuhasz, line 42: Yay, config validation. :tada:

internal/api/auth/token_verifier.go
```diff
@@ -38,26 +38,92 @@ func NewOIDCTokenVerifier(ctx context.Context, cfg *OIDCConfig) (*OIDCTokenVerif
...
 	var claims Claims
 	if err := idToken.Claims(&claims); err != nil {
-		return false, err
+		return false, nil, err
 	}
 
 	// Check that claims are verified.
 	if !claims.EmailVerified {
-		return false, nil
+		return false, nil, nil
```
djjuhasz: Shouldn't this return an error to indicate the email verified claim failed?

internal/api/design/package_.go
```diff
@@ -48,6 +50,9 @@ var _ = Service("package", func() {
 	})
 	Method("list", func() {
 		Description("List all stored packages")
+		Security(JWTAuth, func() {
+			Scope("package:list")
```
djjuhasz, line 64: Adding a security scope is nice and easy. :tada:

internal/package_/goa.go
```diff
@@ -50,6 +51,12 @@ func (w *goaWrapper) JWTAuth(
 		return ctx, ErrInvalidToken
 	}
 
+	if !claims.CheckAttributes(scheme.RequiredScopes) {
+		return ctx, goapackage.Forbidden("invalid ABAC attributes")
```
djjuhasz: Is the "invalid ABAC attributes" message for a system administrator or an end user?  If it's for an end user I think a more generic "access denied" message would be better.

internal/upload/service.go
```diff
@@ -37,7 +36,7 @@ type serviceImpl struct {
 
 var _ Service = (*serviceImpl)(nil)
 
-var ErrInvalidToken error = goastorage.Unauthorized("invalid token")
+var ErrInvalidToken error = goaupload.Unauthorized("invalid token")
```
djjuhasz: Haha, good notice. :mag:

djjuhasz (commented):
@jraddaoui this looks great to me, thanks for all the work! :bow:  I like the way ABAC rules are specified and checked, and adding access attributes seems easy now that you've set up the system.

I added a bunch of questions and some minor suggestions, but nothing that I see as a blocker.

enduro.toml
djjuhasz: P.S I have no idea what an `aud` claim is. :shrug:

.github/workflows/pulumi.yml
jraddaoui: I'm not including Keycloack in the Pulumi project for now. We'll be working on CD soon, we'll revisit this Pulumi project and see if we can use an external provider directly.

Tiltfile
jraddaoui: We did it for Dex because we were recreating its MySQL DB, something we probably didn't need. For Keycloack, I can only see the need to restart it if you make changes to the imported realm.

enduro.toml
jraddaoui:
```
# Enable API authentication. OIDC is the only protocol supported at the
# moment. When enabled the API verifies the access token submitted with 
# each request. The API client is responsible for obtaining an access
# token from the provider.
```

enduro.toml
jraddaoui: Check [here](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest:~:text=aud,single%20case%2Dsensitive%20string.) and [here](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest:~:text=An%20Attacker%20uses,of%20the%20token.).

enduro.toml
jraddaoui: I want to make clear it refers to the previously mentioned claim, "that claim" maybe?

jraddaoui: I'll just remove that line, it will be documented in the dashboard side.

djjuhasz: "such a claim" works, it just needs the article. ;)

enduro.toml
jraddaoui: Just that if you use "attributes.enduro" as the claim path but don't set a separator, it will try to find a field called exactly like that in the top-level fields from the access token. Maybe `Defaults to "", the claim path will be looked as it is in the top-level fields from the access token.`?

enduro.toml
jraddaoui:
```
# Add a prefix to filter the values of the configured claim. If the claim
# contains values unrelated to Enduro's ABAC, the values relevant to Enduro
# should be prefixed so they are the only values used for access control.
# For example, a claim with values ["enduro:*", "unrelated"] will be filtered
# to a value of ["*"] when `claimValuePrefix = "enduro:"`. The default "" will
# not filter any value.
```

internal/api/auth/claims_test.go
jraddaoui: I think the message from `Equal` gives more information about the mismatches.

`Equal` error:

```
assertion failed: &{info@artefactual.com true [*]} (*auth.Claims) != &{info2@artefactual.com false [another]} (&claims2 *auth.Claims)
```

`Assert` error:

```
assertion failed: auth.UserClaimsFromContext(ctx) is not &claims2
```

internal/api/auth/claims_test.go
jraddaoui: The user must have all scopes required in the endpoint.

internal/api/auth/claims_test.go
jraddaoui: Nice, that's a great addition to the tests ;)

enduro.toml
djjuhasz: Maybe `The default value of "" will try to match the claim path as-is to a top-level field from the access token.`?

internal/api/auth/token_verifier.go
jraddaoui: Some bits are copied from another project, where we wanted to differentiate between a verification error and another type of error. That differentiation determines if the error should be logged or not.

internal/api/design/package_.go
jraddaoui: Thanks Goa!

internal/package_/goa.go
jraddaoui: I'll use the default error message for both 401 (Unauthorized) and 403 (Forbidden) responses. Sevein noted at some point that we should be returning proper JSON, even if Goa allows a plain string.

enduro.toml
jraddaoui: A lot better, thanks!

jraddaoui:
Thanks @djjuhasz! I addressed most of your feedback, I'll ping you for another review after adding all the scopes we discussed yesterday and some of the dashboard TODOs.

internal/api/auth/claims.go
```diff
@@ -0,0 +1,67 @@
...
+	return context.WithValue(ctx, contextUserClaimsKey, claims)
+}
+
+// UserClaimsFromContext returns the user claims from the context.
+// An empty string is returned if it was not found.
```
sevein: Outdated comment, i.e. should not read "empty string".

internal/api/auth/config.go
```diff
@@ -26,8 +36,14 @@ type RedisConfig struct {
...
 		return errors.New("Missing OIDC configuration with API auth. enabled.")
 	}
+	if c.OIDC.ABAC.Enabled && c.OIDC.ABAC.ClaimPath == "" {
+		return errors.New("Missing OIDC ABAC claim path with ABAC enabled.")
```
sevein:
```suggestion
		return errors.New("missing OIDC ABAC claim path with ABAC enabled.")
```

https://google.github.io/styleguide/go/decisions.html#error-strings

internal/package_/goa.go
djjuhasz: The thing I found strange about the Goa forbidden type is it's underlying type is a string, but the string value isn't actually used in the `Error()` method. :confused: 

```go
type Forbidden string
...
// Error returns an error description.
func (e Forbidden) Error() string {
	return "Invalid ABAC attributes"
}
```

internal/api/auth/token_verifier.go
```diff
@@ -2,28 +2,28 @@ package auth
...
 	"github.com/coreos/go-oidc/v3/oidc"
 )
 
-type Claims struct {
-	EmailVerified bool `json:"email_verified,omitempty"`
-}
-
 type TokenVerifier interface {
-	Verify(ctx context.Context, token string) (bool, error)
+	Verify(ctx context.Context, token string) (bool, *Claims, error)
```
sevein: Now with this three-value return I'd be inclined to remove the boolean to avoid the ambiguity, i.e.:

| verified | claims    | error     | Interpretation                                |
|----------|-----------|-----------|-----------------------------------------------|
| true     | non-nil   | nil       | Token verified successfully                   |
| false    | nil       | non-nil   | Verification failed with error                |
| false    | nil       | nil       | Token invalid, but no specific error          |
| true     | non-nil   | non-nil   | Ambiguous: verified but with error?           |
| false    | non-nil   | nil       | Ambiguous: invalid token with claims?         |
| true     | nil       | nil       | Ambiguous: verified without claims?           |

If we changed it to return two values:
- Return (claims, nil) for successful verification
- Return (nil, ErrUnauthorized) for unauthorized access
- Return (nil, err) for other errors

If you choose to stick with the three-value return then I think we could document in the Verify method what are unexpected permutations, e.g. if `verified` is true then `claims` must be non-nil and so on.

internal/api/auth/token_verifier.go
```diff
@@ -38,26 +38,92 @@ func NewOIDCTokenVerifier(ctx context.Context, cfg *OIDCConfig) (*OIDCTokenVerif
...
+
+	var keys []string
+	if t.cfg.ABAC.ClaimPathSeparator != "" {
+		keys = strings.Split(t.cfg.ABAC.ClaimPath, t.cfg.ABAC.ClaimPathSeparator)
+	} else {
+		keys = []string{t.cfg.ABAC.ClaimPath}
+	}
+
+	for i, key := range keys {
+		value, exists := data[key]
```
sevein:
```suggestion
		value, ok := data[key]
```

`exists` is clear but often we just write `ok` (the "comma ok" idiom briefly mentioned in [effective go](https://go.dev/doc/effective_go#maps))

internal/api/auth/token_verifier_test.go
```diff
@@ -68,14 +84,20 @@ func TestOIDCTokenVerifier(t *testing.T) {
 		})
 		assert.NilError(t, err)
 
-		verified, err := v.Verify(ctx, token)
+		verified, claims, err := v.Verify(ctx, token)
 		assert.NilError(t, err)
 		assert.Assert(t, verified == false)
+		assert.DeepEqual(t, claims, (*auth.Claims)(nil))
```
sevein: I saw the other conversation with David on this check. This is fine to me but if you're still looking for other alternatives I think this could be one using `gotest.tools/v3/assert/cmp`:

```suggestion
		assert.Assert(t, cmp.IsNil(claims))
```

It would return:

    assertion failed: {foo@bar.com false []} (type *auth.Claims) is not nil

internal/package_/goa.go
```diff
@@ -34,22 +35,31 @@ var patternMatchingCharReplacer = strings.NewReplacer(
 	"_", "\\_",
 )
 
-var ErrInvalidToken error = goapackage.Unauthorized("invalid token")
+var (
+	ErrUnauthorized error = goapackage.Unauthorized("Unauthorized")
+	ErrForbidden    error = goapackage.Forbidden("Forbidden")
+)
 
 func (w *goaWrapper) JWTAuth(
```
sevein, line 43: Would it be useful to write a test for w.JWTAuth given its criticality? i.e. if we're not getting it right here then we're doomed. I realize we don't have yet a mock of TokenVerifier using GoMock but maybe one should be created, so we can do something like:

```go
// File: auth_test.go
package package_

func TestJWTAuth(t *testing.T) {
	w := &goaWrapper{
		packageImpl: &packageImpl{
			tokenVerifier: &auth.NoopTokenVerifier{},  // TODO: use GoMock.
		},
	}
	
	ctx, err := w.JWTAuth(context.Background(), "", &security.JWTScheme{})
	// ...
}
```

sevein (approved):
Looks good! I've added a few minor observations.

> Follow-up work that may or may not happen in this pull request.

I'm leaning towards "may not" here. Smaller pull requests are generally better!

internal/package_/goa.go
jraddaoui: Yeah, this puzzles me too, no matter what I do creating the error in the implementation, it always uses the message from the design in the response. We can think about improving error responses in a following PR.

internal/package_/goa.go
jraddaoui: Yeah, I was thinking about this yesterday, I was not planning to create a mock, but will do ;)

internal/api/auth/token_verifier.go
jraddaoui: Yeah, two values seems better, thanks!

internal/api/auth/token_verifier.go
jraddaoui: Not really a "comma ok" case, but done ;)

internal/package_/goa.go
jraddaoui: I thought we had at least some tests for it, but it looks like we don't, so definitely better with a mock ;)

jraddaoui: This is WIP, I'll get back to it later today.

sbreker:
This looks amazing @jraddaoui! :star_struck:  I do not have anything to add beyond the others' comments. Thanks for your efforts with this!

I agree that follow-on work might work best as a second PR after this is merged...

jraddaoui:
Thanks @sevein and @sbreker!

I think I have addressed all feedback and added a few more commits. I'll leave it disabled for now until we add the dashboard side and the documentation, which as you said will happen in another PR.

djjuhasz (approved):
LGTM! :+1:

djjuhasz:
@jraddaoui I tried testing the authentication out locally, but I'm getting an "NS_ERROR_UNKNOWN_HOST" error from keycloak when I click the Enduro Dashboard "Sign In" button.  The only config change I made was to change `api.auth.oidc.abac.enabled = true` in "enduro.toml".

djjuhasz:
@jraddaoui nevermind, I forgot to add keycloak to my `/etc/hosts`. RTFM moment. :facepalm:

djjuhasz:
After updating my `/etc/hosts` I logged into the Enduro Dashboard fine with the "readonly" account, and I navigated around the read pages without any issues.  I also logged into the Temporal UI and MinIO without any issues.

When I tried to process a SIP, I got an error in the upload activity though:
```
{
  "message": "[storage create]: invalid value expected *storage.CreatePayload, got &{79dc514f-d42e-40b4-a17b-2c2cc325f88f ZippedBag.zip <nil>}",
  "source": "GoSDK",
  "applicationFailureInfo": {
    "type": "ClientError"
  }
}
```

It's possible the error is unrelated to your changes, but I've never seen it before.

jraddaoui:
Thanks @djjuhasz! That's totally related to these changes, I fixed that issue in the tests, but I thought the service was created differently in the main functions. Will fix!

djjuhasz (commented):
Does the order you add the endpoints to the client effect authentication?

jraddaoui:
No, it doesn't. That was just a cosmetic commit:

> Re-order API design methods in storage service
>
> - Keep locations and packages methods together.
> - Move `submit` to indicate `update` is related to that method
>   and not to `create`.

PR #967, opened 2024-06-18
Generate enums with https://github.com/abice/go-enum

djjuhasz:


internal/enums/pkg_status.go
```diff
@@ -1,80 +1,15 @@
...
+/*
+ENUM(
+new         // Unused!
+in progress // Undergoing work.
+done        // Work has completed.
+error       // Processing failed.
+unknown     // Unused!
+queued      // Awaiting resource allocation.
+abandoned   // User abandoned processing.
+pending     // Awaiting user decision.
```
jraddaoui: All other enums use PascalCase, I see this one had that `NewPackageStatus` function that used this values, but we could change that now for consistency. Not a big deal of course, feel free to ignore.

jraddaoui (approved):
Looks good, just one minor comment.

internal/enums/pkg_status.go
djjuhasz: Sounds good to me.  The only hesitation I had was changing the string value of "in progress" to "InProgress" as a user visible value (e.g in the UI).  It looks like the string value of `PackageStatusInProgress` is only used in one event though, which is not user facing: https://github.com/artefactual-sdps/enduro/blob/9ae3c8f1741862cef841c058053ab583bbd53884/internal/package_/package_.go#L185

PR #968, opened 2024-06-18
Fix genpkgs script

djjuhasz:
For generating random package data and loading it into the Enduro package table.

### DIFF
--- a/hack/genpkgs/load.sh
+++ b/hack/genpkgs/load.sh
@@ -6,8 +6,14 @@ __dir=$(cd "$(dirname "$0")" && pwd)
 
 go run ${__dir}/main.go > ${__dir}/data.csv
 
-mysql -h127.0.0.1 -uroot -proot123 -P7450 enduro \
+mysql -h127.0.0.1 -uroot -proot123 -P3306 enduro \
+	-e "SET GLOBAL local_infile=1;"
+
+mysql -h127.0.0.1 -uroot -proot123 -P3306 enduro \
 	-e "DELETE FROM package;"
 
-mysql -h127.0.0.1 -uroot -proot123 -P7450 enduro \
+mysql -h127.0.0.1 -uroot -proot123 -P3306 --local-infile=1 enduro \
 	-e "LOAD DATA LOCAL INFILE '${__dir}/data.csv' INTO TABLE package FIELDS TERMINATED BY ','"
+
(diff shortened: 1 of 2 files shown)


hack/genpkgs/load.sh
```diff
@@ -6,8 +6,11 @@ __dir=$(cd "$(dirname "$0")" && pwd)
 
 go run ${__dir}/main.go > ${__dir}/data.csv
 
-mysql -h127.0.0.1 -uroot -proot123 -P7450 enduro \
+mysql -h127.0.0.1 -uroot -proot123 -P3306 enduro \
+	-e "SET GLOBAL local_infile=1;"
```
jraddaoui, line 10: Really unlikely to be used outside of a development/test environment, but I'd change the global back at the end of the script, just in case.

hack/genpkgs/main.go
```diff
@@ -4,26 +4,23 @@ import (
...
 	"strconv"
-	"time"
 
 	"github.com/google/uuid"
 )
 
 const (
 	// How many records we want to generate.
-	datasetSize = 2500000
+	datasetSize = 55
```
jraddaoui: It should be fast, right? I believe it was initially created to test pagination performance, so maybe keep the default to 10000 or so.

hack/genpkgs/main.go
```diff
@@ -4,26 +4,23 @@ import (
...
 
 	// Size of each batch that we write to the CSV.
 	batchSize = 100
 )
 
 func main() {
-	rand.Seed(time.Now().UnixNano())
-
 	w := csv.NewWriter(os.Stdout)
+	defer w.Flush()
```
jraddaoui: Nice catch! But, instead of using defer, I'd do it right after the for loop, before calling `w.Error()`:

> func (w *csv.Writer) Error() error
>
> Error reports any error that has occurred during a previous [Writer.Write] or [Writer.Flush].

jraddaoui (changes requested):
Nice @djjuhasz! I forgot about these scripts.

hack/genpkgs/main.go
djjuhasz: I only needed 41+ packages for my use case, but I will change the value to 10K. ;-)  I thought about adding a bash script + go script parameter to set the number of records generated, but decided not to go down that rabbit hole. :laughing:

jraddaoui (approved):
Thanks!

PR #973, opened 2024-06-24
Add a pager to the dashboard package list

djjuhasz:


dashboard/src/pages/packages/index.vue
```diff
@@ -20,16 +20,16 @@ const { execute, error } = useAsyncState(() => {
...
-  if (el) tooltip = new Tooltip(el);
+  if (el.value) tooltip = new Tooltip(el.value);
 });
 
-let showLegend = $ref(false);
+let showLegend = ref(false);
 const toggleLegend = () => {
-  showLegend = !showLegend;
+  showLegend.value = !showLegend.value;
   if (tooltip) tooltip.hide();
```
jraddaoui: Why the `$ref` to `ref` change?

dashboard/src/pages/packages/index.vue
```diff
@@ -93,5 +93,33 @@ const toggleLegend = () => {
...
+      <nav aria-label="Package list pages">
+        <ul class="pagination justify-content-center">
+          <li class="page-item">
+            <a
+              href="#"
+              :class="{
+                'page-link': true,
+                disabled: !packageStore.hasPrevPage,
+              }"
+              @click="packageStore.prevPage($event)"
```
jraddaoui: Use `@click.prevent` and don't pass the event.

dashboard/src/stores/package.ts
```diff
@@ -156,6 +174,23 @@ export const usePackageStore = defineStore("package", {
         this.current.status = api.EnduroStoredPackageStatusEnum.InProgress;
       });
     },
+    nextPage(e: Event) {
+      e.preventDefault(); // don't scroll to the top of the page.
```
jraddaoui: And this won't be needed.

jraddaoui (changes requested):
Looks great @djjuhasz! Just a question and a suggestion.

dashboard/src/pages/packages/index.vue
djjuhasz: I think `$ref` is outdated Vue syntax - the VS Code Vue.js linter complains about `$ref`.

djjuhasz: https://vuejs.org/guide/extras/reactivity-transform.html

jraddaoui: Oh, thanks for noticing that! But we use Vue Macros plugin and the VSC extension Vue - Official v2.0.22 should be able to read the project config and it doesn't complain for me. Is that the extension and version you are using?

In any case, I'd leave it as it is for now, it's used in other places so we may want to create an issue to deal with it separately.

jraddaoui: https://github.com/artefactual-sdps/enduro/issues/976

djjuhasz: @jraddaoui Yes, I've got the "Vue Official" VSC extension v2.0.22 and all the `$ref` instances are underlined in red. :confused: 

image
image

djjuhasz: ^ That's a screenshot on main

jraddaoui: I'd like to look at this better in #976 and address it across the entire project.

djjuhasz: > In any case, I'd leave it as it is for now, it's used in other places so we may want to create an issue to deal with it separately.

Do you mean leave it as it is in this PR (ref) or as in the main branch ($ref)?  I assumed the former.

jraddaoui: I mean like it was on main, so we can deal with it in #976.

djjuhasz: Done

dashboard/src/pages/packages/index.vue
```diff
@@ -7,7 +7,7 @@ import { useLayoutStore } from "@/stores/layout";
 import { usePackageStore } from "@/stores/package";
 import { useAsyncState } from "@vueuse/core";
 import Tooltip from "bootstrap/js/dist/tooltip";
-import { onMounted } from "vue";
+import { onMounted, ref } from "vue";
```
jraddaoui: Forgot the import.

dashboard/src/pages/packages/index.vue
```diff
@@ -85,13 +85,47 @@ const toggleLegend = () => {
...
+            </td>
             <td>{{ $filters.formatDateTime(pkg.startedAt) }}</td>
-            <td><UUID :id="pkg.locationId" /></td>
-            <td><StatusBadge :status="pkg.status" /></td>
+            <td>
+              <UUID :id="pkg.locationId" />
+            </td>
+            <td>
+              <StatusBadge :status="pkg.status" />
+            </td>
```
jraddaoui, line 97: Not sure why this formatting changes, not a big deal of course.

jraddaoui (approved):
Thanks @djjuhasz!

dashboard/src/pages/packages/index.vue
djjuhasz: I figured out that I was using the default "Vue.volar" formatter instead of the VSCode "prettier" extension as configured in `.vscode/settings.json`.  After switching to prettier it reformatted.

dashboard/src/pages/packages/index.vue
djjuhasz: Fixed.

jraddaoui (approved):
Thanks @djjuhasz!

PR #975, opened 2024-06-24
Fix dashboard package status badges

### LINKED ISSUE #974: Problem: package status is incorrect or missing in Dashboard
**Describe the bug**

The Enduro Dashboard package status field is not showing the correct value on the package list or package detail screen (see screenshots below).  The package list status column is always blank.  The package detail "Workflow Status" and "Started" values are both blank.

**To Reproduce**

1. Process a SIP with Enduro using a3m as the preservation system
2. Log in to the Enduro Dashboard web UI
3. Click on "Packages" in the navigation pane on the left side of the page
4. Click on a package in the list

**Expected behavior**

The package status field should show the correct value.

**Screenshots**

Screenshot from 2024-06-24 11-23-27
Screenshot from 2024-06-24 11-22-49
Screenshot from 2024-06-24 11-23-11

**Additional context**

The package status was working pretty recently (last week?) so this is a recent regression.

djjuhasz:
Fix #974.


internal/api/design/convert.go
```diff
@@ -0,0 +1,12 @@
...
+// sliceToAny copies the elements of a slice with a specific type (e.g.
+// "[]string") to a slice with type []any. See:
+// https://go.dev/doc/faq#convert_slice_of_interface.
+func sliceToAny[T any](s []T) []any {
+	a := make([]any, len(s))
+	for i, v := range s {
+		a[i] = v
+	}
+	return a
+}
```
jraddaoui: There is an `EnumNameInterfaces()` method available in the enum template already ([here](https://github.com/artefactual-sdps/enduro/blob/main/hack/make/enums.tmpl#L6-L13)), you should be able to use that function directly.

internal/api/design/package_.go
```diff
@@ -187,7 +188,7 @@ var _ = Service("package", func() {
 })
 
 var EnumPackageStatus = func() {
-	Enum("new", "in progress", "done", "error", "unknown", "queued", "pending", "abandoned")
+	Enum(sliceToAny(enums.PackageStatusNames())...)
```
jraddaoui: `Enum(enums.PackageStatusInterfaces()...)`

jraddaoui (changes requested):
Good thing you noticed this @djjuhasz, thanks!

internal/api/design/convert.go
djjuhasz: Ah, nice! Thanks. :)

internal/api/design/package_.go
djjuhasz: Oops. :blush:

internal/enums/pres_action_type.go
```diff
@@ -2,10 +2,10 @@ package enums
...
 /*
 ENUM(
-Unspecified
-CreateAIP
-CreateAndReviewAIP
-MovePackage
+unspecified
+create aip
+create and review aip
+move package
```
jraddaoui, line 8: Just noticed, these were using hyphens: https://github.com/artefactual-sdps/enduro/pull/975/files#diff-5d31faeec1ed87c9e54c8a9f758ddbd6640bd2e1a116020c6466dfcdeb25ef28L264

Could that be an issue?

djjuhasz: It's not an issue for the Enduro or the Dashboard. The strings "create-aip", "create-and-review-aip", and "move-package" don't occur anywhere in this branch anymore.  I've tested the Dashboard and the "Create AIP" and "Move package" titles shows up correctly on the package show page.  As far as I know the "Create and Review AIP" workflow is non-functional in Enduro.

image

jraddaoui (approved):
Thanks!

PR #978, opened 2024-06-26
Add access control checks to dashboard

jraddaoui:
- Add home page.
- Split layout store:
  - Move user related elements to its own `auth` store.
- Allow to disable OIDC authentication:
  - Consider a new environment variable to disable OIDC authentication.
  - Move configuration and user manager to auth store.
  - Remove user menu from sidebar and user name from home page when
    authentication is disabled.
  - Redirect to home page when visiting user signin pages with
    authentication disabled.
- Add access control checks:
  - Consider new environment variables to enable Attribute Based Access
    Control and get a set of user attributes from a configurable claim in
    the access token.
  - Add attributes to auth store state and include actions to parse and
    check those attributes.
  - Configure protected routes and check attributes before each route
    change, redirecting to the home if the check fails.
  - Modify or hide elements (links, tabs, etc.) based on user attributes.

Refs #957.


jraddaoui:
@djjuhasz, @sevein, @sbreker 

I'm struggling with the tests a little more than expected and it's getting late to review and merge, while I work in the documentation. Sorry about that, I'll restore and add more tests in a different PR. Again, I think it makes more sense to review commit by commit.

dashboard/.env.development
```diff
@@ -1,4 +1,9 @@
+VITE_OIDC_ENABLED=true
 VITE_OIDC_AUTHORITY=http://keycloak:7470/realms/enduro
```
sbreker: Nitpick: I would consider changing the realm to "Artefactual" or something to indicate the realm is institutional level for a multi-tenant Keycloak instance. E.g. For the SSO trial we are creating a realm for each customer within a Keycloak instance, and configuring our applications as clients in that realm.

jraddaoui: Good point, I called the realm "enduro" in https://github.com/artefactual-sdps/enduro/blob/main/hack/kube/components/dev/keycloak.yaml#L78, but we have clients for Temporal, MinIO and Enduro in that realm already. This is the default dev configuration, but we can change it ;)

sbreker: Yeah I thought about that after unfortunately... Sorry for not commenting on the previous PR. And yes - it's a nitpick because this is for dev work.

sbreker (approved):
Looks great @jraddaoui :+1:

dashboard/src/pages/index.vue
```diff
@@ -1,8 +1,34 @@
...
+const authStore = useAuthStore();
+const layoutStore = useLayoutStore();
+layoutStore.updateBreadcrumb([]);
 </script>
 
-<template></template>
+<template>
+  <div class="container-xxl">
+    <h1 class="d-flex mb-3">
+      <IconHomeLine class="me-3 text-dark" />Welcome,
```
djjuhasz: Should this be hidden if the user isn't logged in?

dashboard/src/components/PackageLocationCard.vue
```diff
@@ -37,7 +39,13 @@ const choose = async () => {
         >
         <span v-else><UUID :id="packageStore.current.locationId" /></span>
       </p>
-      <div class="actions" v-if="!packageStore.isRejected">
+      <div
+        class="actions"
+        v-if="
+          !packageStore.isRejected &&
```
djjuhasz, line 45: I think the move button should only show if the package status is "done".

dashboard/src/components/__tests__/PackageLocationCard.test.ts
```diff
@@ -38,7 +38,7 @@ describe("PackageLocationCard.vue", () => {
               <!-- Copy icon. --><span><svg viewBox="0 0 24 24" width="1.2em" height="1.2em" aria-hidden="true"><g fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"><path d="M8 4v12a2 2 0 0 0 2 2h8a2 2 0 0 0 2-2V7.242a2 2 0 0 0-.602-1.43L16.083 2.57A2 2 0 0 0 14.685 2H10a2 2 0 0 0-2 2"></path><path d="M16 18v2a2 2 0 0 1-2 2H6a2 2 0 0 1-2-2V9a2 2 0 0 1 2-2h2"></path></g></svg><span class="visually-hidden">Copy to clipboard</span></span>
             </button>
         </div></span></p>
-        <div class="actions"><button type="button" class="btn btn-primary btn-sm">Choose storage location</button></div>
+        <!--v-if-->
```
djjuhasz, line 41: Does this do anything?

djjuhasz (approved):
LGTM @jraddaoui.  Just a few comments.

dashboard/src/pages/index.vue
jraddaoui: When authentication is disabled it will show "Welcome!", when a user is signed in it will show "Welcome, username!".

dashboard/src/components/PackageLocationCard.vue
jraddaoui: Not sure why, but that was like that already, I don't want to change it here.

dashboard/src/components/__tests__/PackageLocationCard.test.ts
jraddaoui: That's the element removed conditionally based in the access control check. Not sure why is rendered like that, but it's needed for the tests to pass.

jraddaoui:
Thanks @djjuhasz!

PR #979, opened 2024-06-28
Add identity and access control docs

jraddaoui:
Create an administrator section in the docs and add an initial identity and access control section.

Refs #957.

### DIFF
--- a/docs/src/admin-manual/iac.md
+++ b/docs/src/admin-manual/iac.md
@@ -1,0 +1,152 @@
+# Identity and access control
+
+Enduro optionally supports external OpenID Connect (OIDC) compatible providers
+for authentication and access control. Users can authenticate against the
+external provider from the dashboard to receive an access token that will be
+sent to the API on each request.
+
+Enduro uses Attribute Based Access Control (ABAC) to determine the actions and
+resources to which an authenticated user has access. It looks for a configurable
+claim in the access token to know the attributes assigned to the user in their
+external provider.
+
+This section explains how to configure the OIDC provider in Enduro's API and
+dashboard.
+
+## API configuration
+
+Below is a self-documented API section from an Enduro configuration file in
+TOML format:
+
+```toml
+[api]
+# TCP address for the server to listen on, in the form "host:port".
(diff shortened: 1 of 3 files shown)


docs/src/admin-manual/iac.md
```diff
@@ -0,0 +1,152 @@
...
+VITE_OIDC_ABAC_ENABLED
+VITE_OIDC_ABAC_CLAIM_PATH
+VITE_OIDC_ABAC_CLAIM_PATH_SEPARATOR
+VITE_OIDC_ABAC_CLAIM_VALUE_PREFIX
+```
+
+They must match the ones configured in the API. `VITE_OIDC_AUTHORITY` has to be
+the same OIDC provider URL and `VITE_OIDC_CLIENT_ID` needs to be the same or a
+trusted client. This client (and the one used in the API configuration, if they
+are not the same) must be included in the `aud` claim from the access token,
```
fiver-watson: ends w a comment instead of a period

docs/src/admin-manual/iac.md
```diff
@@ -0,0 +1,152 @@
...
+
+The authorization flow will request the `openid email profile` scopes by
+default. If needed, `VITE_OIDC_EXTRA_SCOPES` can be used to request additional
+scopes. The ABAC variables will work in the same way as they do in the API,
+they are explained in detail in the API configuration comments above.
+
+These environment variables must be set at build time or they can be replaced
```
fiver-watson: you say "must", and then follow with an "or" - is the or a bad outcome if you don't do the _must_? Or is it an alternative? If it's an alternative, then I would change the "must" to "can" - _must_ is typically used as an imperative (you must do this, or else...) 

> These environment variables can be set at build time, or they can be replaced in the final assets.

fiver-watson (approved):
Overall, looks good - 2 minor grammatical suggestions

jraddaoui:
Thanks @fiver-watson!

@djjuhasz @sbreker, I'll merge so it gets deployed to the docs site. I'll still appreciate your feedback when you have some time, will address it in a follow-up PR.

PR #992, opened 2024-07-11
Allow extra params in the dashboard auth request

jraddaoui:
- Allow to configure and pass extra parameters in the dashboard OIDC
  authorization request.
- Rename enduro-secret to enduro-dashboard-secret and don't use it in
  the enduro manifest as the client id may differ from the one used in the
  dashboard.
- Trim all environment variables.
- Add documentation.

Refs #957.


djjuhasz (approved):
Thanks @jraddaoui, LGTM! :+1:

Issue #980, opened 2024-07-02
Problem: start time is missing on package detail page


sallain:
**Describe the bug**

The start time is missing on the package detail page, which also means that the time to complete is not calculated

**To Reproduce**

1. Process a new package
2. Go to the package detail page
3. Look for the start time and the time to complete!

**Expected behavior**

**Screenshots**

![image](https://github.com/artefactual-sdps/enduro/assets/4612276/21199a98-1117-44e2-9971-4e28ce07987a)

**Additional context**

This worked prior to last week's updates so it seems like a pretty recent regression.

jraddaoui:
I came across this issue while working on #991, I'll take it as @djjuhasz is out this week, I'll have a PR soon ;)

PR #994, opened 2024-07-15
Fix preservation action started at time

jraddaoui:
Mark `sql.NullTime` as valid before sending to persistence.

Refs #980.


mcantelon:
Looks good to me!

mcantelon (approved):
Looks good to me!

Issue #995, opened 2024-07-18
Problem: ingest workflow fails when authentication is enabled in the API


jraddaoui:
**Describe the bug**

Some activities in the ingest workflow need to connect to Enduro's storage API for final storage. If authentication is enabled in that API the ingest workflow will fail.

**To Reproduce**

- Enable auth in Enduro's API
- Make sure Enduro is connected to that API for storage ([here](https://github.com/artefactual-sdps/enduro/blob/main/enduro.toml#L82))
- Trigger an ingest workflow

**Expected behavior**

Workflow finnishes without error and AIP is stored and downloadable.

**Logs**

Working with Archivematica (enduro-am-worker):

```
2024-07-18T15:43:48.801Z	V(2)	enduro-am-worker.temporal-client	log/with_logger.go:84	error	{"Namespace": "default", "TaskQueue": "am", "WorkerID": "1@enduro-am-5f65cbff5c-k7h9z@", "WorkflowID": "processing-workflow-ba2c2d61-283e-4841-a950-cac9cca0c32d", "RunID": "062f88a6-7d83-43e4-9e38-bb653dde03ee", "ActivityType": "create-storage-package-activity", "Attempt": 1, "Error": "create-storage-package-activity: Unauthorized", "error": "Activity error."}
go.temporal.io/sdk/log.(*withLogger).Error
	go.temporal.io/sdk@v1.26.1/log/with_logger.go:84
go.temporal.io/sdk/internal.(*activityTaskHandlerImpl).Execute
	go.temporal.io/sdk@v1.26.1/internal/internal_task_handlers.go:2134
go.temporal.io/sdk/internal.(*activityTaskPoller).ProcessTask
	go.temporal.io/sdk@v1.26.1/internal/internal_task_pollers.go:1001
go.temporal.io/sdk/internal.(*baseWorker).processTask
	go.temporal.io/sdk@v1.26.1/internal/internal_worker_base.go:518
go.temporal.io/sdk/internal.(*baseWorker).processTaskAsync.func1
	go.temporal.io/sdk@v1.26.1/internal/internal_worker_base.go:369
```

Working with Archivematica (enduro):

```
[...]

PR #997, opened 2024-07-23
Serve internal API from enduro binary

jraddaoui:
@djjuhasz @sevein, we discussed splitting the API into an internal and an external implementations, as a less disruptive approach to #995. However, while working on that, having two API packages, designs, clients, etc. feels like a mess and already a disruptive change. I think there is a bigger discussion to have with the team about the ingest vs storage situation. In the meantime, I think this is a less drastic solution.

Refs #995.


djjuhasz (approved):
Thanks @jraddaoui — LGTM! :+1:

PR #999, opened 2024-07-29
Validate bag SIPs (fixes #947)

### LINKED ISSUE #947: Feature: Validate BagIt bags
When Enduro receives a BagIt Bag as a SIP or from pre-processing, it currently converts the bag to an Archivematica Standard Transfer and sends this to preservation as the PIP (preservation information package).  If the bag has been corrupted in delivery to Enduro then the PIP will still be sent to preservation, at the cost of unnecessary system resource usage and network traffic.

It would be better to validate Bags in Enduro to catch errors early (e.g. file corruption) and avoid unnecessary system resource or network usage. The bags should be validated just before they are zipped and delivered to preservation (see diagram below) to validate bags after possible pre-processing changes.  If bag validation fails, halt the Enduro workflow and send the SIP to a "failed SIP" location.

image

1. Validate bags in preservation. As noted, this wastes system and network resource unnecessarily.
2. Validate SIPs that are bags upon receipt. This would require a separate bag validation step for any bags produced by pre-processing.

**Additional context**

See issue #805 for more discussion of the proposed Enduro Bagit workflow.

djjuhasz:
This PR uses https://github.com/artefactual-sdps/temporal-activities/blob/main/bagit/validate_activity.go to validate a SIP that is a BagIt bag.  I've also added a "classify package" activity to determine if a given package is a BagIt bag.

I've kept the individual commits separate to make it easier to review the changes.  I'm inclined to squash the final commit that switches to the Google Distroless Docker image and leave the other commits separate for the merge, so let me know if you think that should change.


internal/enums/package_type.go
```diff
@@ -0,0 +1,10 @@
+package enums
+
+/*
+ENUM(
+Unknown
+BagIt
+Archivematica Standard Transfer
```
jraddaoui, line 7: I thought you were planning to use lower case in all enums, why not this one?

internal/fsutil/fsutil.go
```diff
@@ -72,3 +72,10 @@ func SetFileModes(root string, dirMode, fileMode int) error {
 		},
 	)
 }
+
+func FileExists(path string) bool {
+	if _, err := os.Stat(path); err != nil {
+		return !errors.Is(err, os.ErrNotExist)
```
jraddaoui: We shouldn't asume the file exists if there is a non `os.ErrNotExist` error, we should considerer that error. That's why I never came up with a general solution, but if we find one, it would be nice to have it in https://github.com/artefactual-labs/gotools/blob/main/fsutil/fsutil.go.

internal/workflow/activities/classify_package.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/artefactual-sdps/enduro/internal/bagit"
+	"github.com/artefactual-sdps/enduro/internal/enums"
+)
+
+const ClassifyPackageActivityName = "classify-package-activity"
+
+type (
+	ClassifyPackageActivity struct {
+		logger logr.Logger
```
jraddaoui, line 17: We can get the logger from temporal, as we do in the temporal-activities repo.

internal/workflow/activities/classify_package.go
```diff
@@ -0,0 +1,48 @@
...
+	ctx context.Context,
+	params ClassifyPackageActivityParams,
+) (*ClassifyPackageActivityResult, error) {
+	a.logger.V(1).Info(
+		fmt.Sprintf("Executing %s", ClassifyPackageActivityName),
+		"Path", params.Path,
+	)
+
+	r := ClassifyPackageActivityResult{Type: enums.PackageTypeUnknown}
+	if bagit.IsABag(params.Path) {
```
jraddaoui: So many layers. Not a big deal, but in this case I feel like checking the file existence here would have been better.

internal/workflow/processing.go
```diff
@@ -410,7 +416,92 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+		// Validate the bag.
+		activityOpts := withActivityOptsForLocalAction(sessCtx)
+		var result bagit_activity.ValidateActivityResult
+		err = temporalsdk_workflow.ExecuteActivity(
+			activityOpts,
+			bagit_activity.ValidateActivityName,
+			&bagit_activity.ValidateActivityParams{Path: tinfo.TempPath},
+		).Get(activityOpts, &result)
+		if err != nil || !result.Valid {
+			note := "System error"
```
jraddaoui: Should we differentiate between system and validation errors like we do in preprocessing tasks?

internal/workflow/processing.go
```diff
@@ -410,7 +416,92 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+				err = errors.New(result.Error)
+			}
+
+			e := w.completePreservationTask(
+				sessCtx,
+				datatypes.PreservationTask{
+					ID:     id,
+					Status: enums.PreservationTaskStatusError,
+					CompletedAt: sql.NullTime{
+						Time:  temporalsdk_workflow.Now(sessCtx),
```
jraddaoui: Use `UTC()` like for the StartedAt date.

internal/workflow/processing.go
```diff
@@ -410,7 +416,92 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+						Valid: true,
+					},
+					Note: note,
+				},
+			)
+			if e != nil {
+				return fmt.Errorf("complete validate bag task: %v", err)
```
jraddaoui: Not a big deal as it will be recorded in the notes and the activity logs but, should we join and return both activity errors in this case?

internal/workflow/processing.go
```diff
@@ -410,7 +416,92 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+				Status: enums.PreservationTaskStatusDone,
+				CompletedAt: sql.NullTime{
+					Time:  temporalsdk_workflow.Now(sessCtx),
+					Valid: true,
+				},
+				Note: "Bag is valid",
+			},
+		); err != nil {
+			return fmt.Errorf("complete validate bag task: %v", err)
+		}
```
jraddaoui: It would be nice to create the `datatypes.PreservationTask` based on the activity result and call the `completePreservationTask` only once (in code). Return only the validation activity error if this one doesn't fail or both errors if both activities fail.

internal/workflow/processing.go
```diff
@@ -410,7 +416,92 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+	}
+
+	// Bundle PIP as an Archivematica standard transfer.
```
jraddaoui, line 492: I guess your intention is to do this incrementally? For the AM workflow we should make sure we have a Bag here and start the transfer as Zipped Bag, not sure what was the deal for a3m.

internal/workflow/processing.go
```diff
@@ -435,12 +526,13 @@ func (w *ProcessingWorkflow) SessionHandler(
 		}
 
 		tinfo.Bundle = bundleResult
+		tinfo.PackageType = enums.PackageTypeArchivematicaStandardTransfer
```
jraddaoui, line 517: So we may see this enum value is not needed later on.

jraddaoui (changes requested):
Nice @djjuhasz! 

I'm not sure I understood the squash question, I'd squash all commits except the one that changes the Docker base image. I also think it's okay to squash all of them, we should add a note somewhere about the new requirement (glibc).

internal/enums/package_type.go
djjuhasz: I think in this case "BagIt" and "Archivematica" are more accurate and easier to read than strictly lower case representations, and we aren't using the string values for the dashboard display like we are with some of the previous enum values. I think it's overly restrictive to force all enum "labels" to be lower case, I think we should use best format for the intended use.

internal/fsutil/fsutil.go
djjuhasz: If you can figure out a better way to test for file existence in Go, then I'm all for it - but I don't know a better solution.   Returning an error from this function makes it unnecessarily complicated in my opinion.

internal/workflow/activities/classify_package.go
djjuhasz: I tried using `logger := temporal.GetLogger(ctx)` initially, but it didn't actually log anything to the Tilt UI, whereas this does.  I don't know if `temporal.GetLogger(ctx)` is just broken, or if there is some configuration problem with our dev Tile environment. :confused:

internal/workflow/activities/classify_package.go
djjuhasz: I think the advantage to using `bagit.IsABag` is that we can make the criteria for what is a "Bag" more sophisticated if we want to.  Deciding whether something is a BagIt bag solely based on the existence of "bagit.txt" file seems a bit weak.

internal/workflow/processing.go
djjuhasz: Yeah, I'm not very happy with how the error propagation worked out here, but it's difficult to change with all the layers (bagit_gython, temporal_activities, enduro).  I was just looking at doing a check like `error.Is(err, bagit_gython.ErrInvalid)` in the temporal_activities, but because the validator is injected it seems wrong to check for `bagit_gython.ErrInvalid` specifically. But then checking specifically for `bagit_gython.ErrInvalid` here seems like bad coupling of the validator implementation (bagit_gython) to this layer. :confused: 

I guess maybe I should make a bagit_gython error -> `temporal_activities/bagit` error adapter. :shrug:

internal/fsutil/fsutil.go
jraddaoui: I can't figure out a better way, I still think we should not ignore the error.

internal/workflow/activities/classify_package.go
jraddaoui: Maybe it was the `temporal` import? We do it here: https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/activities/storage.go#L43-L44

internal/workflow/processing.go
djjuhasz: Yes, incremental change. Changing the how Bundle works for AM (i.e. don't convert a bag to an AM transfer) and branching the a3m/AM logic (a3m can't accept bags IIRC) is beyond the scope of this PR in my opinion.

internal/workflow/processing.go
djjuhasz: Yeah, I think we'll probably end up with "Bagi" or "Unknown" for the package type.  I don't know how to test if a package is actually an AMST as there's no formal definition - I only know it's an AMST here because that's what the bundle activity produces.

internal/fsutil/fsutil.go
djjuhasz: Why? What scenario are you envisioning where `os.ErrNotExist` doesn't mean that the file doesn't exist and how would it negatively effect the overall workflow?  In this case if there is some wider filesystem problem (e.g. a failed mount) I believe the workflow will fail before this point.

internal/workflow/processing.go
jraddaoui: I thought it would be possible to do `err != nil` then "System error" or `!result.Valid` then "Validation error", but I'm not fully aware ATM about the final outcome of that discussion about error types you and Sevein had. Feel free to leave it as it is if you think that's okay, we'll see if there is really a need to differentiate. Maybe for now just don't use "System error".

internal/fsutil/fsutil.go
jraddaoui: `os.ErrNotExist` means that the file doesn't exist, but if you need to check it's that type is because there are other types. I'd not ignore those other types, but if you really **really** want to ignore them, I'd do it the other way around, any kind of error means the file doesn't exist.

internal/workflow/processing.go
djjuhasz: The "System error" message is only shown in the dashboard "Preservation Actions" notes column
image

(Imagine that screenshot but with "System error" instead of the "invalid: Payload-Oxum validation failed..." note)

I think in this context all the operator needs to know is that there has been some system error (as opposed to a validation error).  The actual `err.Error()` message is returned on line 484, and will show up in the Temporal UI where a sys admin or developer can see it.

jraddaoui: But that's a validation error, that's what I mean, if we are not able to differentiate them, I'd call them "Validation error" instead of "System error".

internal/fsutil/fsutil.go
djjuhasz: Okay, I've changed `FileExists()` to return false if `os.Stat()` returns any error.

internal/workflow/processing.go
djjuhasz: I need to update https://github.com/artefactual-sdps/temporal-activities/blob/main/bagit/validate_activity.go, but my plan for the returns is:

1. `bagit_activity.ValidateActivityResult{Valid: true}, nil` == validation successful, no errors!
2. `bagit_activity.ValidateActivityResult{Valid: false, Error: "invalid: ..."}, nil` == validation failed, but no system errors
3. `nil, errors.New("some error")` == system error

In case 2 `if result.Error != ""` will evaluate true, and note will be overwritten with `result.Error`.
In case 3 `if result.Error != ""` will evaluate false, and the note will be "System error".

internal/workflow/processing.go
djjuhasz: @jraddaoui in commit 8f1ad231 I've changed the code to only have one call point for `completePreservationTask()`.  Let me know if you prefer the change.

internal/workflow/activities/classify_package.go
djjuhasz: I just changed the code to:

```
logger := temporal.GetLogger(ctx)
logger.V(1).Info(
  fmt.Sprintf("Executing %s", ClassifyPackageActivityName),
  "Path", params.Path,
)
```

The import is:
```
"go.artefactual.dev/tools/temporal"
```

I confirmed that the execution of "classify-package" logs to Tilt a3m worker log before the change, but not after the change.  

I'll bet if you test https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/activities/storage.go#L43-L44 it doesn't log anything either.

internal/workflow/processing.go
jraddaoui: Neat!

internal/workflow/activities/classify_package.go
jraddaoui: Ok, I have created https://github.com/artefactual-sdps/enduro/issues/1000 to check this issue when we have some time.

jraddaoui (approved):
Thanks @djjuhasz!

Issue #1002, opened 2024-08-13
Problem: ingest workflow fails parsing dates from Archivematica


jraddaoui:
**Describe the bug**

```
2024-08-12T20:31:13.005Z	V(2)	enduro-am-worker.temporal-client	log/with_logger.go:84	error	{"Namespace": "default", "TaskQueue": "am", "WorkerID": "1@enduro-am-6c67bbb9fc-b9q9j@", "WorkflowID": "processing-workflow-f1d76fbf-bd21-4115-a582-4b7b3ac0fd3b", "RunID": "8cd48165-5e31-4a57-9c58-b58bde86fb3d", "ActivityType": "poll-transfer-activity", "Attempt": 1, "Error": "save preservation tasks: parsing time \"ul\" as \"2006-01-02T15:04:05\": cannot parse \"ul\" as \"2006\"", "error": "Activity error."}
go.temporal.io/sdk/log.(*withLogger).Error
	go.temporal.io/sdk@v1.26.1/log/with_logger.go:84
go.temporal.io/sdk/internal.(*activityTaskHandlerImpl).Execute
	go.temporal.io/sdk@v1.26.1/internal/internal_task_handlers.go:2134
go.temporal.io/sdk/internal.(*activityTaskPoller).ProcessTask
	go.temporal.io/sdk@v1.26.1/internal/internal_task_pollers.go:1001
go.temporal.io/sdk/internal.(*baseWorker).processTask
	go.temporal.io/sdk@v1.26.1/internal/internal_worker_base.go:518
go.temporal.io/sdk/internal.(*baseWorker).processTaskAsync.func1
	go.temporal.io/sdk@v1.26.1/internal/internal_worker_base.go:369
```

**To Reproduce**

Run an ingest workflow using Archivematica 1.16 or higher as preservation system.

**Expected behavior**

Workflow ends, task dates are parsed and populated when available.

**Additional context**

Issues was already reported and fixed in https://github.com/artefactual-labs/amclient-go/pull/4, but a new version has not being released since.

jraddaoui:
Fixed in #1003.

PR #1003, opened 2024-08-13
Fix issues parsing null task dates from AM

jraddaoui:
Bump amclient to v0.4.1-0.20240705155055-0c5abef5207c.

Refs #1002.


djjuhasz (approved):
Thanks @jraddaoui :+1:

PR #1004, opened 2024-08-13
Add tilt-am-knownhosts rule to Makefile

jraddaoui:
Allows to update the `.known_hosts.secret` file for Archivematica's SFTP connection using `ssh-keyscan` and restarting the required Tilt services after.


djjuhasz (approved):
LGTM :+1:

PR #1007, opened 2024-08-16
Send PIPs to Archivematica as BagIt bags

### LINKED ISSUE #805: Feature: send a BagIt bag to Archivematica for preservation

Currently, all transfers started in Archivematica use the `zipfile` transfer type:

https://github.com/artefactual-sdps/enduro/blob/main/internal/am/start_transfer.go#L49

This is not an issue in the current implementation where the transfer is always bundled as a ZIP file. However, it limits the extensibility of the workflow; thinking in the particular case of the SFA fork, where the transfer is transformed into a zipped bag in the pre-processing activities:

https://github.com/artefactual-sdps/enduro-sfa/pull/4/files#diff-ae98fc39bbc9e053ec8d1d2ed56184cd9ba7ea280d3e72975617da81c3cfadd3 

Provide a configuration setting like the one used for the processing configuration:

https://github.com/artefactual-sdps/enduro/blob/main/enduro.toml#L99

Allow changing the transfer type value in workflow. Thinking about the possibility of using child workflows to manage that extensibility, another option could be to indicate the transfer type in the child workflow result.

djjuhasz:
Fixes #805

- Remove the bundle activity from the AM processing workflow
- Change the package type to "zipped bag" when starting a transfer via
  the Archivematica API
- Bag the PIP before sending it to Archivematica (if it's not already a
  bag)
- Add a "TransferSourcePath" config value to specify the API path to the
  Transfer Source directory where PIPs are uploaded


hack/kube/overlays/dev-am/enduro-am.yaml
```diff
@@ -81,6 +81,11 @@ spec:
                 secretKeyRef:
                   name: enduro-am-secret
                   key: api_key
+            - name: ENDURO_AM_TRANSFERSOURCEPATH
+              valueFrom:
+                secretKeyRef:
+                  name: enduro-am-secret
+                  key: transfer_source_path
```
jraddaoui, line 88: If you are planning to include it in the secret, please, add it in the documentation:

https://github.com/artefactual-sdps/enduro/blob/main/docs/src/dev-manual/archivematica.md?plain=1#L78

I also think it's okay not to consider it in the secret, as it's done with some of the other config options.

internal/workflow/processing.go
```diff
@@ -803,51 +772,98 @@ func (w *ProcessingWorkflow) waitForReview(ctx temporalsdk_workflow.Context) *pa
...
+			&bagit_activity.CreateBagActivityParams{SourcePath: tinfo.TempPath},
+		).Get(lctx, &zipResult)
+		if err != nil {
+			return err
+		}
+		tinfo.PackageType = enums.PackageTypeBagIt
+	}
+
+	// Zip PIP.
+	activityOpts := withActivityOptsForLongLivedRequest(ctx)
```
jraddaoui: Should we use `withActivityOptsForLocalAction(ctx)` instead of `withActivityOptsForLongLivedRequest(ctx)` for the zip activity too?

internal/workflow/processing_test.go
```diff
@@ -663,21 +666,37 @@ func (s *ProcessingWorkflowTestSuite) TestAMWorkflow() {
...
+		pkgsvc,
+		mock.AnythingOfType("*workflow.updatePackageLocalActivityParams"),
+	).Return(nil, nil)
+	// s.env.OnActivity(
+	// 	setStatusLocalActivity,
+	// 	mock.Anything,
+	// 	mock.Anything,
+	// 	mock.Anything,
+	// 	mock.Anything,
+	// ).Return(nil, nil).Never()
```
jraddaoui: ?

jraddaoui (changes requested):
Thanks @djjuhasz, this looks great!

Some minor comments below and a bigger request after seeing the changes from both PRs (more related to #999 than to this one).

We decided at some point that preprocessing child workflows MUST return a valid Bag. If a preprocessing child workflow has been executed and the package is not identified as a Bag, we should fail the workflow instead of making it a Bag.

Check the following diagram that @sallain wrote a while ago, it doesn't include the identify activity but it tries to represent that, if a child workflow was run, we must have a valid Bag.

https://github.com/artefactual-sdps/enduro/issues/805#issuecomment-2040656506

internal/workflow/processing_test.go
djjuhasz: @jraddaoui oh yeah, I meant to delete that call.  That was at line 669 before I re-wrapped the lines — it wasn't commented, but as far as I can tell it doesn't do anything because of the `Never()` method at the end there. :shrug:

hack/kube/overlays/dev-am/enduro-am.yaml
djjuhasz: Okay, I'll add it to the docs.  

I don't think it needs to be a secret, except that I want to be able to change the transfer source path in dev without modifying `enduro.toml` (or any other files that might get commited) and there's no other way to do that AFAIK.

internal/workflow/processing.go
djjuhasz: Yeah, good idea. I can't imagine zipping a PIP to take longer than an hour. :sweat_smile:

djjuhasz:
@jraddaoui okay, I've addressed your comments and the workflow will now stop if preprocessing returns a SIP that is not a bag.  I also had to fight with prettier a bit about formatting markdown, and made some changes to the preprocessing_test script to get the preprocessing test working and for my sanity trying to figure out where preservation tasks are being created and completed.

I'll squash all the changes before merging.

jraddaoui (approved):
Thanks @djjuhasz!

PR #1009, opened 2024-08-23
Open localhost:9002 in Tilt

djjuhasz:
Open the internal API port (localhost:9002) in the Tilt dev deployment to allow access to the localhost:9002/swagger/swagger.json resource.


jraddaoui (approved):
Thanks!

Issue #1000, opened 2024-08-05
Problem: unable to get the temporal logger from context in workflows/activities


jraddaoui:
Using "go.artefactual.dev/tools/temporal", `temporal.GetLogger(ctx)` returns a logger that doesn't log anything.

https://github.com/artefactual-sdps/enduro/pull/999#discussion_r1697729840

PR #1015, opened 2024-09-02
Use logger interceptor in Temporal workers

jraddaoui:
Follow https://enduro.readthedocs.io/dev-manual/logging recommendations.

- Use `go.artefactual.dev/tools/temporal` to set the clients logger.
- Use replay-aware Temporal SDK logger from processing workflow.
- Add logger interceptor to all workers.
- Use `go.artefactual.dev/tools/temporal` `GetLogger` in activities.
- Remove logger usage from local activities.

Refs #1000.


cmd/enduro-a3m-worker/main.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/pflag"
 	"go.artefactual.dev/tools/log"
+	temporal_tools "go.artefactual.dev/tools/temporal"
```
sevein, line 25: If you like, you can enforce this alias here: https://github.com/artefactual-sdps/enduro/blob/8aa0da8a1c1afd7cceaabfeac1abdbfd7393f9f8/.golangci.yml#L41-L47

cmd/enduro-a3m-worker/main.go
```diff
@@ -212,6 +213,9 @@ func main() {
 			EnableSessionWorker:                true,
 			MaxConcurrentSessionExecutionSize:  cfg.A3m.Capacity,
 			MaxConcurrentActivityExecutionSize: 1,
+			Interceptors: []temporalsdk_interceptor.WorkerInterceptor{
+				temporal_tools.NewLoggerInterceptor(logger.WithName("a3m-worker")),
```
sevein: The logger was already named `enduro-a3m-worker` (using `appName`) so I believe that the name of this child logger would be `enduro-a3m.worker.a3m-worker`, is that what you want?

internal/am/upload_transfer.go
```diff
@@ -57,7 +51,8 @@ func (a *UploadTransferActivity) Execute(
 	ctx context.Context,
 	params *UploadTransferActivityParams,
 ) (*UploadTransferActivityResult, error) {
-	a.logger.V(1).Info("Execute UploadTransferActivity", "SourcePath", params.SourcePath)
+	logger := temporal_tools.GetLogger(ctx)
+	logger.V(1).Info("Execute UploadTransferActivity", "SourcePath", params.SourcePath)
```
sevein, line 55: A thought for the future: I've always seen this logging entry as unnecessary becuse the SDK produces a similar event automatically if I remember correctly, it does not log local variables or parameters but inputs are available in the Temporal workflow history anyways.

sevein (approved):
Nice, I had not realized we were not doing this!

cmd/enduro-a3m-worker/main.go
jraddaoui: I don't like to enforce it.

cmd/enduro-a3m-worker/main.go
jraddaoui: Good point, maybe not appending anything works better in this case.

internal/am/upload_transfer.go
jraddaoui: I said the same in https://github.com/artefactual-sdps/temporal-activities/pull/3#discussion_r1566247913. But @djjuhasz likes it.

jraddaoui:
Thanks @sevein!

internal/am/upload_transfer.go
djjuhasz: @sevein I find the log entry helpful for debugging failures where Temporal schedules the activity (so it shows up in the workflow history) but the worker doesn't actually run the `Execute()` method.  Echoing the parameters is probably unnecessary in that case though.

sevein: Perhaps that can be done from the worker interceptor too (ActivityInboundInterceptor), I will investigate.

Issue #977, opened 2024-06-25
Problem: package upload API endpoint is in an unnecessary service


jraddaoui:
**Is your feature request related to a problem? Please describe.**

There is an entirely different service in the API called `upload`, with a single endpoint `/upload` used to upload SIPs to the MinIO bucket that triggers the ingest workflow.

**Describe the solution you'd like**

I'd like to remove that entire service and move that endpoint to the existing package service.

**Describe alternatives you've considered**

None.

**Additional context**

When we discussed the scopes/attributes needed for access control we decided to use `package:upload` in this endpoint, which evidences the need to move it to the package service.

PR #1016, opened 2024-09-03
Move API upload endpoint to package service

jraddaoui:
- Remove the upload service and package.
- Set up and close the upload bucket in main.
- Update configuration:
  - Make upload max size configurable.
  - Use `go.artefactual.dev/tools/bucket` config struct.

Refs #977.


sevein (approved):
Nice cleanup!

sevein (commented):
For the record, I'm now remembering that a missing piece for this upload method to become super useful is to return an identifier, e.g. the workflow identifier, so a client can pull processing state. I think that we had an issue describing this limitation, but I can't find it.

jraddaoui:
This endpoint only proxies the upload to a bucket, the workflow is then triggered like a normal MinIO deposit, if properly configured.

Issue #951, opened 2024-05-31
Problem: PREMIS files are not validated


sallain:
**Is your feature request related to a problem? Please describe.**

Whether generated by Enduro (through a child workflow) or included in a SIP, PREMIS XML files should be validated before the package is sent to preservation. Archivematica/a3m can parse a PREMIS file to add the file's events to the AIP METS, which happens quite late in the AM/a3m workflow - ensuring that the PREMIS file is valid will hopefully avoid errors at this late point.

**Describe the solution you'd like**

Add a new activity to validate the premis.xml file against the [PREMIS v3 schema](https://www.loc.gov/standards/premis/v3/premis.xsd) before sending to AM/a3m, ensuring that it's well-formed and valid.

PREMIS files generated by Enduro child workflows should always be validated. A PREMIS file included in a transfer may have been validated in advance, so it might not be necessary to validate these. A reasonable approach might be to validate any PREMIS file in the SIP's `metadata` directory, regardless of origin, as this is the file that will be picked up by Archivematica/a3m.

**Describe alternatives you've considered**

None

**Additional context**

* There is a related task to generate a PREMIS file based on child workflow tasks - [preprocessing-sfa#19](https://github.com/artefactual-sdps/preprocessing-sfa/issues/19)
* Information about this activity should be visible in the Enduro UI, as described in #949 
* I've added a hand-tooled, valid, multi-object sample PREMIS file in the comments

sallain:
[premis-annotated-multi.zip](https://github.com/user-attachments/files/15516826/premis-annotated-multi.zip)

sallain:
Note: I've only listed validating against the schema as a first iteration. Other checks might include:

* Ensuring that each object listed in the PREMIS file exists in the transfer
* Ensuring that identifiers (for objects, events, and agents) are unique, and replacing them with unique IDs if needed
* Ensuring that there are no extra agents
* and possibly others!

fiver-watson:
Note additionally that this is something that will be used repeatedly for any Enduro user performing custom ingest activities that might generate PREMIS, and/or anyone submitting their own PREMIS files with a SIP.  For this reason, ideally this will be implemented as a reusable temporal activity, rather than a client-specific child workflow.

fiver-watson:
@mcantelon also, as discussed in the meeting today: 

Let's make this a general "Validate XML" task for its first pass, that can accept both a file to validate and a schema file to use for the validation as inputs.

mcantelon:
[...]

PR #1023, opened 2024-09-13
Check PREMIS validity after preprocessing

mcantelon:
Refs #951.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,9 @@
 ARG TARGET=enduro
 ARG GO_VERSION
 
+FROM alpine:3.20 AS build-libxml
+RUN apk add --no-cache libxml2-utils
+
 FROM golang:${GO_VERSION}-alpine AS build-go
 WORKDIR /src
 ENV CGO_ENABLED=0
@@ -61,10 +64,22 @@ FROM base AS enduro-a3m-worker
 COPY --from=build-enduro-a3m-worker --link /out/enduro-a3m-worker /home/enduro/bin/enduro-a3m-worker
 COPY --from=build-enduro-a3m-worker --link /src/enduro.toml /home/enduro/.config/enduro.toml
 CMD ["/home/enduro/bin/enduro-a3m-worker", "--config", "/home/enduro/.config/enduro.toml"]
+COPY hack/xsd/premis.xsd /home/enduro/premis.xsd
+COPY --from=build-libxml /usr/bin/xmllint /usr/bin/xmllint
+COPY --from=build-libxml /usr/lib/libxml2.so.2 /usr/lib/libxml2.so.2
+COPY --from=build-libxml /lib/ld-musl-x86_64.so.1 /lib/ld-musl-x86_64.so.1
+COPY --from=build-libxml /lib/libz.so.1 /lib/libz.so.1
+COPY --from=build-libxml /usr/lib/liblzma.so.5 /usr/lib/liblzma.so.5
 
 FROM base AS enduro-am-worker
 COPY --from=build-enduro-am-worker --link /out/enduro-am-worker /home/enduro/bin/enduro-am-worker
 COPY --from=build-enduro-am-worker --link /src/enduro.toml /home/enduro/.config/enduro.toml
 CMD ["/home/enduro/bin/enduro-am-worker", "--config", "/home/enduro/.config/enduro.toml"]
+COPY hack/xsd/premis.xsd /home/enduro/premis.xsd
+COPY --from=build-libxml /usr/bin/xmllint /usr/bin/xmllint
+COPY --from=build-libxml /usr/lib/libxml2.so.2 /usr/lib/libxml2.so.2
+COPY --from=build-libxml /lib/ld-musl-x86_64.so.1 /lib/ld-musl-x86_64.so.1
+COPY --from=build-libxml /lib/libz.so.1 /lib/libz.so.1
+COPY --from=build-libxml /usr/lib/liblzma.so.5 /usr/lib/liblzma.so.5
 
 FROM ${TARGET}
--- a/internal/premis/config_test.go
+++ b/internal/premis/config_test.go
@@ -1,0 +1,68 @@
+package premis_test
+
+import (
+	"fmt"
+	"testing"
+
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/fs"
+
+	"github.com/artefactual-sdps/enduro/internal/premis"
+)
+
+func TestConfig(t *testing.T) {
+	t.Parallel()
+
+	type test struct {
+		name    string
+		config  *premis.Config
+		wantErr string
+	}
+
+	// Non-existent XSD path.
+	badXSDfs := fs.NewDir(t, "", fs.WithFile("missing.xsd", ""))
+	badXSDPath := badXSDfs.Join("missing.xsd")
+	badXSDfs.Remove()
+
+	for _, tt := range []test{
+		{
+			name: "Passes validation (disabled)",
+			config: &premis.Config{
+				Enabled: false,
+			},
+		},
+		{
+			name: "Fails validation (missing XSD path)",
+			config: &premis.Config{
+				Enabled: true,
+			},
+			wantErr: "xsdPath is required in the [validatePremis] configuration when enabled",
+		},
+		{
+			name: "Fails validation (missing XSD file)",
+			config: &premis.Config{
+				Enabled: true,
+				XSDPath: badXSDPath,
+			},
+			wantErr: fmt.Sprintf("xsdPath in [validatePremis] not found: stat %s: no such file or directory", badXSDPath),
+		},
+		{
+			name: "Passes validation (enabled)",
+			config: &premis.Config{
+				Enabled: true,
+				XSDPath: fs.NewDir(t, "", fs.WithFile("empty.xsd", "")).Join("empty.xsd"),
+			},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			err := tt.config.Validate()
+			if tt.wantErr != "" {
+				assert.Error(t, err, tt.wantErr)
+				return
+			}
+			assert.NilError(t, err)
+		})
+	}
+}
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"path/filepath"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/artefactual-sdps/temporal-activities/archiveextract"
@@ -20,6 +21,7 @@ import (
 	"github.com/artefactual-sdps/temporal-activities/bagvalidate"
 	"github.com/artefactual-sdps/temporal-activities/bucketupload"
 	"github.com/artefactual-sdps/temporal-activities/removepaths"
+	"github.com/artefactual-sdps/temporal-activities/xmlvalidate"
 	"github.com/google/uuid"
 	"go.artefactual.dev/tools/ref"
 	temporal_tools "go.artefactual.dev/tools/temporal"
@@ -834,6 +836,15 @@ func (w *ProcessingWorkflow) transferA3m(
 	tinfo.SendToFailed.ActivityName = activities.SendToFailedPIPsName
 	tinfo.SendToFailed.NeedsZipping = true
 
+	err := w.validatePREMIS(
+		sessCtx,
+		filepath.Join(tinfo.Bundle.FullPath, "metadata", "premis.xml"),
+		tinfo.PreservationActionID,
+	)
+	if err != nil {
+		return err
+	}
+
 	// Send PIP to a3m for preservation.
 	{
 		activityOpts := temporalsdk_workflow.WithActivityOptions(sessCtx, temporalsdk_workflow.ActivityOptions{
@@ -881,6 +892,15 @@ func (w *ProcessingWorkflow) transferAM(ctx temporalsdk_workflow.Context, tinfo 
 			return err
 		}
 		tinfo.PackageType = enums.PackageTypeBagIt
+	}
+
+	err = w.validatePREMIS(
+		ctx,
+		filepath.Join(tinfo.TempPath, "data", "metadata", "premis.xml"),
+		tinfo.PreservationActionID,
+	)
+	if err != nil {
+		return err
 	}
 
 	// Zip PIP.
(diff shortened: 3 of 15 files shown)


Dockerfile
```diff
@@ -3,6 +3,9 @@
 ARG TARGET=enduro
 ARG GO_VERSION
 
+FROM golang:${GO_VERSION}-bookworm AS build-libxml
```
jraddaoui: Why a Go image and not just basic Alpine to install and copy the `xmllint` binary and deps.

internal/workflow/processing.go
```diff
@@ -439,6 +440,16 @@ func (w *ProcessingWorkflow) SessionHandler(
...
 
+	// Validate PREMIS.
+	activityOpts := withActivityOptsForRequest(sessCtx)
+	err := temporalsdk_workflow.ExecuteActivity(activityOpts, xmlvalidate.Name, xmlvalidate.Params{
+		XMLPath: filepath.Join(tinfo.TempPath, "data", "metadata", "premis.xml"),
+		XSDPath: "/home/enduro/premis.xsd",
+	}).Get(activityOpts, nil)
+	if err != nil {
+		return err
+	}
```
jraddaoui: I think this should happen in two different points in the workflow instead of right after preprocessing:

- For AM, this should happen after bag creation/validation (to be sure of the PREMIS XML path inside the SIP).
- For a3m, it should occur after the bundle activity for the same reason, as that will be the PIP received by a3m. In this case it won't be a bag, so the XMLPath will need to change.

internal/workflow/processing.go
```diff
@@ -439,6 +440,16 @@ func (w *ProcessingWorkflow) SessionHandler(
 		return err
 	}
 
+	// Validate PREMIS.
+	activityOpts := withActivityOptsForRequest(sessCtx)
+	err := temporalsdk_workflow.ExecuteActivity(activityOpts, xmlvalidate.Name, xmlvalidate.Params{
+		XMLPath: filepath.Join(tinfo.TempPath, "data", "metadata", "premis.xml"),
+		XSDPath: "/home/enduro/premis.xsd",
```
jraddaoui: This should be configurable:

```
[validatePremis]
xsdPath = "/home/enduro/premis.xsd"
```

Or something like that. It will require changes in the config package.

internal/workflow/processing.go
jraddaoui: We should check if the error is a missing `premis.xml` file (shouldn't be considered as an error in workflow) or other type of error (should fail the workflow).

internal/workflow/processing.go
```diff
@@ -439,6 +440,16 @@ func (w *ProcessingWorkflow) SessionHandler(
 		return err
 	}
 
+	// Validate PREMIS.
+	activityOpts := withActivityOptsForRequest(sessCtx)
+	err := temporalsdk_workflow.ExecuteActivity(activityOpts, xmlvalidate.Name, xmlvalidate.Params{
```
jraddaoui: We also need to record a preservation task for this activity and persist the errors from it, to be able to report them in the API/dashboard.

internal/workflow/processing.go
jraddaoui: As noted in Slack, it should also be optional, so maybe:

```
[validatePremis]
enabled = true
xsdPath = "/home/enduro/premis.xsd"
```

It could also be checked on the `xsdPath` var directly if you don't want to add an extra var. If `xsdPath` is empty, validation is disabled.

jraddaoui (changes requested):
Thanks @mcantelon! Check my comments below.

internal/workflow/processing.go
jraddaoui: Not sure if it will require changes in https://github.com/artefactual-sdps/temporal-activities/blob/main/xmlvalidate/activity.go#L45. You may be able to check if it is an `os.IsNotExist(err)` error, but need to be careful about the file (it could be a missing XSD file).

internal/workflow/processing.go
jraddaoui: Thinking about possible errors from the activity, we should check the existence of the XSD file when reading the config. We don't want to let the workers start, knowing all workflows will fail in that activity.

Right now, we use the same config file for all services/containers, so this check will fail for the `enduro` container, as you are only copying the XSD file in the `enduro-am-worker` and the `enduro-a3m-worker` containers. You should be able to solve that issue changing that config value in the environment:

https://github.com/artefactual-sdps/enduro/blob/main/hack/kube/base/enduro.yaml#L30

Adding a new env. var., based on the example:

```
- name: ENDURO_VALIDATE_PREMIS_ENABLED
  value: "false"
```

If the config validation checks for `Enabled` first before validating the existence of the XSD file, then you don't need to overwrite the XSD path.

Similar, if you decide to use a single var:

```
- name: ENDURO_VALIDATE_PREMIS_XSD_PATH
  value: ""
```

If the config validation sees an empty value it doesn't check of its existence.

jraddaoui: Config validation example:

https://github.com/artefactual-sdps/enduro/blob/main/internal/config/config.go#L50
https://github.com/artefactual-sdps/enduro/blob/main/internal/config/config.go#L68
https://github.com/artefactual-sdps/enduro/blob/main/internal/preprocessing/preprocessing.go#L48-L62

internal/workflow/processing.go
jraddaoui: No need to worry if you already validated the config XSD path, it must be a missing XML.

internal/workflow/processing.go
```diff
@@ -1188,3 +1207,61 @@ func (w *ProcessingWorkflow) sendToFailedBucket(
...
+	if err != nil {
+		pt.Status = enums.PreservationTaskStatusError
+		pt.Note = "System error"
+	}
+	if len(xmlvalidateResult.Failures) > 0 {
+		pt.Status = enums.PreservationTaskStatusError
+		var note string
+		for _, failure := range xmlvalidateResult.Failures {
+			note = fmt.Sprintf("%s\n", failure)
+		}
```
jraddaoui: @mcantelon This is not appending to the note, it will just use the last failure. Also, we should be able to use `pt.Note` directly instead of creating `note`.

Dockerfile
mcantelon: I thought Alpine wasn't working when I'd tried it earlier, at your suggestion, but I'd misunderstood something and it actually *was* working. :facepalm:

internal/workflow/processing.go
mcantelon: I've fixed that and cleaned it up slightly. The display of preservation task notes in Enduro doesn't yet reflect line breaks in the text so the results appear as a jumble, but the info gets there at least.

jraddaoui: It should reflect those line breaks since last week ;) Maybe this needs a rebase?

Dockerfile
jraddaoui: Nice, a lot cleaner, thanks!

internal/premis/config_test.go
```diff
@@ -0,0 +1,45 @@
...
+				Enabled: false,
+			},
+		},
+		{
+			name: "Fails validation (missing XmlPath)",
+			config: &premis.Config{
+				Enabled: true,
+			},
+			wantErr: "xsdPath is required in the [validatePremis] configuration when enabled",
+		},
```
jraddaoui, line 40: We should add a couple of tests that check the XSD file existence. Not sure if you ever used `gotest.tools/v3/fs`, but you could create the XSD file directly in the config. For example:

For the missing XSD, just pass a fake path:

```
{
	name: "Fails validation (missing XSD file)",
	config: &premis.Config{
		Enabled: true,
		XSDPath: "/fake/path/to.xsd",
	},
	wantErr: "xsdPath in [validatePremis] not found",
},
```

To check that it actually exists:

```
{
	name: "Passes validation (enabled)",
	config: &premis.Config{
		Enabled: true,
		XSDPath: fs.NewDir(t, "", fs.WithFile("empty.xsd", "")).Join("empty.xsd"),
	},
},
```

internal/premis/config_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	for _, tt := range []test{
+		{
+			name: "Passes validation (disabled)",
+			config: &premis.Config{
+				Enabled: false,
+			},
+		},
+		{
+			name: "Fails validation (missing XmlPath)",
```
jraddaoui: Should read `Fails validation (missing XSD path)`.

internal/workflow/processing.go
```diff
@@ -1188,3 +1208,72 @@ func (w *ProcessingWorkflow) sendToFailedBucket(
...
+
+	// Attempt to validate PREMIS.
+	var xmlvalidateResult xmlvalidate.Result
+	activityOpts := withActivityOptsForLocalAction(ctx)
+	err = temporalsdk_workflow.ExecuteActivity(activityOpts, xmlvalidate.Name, xmlvalidate.Params{
+		XMLPath: xmlPath,
+		XSDPath: w.cfg.ValidatePREMIS.XSDPath,
+	}).Get(activityOpts, &xmlvalidateResult)
+
+	// Allow PREMIS XML to not exist without failing workflow.
```
jraddaoui: Move this comment to the second if.

internal/workflow/processing.go
```diff
@@ -1188,3 +1208,72 @@ func (w *ProcessingWorkflow) sendToFailedBucket(
...
+		}
+	}
+
+	// Set preservation task status to error and add PREMIS validation failures to note.
+	if len(xmlvalidateResult.Failures) > 0 {
+		pt.Status = enums.PreservationTaskStatusError
+		pt.Note = "PREMIS is invalid\n"
+
+		for _, failure := range xmlvalidateResult.Failures {
+			pt.Note += fmt.Sprintf("\n%s\n", failure)
```
jraddaoui: These seem like a lot of line breaks, better just `%s\n`. Ideally, the last failure should just be`%s`, whitout line break.

internal/workflow/processing_test.go
```diff
@@ -171,6 +177,10 @@ func (s *ProcessingWorkflowTestSuite) setupAMWorkflowTest(
 		bagcreate.New(bagcreate.Config{}).Execute,
 		temporalsdk_activity.RegisterOptions{Name: bagcreate.Name},
 	)
+	s.env.RegisterActivityWithOptions(
+		xmlvalidate.New(xmlvalidate.NewXMLLintValidator()).Execute,
+		temporalsdk_activity.RegisterOptions{Name: xmlvalidate.Name},
+	)
```
jraddaoui: This shouldn't be needed, it's already registered for both workflow branches above.

internal/workflow/processing_test.go
```diff
@@ -221,6 +231,10 @@ func (s *ProcessingWorkflowTestSuite) setupA3mWorkflowTest(
 		activities.NewBundleActivity().Execute,
 		temporalsdk_activity.RegisterOptions{Name: activities.BundleActivityName},
 	)
+	s.env.RegisterActivityWithOptions(
+		xmlvalidate.New(xmlvalidate.NewXMLLintValidator()).Execute,
+		temporalsdk_activity.RegisterOptions{Name: xmlvalidate.Name},
+	)
```
jraddaoui: This shouldn't be needed, it's already registered for both workflow branches above.

internal/workflow/processing_test.go
```diff
@@ -321,6 +335,12 @@ func (s *ProcessingWorkflowTestSuite) TestPackageConfirmation() {
 		nil,
 	)
 
+	s.env.OnActivity(xmlvalidate.Name, sessionCtx,
+		&xmlvalidate.Params{XMLPath: filepath.Join(transferPath, "metadata/premis.xml"), XSDPath: ""},
+	).Return(
+		&xmlvalidate.Result{Failures: []string{}}, nil,
+	)
+
```
jraddaoui, line 365: It's rare that we don't need to add the create/complete preservation action activities calls.

internal/workflow/processing_test.go
```diff
@@ -321,6 +335,12 @@ func (s *ProcessingWorkflowTestSuite) TestPackageConfirmation() {
 		nil,
 	)
 
+	s.env.OnActivity(xmlvalidate.Name, sessionCtx,
+		&xmlvalidate.Params{XMLPath: filepath.Join(transferPath, "metadata/premis.xml"), XSDPath: ""},
```
jraddaoui: Add the PREMIS validation config including the XSD path, to check that value here.

jraddaoui: Also, `filepath.Join(transferPath, "metadata", "premis.xml")`.

internal/workflow/processing_test.go
```diff
@@ -482,6 +502,42 @@ func (s *ProcessingWorkflowTestSuite) TestAutoApprovedAIP() {
...
+			PkgSvc: pkgsvc,
+			RNG:    s.workflow.rng,
+			PreservationTask: datatypes.PreservationTask{
+				Name:   "Validate PREMIS",
+				Status: enums.PreservationTaskStatusInProgress,
+				StartedAt: sql.NullTime{
+					Time:  startTime,
+					Valid: true,
+				},
+				PreservationActionID: 0,
```
jraddaoui: It would be nice to create a variable (`paID = 1`) at the start of the test, return that var instead of `0` from the create preservation action activity and reuse that var and all the create preservation task calls.

internal/workflow/processing.go
```diff
@@ -1188,3 +1208,72 @@ func (w *ProcessingWorkflow) sendToFailedBucket(
 
 	return nil
 }
+
+func (w *ProcessingWorkflow) validatePREMIS(
+	ctx temporalsdk_workflow.Context,
+	xmlPath string,
+	paID int,
+) error {
+	id, err := w.createPreservationTask(
```
jraddaoui, line 1222: **Now that I see the tests, we forgot to check the enabled config before running all this. It should be easy, just check the config at the start and return nil directly if it is not enabled.**

After adding that, all the tests won't run this activities unless you enabled the validation in each test configuration. You can only enable it in one of the tests, to avoid making changes to all of them.

jraddaoui (changes requested):
Pretty close, thanks @mcantelon!

internal/workflow/processing_test.go
```diff
@@ -258,6 +265,7 @@ func (s *ProcessingWorkflowTestSuite) TestPackageConfirmation() {
 		Storage: storage.Config{
 			DefaultPermanentLocationID: locationID,
 		},
+		ValidatePREMIS: premis.Config{Enabled: true},
```
jraddaoui: Add an XSDPath too, that way you can check it in the activity params.

mcantelon:
This should be ready for another review. Thanks!

internal/workflow/processing_test.go
```diff
@@ -258,10 +265,16 @@ func (s *ProcessingWorkflowTestSuite) TestPackageConfirmation() {
 		Storage: storage.Config{
 			DefaultPermanentLocationID: locationID,
 		},
+		ValidatePREMIS: premis.Config{
+			Enabled: true,
+			XSDPath: "/home/enduro/premis.xsd",
+		},
```
jraddaoui, line 271: These test are getting really long. I'd only enable PREMIS validation here ...

internal/workflow/processing_test.go
```diff
@@ -571,6 +666,10 @@ func (s *ProcessingWorkflowTestSuite) TestAMWorkflow() {
 		A3m:          a3m.Config{ShareDir: s.CreateTransferDir()},
 		Preservation: pres.Config{TaskQueue: temporal.AmWorkerTaskQueue},
 		Storage:      storage.Config{DefaultPermanentLocationID: amssLocationID},
+		ValidatePREMIS: premis.Config{
+			Enabled: true,
+			XSDPath: "/home/enduro/premis.xsd",
+		},
```
jraddaoui, line 636: ... and here. That way it will be tested in the a3m and the AM branches from the workflow. All other test cases could remain as they are.

Keep the preservation action/task IDs changes in all of them, those are great. But remove the config and the activity call from all except these two.

jraddaoui (changes requested):
Thanks @mcantelon! Just a minor cleanup missed from the last feedback, to not increase a lot the workflow tests length.

internal/workflow/processing_test.go
jraddaoui: Thanks for doing it across all tests.

jraddaoui:
And for [the failing check](https://github.com/artefactual-sdps/enduro/actions/runs/11741933363/job/32711630401?pr=1023) just run `make golines`.

mcantelon:
Thanks again @radda! Ready for another look!

internal/workflow/processing_test.go
```diff
@@ -372,10 +426,16 @@ func (s *ProcessingWorkflowTestSuite) TestAutoApprovedAIP() {
 		Storage: storage.Config{
 			DefaultPermanentLocationID: locationID,
 		},
+		ValidatePREMIS: premis.Config{
+			Enabled: false,
+		},
```
jraddaoui: We can remove these three lines too, it's the default.

internal/workflow/processing_test.go
```diff
@@ -717,6 +824,9 @@ func (s *ProcessingWorkflowTestSuite) TestPackageRejection() {
...
```
jraddaoui: Same

internal/workflow/processing_test.go
```diff
@@ -832,10 +942,15 @@ func (s *ProcessingWorkflowTestSuite) TestPreprocessingChildWorkflow() {
...
```
jraddaoui: Same.

internal/workflow/processing_test.go
```diff
@@ -1203,10 +1318,14 @@ func (s *ProcessingWorkflowTestSuite) TestFailedPIPA3m() {
 		A3m:          a3m.Config{ShareDir: s.CreateTransferDir()},
 		Preservation: pres.Config{TaskQueue: temporal.A3mWorkerTaskQueue},
 		Storage:      storage.Config{DefaultPermanentLocationID: locationID},
+		ValidatePREMIS: premis.Config{
+			Enabled: false,
+		},
```
jraddaoui: Same.

internal/workflow/processing_test.go
```diff
@@ -1366,10 +1485,14 @@ func (s *ProcessingWorkflowTestSuite) TestFailedPIPAM() {
 	cfg := config.Configuration{
 		Preservation: pres.Config{TaskQueue: temporal.AmWorkerTaskQueue},
 		Storage:      storage.Config{DefaultPermanentLocationID: amssLocationID},
+		ValidatePREMIS: premis.Config{
+			Enabled: false,
+		},
```
jraddaoui: Same.

jraddaoui (approved):
Thanks @mcantelon!

To save 15 more line in the tests, we could remove the default PREMIS config from them. Make sure you squash and change the commit message, maybe "before preservation" works better than "after preprocessing" now.

mcantelon:
Thanks @jraddaoui !

PR #1024, opened 2024-09-19
Upgrade entgo version 0.13.1 -> 0.14.1

djjuhasz:
- Upgrade go version 1.22.6 -> 1.23.1 (required by entgo 0.14.1)
- Upgrade entgo version 0.13.1 -> 0.14.1
- Upgrade golangci-lint version 1.58.1 -> 1.61.0


jraddaoui (changes requested):
Thanks @djjuhasz!

I had to upgrade `golangci-lint` to at least 1.60.1 to make it work with Go 1.23.x, and there seems to be some Go mod issues too.

jraddaoui (approved):

Issue #988, opened 2024-07-09
Problem: Packages count only ever counts packages listed on current page


fiver-watson:
## Description

The Packages browse page in Enduro includes a count of page results and total packages, just below the page header. However, this counter constantly only ever reads `20/20` - it is currently only ever listing the total package count as the current total page count. 

**To Reproduce**

1. Log into enduro
2. Make sure you have more than 20 packages processed
3. Navigate to the Packages page
4. Look at the count below the Packages header

**Error encountered**

Package count reads: 

> Showing 20/20

## Expected outcome

This counter should: 

* List the number of results on the current page as the first value, and
* List the TOTAL number of ingested packages as the second value

djjuhasz:
Fixing the total package count is going to require changes to the Enduro API - the current `/package` GET endpoint doesn't return a count of the total number of packages stored in Enduro.  This change will require changes to the Goa API layer, the package service layer, and the persistence layer so this is not a good first issue to tackle.

sallain:
Thanks for the context, @djjuhasz!

fiver-watson:
👍 LGTM! Also, I personally did end up liking have the results count both at the top and the bottom of the page. Even on a relatively large monitor, , the 20 results regularly went beyond my screen, so having that count repeated at the bottom by the pager (esp given that the pager doesn't reload the page fully and bring you back to the top) works well.

Issue #514, opened 2023-06-16
Problem: we don't have a good pattern to implement filtering, ordering and pagination in persistence


Diogenesoftoronto:
sdps_preprocessing implements a generic "Filter" type that can perform filtering of Ent queriers based on user parameters, e.g. filtering, ordering, etc... being generic means that the same pattern can be applied across multiple entities effortlessly.

djjuhasz:
Commit e2cc8b34dfb7aff4270a83fd0f5d47c289c94d29 adds a general query filter implementation to the persistence and entclient layers.

PR #1025, opened 2024-09-24
Add a persistence query filter implementation

djjuhasz:
Refs #988, #514

To address #988 and provide a count of the total number of packages in the dashboard the API `GET /package` endpoint to provide total number of packages in a result set. Because we need to update the API and the underlying database calls, this is a good opportunity to also address #514 and add a general search implementation that allows filtering, ordering and paging results. This PR adds this general search filtering implementation with paging and ordering, and also provides a count of the total number of records matching the search criteria before paging.

- Add a persistence package filter stub implementation
- Add persistence Page and Sort structs
- Add an entgo query filter implementation
- Add methods to add page, sort, and column value filters to the ent filter implementation
- Add a timerange package for representing time ranges (i.e. a time period with a start and end time)

### DIFF
--- a/internal/persistence/filter.go
+++ b/internal/persistence/filter.go
@@ -1,0 +1,60 @@
+package persistence
+
+import (
+	"github.com/google/uuid"
+
+	"github.com/artefactual-sdps/enduro/internal/enums"
+	"github.com/artefactual-sdps/enduro/internal/timerange"
+)
+
+type (
+	// Sort determines how the filtered results are sorted by specifying a
+	// slice of sort columns.  The first SortColumn has the highest sort
+	// precedence, and the last SortColumn the lowest precedence.
+	Sort []SortColumn
+
+	// SortColumn specifies a column name on which to sort results, and the
+	// direction of the sort (ascending or descending).
+	SortColumn struct {
+		// Name of the column on which to sort the results.
+		Name string
+
+		// Desc is true if the sort order is descending.
+		Desc bool
+	}
+)
+
+// NewSort returns a new sort instance.
+func NewSort() Sort {
+	return Sort{}
+}
+
+// AddCol adds a SortColumn to a Sort then returns the updated Sort.
+func (s Sort) AddCol(name string, desc bool) Sort {
+	s = append(s, SortColumn{Name: name, Desc: desc})
+	return s
+}
+
+// Page represents a subset of results within a search result set.
+type Page struct {
+	// Limit is the maximum number of results per page.
+	Limit int
+
+	// Offset is the ordinal position, relative to the start of the unfiltered
+	// set, of the first result of the page.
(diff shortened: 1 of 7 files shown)


jraddaoui:
@sevein I am currently at SFA, if you have some time before I get to this, I'd appreciate if you could take a look.

internal/timerange/timerange.go
```diff
@@ -0,0 +1,23 @@
...
+	"errors"
+	"time"
+)
+
+type Range struct {
+	Start *time.Time
+	End   *time.Time
+}
+
+func New(start, end *time.Time) (Range, error) {
```
sevein: Why are pointers used for time.Time instead of values?

internal/timerange/timerange.go
```diff
@@ -0,0 +1,23 @@
...
+	// The end time must be greater than or equal to the start time.
+	if end != nil && start != nil && end.Before(*start) {
+		return Range{}, errors.New("new time range: end time cannot be before start time")
+	}
+	return Range{Start: start, End: end}, nil
+}
+
+func NewInstant(t *time.Time) Range {
```
sevein: // NewInstant returns a Range where the Start and End times are both set to the given time.

internal/persistence/filter.go
```diff
@@ -0,0 +1,59 @@
...
+	return Sort{}
+}
+
+func (s Sort) AddCol(name string, desc bool) Sort {
+	s = append(s, SortColumn{Name: name, Desc: desc})
+	return s
+}
+
+// Page filters search results to a maximum Limit at an Offset from the
+// beginning of the unfiltered result set
```
sevein: Missing period.

sevein (approved):
It looks good to me; there are some nice improvements compared to the original implementation.

Quick thought on the pull request description: it summarizes the changes well, but it might help to add a bit more background. The "what" is clear from the code itself, but it's harder to see how this addresses issue #988, especially with the query filter. A little extra context would make it easier for reviewers to follow. For example, you could mention that this PR introduces filtering capabilities like sorting and pagination in the persistence layer. The PackageFilter struct is one of the filters that will be accepted by certain methods, but that part will come in future PRs.

internal/timerange/timerange.go
djjuhasz: @sevein I'm using nil values to represent half-open ranges - a nil start or end time represents any time.  So a range with a start time of "2024-09-25 15:15:25 UTC" and a nil end time would represent any time equal to or after "2024-09-25 15:15:25 UTC".

Passing pointers to the `New()` constructor is awkward though, and I would like to improve the ergonomics. I suppose a half open range could be approximated by choosing a very early start time or a very late end time instead of using pointers, and using values would simplify things.

PR #1031, opened 2024-09-27
Add `ListPackages()` to persistence service

djjuhasz:
Refs #988

This continues the work toward #988 by adding a `ListPackages()` method to the persistence service that includes a count of the total search results returned (before paging), as well as allowing filtering, sorting, and paging of the search.

Other changes:
- Add a telemetry wrapper for the `ListPackages()` method to provide extra performance and debugging data
- Add additional commentary for the `Filter.Page()` method
- Add many unit tests for the `ListPackages()` functionality

### DIFF
--- a/internal/persistence/ent/client/package.go
+++ b/internal/persistence/ent/client/package.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/artefactual-sdps/enduro/internal/datatypes"
 	"github.com/artefactual-sdps/enduro/internal/persistence"
+	"github.com/artefactual-sdps/enduro/internal/persistence/ent/db"
+	"github.com/artefactual-sdps/enduro/internal/persistence/ent/db/pkg"
 )
 
 // CreatePackage creates and persists a new package using the values from pkg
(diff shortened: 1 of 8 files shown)


internal/persistence/ent/client/package.go
```diff
@@ -130,3 +132,56 @@ func (c *client) UpdatePackage(
...
+		Offset: f.Offset,
+		Total:  total,
+	}
+
+	return res, pp, err
+}
+
+// filterPackages filters a package query based on filtering inputs.
+func filterPackages(q *db.PkgQuery, f *persistence.PackageFilter) (page, whole *db.PkgQuery) {
+	h := NewFilter(q, SortableFields{
```
jraddaoui: Why `h`?

internal/persistence/ent/client/package_test.go
```diff
@@ -10,14 +10,26 @@ import (
...
 
+func nullUUIDToPtr(u uuid.NullUUID) *uuid.UUID {
+	var p *uuid.UUID
+	if u.Valid {
+		p = &u.UUID
+	}
+
+	return p
+}
+
```
jraddaoui: Only used twice with UUIDs already known to be valid. I's use `&u.UUID` directly in those two cases.

jraddaoui (approved):
LGTM, thanks @djjuhasz!

internal/persistence/ent/client/package.go
djjuhasz: I'm not sure — I copied it over from the STE implementation as-is.

internal/persistence/ent/client/package_test.go
djjuhasz: Removed.

internal/persistence/ent/client/package.go
djjuhasz: I'm open to other suggestions.

jraddaoui: Maybe changing the parameter var name too? 

```
func filterPackages(q *db.PkgQuery, pf *persistence.PackageFilter) (page, whole *db.PkgQuery) {
	f := NewFilter(q, SortableFields{...
```

djjuhasz: I tried changing the `f *persistence.PackageFilter` param to `pf *persistence.PackageFilter` but I didn't like that the new name didn't agree with the `ListPackages(ctx context.Context, f persistence.PackageFilter)` name — I think passing through PackageFilter with the same variable name is easier to track.

I've gone with changing the query filter name from `h` to `qf` — it's a bit longer but I kind of like that the parameters `q` and `f` become query filter `qf`. Math nerd joy :nerd_face:

Issue #1032, opened 2024-10-01
Regression: API authentication is not working in the public API


jraddaoui:
**Describe the bug**

Authentication does not work in the public API.

**To Reproduce**

Enable authentication in the API configuration and hit any endpoint without providing an access token.

**Expected behavior**

A 401 (Unauthorized) response should be returned by the API.

**Additional context**

Regression introduced in https://github.com/artefactual-sdps/enduro/commit/111aa0f5cb5dcbf6ea93a000bff7a2bef5b5fbda.

To reproduce in the development environment, change the exposed port in the Tiltfile, from 9002 to 9000:

https://github.com/artefactual-sdps/enduro/blob/main/Tiltfile#L102

jraddaoui:
Fixed in #1033.

PR #1033, opened 2024-10-01
Do not overwrite services without authentication

jraddaoui:
Recreate package and storage services for the internal API without overwriting the existing ones used in the public API.

Refs #1032.


djjuhasz (approved):
LGTM :+1:

PR #1036, opened 2024-10-02
Add total result count to package list API

djjuhasz:
Refs #988

This commit connects the API "GET /package" endpoint to the persistence layer `ListPackages()` method. The cursor request parameter is replaced by the optional limit and offset parameters. The response now returns the current page limit and offset, and the total number of results found before paging, instead of a cursor value.

These changes will break the existing cursor-based pagination in the Enduro Dashboard. I will fix the Dashboard paging in a subsequent commit.

- Add optional `limit` and `offset` parameters to the GET /package request
- Add page `limit`, `offset`, and `total` fields to GET /package response
- Use `persistence.ListPackages()` to populate API results
- Add an adapter to convert goa `package_.ListPayload` search filters to `persistence.Filter` filters
- Add tests for the conversion of the package API parameters to persistence layer parameters, and the persistence search results to API response results

### DIFF
--- a/internal/api/design/package_.go
+++ b/internal/api/design/package_.go
@@ -78,10 +78,12 @@ var _ = Service("package", func() {
 			Attribute("status", String, func() {
 				EnumPackageStatus()
 			})
-			Attribute("cursor", String, "Pagination cursor")
+			Attribute("limit", Int, "Limit number of results to return")
+			Attribute("offset", Int, "Offset from the beginning of the found set")
+
 			Token("token", String)
 		})
-		Result(PaginatedCollectionOf(StoredPackage))
+		Result(PackageList)
 		HTTP(func() {
 			GET("/")
 			Response(StatusOK)
@@ -92,7 +94,8 @@ var _ = Service("package", func() {
 				Param("latest_created_time")
 				Param("location_id")
 				Param("status")
-				Param("cursor")
+				Param("limit")
+				Param("offset")
 			})
 		})
 	})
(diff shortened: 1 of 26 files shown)


internal/api/design/package_.go
```diff
@@ -319,6 +322,20 @@ var StoredPackage = ResultType("application/vnd.enduro.stored-package", func() {
 	Required("id", "status", "created_at")
 })
 
+var Page = ResultType("application/vnd.enduro.page", func() {
+	Description("Page represents a subset of search results.")
+	Attribute("limit", Int, "Maximum items per page")
+	Attribute("offset", Int, "Offset from first result to start of page")
+	Attribute("total", Int, "Total result count before paging")
+	Required("limit", "offset", "total")
+})
```
jraddaoui: This is probably going to be reused, I'd move it to https://github.com/artefactual-sdps/enduro/blob/main/internal/api/design/pagination.go.

internal/api/design/package_.go
```diff
@@ -319,6 +322,20 @@ var StoredPackage = ResultType("application/vnd.enduro.stored-package", func() {
...
+
+var PackageList = ResultType("application/vnd.enduro.packages", func() {
+	Attribute("items", CollectionOf(StoredPackage))
+	Attribute("page", Page)
+	Required("items")
```
jraddaoui: The way you set this up, we may want to require the `page` attribute too.

jraddaoui (approved):
Really nice @djjuhasz!

PR #1043, opened 2024-10-14
Use Node.js 22 (LTS)

sevein:
Node.js 22 was released back in April and will soon be LTS.

See https://github.com/nodejs/Release.


jraddaoui (approved):
Thanks!

Issue #1035, opened 2024-10-01
Problem: it's not possible to use roles for access control


jraddaoui:
**Is your feature request related to a problem? Please describe.**

My OIDC provider doesn't allow to configure Enduro attributes in their system and it can only return a list of predefined roles on the access token.

**Describe the solution you'd like**

I'd like to be able to map our existing OIDC roles to Enduro attributes in Enduro through configuration.

PR #1044, opened 2024-10-15
Add ABAC roles mapping configuration

jraddaoui:
Allows to configure a role mapping to map roles included in the access
token to Enduro attributes for access control.

Refs #1035.


dashboard/src/stores/auth.ts
```diff
@@ -132,6 +136,19 @@ export const useAuthStore = defineStore("auth", {
         this.config.abac.claimValuePrefix =
           env.VITE_OIDC_ABAC_CLAIM_VALUE_PREFIX.trim();
       }
+      if (env.VITE_OIDC_ABAC_USE_ROLES) {
+        this.config.abac.useRoles =
+          env.VITE_OIDC_ABAC_USE_ROLES.trim().toLowerCase() === "true";
+      }
+      if (this.config.abac.useRoles && env.VITE_OIDC_ABAC_ROLES_MAPPING) {
```
djjuhasz, line 143: Is having a separate "useRoles" variable necessary, or could we just treat an empty `roles mapping` as disabling role based authentication?

djjuhasz (commented):
Looks good @jraddaoui.  I did have one question about the need for the "use roles" variable.

The operator permissions also seem fairly limited (e.g. no storage attributes), which is fine for testing but makes me wonder if it's possible to create a permission list by exclusion, e.g. `"operator": ["package:*", "storage:*", "^storage:location:create"]`.  This example is very arbitrary, I'm sure a real world "operator" permissions would be quite different.  I'm also not sure if ["*"] is equivalent to ["package:*", "storage:*"], and could be substituted in the example.

jraddaoui:
Thanks @djjuhasz!

I noted the option to exclude attributes at the bottom of [the issue description](https://github.com/artefactual-sdps/enduro/issues/957) but never got to implement it. **The attributes included in this PR are just examples and for testing purposes**, they are based on a list provided by Dan (see the [Keycloack configuration](https://github.com/artefactual-sdps/enduro/blob/main/hack/kube/components/dev/keycloak.yaml#L80-L161)), I just removed the storage attributes to simplify.

dashboard/src/stores/auth.ts
jraddaoui: I had my conflicts about this too. In the end I decided to be explicit, following the same pattern we have for the `enabled` settings for auth and ABAC.

djjuhasz (approved):
Okay, thanks @jraddaoui. LGTM. :+1:

PR #1045, opened 2024-10-15
Remove multi-module workspace

sevein:
Previously, we used a multi-module workspace (via the go.work file) to address issues with gopls detecting nested modules. However, these issues now appear to be resolved. Since the go.work file no longer provides any significant benefit, we're removing it from the project.

I've used `go mod tidy -diff` to check that the mod files are tidy. `-diff` is a new flag introduced in go1.23.


Makefile
```diff
@@ -191,13 +191,17 @@ mod-tidy:
 	go mod download
 	go mod tidy
 
+check-mod-tidy: # @HELP Check that mod files are tidy.
+	go mod tidy -diff
+	cd hack/pulumi && go mod tidy -diff
+
```
jraddaoui: At some point these rules/targets got sorted alphabetically, I'd rename it to `mod-tidy-check` to keep them together.

jraddaoui (approved):
Works for me, thanks!

Makefile
jraddaoui: I tweaked that commit to rename it ;)

PR #1046, opened 2024-10-15
Optimize Makefile with lazy evaluation of packages

sevein:
This commit updates the declarations of `PACKAGES` and related variables in the Makefile to implement lazy loading, so the computation is deferred until they are actually needed.

Running commands like `make help` should now be faster (~80ms compared to ~500ms).


djjuhasz (approved):
LGTM :+1:

jraddaoui (approved):
Nice, thanks!

PR #1047, opened 2024-10-16
Enable Tempo metrics generator

sevein:
This pull request installs Prometheus in the development environment so we can start capturing application metrics (as opposed to logs or traces). It also enables the [metrics generator](https://grafana.com/docs/tempo/latest/metrics-generator/) in Grafana Tempo, which is a mechanism that derives metrics from ingested traces and store them in Prometheus. These metrics are useful in a couple of main ways: [span metrics](https://grafana.com/docs/tempo/latest/metrics-generator/service_graphs/) and [service graphs](https://grafana.com/docs/tempo/latest/metrics-generator/service_graphs/).

The Prometheus datasource in Grafana has been configured with [exemplars](https://grafana.com/docs/grafana/latest/fundamentals/exemplars/) enabled, so you should be able to associate trace data with metric data, something that we can start exploring further once we start producing application metrics.

I've updated the dashboard with a simple service view panel:

image

### DIFF
--- a/hack/kube/components/dev/grafana-alloy.yaml
+++ b/hack/kube/components/dev/grafana-alloy.yaml
@@ -1,25 +1,26 @@
-# Generated with `helm template grafana/grafana-agent --set=agent.mode=flow --set=controller.type=statefulset` and modified.
+# Generated with `helm template grafana/alloy --name-template=grafana-alloy --set=controller.type=statefulset --set=configReloader.enabled=false --set=crds.create=false`.
 ---
-# Source: grafana-agent/templates/serviceaccount.yaml
+# Source: alloy/templates/serviceaccount.yaml
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: grafana-agent
+  name: grafana-alloy
+  namespace: enduro-sdps
   labels:
-    app.kubernetes.io/name: grafana-agent
+    app.kubernetes.io/name: grafana-alloy
 ---
-# Source: grafana-agent/templates/configmap.yaml
+# Source: alloy/templates/configmap.yaml
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: grafana-agent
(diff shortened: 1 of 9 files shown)


hack/kube/components/dev/grafana-alloy.yaml
```diff
@@ -1,25 +1,26 @@
...
+# Source: alloy/templates/serviceaccount.yaml
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: grafana-agent
+  name: grafana-alloy
+  namespace: enduro-sdps
   labels:
-    app.kubernetes.io/name: grafana-agent
+    app.kubernetes.io/name: alloy
```
jraddaoui: Not sure is this even shown/used but, should we match this label with the metadata name above? There are many repetitions below, also for the prometheus-server.

jraddaoui (approved):
Really nice, thanks @sevein!

hack/kube/components/dev/grafana-alloy.yaml
sevein: Yeah good point, I wasn't really paying attention to that, thanks.

sevein:
```
$ kubectl get all -A -l app.kubernetes.io/name=grafana-alloy
NAMESPACE     NAME                  READY   STATUS    RESTARTS   AGE
enduro-sdps   pod/grafana-alloy-0   1/1     Running   0          28m

NAMESPACE     NAME                    TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                       AGE
enduro-sdps   service/grafana-alloy   ClusterIP   10.43.183.218   <none>        12345/TCP,4317/TCP,4318/TCP   28m

NAMESPACE     NAME                             READY   AGE
enduro-sdps   statefulset.apps/grafana-alloy   1/1     28m
```

PR #1048, opened 2024-10-16
Implement CreatePreservationTasks with CreateBulk

sevein:
This PR explores a new pattern for handling bulk database operations using [Go 1.23's range-over-function feature](https://tip.golang.org/doc/go1.23#language). It introduces a new `CreatePreservationTasks` method in the persistence package, which accepts a generator of preservation tasks ([`iter.Seq[*datatype.PreservationTask]`](https://pkg.go.dev/iter#Seq)). The method utilizes the `batch` function to group tasks into manageable batches and leverages [`CreateBulk`]((https://entgo.io/docs/crud/#create-many)) to perform efficient bulk inserts in a single INSERT operation.

To view the main idea of this proposal in isolation, visit: https://go.dev/play/p/y7tYvD7NO28.

In future PRs, this new function could be used in areas like CreateAIPActivity and JobTracker to improve bulk task creation.

### DIFF
--- a/internal/persistence/ent/client/batch.go
+++ b/internal/persistence/ent/client/batch.go
@@ -1,0 +1,31 @@
+package entclient
+
+import "iter"
+
+var defaultBatchSize = 100
+
+// batch groups elements from the provided sequence into batches of size
+// defaultBatchSize, yielding a stream of slices, each containing at most
+// defaultBatchSize elements, except for the last batch which may contain fewer.
+// Note that the slices are reused internally to minimize allocations, so they
+// should not be retained or modified after being yielded to avoid unexpected
+// behavior.
+func batch[T any](seq iter.Seq[T]) iter.Seq[[]T] {
+	return func(yield func([]T) bool) {
+		b := make([]T, 0, defaultBatchSize)
+		seq(func(a T) bool {
+			b = append(b, a)
+			if len(b) == defaultBatchSize {
+				if !yield(b) {
+					return false
+				}
+
+				b = b[:0]
+			}
+			return true
+		})
+		if len(b) != 0 {
+			yield(b)
+		}
+	}
+}
--- a/internal/persistence/ent/client/batch.go
+++ b/internal/persistence/ent/client/batch.go
@@ -13,19 +13,19 @@ var defaultBatchSize = 100
 func batch[T any](seq iter.Seq[T]) iter.Seq[[]T] {
 	return func(yield func([]T) bool) {
 		b := make([]T, 0, defaultBatchSize)
-		seq(func(a T) bool {
+		for a := range seq {
 			b = append(b, a)
 			if len(b) == defaultBatchSize {
 				if !yield(b) {
-					return false
+					return
 				}
-
 				b = b[:0]
 			}
-			return true
-		})
+		}
 		if len(b) != 0 {
-			yield(b)
+			if !yield(b) {
+				return
+			}
 		}
 	}
 }
--- a/internal/persistence/ent/client/preservation_task_test.go
+++ b/internal/persistence/ent/client/preservation_task_test.go
@@ -5,10 +5,12 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"slices"
 	"testing"
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/uuid"
 	"gotest.tools/v3/assert"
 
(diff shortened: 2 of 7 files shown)


jraddaoui:
Sorry for the late update @sevein. I took a quick look a while ago, really nice and good learnings! I think @djjuhasz is also looking at it and will do the actual code review.

djjuhasz:
@djjuhasz yes, I've been looking at it and will provide feedback.  I'm having a hard time wrapping my head around the iterator code flow.

sevein:
> I've been looking at it and will provide feedback. I'm having a hard time wrapping my head around the iterator code flow.

The syntax is a bit confusing :cry:... I found the docstring of the compiler's rangefunc package somewhat useful: https://github.com/golang/go/blob/master/src/cmd/compile/internal/rangefunc/rewrite.go.

There are a couple of areas where I'm unsure if I'm doing things correctly:

- Making `CreatePreservationTasks` accept an iterator seemed convenient because it allows the user to simply provide a function that yields objects, without having to worry about chunking at all. I'm thinking in [`savePreservationTasks`](https://github.com/artefactual-sdps/enduro/blob/cfb6643f7042bc75eb5e619f22d002ee0c2e0664/internal/a3m/a3m.go#L167-L203), you'd only need to write a function that walks the slice of a3m jobs and produces preservation tasks or use some kind of map & transform helper like [this](https://pkg.go.dev/github.com/pamburus/go-mod/gi#Map). It seems a better option if compared with accepting a slice I think?
- `batch` accepts and returns an iterator which is a pattern I haven't seen a lot in other projects, e.g. [`pkg.go.dev/slices`](https://pkg.go.dev/slices) has functions where they either accept or return an iterator but not both. But I've found at least one example: [ChunkToSlices](https://github.com/pamburus/go-mod/blob/bf6181e3ef0c7293c500e79a79dae8e8117b980a/gi/chunk.go#L54-L85).

I haven't found yet an implementation like this in other codebases, I should probably look around more. I assume other poeple are also trying to solve the problem of ublk inserts with iterators? A project that is slightly related is https://github.com/achille-roussel/sqlrange.

internal/persistence/ent/client/preservation_task_test.go
```diff
@@ -154,6 +156,120 @@ func TestCreatePreservationTask(t *testing.T) {
...
+						PreservationActionID: pa1.ID,
+					},
+				}
+			},
+		},
+		{
+			name: "Creates multiple preservation tasks exceeding batch size",
+			tasks: func(entc *db.Client) []*datatypes.PreservationTask {
+				pa1, _ := addDBFixtures(t, entc)
+				pts := make([]*datatypes.PreservationTask, 0, 300) // Three batches.
```
djjuhasz, line 197: This seems like a good place to use a generator instead of creating a big slice of PreservationTasks.

internal/persistence/ent/client/batch.go
```diff
@@ -0,0 +1,31 @@
...
+
+var defaultBatchSize = 100
+
+// batch groups elements from the provided sequence into batches of size
+// defaultBatchSize, yielding a stream of slices, each containing at most
+// defaultBatchSize elements, except for the last batch which may contain fewer.
+// Note that the slices are reused internally to minimize allocations, so they
+// should not be retained or modified after being yielded to avoid unexpected
+// behavior.
+func batch[T any](seq iter.Seq[T]) iter.Seq[[]T] {
```
djjuhasz, line 13: It doesn't seem like there is any code specific to ent in the batch function — I think it would be nice to make it a separate batch package that could be used other places in Enduro (or other projects).  The only change I would make it more reusable as package would be the ability to override the default batch size when calling the function.

djjuhasz (approved):
Thanks @sevein, this is pretty cool. I made a couple of comments, but neither is a blocker to merging.

I think passing an iterator to `CreatePreservationTasks()` instead of a slice is going to make it a bit harder to understand for those unfamiliar with iterators, but it's more flexible so I think it's worth it.  Adding a call to `CreatePreservationTasks()` in the code would be helpful as an example, but that can come in a later commit.

It would be cool to add some benchmark tests at some point to compare batch vs. single inserts, though I guess the performance will depend a lot on the DBMS used and deployment particulars.

internal/persistence/ent/client/batch.go
sevein: One option could be `github.com/artefactual-labs/gotools/iterutil`, what do you think?

internal/persistence/ent/client/preservation_task_test.go
sevein: I need the full slice for assert.DeepEqual - what testing approach did you have in mind with an iterator?

internal/persistence/ent/client/batch.go
djjuhasz: Sure, sounds good to me. :)

internal/persistence/ent/client/preservation_task_test.go
djjuhasz: Hmm, I didn't realize `CreatePreservationTasks()` returned the full slice of created preservation tasks. I guess my counter question is: what's the advantage of accepting an iterator when you return the full slice of results anyway?  I'm trying to imagine a scenario where you would want to generate 3 tranches of data 100 rows at a time (for example), yield the rows to `CreatePreservationTasks()`, but then get back all three hundred results in one slice. :thinking:

djjuhasz: It also just seems weird to me to create a slice of 300 rows, convert that to an iterator (with `slices.Values(tasks)`), then get back a slice of 300 results.  I get that this is just a test, so the data flow is necessarily simplified, but I just wonder if we're going to end up using this same flow in production. :thinking:

sevein: > What's the advantage of accepting an iterator when you return the full slice of results anyway? I'm trying to imagine a scenario where you would want to generate 3 tranches of data 100 rows at a time (for example), yield the rows to CreatePreservationTasks(), but then get back all three hundred results in one slice.

Fair point. My use case was [savePreservationTasks](https://github.com/artefactual-sdps/enduro/blob/c0afd7699076adb9012b15c7e89ac9c8ee42a8db/internal/a3m/a3m.go#L167) where I thought thatit would be preferable to eventually load tasks from the data source (a3m) in batches while doing bulk inserts, but I was probably overthinking. I also realized that `packageImpl.CreatePreservationTasks` needs the slice back from MySQL so we can publish the corresponding events with the MySQL-generated identifiers.
I'm not sure if it's a good idea to accept an iterator as a generic form of an iterable, so the user gets to decide if they want to pass a slice wrapped by slices.Values or an iterator function. Honestly I don't know what ar ethe performance implications, I've been assuming that iterators are fast because the compiler is inlining all the code but that may not be true.
I can append new commits to this PR bringing the simpler form accepting a slice, and doing all the work end to end (a3m activity » package service » persistence » ent) to see how it feels.

There are all sorts of packages emerging using rangefunc and generics, e.g. https://github.com/BooleanCat/go-functional introduces the idea of chainable methods.  I need to keep exploring this space.

djjuhasz: Yeah, iterators are quite interesting. Because I'm new to coroutines and generators, I keep referring back to https://research.swtch.com/coro, which I've found really helpful for conceptualizing the mechanics and use-cases of coroutines, but I'm still not very confident of my understanding.

To me the `batch()` function fits well with my concept of a coroutine - the batch function receives data from an iterator until it hits the batch threshold, then yields a tranche of records back to the caller which then does some stuff with that tranche, before resuming the batch coroutine.

I think the use case for `CreatePreservationTasks()` where the iterator parameter is sending batched rows from a MySQL query is an interesting one. I guess in that case the hypothetical `ListPreservationTasks()` would do a SELECT query with a limit on the rows returned, and then would return the results as an iterator. When the batching function in `ListPreservationTasks()` got to the last row in the tranche would return control back to it's caller to SELECT the next tranche of rows.  Meanwhile `CreatePreservationTasks()` would be batching the rows from it's input iterator and inserting them into the database when it hit the batch limit. I guess my naive expectation at that point is that `CreatePreservationTasks()` would yield back the results from that tranche of inserted rows to it's caller so the caller could do whatever it needs to do with those results (e.g. publish events), then the results could be expired before receiving the next tranche of results (and collected by the garbage collector whenever that ran).  This scenario took a while for me to puzzle through because it's a pretty long chain of coroutines to switch between, with batching on both the SELECT and INSERT operations. :exploding_head: 

@sevein is this the close to the flow you were envisioning, or were you thinking of something very different?

sevein: What you're saying reasonates with me as well but I am avoiding to adopt the mental model where we think of iterators as coroutines because that's not how they ended up implementing it - Russ' article explores the broader coroutine space, which differs from go1.23's implementation. Unlike Python (run this [example](https://gist.github.com/sevein/0f26279744f27408711d3143e555be8a) if interested), in Go iterators we have a single stack and there is no pausing/resuming, the compiler is inlining the code (these [examples](https://go.dev/wiki/RangefuncExperiment#can-range-over-a-function-perform-as-well-as-hand-written-loops) are clarifying to me).

> I guess in that case the hypothetical ListPreservationTasks() would do a SELECT query with a limit on the rows returned, and then would return the results as an iterator. When the batching function in ListPreservationTasks() got to the last row in the tranche would return control back to it's caller to SELECT the next tranche of rows. 

Depending on the underlying database driver, I suspect you could leverage server-side cursors so the database can stream the result set in chunks efficiently. See [sqlrange](https://github.com/achille-roussel/sqlrange), the Query function can be a good example. That probably means that you may not need to handle the LIMIT clause manually. But they're not supported in MySQL as far as I know.

> I guess my naive expectation at that point is that CreatePreservationTasks() would yield back the results from that tranche of inserted rows to it's caller so the caller could do whatever it needs to do with those results (e.g. publish events), then the results could be expired before receiving the next tranche of results (and collected by the garbage collector whenever that ran).

Yeah, that make sense. I chose not to return an iterator, concerned that I wouldn't be able to manage the complexity or that the API would become too cryptic. But it looks like that approach would result in less allocations and less pressure in the garbage collector? But I'll start the simpler version where we just work with slices, then we can explore other options further.

djjuhasz: > What you're saying reasonates with me as well but I am avoiding to adopt the mental model where we think of iterators as coroutines because that's not how they ended up implementing it - Russ' article explores the broader coroutine space, which differs from go1.23's implementation. Unlike Python (run this [example](https://gist.github.com/sevein/0f26279744f27408711d3143e555be8a) if interested), in Go iterators we have a single stack and there is no pausing/resuming, the compiler is inlining the code (these [examples](https://go.dev/wiki/RangefuncExperiment#can-range-over-a-function-perform-as-well-as-hand-written-loops) are clarifying to me).

Man https://go.dev/wiki/RangefuncExperiment#can-range-over-a-function-perform-as-well-as-hand-written-loops is interesting, but am I the only one that has a really hard time parsing functions used function params or return values? :exploding_head: I had to read those examples at least three times to figure out what was going on. :laughing: 

I'm not really clear on the practical differences between using Go's inlining iterator functions and Python's coroutines, but I did try to reimplement your Python coroutine in Go.  I did get the Go coroutine output to match the Python output, but it was much harder in Go, and I had to use `iter.Pull` to get access to a `next()` function.  The exercise highlighted just how shaky my understanding is on how to write and use iterators in Go. :grimacing: 

> Depending on the underlying database driver, I suspect you could leverage server-side cursors so the database can stream the result set in chunks efficiently. See [sqlrange](https://github.com/achille-roussel/sqlrange), the Query function can be a good example. That probably means that you may not need to handle the LIMIT clause manually. But they're not supported in MySQL as far as I know.

Hmm, server-side cursors sound cool.  We could use whatever method works best for chunking the query results here.

> Yeah, that make sense. I chose not to return an iterator, concerned that I wouldn't be able to manage the complexity or that the API would become too cryptic. But it looks like that approach would result in less allocations and less pressure in the garbage collector? But I'll start the simpler version where we just work with slices, then we can explore other options further.

Doing the simpler version of passing slices sounds good to me. :+1:

sevein: > I did get the Go coroutine output to match the Python output, but it was much harder in Go, and I had to use iter.Pull to get access to a next() function.

Thanks for doing this, it made me realize that I misunderstood a few things, and this conversation has been helpful in clarifying them. It turns out that Go used [a lightweight/constrained version of coroutines](https://github.com/golang/go/blob/go1.23.3/src/runtime/coro.go) to implement [iter.Pull](https://github.com/golang/go/blob/c390a1c22e8951263e6c01346a4281d604b25062/src/iter/iter.go#L241). This post was very clarifying: https://unskilled.blog/posts/coroutines-in-go/.
I feel like I've finally reached the point where I know I know nothing, so now I can get back to the task I was supposed to be doing in the first place and debug it like a monkey, haha.

sevein:
Replaced by https://github.com/artefactual-sdps/enduro/pull/1438.

PR #1054, opened 2024-10-24
Add a package list status filter

### LINKED ISSUE #989: Problem: There is no way to filter package results by workflow status
## Description

One of the feedback elements from the recent Talk-Aloud usability tests. Currently users have no way of filtering package results by status. This can make it hard to find specific package when there are many results. 

**To Reproduce**

1.Log in
2. Navigate to Packages
3. try to see only Done packages, or only those with Errors

**Resulting error**

No method available to filter results by workflow status

**Expected behavior**

Users can easily filter results by workflow status to better discover relevant packages.

## Proposed solution

Following this feedback, we drafted 2 mockups to cover potential implementation solutions - one using tabs, and the other using a drop-down menu. After reviewing it with the team, no team members had strong preferences one way or the other. We then decided to get input from another Enduro user in a different context - Maaike, who operates the MoMA Enduro instance as part of their managed service. 

Here is Maaike's feedback: 

> In my personal opinion, I prefer the Tabs option - I would find this a more efficient way to monitor the progress of a batch and view errors. The current status filter in [MoMA's version of] Enduro is similar to the drop-down filter option, which while functional, is slightly more tedious to manage than a series of tabs that I could quickly flip between.

Consequently we propose that the first implementation try using the **tabs**. A mockup is included below: 

enduro-pkg-status-filter-tabs

djjuhasz:
Fixes #989.

Add a package status filter select box to the Dashboard package list page to allow filtering packages by status.

Changes:
- Add a "filters" section to the package list page
- Add a "status" selector to the filters section
- Filter package list results when the "status" filter is changed

### DIFF
--- a/dashboard/src/pages/packages/index.vue
+++ b/dashboard/src/pages/packages/index.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { PackageListStatusEnum } from "@/openapi-generator";
 import PackageListLegend from "@/components/PackageListLegend.vue";
 import PageLoadingAlert from "@/components/PageLoadingAlert.vue";
 import StatusBadge from "@/components/StatusBadge.vue";
@@ -50,6 +51,24 @@ const toggleLegend = () => {
       Showing {{ packageStore.page.offset + 1 }} -
       {{ packageStore.lastResultOnPage }} of
       {{ packageStore.page.total }}
+    </div>
+
+    <div class="d-flex flex-wrap gap-3 mb-3 p-3 border bg-light">
+      <h3 class="mb-0 pe-3 border-end">Filters</h3>
+      <div class="d-flex gap-2">
+        <label for="filter-status" class="align-self-center">Status</label>
+        <select
+          id="filter-status"
+          v-model="packageStore.filters.status"
+          @change="packageStore.fetchPackages(1)"
+          class="form-select"
+        >
+          <option value="">any</option>
+          <option v-for="item of PackageListStatusEnum" :value="item">
+            {{ item }}
+          </option>
+        </select>
+      </div>
     </div>
 
     <PageLoadingAlert :execute="execute" :error="error" />
--- a/dashboard/src/stores/package.ts
+++ b/dashboard/src/stores/package.ts
@@ -1,5 +1,8 @@
 import { api, client } from "@/client";
-import { MonitorEventEventTypeEnum } from "@/openapi-generator";
+import {
+  MonitorEventEventTypeEnum,
+  PackageListStatusEnum,
+} from "@/openapi-generator";
 import { useLayoutStore } from "@/stores/layout";
 import router from "@/router";
 import { defineStore, acceptHMRUpdate } from "pinia";
@@ -43,6 +46,10 @@ export const usePackageStore = defineStore("package", {
     // User-interface interactions between components.
     ui: {
       download: new UIRequest(),
+    },
+
+    filters: {
+      status: "" as PackageListStatusEnum,
     },
   }),
   getters: {
@@ -164,6 +171,7 @@ export const usePackageStore = defineStore("package", {
       const resp = await client.package.packageList({
         offset: page > 1 ? (page - 1) * this.page.limit : undefined,
         limit: this.page?.limit || undefined,
+        status: this.filters.status ? this.filters.status : undefined,
       });
       this.packages = resp.items;
       this.page = resp.page;
--- a/dashboard/src/styles/bootstrap.scss
+++ b/dashboard/src/styles/bootstrap.scss
@@ -25,6 +25,7 @@
 @import "bootstrap/scss/tooltip";
 @import "bootstrap/scss/offcanvas";
 @import "bootstrap/scss/pagination";
+@import "bootstrap/scss/forms";
 
 // Utilities API last to generate classes based on the Sass map in `_utilities.scss`.
 @import "bootstrap/scss/utilities/api";


djjuhasz:
Here's what the "status" filter looks like with these changes:
image

djjuhasz:
Alternate style:
image

fiver-watson:
I do prefer the alternate style and say we go with that for now. 

That said, regarding the tabs in the original mockup, some thoughts on my rationale for doing them that way: 

1. I did a UX interview with Maaike, because there apparently already _is_ a drop-down status filter in the MoMA Enduro instance - she told me she liked the proposed design of the tabs much better and thought it would be a lot less clicks than she currently makes when reviewing failures, for example
2. I showed those designs to some folks at SFA earlier, who liked them
3. Based on the fields we show in the packages page, I don't see there being many other filters. Columns currently are ID, Name, UUID, Started, Location, and Status. 
4. In terms of other ways to surface / discover content, my thought was:
   * Making the columns sortable  (and reversible ascending / descending) will go a long way in helping people find specific content fast, particularly if a fast simple search is added too
   * If we so add search in the future, it would likely be something simple and fast, like Typesense, filtering package results as you type
   * If we want something fancier, like an advanced search, that is where more complex filters (like a date range search) would make sense - but i suspect that could be hidden in a collapsible panel, as such filters would likely not be needed on a daily basis

Perhaps something we could try if we DO iterate on this in the future and try some tabs would be to keep and deploy both versions, and do some A/B testing with users?

djjuhasz:
@fiver-watson I kept the initial design very basic to focus on getting the filter working and with the intention of refining the UI later. I like the idea of doing some A/B usability testing in the future.

I have a few thoughts about the tab design:
1. There are eight different possible statuses for a package right now in Enduro (new, "in progress", done, error, unknown, queued, abandoned, pending), though there is a code comment that "new" is unused.  I'm not sure if your design of four status tabs (plus the "all" tab) is intentionally omitting some of the possible statuses, or if you were unaware of the additional possible status values.  I looked up the "pending" status to see how it's different from "queued" and "pending" indicates a package that is waiting on a user decision.
2. The tab design takes a lot of space, especially if we add the three additional active status values (unknown, abandoned, pending), which may be a problem for small screens or if we want to add other filters, a search bar, etc.  I take your point that are not a lot of other columns in the table that we would want to filter on, but I do think a date range filter is one that would probably be useful for large data sets.
3. The tabs are mutually exclusive (aside from the "All" tab), so there's no way to filter (or show) multiple statuses at the same time (e.g. viewing unknown, abandoned, error packages). Of course the current select box doesn't allow selecting multiple statuses either, but there are designs that would (e.g. dropdown with checkboxes, a filter list with checkboxes).

All that said, I can start a new PR to implement the tab filters so we can try that out.

djjuhasz:
@fiver-watson oh, there's a code comment that the "unknown" status is unused too.  I'm not sure why we have two statuses defined that we aren't using. :shrug:

djjuhasz:
For reference, the full list of package statuses, with brief descriptions:
```
new          // Unused!
in progress  // Undergoing work.
done         // Work has completed.
error        // Processing failed.
unknown      // Unused!
queued       // Awaiting resource allocation.
abandoned    // User abandoned processing.
pending      // Awaiting user decision.
```

From https://github.com/artefactual-sdps/enduro/blob/main/internal/enums/pkg_status.go

dashboard/src/styles/bootstrap-base.scss
```diff
@@ -7,9 +7,6 @@ $offcanvas-horizontal-width: 300px;
 $offcanvas-padding-x: 0;
 $headings-color: $primary;
 $headings-font-weight: 400;
-$border-radius: 0;
-$border-radius-sm: 0;
-$border-radius-lg: 0;
```
jraddaoui: This is changing many parts of the dashboard where we decided to remove the border radius (badges, buttons, tabs, etc.). I'd like to keep the boxy look everywhere, even in the input fields.

dashboard/src/pages/packages/index.vue
```diff
@@ -52,6 +66,26 @@ const toggleLegend = () => {
       {{ packageStore.page.total }}
     </div>
 
+    <div class="row mx-0 mb-3 align-items-left border bg-light">
+      <div class="col-auto pb-1 align-self-center border-end">
+        <h3 class="mb-0">Filters</h3>
+      </div>
+      <div class="col-auto pb-1">
+        <label for="filter-status" class="form-label mb-0"> Status </label>
```
jraddaoui:
```html
<div class="d-flex flex-wrap gap-3 mb-3 p-3 border bg-light">
      <h3 class="mb-0 pe-3 border-end">Filters</h3>
      <div class="d-flex gap-2">
        <label for="filter-status" class="align-self-center">Status</label>
```

- On the first div:
  - Used `d-flex` instead of `row mx-0` and `col-` classes in the children.
  - Added `flex-wrap` for responsiveness.
  - Added `gap-3` for separation between elements (also works spacing vertically when the wrap kicks in).
  - Removed `align-items-left`, it should be `align-items-start` and it's the default.
  - Added padding `p-3`.
- Removed the div wrapping the heading, it's not needed.
- On the heading, just moved the border there and added padding to the end.
- On the div wrapping the label and select:
  - Added `d-flex gap-2`, same as the first div, smaller gap and not wrapping.
- On the label:
  - Removed `form-label mb-0`, the first one only adds margin bottom and the second one removes it.
  - Added `align-self-center` to center it vertically.


image
image

dashboard/src/pages/packages/index.vue
```diff
@@ -52,6 +66,26 @@ const toggleLegend = () => {
...
+        <select
+          id="filter-status"
+          v-model="statusFilter"
+          @change="changeStatusFilter"
+          class="form-select border rounded-2"
```
jraddaoui: `class="form-select"` should be enough considering my previous comment about the boxy look.

dashboard/src/pages/packages/index.vue
```diff
@@ -52,6 +66,26 @@ const toggleLegend = () => {
...
+        >
```
jraddaoui: I think you can assign properties from the store directly as the model value and the change function.

```vue
<select
  id="filter-status"
  v-model="packageStore.filters.status"
  @change="packageStore.fetchPackages(1)"
  class="form-select"
>
```

This will remove the need of all the changes on the script part.

dashboard/src/stores/package.ts
```diff
@@ -44,6 +47,10 @@ export const usePackageStore = defineStore("package", {
     ui: {
       download: new UIRequest(),
     },
+
+    filters: {
+      status: undefined as PackageListStatusEnum | undefined,
```
jraddaoui: With the changes mentioned above, if you modify this line to read `status: "" as PackageListStatusEnum`, the select will be populated with `any` on page load.

jraddaoui (changes requested):
Nice first pass, thanks @djjuhasz!

jraddaoui:
> Alternate style: image

I prefer the label and input inline, check the screenshots from a review comment above.

fiver-watson:
@djjuhasz Abandoned is also unused - there is currently no way for a user to "Abandon" a package once in-flight. Pending is not currently in use, but may be soon with upcoming development discussions . 

Point taken about the mutual exclusivity, but TBH that was kind of intentional, to make their use easy, fast, and intuitive.  We haven't yet heard a use case where an operator would want or need to limit content in this particular way - and if we continue to add additional discovery methods (like a search, sortable columns, etc) then it may not be needed... or better yet, might be more suited to a collapsible Advanced Search panel / page, that is handy when available but less regularly used. 

In any case, we have a first implementation and the alternative design looks good to me - let's get some feedback from our users and see where we should go next.

dashboard/src/styles/bootstrap-base.scss
djjuhasz: I dislike the boxy look for the select, but okay. :shrug:

I think it was a mistake to set the the `$border-radius` variables to 0, because you can't set a border radius for an individual HTML element with something like: `<select class="form-select rounded-2">`.  Setting `$border-radius: 0` seems to cause all the "rounded" classes to be multiplied by zero and result in a zero `border-radius`. :frowning:  I think it would have been better to set the style `rounded-0` on the applicable elements (e.g. button, .badge) in `main.scss`.

jraddaoui (approved):
Thanks @djjuhasz!

PR #1055, opened 2024-10-25
Use tabs to filter the package list by status

djjuhasz:
Refs #989.

### DIFF
--- a/dashboard/src/components/PackageListSelectFilters.vue
+++ b/dashboard/src/components/PackageListSelectFilters.vue
@@ -1,0 +1,26 @@
+<script setup lang="ts">
+import { usePackageStore } from "@/stores/package";
+import { PackageListStatusEnum } from "@/openapi-generator";
+
+const packageStore = usePackageStore();
+</script>
+
+<template>
+  <div class="d-flex flex-wrap gap-3 mb-3 p-3 border bg-light">
+    <h3 class="mb-0 pe-3 border-end">Filters</h3>
+    <div class="d-flex gap-2">
+      <label for="filter-status" class="align-self-center">Status</label>
+      <select
+        id="filter-status"
+        v-model="packageStore.filters.status"
+        @change="packageStore.fetchPackages(1)"
+        class="form-select"
+      >
+        <option value="">any</option>
+        <option v-for="item of PackageListStatusEnum" :value="item">
+          {{ item }}
+        </option>
+      </select>
+    </div>
+  </div>
+</template>
--- a/dashboard/src/components/Tabs.vue
+++ b/dashboard/src/components/Tabs.vue
@@ -1,15 +1,25 @@
 <script setup lang="ts">
 import type { FunctionalComponent, SVGAttributes } from "vue";
 import type { RouteLocationResolved } from "vue-router/auto";
+import { useRoute } from "vue-router/auto";
+import { isEqual } from "lodash-es";
+
+const route = useRoute();
+
+type Tab = {
+  icon?: FunctionalComponent<SVGAttributes, {}>;
+  text: string;
+  route: RouteLocationResolved;
+  show: boolean;
+};
 
 const { tabs } = defineProps<{
-  tabs: {
-    icon: FunctionalComponent<SVGAttributes, {}>;
-    text: string;
-    route: RouteLocationResolved;
-    show: boolean;
-  }[];
+  tabs: Tab[];
 }>();
+
+function isActive(tab: Tab): boolean {
+  return tab.route.path == route.path && isEqual(tab.route.query, route.query);
+}
 </script>
 
 <template>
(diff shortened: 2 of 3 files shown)


djjuhasz:
Screenshot: 
image

jraddaoui:
@djjuhasz, I'm out tomorrow but some early feedback.

I just realized we already have a tabs component with links and icons that is even "responsive", it's used from the location index page. It would be nice to tweak that one a bit to make it work in this case, maybe accepting a parameter in the URLs or something like that:

https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/components/Tabs.vue
https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/pages/locations/%5Bid%5D.vue#L22-L41
https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/pages/locations/%5Bid%5D.vue#L54

image
image

djjuhasz:
Having to scroll horizontally on a phone to click a tab seems like a terrible UI experience to me. :confused:

djjuhasz:
@jraddaoui this is ready for another look. :pray:

dashboard/src/components/PackageListSelectFilters.vue
```diff
@@ -0,0 +1,26 @@
...
+        class="form-select"
+      >
+        <option value="">any</option>
+        <option v-for="item of PackageListStatusEnum" :value="item">
+          {{ item }}
+        </option>
+      </select>
+    </div>
+  </div>
+</template>
```
jraddaoui, line 26: We can remove this component now, right?

dashboard/src/components/Tabs.vue
```diff
@@ -1,15 +1,25 @@
...
-    icon: FunctionalComponent<SVGAttributes, {}>;
-    text: string;
-    route: RouteLocationResolved;
-    show: boolean;
-  }[];
+  tabs: Tab[];
 }>();
+
+function isActive(tab: Tab): boolean {
+  return tab.route.path == route.path && isEqual(tab.route.query, route.query);
```
jraddaoui, line 21: Why is `lodash` needed for this comparison?

dashboard/src/pages/packages/index.vue
```diff
@@ -1,44 +1,116 @@
...
 import IconSkipEndFill from "~icons/bi/skip-end-fill";
 import IconSkipStartFill from "~icons/bi/skip-start-fill";
 import IconCaretRightFill from "~icons/bi/caret-right-fill";
 import IconCaretLeftFill from "~icons/bi/caret-left-fill";
 
+// Tab icons.
+import RawIconCheckCircleLine from "~icons/clarity/check-circle-line?raw&font-size=20px";
+import RawIconTimesCircleLine from "~icons/clarity/times-circle-line?raw&font-size=20px";
+import RawIconPlayLine from "~icons/clarity/play-line?raw&font-size=20px";
+import RawIconBarsLine from "~icons/clarity/bars-line?raw&font-size=20px";
```
jraddaoui, line 32: I have been thinking this for a while, we may want to look for some import sorting tool.

jraddaoui (approved):
Looks great, thanks @djjuhasz!

dashboard/src/components/Tabs.vue
djjuhasz: `tab.route.query` and `route.query` are objects, and when you compare objects in Typescript / Javascript it checks if they are the identical object not that the properties of the object are equivalent, i.e. "(tab.route.query == route.query) => false".

dashboard/src/components/PackageListSelectFilters.vue
djjuhasz: From @fiver-watson's comments I thought we might do so A / B testing of the dropdown vs tab status filters, so I kept this component for that.

dashboard/src/pages/packages/index.vue
djjuhasz: Yeah, the imports are pretty messy.

dashboard/src/components/Tabs.vue
djjuhasz: https://www.typescriptlang.org/play/?#code/MYewdgzgLgBCBGArAjDAvDA3jaBDKArhAFwwDkAJuAKZkwC+A3AFCiSwKIBM6WOU+IqUo06TZq3AQQAG2oA6GSADmACjKrOqNBk5cAlMJgBqGJqTbdSA-sZA

dashboard/src/components/PackageListSelectFilters.vue
jraddaoui: Okay, it will be nice to add some tests if we keep it. Doesn't need to happen in this PR.

jraddaoui: And make it a real component ;)

Issue #1039, opened 2024-10-04
Feature: convert Unix linebreaks to HTML linebreaks in front-end code


sallain:
**Is your feature request related to a problem? Please describe.**

Error messages are currently a bit dense to read. In order to make pretty error messages, it would be nice to be able to implement linebreaks. However, they should look nice in both the terminal and in the UI. 

**Describe the solution you'd like**

To address this, the front-end code should convert Unix linebreaks to HTML linebreaks when it encounters them in error messages.

**Describe alternatives you've considered**

Just use HTML linebreaks, but then error messages will be more cluttered in non-HTML contexts.

**Additional context**

This is a child issue of https://github.com/artefactual-sdps/preprocessing-sfa/issues/43

sallain:
LGTM!

PR #1056, opened 2024-10-25
Render line breaks in notes column for packages

DanielCosme:
Solves: #1039

### DIFF
--- a/dashboard/src/components/PreservationActionCollapse.vue
+++ b/dashboard/src/components/PreservationActionCollapse.vue
@@ -138,7 +138,7 @@ watch($$(expandCounter), () => show());
             <td>
               <StatusBadge :status="task.status" />
             </td>
-            <td class="line-wrap">{{ task.note }}</td>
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
LGTM! Just a note, I'd like to use `pre` tags when we move to an expandable section, but this is better inside the table display. Thanks @DanielCosme!

dashboard/src/components/PreservationActionCollapse.vue
```diff
@@ -138,7 +138,7 @@ watch($$(expandCounter), () => show());
             <td>
               <StatusBadge :status="task.status" />
             </td>
-            <td>{{ task.note }}</td>
+            <td class="line-wrap">{{ task.note }}</td>
```
djjuhasz: @DanielCosme I know I suggested "line-wrap" the other day, but I think "line-break" is actually more accurate.

DanielCosme: Ok

Issue #886, opened 2024-03-12
Problem: Enduro Ingest workflows are not easily extensible


fiver-watson:
Enduro Ingest aims to help the keepers of cultural memory preserve their content. However, Enduro is also an open source project trying to reach an international user base, which spreads across many different jurisdictions, uses different national and regional standards, follows many varying local policies, are trying to address many different use cases, and correspondingly can have many different submission requirements from upstream content creators.

At present, a3m and Archivematica  - the underlying preservation engines of the system - accept a very limited number of variations in the expected incoming SIP structure. These limitations do not align with the many variations in Transfer structures that might be submitted by Enduro users and upstream creators. 

Currently there is no established method of customizing and/or extending Enduro's functionality, short of forking the project. This makes the ongoing maintenance of production Enduro instances across different users and jurisdictions difficult. 

In short - There is no well-defined and project-supported method to extend default workflows to account for specific user needs, particularly in regards to pre-preservation processing activities. Enduro would greatly benefit from a standardized and documented method of extending the existing workflows locally.

sallain:
Still needs documentation and a little more work but this can be closed in general

PR #1061, opened 2024-11-07
Add poststorage child workflows

jraddaoui:
Allow to configure a set of poststorage child workflows that will be
started after AIP storage. These workflows will receive the AIPUUID
as a parameter and the parent workflow will only wait for them to be
started by Temporal. They are started with a disconnected context and
using the abandon parent close policy, so they can continue running
after the parent workflow finishes, therefore their results are ignored.

Refs #886.

### DIFF
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -37,6 +37,7 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/enums"
 	"github.com/artefactual-sdps/enduro/internal/fsutil"
 	"github.com/artefactual-sdps/enduro/internal/package_"
+	"github.com/artefactual-sdps/enduro/internal/poststorage"
 	"github.com/artefactual-sdps/enduro/internal/preprocessing"
 	"github.com/artefactual-sdps/enduro/internal/temporal"
 	"github.com/artefactual-sdps/enduro/internal/watcher"
@@ -762,6 +763,10 @@ func (w *ProcessingWorkflow) SessionHandler(
 				return err
 			}
 		}
+
+		if err := w.poststorage(sessCtx, tinfo.SIPID); err != nil {
+			return err
+		}
 	} else if !tinfo.req.AutoApproveAIP {
 		// Record package rejection in review preservation task
 		{
@@ -1040,6 +1045,10 @@ func (w *ProcessingWorkflow) transferAM(ctx temporalsdk_workflow.Context, tinfo 
 		}
 	}
 
+	if err := w.poststorage(ctx, tinfo.SIPID); err != nil {
+		return err
+	}
+
 	// Delete transfer.
 	activityOpts = withActivityOptsForRequest(ctx)
 	err = temporalsdk_workflow.ExecuteActivity(activityOpts, am.DeleteTransferActivityName, am.DeleteTransferActivityParams{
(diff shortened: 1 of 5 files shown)


internal/workflow/processing.go
```diff
@@ -1094,6 +1103,36 @@ func (w *ProcessingWorkflow) preprocessing(ctx temporalsdk_workflow.Context, tin
 	}
 }
 
+// poststorage executes the configured poststorage child workflows. It uses
+// a disconnected context, abandon as parent close policy and only waits
+// until the workflows are started, ignoring their results.
+func (w *ProcessingWorkflow) poststorage(ctx temporalsdk_workflow.Context, aipUUID string) error {
```
djjuhasz, line 1129: I think we will want to capture the results of the poststorage workflow for display in the Dashboard at some point, which may require rethinking the child workflow policy, but I think this is fine for now.

internal/workflow/processing.go
```diff
@@ -760,6 +761,10 @@ func (w *ProcessingWorkflow) SessionHandler(
 				return err
 			}
 		}
+
+		if err := w.poststorage(sessCtx, tinfo.SIPID); err != nil {
```
djjuhasz, line 767: I think the AIS poststorage workflow is just going to fail if we aren't using AM for preservation right now. I think you should either leave this out, or add some sort of escape path in the AIS workflow if a3m is configured as preservation system.

djjuhasz (changes requested):
Thanks @jraddaoui. I had a couple of thoughts, but one doesn't require any changes.  I think you need to address what happens if the poststorage workflow is called with an a3m vs. AM preservation system.

internal/workflow/processing.go
jraddaoui: I'm not sure I get your point. If they are not using AM it will probably fail before reaching this point for them. But we still want to have the option to run poststorage child workflows after a3m and storage as a general solution. Then the child workflow could use Enduro's storage API to fetch the AIP, instead of AMSS as the AIS workflow does.

The child workflow must be developed based on the custom needs, because they will only use AM we don't need to consider a3m at this point in the AIS workflow. Which would require a different implementation, as Enduro doesn't have an extract file endpoint.

internal/workflow/processing.go
jraddaoui: We may do that at some point. I initially considered that and I was planning to add a "parentClose" and "failParentWorkflow" config vars for each configured workflow. Something for the future, it will require to change the policy, how we wait for each workflow, add an expected result type, etc.

internal/workflow/processing.go
djjuhasz: Okay, but if a system administrator configured Enduro to run the AIS poststorage workflow and forgot to configure AM as the preservation system, then the poststorage workflow would error out trying to create the AIS package. Because there's no reporting on poststorage events in the Dashboard the operator would have to look at the Temporal UI to try and figure out why AIS packages aren't being delivered.

jraddaoui: Sure, we may want to add some information about them in the API/UI in the future. But I don't see that as a reason to remove this call or to change the AIS child workflow.

djjuhasz: My reservations come from a few things:

1. The future of Enduro with a3m is uncertain. We may never exercise the code in the a3m branch. It seems wrong to add code that is not used or tested.
2. The poststorage child workflow has no way to know which preservation system Enduro is using. This seems like a  problem when implementing poststorage workflows that depend on one particular preservation system.
3. Having two call points for the same poststorage workflow where one works and the other fails seems like a potential source for weird and confusing bugs.

I don't understand your argument that "If they are not using AM it will probably fail before reaching this point for them."  I think the a3m workflow should work fine up until the poststorage call (I may be wrong , I haven't tested this branch in a while).  At this point, if Enduro is configured to run with a3m and a poststorage workflow is configured to run that only works with AM, then this call will return an error.

jraddaoui: Welcome back @djjuhasz!

1. I was just talking about this with @aseles13, for now the a3m branch is still there, so we have to consider it. Not for them, but as a general solution. Same as with [the PREMIS validation work](https://github.com/artefactual-sdps/enduro/pull/1023#discussion_r1819505549).
2. Child workflows are a way to extend the workflow with custom activities based on the user needs. If the user needs to consider both preservation/storage systems, the child workflow should consider that too. Because both things are configurable, you can have a poststorage child workflow that only works with AM.
3. Same as above, if you are using a3m in one Enduro instance and AM in the other, link each of them with their corresponding child workflow (or make the child workflow work with both).
4. I test the a3m workflow branch from time to time, if you enable their preprocessing child workflow, a3m fails on "Verify metadata directory checksums", if you don't use it it works (tested with `little_vecteur_aip.zip`). Again, I'd consider a configuration issue.

djjuhasz: Okay. I'd prefer to have a way to have a poststorage workflow to exit cleanly (without an error) if it only works with AM, and Enduro is configured for a3m, rather than relying on correct configuration, but you obviously disagree. I'll approve the PR.

djjuhasz (approved):

jraddaoui:
Yes, if a child workflow only works with AM and Enduro is configured to run with a3m, I'd prefer that child workflow to fail to know I misconfigured the system.

As I said, we may want to improve that at some point to be able to report the child workflows in the API/UI. For now they are just like webhooks to be able to trigger custom workflows after storage, and we only make sure those workflows are started by Temporal. Those workflows can be monitored elsewhere for now, until we decide if we want to add information about them in the API/UI.

I'll add some tests and ping you for a final review.

djjuhasz (approved):
LGTM :+1:

Issue #950, opened 2024-05-31
Problem: AIP creation shows too many jobs in the Enduro UI that shouldn't be visible


fiver-watson:
**Describe the bug**

When running an AIP creation workflow, Archivematica currently displays 124 tasks. Many of these are for decisions the user is not prompted to make (e.g. "Run File format identification?") based on the configuration, or as steps that a user does not need to see (e.g. "Moving to processing directory"). This clutters the interface and makes it harder for operators to easily glean useful workflow information.  

**To Reproduce**

1. Start an ingest
2. Wait for it to complete processing
3. Examine the result on the package details page of the related SIP

**Expected behavior**

Operators should only see key tasks shown in the Enduro UI when running preservation workflows, that impart useful information on what tasks are currently being performed.

**Additional context**

This issue is the first in what will hopefully be a multi-part development effort, to eventually add a more curated and customized list of tasks in a preservation / AIP creation workflow, with useful outcome information (and/or error information) shown for each. New issues will be filed for future increments. 

Analysis work on this topic is being done in this spreadsheet: 

* https://docs.google.com/spreadsheets/d/10jByMHs6kTl4B22VCnwlEd2tLPb5tdJ595wZg-J96ts/edit#gid=1083719011

For now / for this issue, any row marked in red in the "AM-Enduro list (2024)" sheet should be prevented from displaying in the Enduro user interface.

sallain:
Note: even though this isn't really specific to error messages, it's a precursor to error message work that we will do in future sprints.

fiver-watson:
Hi @jraddaoui !

Overall, I think this is looking good. 

Of the 3 activities you identified that we didn't have included originally  in the spreadsheet - i.e.:

* Scan for viruses in directories (row 17)
* Scan for viruses in submission documentation (row 94)
* Scan for viruses in metadata (row 106)

I think that if these activities ARE configured in AM to run, then it makes sense to show them in the Enduro UI. It's comforting to know that all of these things are being scanned, lol! So, unless Sara disagrees when she reviews this, I would say no changes needed here. 

A couple quick questions for you - in reviewing the spreadsheet against a `little_borndigital_SIP_20111020_BFB_v60.zip` sample in Enduro, I did find 2 rows in the spreadsheet that were not present in the UI: 

* Validate preservation derivatives (row 75)
* Check for manual normalized files (row 81)

[...]

PR #1063, opened 2024-11-12
Reduce preservation tasks created from AM jobs

jraddaoui:
Create preservation tasks from only a set of the Archivematica transfer and ingest jobs. Looking at the initial data and the UUIDs used in the default `workflow.json` chains from Archivematica source code, define the map of jobs we want to keep using their `LinkID`.

Refs #950.

### DIFF
--- a/internal/am/job_tracker.go
+++ b/internal/am/job_tracker.go
@@ -12,6 +12,53 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/package_"
 )
 
+var keepJobs = map[string]struct{}{
+	// Transfer jobs.
+	"3229e01f-adf3-4294-85f7-4acb01b3fbcf": {}, // Extract zipped bag transfer
+	"154dd501-a344-45a9-97e3-b30093da35f5": {}, // Rename with transfer UUID
+	"3c526a07-c3b8-4e53-801b-7f3d0c4857a5": {}, // Assign file UUIDs to objects
+	"c77fee8c-7c4e-4871-a72e-94d499994869": {}, // Assign checksums and file sizes to objects
+	"5415c813-3637-49ab-afec-9b435c2e4d2c": {}, // Assign UUIDs to directories
+	"1c2550f1-3fc0-45d8-8bc4-4c06d720283b": {}, // Scan for viruses in directories
+	"56eebd45-5600-4768-a8c2-ec0114555a3d": {}, // Generate transfer structure report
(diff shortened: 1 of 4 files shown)


internal/am/job_tracker.go
```diff
@@ -12,6 +12,53 @@ import (
...
+	"873b428f-2c86-42b6-b463-aeda925bf559": {}, // Load persistent identifiers from external file
+	"ccf8ec5c-3a9a-404a-a7e7-8f567d3b36a0": {}, // Generate METS.xml document
+	"523c97cc-b267-4cfb-8209-d99e523bf4b3": {}, // Add README file
+	"3e25bda6-5314-4bb4-aa1e-90900dce887d": {}, // Prepare AIP
+	"d55b42c8-c7c5-4a40-b626-d248d2bd883f": {}, // Compress AIP
+	"3f543585-fa4f-4099-9153-dd6d53572f5c": {}, // Verify AIP
+	"20515483-25ed-4133-b23e-5bb14cab8e22": {}, // Store the AIP
+	"48703fad-dc44-4c8e-8f47-933df3ef6179": {}, // Index AIP
+	"b7cf0d9a-504f-4f4e-9930-befa817d67ff": {}, // Clean up after storing AIP
+}
```
jraddaoui, line 60: I was initially planning to split this map and pass two different maps from the activities (one for the transfer jobs and the other for the ingest jobs), adding a new field to the `JobTracker` struct, setting it on the `NewJobTracker` and checking it while filtering. But in the end I thought that having all the jobs here may be better. Happy to change it back if you think that's better.

internal/am/job_tracker.go
```diff
@@ -105,16 +152,18 @@ func (jt *JobTracker) savePreservationTasks(ctx context.Context, jobs []amclient
 	return count, nil
 }
 
-// filterSavedJobs filters out jobs that have an ID in saved then returns the
-// filtered job list.
-func filterSavedJobs(jobs []amclient.Job, saved map[string]struct{}) []amclient.Job {
-	var unsaved []amclient.Job
+// filterSavedJobs filters out jobs that have an ID in saved and jobs without
+// LinkID in the jobs we want to keep, then returns the filtered job list.
+func (jt *JobTracker) filterSavedJobs(jobs []amclient.Job) []amclient.Job {
```
djjuhasz: I think this method should be renamed to `filterJobs()` or something else that doesn't imply that it's only filtering out saved jobs.

djjuhasz (approved):
LGTM :+1: I did suggest renaming the `filterSavedJobs()` method because its scope has expanded.

PR #1064, opened 2024-11-13
Copy premis.xsd into enduro image

jraddaoui:
This file should only be needed by the AM and a3m workers to do PREMIS validation when it's enabled. However, config validation is the same for all commands and, when PREMIS validation is enabled, it checks the existence of the XSD file in the filesystem. We tried to disable PREMIS validation only in the enduro container, but that config value is read in the workflow context, disabling the scheduling of the activity from the workflow worker.

Refs #951.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,30 +56,31 @@ FROM gcr.io/distroless/base-debian12:latest AS base
 USER 1000
 
 FROM base AS enduro
-COPY --from=build-enduro --link /out/enduro /home/enduro/bin/enduro
-COPY --from=build-enduro --link /src/enduro.toml /home/enduro/.config/enduro.toml
+COPY --link --from=build-enduro /out/enduro /home/enduro/bin/enduro
+COPY --link --from=build-enduro /src/enduro.toml /home/enduro/.config/enduro.toml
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -56,30 +56,31 @@ FROM gcr.io/distroless/base-debian12:latest AS base
 USER 1000
 
 FROM base AS enduro
-COPY --from=build-enduro --link /out/enduro /home/enduro/bin/enduro
-COPY --from=build-enduro --link /src/enduro.toml /home/enduro/.config/enduro.toml
+COPY --link --from=build-enduro /out/enduro /home/enduro/bin/enduro
+COPY --link --from=build-enduro /src/enduro.toml /home/enduro/.config/enduro.toml
+COPY --link hack/xsd/premis.xsd /home/enduro/premis.xsd
```
jraddaoui, line 61: This line is the main diff, all others are just adding missing `--link` to the `COPY` layers and reorganizing them a little.

jraddaoui:
We should try to find a better solution to how we deal with the single configuration for multiple commands in distributed environments, but this makes PREMIS validation work for now in the dev environment. Other deployments are running everything together, so it shouldn't be an issue there.

mcantelon (approved):
LGTM!

PR #1065, opened 2024-11-13
Update dashboard status filter tab icons

djjuhasz:
Update status filter tab icons for better representation and consistency.


jraddaoui (approved):
Thanks!

Issue #1066, opened 2024-11-14
Problem: API access control allows tokens without Enduro attributes


jraddaoui:
**Describe the bug**

A user without Enduro attributes in the access token given by the OIDC provider may be able to access all API endpoints with ABAC enabled in the configuration.

**To Reproduce**

* Enable auth and ABAC in the API.
* Make sure the OIDC provider gives an access token without Enduro attributes:
  * The configured claim must be included in the token (some providers don't return an empty claim).
  * Parsing the attributes from the claim may involve filtering and roles mapping based on configuration (which could help to create an empty attributes slice).
* Make a request to the API endpoint with that token.

**Expected behavior**

All API endpoints at the moment should return a `403 Forbidden` error code.

**Additional context**

The [parsing function](https://github.com/artefactual-sdps/enduro/blob/main/internal/api/auth/token_verifier.go#L76-L148) returns a `nil` slice instead of an empty one if there are no Enduro attributes after parsing, filtering and mapping roles. That `nil` attributes slice is considered as ABAC disabled in the [check attributes function](https://github.com/artefactual-sdps/enduro/blob/main/internal/api/auth/claims.go#L15-L44) used from the endpoints.

sallain:
Closing for now, client will let us know if there are any issues.

PR #1067, opened 2024-11-14
Fix API ABAC for access tokens without attributes

jraddaoui:
Make sure we return an empty slice of attributes in the claim if no Enduro attributes are found in the token after parsing, filtering and mapping roles. A nil attributes slice is considered as ABAC disabled in the API endpoints check.

Refs #1066.

### DIFF
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -1284,7 +1284,6 @@ func (w *ProcessingWorkflow) validatePREMIS(
 		XMLPath: xmlPath,
 		XSDPath: w.cfg.ValidatePREMIS.XSDPath,
 	}).Get(activityOpts, &xmlvalidateResult)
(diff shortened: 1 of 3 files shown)


internal/workflow/processing.go
```diff
@@ -1284,7 +1284,6 @@ func (w *ProcessingWorkflow) validatePREMIS(
 		XMLPath: xmlPath,
 		XSDPath: w.cfg.ValidatePREMIS.XSDPath,
 	}).Get(activityOpts, &xmlvalidateResult)
-
```
jraddaoui, line 1287: Added by `make lint`, I didn't want to create another commit just for it ;)

djjuhasz (approved):
LGTM @jraddaoui. :+1:

Issue #1062, opened 2024-11-08
Enhancement: Add "About" information to user menu with application and child workflow versions


fiver-watson:
**Is your feature request related to a problem? Please describe.**

When performing iterative development and testing with clients, it can be difficult for end user to be able to know (and tell us) what version of Enduro they are using - for minor changes in validation between subsequent deployments, this can be critical for QA and troubleshooting in general, as well as general product development. 

It would be nice for there to be an easy way for users to learn more about the application version - and ideally, the version (or branch) of any child workflow activities they have loaded as well. 

**Describe the solution you'd like**

I propose adding an "About" link to the User Menu (i.e. where users can logout once authenticated - currently in the bottom left of the navigation sidebar, but likely to be moved to the upper right corner of the header). 

This would be have similar to the "About" links available in browsers - It would list the full release version (and/or branch name if there are no releases tagged for that particular version?) for the application, and ideally, the  release version of any loaded child workflow activities as well. 

I would finally suggest that it include a link for help, that links to the current version's documentation index page  at `/docs/src/index.md` (e.g. for the 0.3.0 release, it would link to: `https://github.com/artefactual-sdps/enduro/blob/v0.3.0/docs/src/index.md`), and a link to the project's LICENSE and CONTRIBUTING markdown files. 

This could be mocked up further as needed, but I am personally not picky about the implementation - on click, the "About" link could open a modal (similar to Firefox's About page), or a different tab (similar to Chrome's About page), or the information could just be listed right there below the Logout link in the user menu.  The goal is to make information about the application and the project overall more easily discoverable in the user interface.  

**Describe alternatives you've considered**

* It could be included on the Welcome / Landing page - a fine option I would be happy with, but one that's not accessible from anywhere and requires leaving the current page
* It could be that the About link merely takes the user to the relevant release tag page in the Enduro repo - but this is a bunch of extra clicks that doesn't provide more information, and wouldn't include any child workflow version information. 
* If and when we have a dedicated Settings page, it could be there, but we don't at present

fiver-watson:
Some initial wireframes, going with a modal (and assuming we will use this opportunity to move the user menu to the top right): 

![Image](https://github.com/user-attachments/assets/66c7606c-c2b5-434b-9186-c8017fe12d7e)
![Image](https://github.com/user-attachments/assets/cfb7d069-d1a8-4677-b431-6011b599f470)
[...]

PR #1075, opened 2024-11-22
Add about service/endpoint to API

jraddaoui:
Add a new service and endpoint to the API to get system information. Include short version, preservation system, and preprocessing and poststorage child workflows configuration.

Refs #1062.


djjuhasz (approved):
LGTM @jraddaoui! :+1:

Issue #976, opened 2024-06-25
Problem: Vue's Reactivity Transform experimental feature has been removed


jraddaoui:
**Is your feature request related to a problem? Please describe.**

https://vuejs.org/guide/extras/reactivity-transform.html

> Reactivity Transform was an experimental feature, and has been removed in the latest 3.4 release. Please read about [the reasoning here](https://github.com/vuejs/rfcs/discussions/369#discussioncomment-5059028).
>
> If you still intend to use it, it is now available via the [Vue Macros](https://vue-macros.sxzz.moe/features/reactivity-transform.html) plugin.

**Describe the solution you'd like**

While we use it via the Macros plugin, we should reconsider if we really need to use it.

**Describe alternatives you've considered**

Keep it.

**Additional context**

@djjuhasz noted this issue in https://github.com/artefactual-sdps/enduro/pull/973

PR #1078, opened 2024-11-23
Stop using Vue's reactivity transform

jraddaoui:
Refs #976.

### DIFF
--- a/dashboard/src/components/PreservationActionCollapse.vue
+++ b/dashboard/src/components/PreservationActionCollapse.vue
@@ -4,54 +4,56 @@ import PackageReviewAlert from "@/components/PackageReviewAlert.vue";
 import StatusBadge from "@/components/StatusBadge.vue";
 import { useAuthStore } from "@/stores/auth";
 import Collapse from "bootstrap/js/dist/collapse";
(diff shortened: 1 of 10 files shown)


dashboard/src/components/PreservationActionCollapse.vue
```diff
@@ -4,54 +4,58 @@ import PackageReviewAlert from "@/components/PackageReviewAlert.vue";
...
-const { action, index, toggleAll } = defineProps<{
+const props = defineProps<{
   action: api.EnduroPackagePreservationAction;
   index: number;
   toggleAll: boolean | null;
 }>();
 
+const action = toRef(props, "action");
+const index = toRef(props, "index");
+const toggleAll = toRef(props, "toggleAll");
```
sevein: Would this work?

    const { action, index, toggleAll } = toRefs(props);

sevein (approved):
:+1:

dashboard/src/components/PreservationActionCollapse.vue
jraddaoui: 😲 `toRefs`, it does.

PR #1086, opened 2024-11-30
Change preservation tasks from table to cards

### LINKED ISSUE #1077: Problem: Preservation activity task list is too condensed, not very accessible as a table
### The problem

* Table view of task list does not have enough room to include task STDOUT / STDERR information
* Table is not very responsive (horizontal scrollbar added on narrow devices is a sub-optimal user experience)
* Table view will not easily accommodate buttons required for input on Pending tasks, such as reviewing a deletion request
* If we ever decide to add more information to the task view, there is currently no room - table is already pretty overloaded with the current number of columns
* Tables are less accessible overall

### Proposed solution

After some rounds of discussion and wireframing, we have decided to try implementing [cards](https://bootstrap-vue.org/docs/components/card) for tasks, instead of a table. This should: 

* Be more responsive and accessible
* Allow for additional information to be added in the future
* Better support potential task inputs, such as approving or rejecting Pending deletion requests
* Allow us to add Archivematica task STDOUT / STDERR log info to tasks in a visually manageable way

**Mockup**: 

Image

We will add the expandable sections with AM STDOUT/ERR information in a separate ticket. The mockup above should provide a general sense of the card design. 

I also included some initial proposals for how these cards would display on narrow device widths - for example: 

* On completed tasks, show only the completed time. If tasks are still in progress, show when it was started
* When the card is collapsed, use overflow / ellipses for task titles and notes. When expanded, we can switch to wrapping instead to show the full content (if that is possible!) 
* For error and pending messages - as we slowly and incrementally implement updates to these (see Epic, [preprocessing-sfa#43](https://github.com/artefactual-sdps/preprocessing-sfa/issues/43)), error and pending messages will have a bolded summary first, and then details below. In a condensed view, only show the first line

djjuhasz:
I've switched the preservation tasks over to cards (note: the [Bootstrap 5 card syntax](https://getbootstrap.com/docs/5.0/components/card/) has changed from the docs @fiver-watson linked in #1077).

I think the current implementation looks pretty good at desktop screen sizes, e.g:
Screenshot from 2024-11-29 16-36-18

But the card "columns" are overflowing on narrow screen widths :(
Screenshot from 2024-11-29 16-34-13

I'll do some more reading on how to fix the overflow problem and maybe @jraddaoui will have some ideas too.

### DIFF
--- a/dashboard/src/components/PreservationActionCollapse.vue
+++ b/dashboard/src/components/PreservationActionCollapse.vue
@@ -1,6 +1,10 @@
 <script setup lang="ts">
 import type { api } from "@/client";
 import PackageReviewAlert from "@/components/PackageReviewAlert.vue";
+import type {
+  EnduroPackagePreservationTask,
+  EnduroPackagePreservationTaskStatusEnum,
+} from "@/openapi-generator";
 import StatusBadge from "@/components/StatusBadge.vue";
 import { useAuthStore } from "@/stores/auth";
 import Collapse from "bootstrap/js/dist/collapse";
@@ -54,6 +58,10 @@ watch(toggleAll, () => {
 
 let expandCounter = ref<number>(0);
 watch(expandCounter, () => show());
+
+function isComplete(task: EnduroPackagePreservationTask) {
+  return task.status == "done" || task.status == "error";
+}
 </script>
 
 <template>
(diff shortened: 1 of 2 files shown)


djjuhasz:
The columns also vary in width when the screen is too narrow :(
Screenshot from 2024-11-29 16-35-35

djjuhasz:
Here's an example with errors:
Screenshot from 2024-11-29 16-47-12

djjuhasz:
Oops, I fixed the line breaks on the error message text:
Screenshot from 2024-11-29 16-50-00

djjuhasz:
With @jraddaoui's help I've fixed the problem with the card values breaking out of the card body
Screenshot from 2024-12-02 17-22-22

fiver-watson:
@djjuhasz No idea how many of my suggestions are actually viable in terms of responsiveness within the BS5 framework etc, but I did include some initial suggestions to manage narrow screens on the original ticket, including some mockups. Those included: 

> - On completed tasks, show only the completed time. If tasks are still in progress, show when it was started
> - When the card is collapsed, use overflow / ellipses for task titles and notes. When expanded, we can switch to wrapping instead to show the full content (if that is possible!)
> - For error and pending messages - as we slowly and incrementally implement updates to these (see Epic, https://github.com/artefactual-sdps/preprocessing-sfa/issues/43), error and pending messages will have a bolded summary first, and then details below. In a condensed view, only show the first line

Also added a response w some other relevant ideas on the ticket conversation, [here](https://github.com/artefactual-sdps/enduro/issues/1077#issuecomment-2514515040)

Either way, great progress so far!

djjuhasz:
@fiver-watson Thank, I missed your mockup and suggestions on how to display cards on a narrow screen. I've responded to your suggestions below:

> - On completed tasks, show only the completed time. If tasks are still in progress, show when it was started

I think this is a good idea that should be extended to the full width display. I think showing both the start and completed time for each task in the collapsed view is unnecessary and adds extra visual noise.  If an operator wants to see the started time for a completed task I think it's reasonable that they click on the card to expand it to get that information.

> - When the card is collapsed, use overflow / ellipses for task titles and notes. When expanded, we can switch to wrapping instead to show the full content (if that is possible!)

I think we implemented something like this in AtoM and I think it's a good idea. I'm not sure if eliding the text can be accomplished with pure CSS or if it would require Javascript to implement.  Either way, I believe it's possible but I will have to do some reading to figure out how to implement it.

> - For error and pending messages - as we slowly and incrementally implement updates to these (see [Epic #43](https://github.com/artefactual-sdps/preprocessing-sfa/issues/43), error and pending messages will have a bolded summary first, and then details below. In a condensed view, only show the first line

I think we should only show the first line in the collapsed card at all screen widths.

jraddaoui:
For overflow/ellipses, check https://getbootstrap.com/docs/5.3/helpers/text-truncation/, we use it in a few places already.

jraddaoui:
One easy thing that could also help in small widths, I think the title font size is too big. If we use strong text, it looks better to me with the default font size.

djjuhasz:
> For overflow/ellipses, check https://getbootstrap.com/docs/5.3/helpers/text-truncation/, we use it in a few places already.

@jraddaoui unfortunately "text-truncate" doesn't work with `display: flex` so it isn't working with the card flexbox layout. :(

djjuhasz:
I made some progress today, but moving the "status" badge to it's own column in flexbox layout causes it to break out of the card on narrow screen widths again :(
image

I did manage to fix this at one point by allowing the task label to wrap to the next line.  The ideal fix would be truncating the text (with ellipsis) but text truncation isn't working with the the flex divs. :(

fiver-watson:
Hi @djjuhasz, thanks so much for your persistence on this issue (i know UI design stuff can be really finnicky) and for your very helpful thoughts! 

I am fine with all of your suggestions in your  [comment above](https://github.com/artefactual-sdps/enduro/pull/1086#issuecomment-2518605437) - they make good sense and I agree that they will improve the user experience. Thanks! 

> I think we should only show the first line in the collapsed card at all screen widths.

This totally makes sense - I guess I just wanted to make sure that regardless of screen width, AT LEAST the first line is shown for errors and warnings, where it's important to be able to get an initial sense of why / what's going on at a glance. In most cases, I think that success messages are not going to be more than 1 line anyway.

djjuhasz:
I've reached a good stopping point where the cards nicely display on a large screen or a small screen and I think this is a good first iteration.

image

I will come back to this later to add the ability to expand and collapse the cards to show detailed log and error information.

mcantelon (approved):
Kudos for taming the eternal dragon of responsive layout! My understanding of this realm is limited, but seems good to me!

PR #1087, opened 2024-12-06
Allow relation between child workflows

jraddaoui:
Consider a `PreprocessingID`received in the preprocessing child workflow result and pass it as parameter to poststorage child workflows.

Refs #886.

### DIFF
--- a/internal/poststorage/poststorage.go
+++ b/internal/poststorage/poststorage.go
@@ -7,5 +7,6 @@ type Config struct {
 }
 
 type WorkflowParams struct {
-	AIPUUID string
+	AIPUUID         string
+	PreprocessingID string
 }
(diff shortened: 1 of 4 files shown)


jraddaoui:
@djjuhasz @mcantelon, this could be merged if you think it's okay. I'm still working on the related PR (https://github.com/artefactual-sdps/preprocessing-sfa/pull/99) which explains a little more the purpose of these changes. I'll work on documentation for all child workflow related changes done recently when I'm back in January.

djjuhasz:
@jraddaoui I left a comment on https://github.com/artefactual-sdps/preprocessing-sfa/pull/99 about this, but I'll repeat it here — why don't you use the WorkflowID? It's a unique identifier for the processing workflow execution that's already available in the workflow code and tests.

jraddaoui:
Thanks @djjuhasz, check https://github.com/artefactual-sdps/preprocessing-sfa/pull/99#discussion_r1906046652.

PR #1091, opened 2024-12-19
Add glossary to user manual docs

fiver-watson:
Modified the glossary developed for SFA for general use in the documentation. I mentioned this to @sallain before pushing, but a couple notes on some of the decisions made here: 

* Markdown doesn't have native support for glossaries, and definition lists are also part of extended markdown. I wasn't sure if Mkdocs supports definition lists, but you also can't hyperlink to a definition anyways. In the end, I decided to use level 3 headers for the terms, with the definitions just added in text below. 
* I have added this to the User Manual, and not as a glossary for all manuals, because this is a glossary of highly-specific domain language, intended primarily for operators with familiarity in digital preservation concepts and terminology. Some definitions may not match exactly how terms are used in a developer's context, and I wanted to keep the focus on the domain. We can always change this decision later, or add different glossaries to the other manuals as needed. 
* I have also chosen NOT to go through all existing user manual pages to find glossary terms used in text and link them back to the glossary. Sara and I propose we not do this throughout unless there are specific instances where it seems useful or appropriate - doing so every time a domain term from the glossary appears will be a writing and maintenance nightmare. Also, since markdown anchors are page-specific, you can't call them from anywhere, meaning the links would first need to link to the relevant md file, and then the anchor within that file - meaning each link would be annoying to look up, and the links would be long and intrusive when trying to write or edit documentation. We can always make more specific mention of the glossary on the User Manual README / landing page, but in general, the idea is that if readers encounter a term they don't understand or would like clarity on how it is used in the Enduro docs, they can go find it in the glossary.
* Similarly, in the glossary itself, I have **bolded** the first time a different glossary term is used in a given definition, rather than hyperlinking it. I mentioned this in the glossary's intro text - readers will get an indication when part of the definition uses another term they can look up or cross-reference. This was for consistency with the broader decision not to hyperlink every instance of a term throughout our documentation, and for ease of writing up this draft for review. Again, happy to change this based on feedback or in the future. 
* Finally - another thing basic markdown doesn't support is footnotes. We had one footnote referencing the PREMIS Data Dictionary in our original version of the glossary, which I recreated here with headings. I originally used a lower level 4 heading for the Footnotes section at the bottom of the page, but the Readthedocs theme then nested the "Footnotes" header under the last glossary term in the sidebar navigation menu, so I moved it all up a level. We could possibly just remove this one footnote, but I thought I would leave it for review and discussion in this PR. 

Looking forward to feedback!


fiver-watson:
goddamnit, that was premature apparently. fixing now...

fiver-watson:
Okay, ready to go!

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+  objects to increase the likelihood of successful future access, in case the
+  original objects use proprietary and/or rare formats that may not be easily
+  rendered in the future.
+* **Access derivative**: a derived version of a preservation object using file
+  formats deemed more suitable for dissemination and access across space and
+time (particularly over networked environments), based on factors such as: open
+format specifications, smaller file sizes, widespread adoption and support
+(particularly in web browsers), high compression rates, etc. Access derivatives
+are typically used in the creation of **Dissemination Information Packages
+(DIPs)**.
```
sallain: Bullet formatting went a bit funny here

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Document
+
+An **object** consisting of information or data fixed to some media.
+
+## Event
+
+A preservation-relevant action that involves at least one **object** and/or
+**agent**. Events are typically captured in the preservation **metadata** of a
+**package** during a preservation work**flow using the
```
sallain: `work**flow` should be `**workflow**`

sallain (approved):
Great work!

docs/src/user-manual/glossary.md
fiver-watson: Good catch, fixed

docs/src/user-manual/glossary.md
fiver-watson: thanks!

sallain (approved):

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
+# Glossary
+
+**NOTE**: This glossary of terms outlines the domain-specific language used when
```
djjuhasz, line 3: I don't think **NOTE:** is necessary as the text nicely explains the purpose of the Glossary.

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+
+-----
+
+## Activity
+
+See [Task](#task).
+
+## Agent
+
+An actor (human, machine, or software) associated with one or more **events**
```
djjuhasz, line 21: I thought the agent actively *initiates* an event, rather than just being associated with the event?

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+in a preservation **workflow**.
+
+## Archival Information Package (AIP)
+
+A type of package derived from a **Processing Information Package (PIP)** that
+is transformed during a preservation **workflow** into one or more AIPs,
```
djjuhasz, line 27: This definition is circular. I think "...into one or more AIPs," should be cut from this paragraph.  The detail that more than one AIP can be produced from a single PIP can be added to one of the subsequent paragraphs.

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+depending on the **preservation policies** defined in the **preservation engine**.
+
+An AIP is the output of the **preservation engine**, consisting of one or more
```
djjuhasz, line 30: Hmm, maybe this general definition of an AIP should be the first paragraph?

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+[OAIS](https://en.wikipedia.org/wiki/Open_Archival_Information_System) Reference
+Model created by the Consultative Committee for Space Data Systems (and since
+formally recognized as an international standard in ISO 14721). A detailed
+description of the AIP and its components is available in the Archivematica
+documentation - see: [AIP
+structure](https://www.archivematica.org/docs/latest/user-manual/archival-storage/aip-structure/#aip-structure).
+
+## Child workflow
+
+Child workflows are a feature of the [workflow engine](#workflow-engine),
```
djjuhasz, line 54: I think it would be better to start with a general definition of a child workflow, without limiting the definition to a feature of the workflow engine.  Something like "A child workflow is an ancillary **workflow** that is spawned by a parent workflow."  There are other ways to create and run a child workflow besides Temporal's child workflow mechanism.

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+allowing a **workflow** to spawn another workflow. This capability allows
+breaking down complex tasks into manageable units, improving maintainability,
+extensibility and scaling operations.
+
+In Enduro, we recommend using Child workflows to manage user-specific **tasks**
```
djjuhasz, line 59: If "Child Workflows" is a proper noun then I think both words should be capitalized, and otherwise neither word should be capitalized.

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+and **preservation actions** in a workflow, such as SIP validation against a set
```
djjuhasz, line 60: bold **SIP**?

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+of institution-specific requirements. This keeps locally specific
+implementations nicely separated from more general, reusable workflow activities
```
djjuhasz, line 62: "from *the* more general, reusable workflow activities..." (without emphasis).

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+in a workflow.
+
+## File
+
+A named and ordered sequence of binary data, organized as a content bitstream
+that can be stored and transmitted, and typically bearing a specific **format**.
+
+## File format
+
+A standardized method of specifying how bits are
```
djjuhasz, line 136: This line is broken before it's necessary.

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+used to encode information in a digital storage medium such as a **file**. A
+format's associated specification may be proprietary or open.
+
+## Ingest
+
+A phase in a preservation **workflow** describing all the **preservation
+policy**-defined **tasks** performed on a **SIP** when it is received from a
```
djjuhasz, line 143: The hyphen before "-defined" seems unnecessary.

djjuhasz: "...after it's received from a **producer**..."?

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+**producer**, prior to preservation. Ingest task examples include checking the
+file format against a list of acceptable formats and checking for hidden files.
+Typically, ingest concludes by finalizing the transformation of a SIP into a
+**PIP** for preservation processing by the **preservation engine**.
+
+## Intellectual entity
+
+The [PREMIS](https://www.loc.gov/standards/premis/) 3 Data Dictionary defines an
```
djjuhasz, line 151: The "3" looks so lonely outside the hyperlink. :pleading_face: Maybe "The [PREMIS v3 Data Dictionary](https://www.loc.gov/standards/premis/v3/premis-3-0-final.pdf) defines..."?

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+consideration for preservation. There is at least one object in every
+**package**, and each object can contain zero or more **files**. An object may
+be described by a **producer** using any desired descriptive **metadata** that
+is meaningful to the designated community.
+
+## Package
+
+An **intellectual entity** describing a type of collection, composed of a set of
+**files** and related **metadata**, assembled together for a particular purpose.
+A package will typically contain one or more **objects**, zero or more
```
djjuhasz, line 218: By the definition of "Object" above a package must *always* contain one or more objects.

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+## Preservation policy
+
+A policy or business rule defining the target inputs and/or outcomes of a
+preservation action or **workflow** **task**, and any supporting contextual
+information. For example, setting a policy in the **preservation engine** to
+**normalize** a JPEG image file to a JPEG2000 during **AIP** creation.
+
+## Preservation processing
+
+A phase in a preservation
```
djjuhasz, line 264: Short line.

djjuhasz: Also should "preservation workflow" be bold?

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+workflow describing all the tasks that occur after Ingest, when a PIP is sent to
+the preservation engine for transformation into one or more AIPs. Prior to final
+bagging, compression, and storage, any AIPs created during ingest may also
+undergo any additional Post-ingest activities defined in the system. Examples of
```
djjuhasz, line 268: "post-ingest" shouldn't be capitalized.

docs/src/user-manual/glossary.md
```diff
@@ -0,0 +1,368 @@
...
+An operation performed on a **file** or **package** in the context of a
+**workflow**. Also called an **activity**.
+
+## Transfer
+
+See: [Submission Information Package (SIP)](#submission-information-package-sip)
+
+## Watched directory
+
+A filesystem directory used that is configured with a small command-line utility
```
djjuhasz, line 337: I think this can be shortened to "A filesystem directory that is monitored for changes (e.g. adding, deleting, or renaming a file), and where such a change may trigger one or more subsequent actions (e.g. preservation processing)."

djjuhasz (commented):
@fiver-watson I have requested many changes, but do not despair for they are all small changes. :)

djjuhasz:
Oh,  this was merged already. Nevermind. :(

fiver-watson:
Sorry @djjuhasz - wasn't sure if you wanted to get into the nitty gritty of domain stuff, so after sara reviewed and approved, I merged it... but it's good feedback - i can use it on a new PR. Thanks for taking the time!

docs/src/user-manual/glossary.md
fiver-watson: This one I am less in agreement with. I was trying to avoid getting too detailed, but in the PREMIS Data Dictionary for example, An Agent: 

> May hold or grant one or more Rights. 
> May carry out, authorize, or compel one or more Events. 
> May create or act upon one or more Objects through an Event or with respect to a Rights statement.

So, much like in archival descriptions, an actor is not always the creator.

docs/src/user-manual/glossary.md
fiver-watson: :+1:

docs/src/user-manual/glossary.md
fiver-watson: ok, i have tried rewording the whole AIP definition to avoid tautology a bit more.

docs/src/user-manual/glossary.md
fiver-watson: ok, i have taken your advice and tried rewording the definition a bit. That said, while i agree that it's not just a _Temporal_ feature, this was part of the whole reason I reworked the Components docs to be more general. Workflows in general are going to be managed by some kind of "workflow engine" in Enduro, regardless of whether it's Cadence, Temporal, or something else.  If we ever found a need and a way to run workflows and child workflows in the Enduro stack outside of the workflow engine, then I had thought that would be a big enough change to warrant updating the definition. 

In any case, I removed the reference to the workflow engine from this definition.

docs/src/user-manual/glossary.md
fiver-watson: :+1:

docs/src/user-manual/glossary.md
fiver-watson: sure... but adding "must" and "may" gets a bit complex. I just removed the word "typically" - does that work?

docs/src/user-manual/glossary.md
fiver-watson: lol i missed a lot of terms that should have been bolded in that one. Fixed.

docs/src/user-manual/glossary.md
fiver-watson: :+1:

docs/src/user-manual/glossary.md
fiver-watson: updated!

fiver-watson:
Ok @djjuhasz I have replied to your comments above, but implemented the changes in a new PR - see #1097

docs/src/user-manual/glossary.md
djjuhasz: Sure, sounds good. :+1:

docs/src/user-manual/glossary.md
djjuhasz: Hmm, interesting. I always assumed "actor" implied "action". Association seems like a weaker criteria then the word "actor" implies. Also the general definitions of "agent" that I've seen all include the concept of taking an active role or taking action.

PR #1092, opened 2024-12-29
WIP: Restrict transfer uploads to ZIP files

mcantelon:


jraddaoui:
I did something similar in https://github.com/artefactual-sdps/enduro/commit/761726cbbf143e85785a214bfdad23398467c058. We may want to prevent directory uploads in the dev. env., but the extraction task should work with multiple compression formats, not just ZIP.

Issue #986, opened 2024-07-09
Problem: Preservation action tasks are difficult to find in the Enduro UI


fiver-watson:
**Describe the bug**

Based on user observation during Talk-Aloud tests conducted with members of SFA during PoC#2, many users did not initially realize that the Preservation actions could be expanded to show tasks. Consequently, they had difficulty in using Enduro to learn more about the tasks being performed as part of a preservation workflow, such as SIP identification outcomes, or finding errors during failed ingests. 

**To Reproduce**

Notes about this testing are captured in the following document: 

* [SFA PoC#2 - Talk-Aloud Testing: Tasks and observation notes (2024-07-02/03)](https://docs.google.com/document/d/1lt4woB-TzOJJuoXR-L8gE_6SFCfG1rVnZ1bcno23Dk0/edit)

Essentially, participants were asked to ingest a SIP, and then use the Enduro UI to determine the SIP type. One group took a long time to find that the Preservation actions could be expanded, finding it almost by accident. The other group did not find it at first, until participants switched. In both cases, had participants not been motivated by the artificial nature of the test to continue looking, they likely would have given up before finding the information they needed. 

**Expected behavior**

It should be more apparent to users that the Preservation actions can be expanded, and that potentially useful task information can be found below. 

**Additional context**

After some internal analysis and discussion, our team has decided to try implementing 2 different UI changes to address this. 

First, is to make it more obvious that the preservation actions are expandable. At present, users must click directly on the caret to expand. The caret does not change on hover, and the rest of the area is not clickable. 

Our proposal is to: 

* Make the entire preservation actions area clickable
* Give the entire area hover effects: 
  * Outline the clickable area in purple
  * Lightly grey the enitre clickable area
  * Change the caret icon to a solid fill (or similar)

Mockup: 
![enduro-pres-actions-hover-effects](https://github.com/artefactual-sdps/enduro/assets/5068277/5d9809c9-f636-44b3-a379-6c11b38bee38)

Second, we propose adding more help information to the UI. At present, the Enduro UI uses a mix of domain-specific and application-specific language, such as "Preservation actions," "Tasks," "Workflows" etc. Users will have varying levels of familiarity with domain terminology, and new users will definitiely not know Enduro-specific terminology on first use. 

[...]

PR #1103, opened 2025-01-15
Implement accordion preservation action task lists

djjuhasz:
Refs issue #986

Use the Bootstrap 5 accordion component to implement collapsable preservation task lists.  The accordion component supports using a large button as a section header and closing open sections when a new section is opened.

The specific #986 features implemented are:
- Add a border around the preservation action header to hint that it is a control widget
- Make the entire preservation action header clickable to collapse or expand the related task list
- Change the preservation action header border and background colors on hover


djjuhasz:
Here's how it looks in action.
[Screencast from 2025-01-14 17-53-00.webm](https://github.com/user-attachments/assets/3a7401fc-226a-4a99-948c-5f0f608bb608)

dashboard/src/styles/main.scss
```diff
@@ -33,3 +33,18 @@ a:not(.btn),
...
+.accordion-button {
+  background-color: $light;
+  border: solid 1px $gray-500;
+}
+
+.accordion-button:hover {
+  background-color: $primary-bg-subtle;
+  border-color: $primary;
+  border-width: 2px;
+}
```
jraddaoui: We should also target the focus state.

dashboard/src/pages/packages/[id]/index.vue
```diff
@@ -71,38 +71,19 @@ let toggleAll = ref<boolean | null>(false);
...
-        v-for="(action, index) in packageStore.current_preservation_actions
-          ?.actions"
-      />
+      <hr />
+
+      <div class="accordion mb-2" id="preservation-actions">
+        <PreservationActionCollapse
+          :action="action"
+          :index="index"
+          v-model:toggleAll="toggleAll"
```
Comment (jraddaoui) [inline dashboard/src/pages/packages/[id]/index.vue]: I think this and the variable from the script could be removed.

jraddaoui (approved):
Really nice, great work @djjuhasz!

dashboard/src/styles/main.scss
djjuhasz: Adding the focus state actually changes the way the highlighting works.   The current behaviour is that the "
active" (i.e. open) section has a different look (purple header text, purple arrow icon, white background) then the closed sections (black text, black arrow icon, light gray background), which is different from the hover style above.  Because the accordion button gets focus when it's clicked, and it doesn't lose focus until you click somewhere else, it effectively becomes the "active" style for the section unless you tab or change the focused element some other way.  This change isn't bad, but I think it makes it a bit less obvious that preservation action header is an interactive element when another section is clicked.

Here's an screencast of how it the accordion works with the focus state added: 
[Screencast from 2025-01-15 10-19-54.webm](https://github.com/user-attachments/assets/53368a30-121d-4a1e-979c-a8ebf37bee1c)

jraddaoui: I think you could set in the selector focus + not active. Not a big deal in any case, I just think it's nice to also have a visual hint when doing keyboard navigation.

PR #1107, opened 2025-01-21
Make preservation task cards expandable

### LINKED ISSUE #1077: Problem: Preservation activity task list is too condensed, not very accessible as a table

djjuhasz:
Fixes #1077.

- Show only the first line of a multi-line task note by default
- Show an icon on task cards with multi-line notes to toggle display of the subsequent lines of the note
- Add css animations for the expansion and contraction of the additional note content

### DIFF
--- a/dashboard/src/components/PreservationActionCollapse.vue
+++ b/dashboard/src/components/PreservationActionCollapse.vue
@@ -1,13 +1,21 @@
 <script setup lang="ts">
-import { ref, toRefs } from "vue";
+import { computed, ref, toRefs } from "vue";
 
 import type { api } from "@/client";
 import PackageReviewAlert from "@/components/PackageReviewAlert.vue";
+import PreservationTask from "@/components/PreservationTask.vue";
 import StatusBadge from "@/components/StatusBadge.vue";
-import type { EnduroPackagePreservationTask } from "@/openapi-generator";
 import { useAuthStore } from "@/stores/auth";
 
 const authStore = useAuthStore();
+const tasks = computed<api.EnduroPackagePreservationTask[]>(() => {
+  if (!props.action.tasks) {
+    return [];
+  }
+
+  // Show the last task first.
+  return props.action.tasks.slice().reverse();
+});
 
 const props = defineProps<{
   action: api.EnduroPackagePreservationAction;
@@ -17,10 +25,6 @@ const props = defineProps<{
 const { action, index } = toRefs(props);
 
 let expandCounter = ref<number>(0);
-
-function isComplete(task: EnduroPackagePreservationTask) {
-  return task.status == "done" || task.status == "error";
-}
 </script>
 
 <template>
@@ -59,70 +63,24 @@ function isComplete(task: EnduroPackagePreservationTask) {
     <div
       v-if="action.tasks"
       :id="'pa-body-' + index"
-      class="accordion-collapse collapse bg-light"
+      class="accordion-collapse collapse"
       :aria-labelledby="'pa-heading-' + index"
       data-bs-parent="#preservation-actions"
     >
-      <div class="accordion-body d-flex flex-column gap-1">
-        <PackageReviewAlert
-          v-model:expandCounter="expandCounter"
-          v-if="authStore.checkAttributes(['package:review'])"
-        />
-        <div
-          v-for="(task, index) in action.tasks.slice().reverse()"
-          :key="action.id"
-          class="card"
+      <PackageReviewAlert
+        v-model:expandCounter="expandCounter"
+        v-if="authStore.checkAttributes(['package:review'])"
+      />
+      <ul class="accordion-body d-flex flex-column gap-1">
+        <li
+          v-for="(task, index) of tasks"
+          :id="'prestask-' + (tasks.length - index)"
+          :key="task.id"
+          class="mb-2 card bg-light"
         >
-          <div class="card-body">
-            <div class="d-flex flex-row align-start gap-3">
-              <div class="fd-flex">
-                <span
-                  class="fs-6 badge rounded-pill border border-primary text-primary"
-                >
-                  {{ action.tasks.length - index }}
-                </span>
-              </div>
-              <div
-                class="d-flex flex-column flex-grow-1 align-content-stretch min-w-0"
-              >
-                <div class="d-flex flex-wrap pt-1">
-                  <div class="me-auto text-truncate fw-bold">
-                    {{ task.name }}
-                  </div>
-                  <div class="me-3">
-                    <span
-                      v-if="
-                        !isComplete(task) &&
-                        $filters.formatDateTime(task.startedAt)
-                      "
-                    >
-                      Started:
-                      {{ $filters.formatDateTime(task.startedAt) }}
-                    </span>
-                    <span
-                      v-if="
-                        isComplete(task) &&
-                        $filters.formatDateTime(task.completedAt)
-                      "
-                    >
-                      Completed:
-                      {{ $filters.formatDateTime(task.completedAt) }}
-                    </span>
-                  </div>
-                </div>
-                <div class="d-flex flex-row gap-4">
-                  <div class="flex-grow-1 line-break">
-                    {{ task.note }}
-                  </div>
-                </div>
-              </div>
-              <div class="d-flex pt-1">
-                <StatusBadge :status="task.status" />
-              </div>
-            </div>
-          </div>
-        </div>
-      </div>
+          <PreservationTask :index="tasks.length - index" :task="task" />
+        </li>
+      </ul>
     </div>
   </div>
 </template>
--- a/dashboard/src/components/PreservationTask.vue
+++ b/dashboard/src/components/PreservationTask.vue
@@ -1,0 +1,105 @@
+<script setup lang="ts">
+import { ref } from "vue";
+
+import StatusBadge from "@/components/StatusBadge.vue";
+import { FormatDateTime } from "@/composables/dateFormat";
+import type { EnduroPackagePreservationTask } from "@/openapi-generator";
+
+class Card {
+  isOpen: boolean;
+  note: string;
+  more: string;
+
+  constructor(task: EnduroPackagePreservationTask) {
+    this.isOpen = false;
+
+    if (task.note?.includes("\n")) {
+      const [firstLine, ...remainingLines] = task.note.split("\n");
+      this.note = firstLine;
+      this.more = remainingLines.join("\n");
+    } else {
+      this.note = task.note ? task.note : "";
+      this.more = "";
+    }
+  }
+
+  toggle() {
+    if (!this.more) {
+      return;
+    }
+
+    this.isOpen = !this.isOpen;
+  }
+}
+
+const isComplete = (task: EnduroPackagePreservationTask) => {
+  return task.status == "done" || task.status == "error";
+};
+
+const props = defineProps<{
+  index: number;
+  task: EnduroPackagePreservationTask;
+}>();
+
+const card = ref(new Card(props.task));
+</script>
+
+<template>
+  <div :id="'pt-' + props.index + '-body'" class="card-body">
+    <div class="d-flex flex-row align-start gap-3">
+      <div class="fd-flex">
+        <span
+          class="fs-6 badge rounded-pill border border-primary text-primary"
+        >
+          {{ index }}
+        </span>
+      </div>
+      <div
+        class="d-flex flex-column flex-grow-1 align-content-stretch min-w-0 gap-2"
+      >
+        <div class="d-flex flex-wrap pt-1">
+          <div class="me-auto text-truncate fw-bold">
+            {{ task.name }}
+          </div>
+          <div :id="'pt-' + index + '-time'" class="me-3">
+            <span v-if="!isComplete(task) && FormatDateTime(task.startedAt)">
+              Started:
+              {{ FormatDateTime(task.startedAt) }}
+            </span>
+            <span v-if="isComplete(task) && FormatDateTime(task.completedAt)">
+              Completed:
+              {{ FormatDateTime(task.completedAt) }}
+            </span>
+          </div>
+        </div>
+        <div class="flex-grow-1">
+          <span :id="'pt-' + index + '-note'">{{ card.note }}</span>
+          <span v-if="card.more">
+            <span v-show="!card.isOpen">... </span>
+            <Transition name="fade">
+              <p
+                v-show="card.isOpen"
+                :id="'pt-' + index + '-note-more'"
+                class="line-break"
+              >
+                {{ card.more }}
+              </p>
+            </Transition>
+            <a
+              :id="'pt-' + index + '-note-toggle'"
+              :aria-controls="'pt-' + index + '-note-more'"
+              aria-label="Toggle display of additional notes"
+              @click.prevent="card.toggle"
+              href="#"
+            >
+              {{ card.isOpen ? "Show less" : "Show more" }}
+            </a>
+          </span>
+        </div>
+      </div>
+      <div class="d-flex pt-1">
+        <StatusBadge :status="task.status" />
+      </div>
+    </div>
+  </div>
+</template>
--- a/dashboard/src/components/__tests__/PreservationTask.test.ts
+++ b/dashboard/src/components/__tests__/PreservationTask.test.ts
@@ -1,0 +1,106 @@
+import { VueWrapper, mount } from "@vue/test-utils";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import PreservationTask from "../PreservationTask.vue";
+
+describe("PreservationTask.vue", () => {
+  let wrapper: VueWrapper;
+
+  beforeEach(() => {
+    wrapper = mount(PreservationTask, {
+      attachTo: document.body,
+      props: {
+        index: 1,
+        task: {
+          id: 1,
+          name: "Task 1",
+          startedAt: new Date("2020-02-25T17:21:03Z"),
+          completedAt: new Date("2020-02-25T17:22:38Z"),
+          status: "done",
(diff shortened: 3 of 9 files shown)


djjuhasz:
This commit hides the second and subsequent lines of a multi-line note (i.e. a note with multiple line break character) by default.  When the expand icon is clicked the additional lines of the note are displayed.

To test this feature you will need to have a task with a multi-line note. I manually inserted a multi-line note into the database to test locally.

Here's how it looks in action:
[Screencast from 2025-01-20 18-04-51.webm](https://github.com/user-attachments/assets/76626798-9858-4382-aba5-ffcca7dc4f37)

fiver-watson:
Hi @djjuhasz, 

Looking good! Though I do have a few feedback thoughts to share, but given that we don't actually have many tasks with long or multi-line text at the moment, it could always wait for a future iteration if we prefer. Anyway, some thoughts: 

* First, it's hard to tell from a short vid, but is the caret direction the opposite of the larger Preservation Actions one above it? Want to ensure that they are consistent. Seems to me like the caret should point down when there is something to expand (click to show more below), and up when you want to click to collapse it again
* Also, with this specific example in particular (full sentence and linebreak before the hidden text), the ONLY way you can tell there's anything to expand is the little caret icon, which is a bit hard to see. I wonder if:
   *  we want to add the same hover and click effects we added to the main Preservation Actions section, for tasks that have additional content below the fold?
   * there's some way, like a "more" link or  a set of ellipses or something, we want to indicate in the text area that there might be more to see?

Thanks! Thoughts and alternatives welcome :)

dashboard/src/components/PreservationActionCollapse.vue
```diff
@@ -15,11 +18,24 @@ const props = defineProps<{
...
+      if (card.note && card.note.includes("\n")) {
+        const lines = card.note.split("\n");
+        card.note = lines[0];
+        card.more = lines.slice(1).join("\n");
+      }
+      tasks.value.push(card);
+      i++;
+    }
+  }
+});
```
sevein: I was reviewing the [one-way data flow rules](https://vuejs.org/guide/components/props#one-way-data-flow), and I'm wondering if it's better to avoid mutating `props.action.tasks` (is it causing any console errors?). If you want the component to be reactive, maybe using a [computed property](https://vuejs.org/guide/essentials/computed) could help? I’m not completely sure so let me know what you think.

sevein: ChatGPT suggested the following but I'm not sure if it's going too far:

```ts
const processedTasks = computed(() => {
  const tasks = props.action.tasks || [];
  return tasks.slice().reverse().map((task, index) => {
    const card: TaskCard = { ...task, index: tasks.length - index };
    if (card.note?.includes("\n")) {
      const [firstLine, ...remainingLines] = card.note.split("\n");
      card.note = firstLine;
      card.more = remainingLines.join("\n");
    }
    return card;
  });
});

onMounted(() => {
  tasks.value = processedTasks.value;
});
```

djjuhasz: Oh, oops. I didn't intend to mutate `props.action.tasks`, good catch @sevein. 

It looks like `const rev = props.action.tasks.toReversed();` should return a copy of the array in reverse order. Do you see any issues with that?

djjuhasz: Huh, it's interesting that ChatGPT suggests the destructuring operator in ` const card: TaskCard = { ...task, index: tasks.length - index };`  I didn't know you could do that to cast one type to another. :sunglasses:

sevein: > Do you see any issues with that?

Looks good to me!

djjuhasz: Oh, except `array.toReversed()` was added in ES2023, and we seem to be running an older version of ECMAScript.  It looks like `const rev = tasks.slice().reverse()` will create a reversed copy of `tasks` and not mutate the original.

djjuhasz: I tried the ChatGPT solution, but `const card: TaskCard = { ...task, index: tasks.length - index };` doesn't compile because the created object is not a valid TaskCard (it's missing `isOpen` and `more` properties). :thinking:

djjuhasz:
@fiver-watson (cc @jraddaoui @mcantelon) I've implemented your feedback by making the entire preservation task card clickable to expand the card and show additional lines of the task note. 

[Screencast](https://github.com/user-attachments/assets/89a2c4d1-099c-49b1-8836-77bb361f8341).

Unfortunately making the preservation task card clickable has some problems that I'm struggling to resolve.

### Problems

I first attempted to make the preservation tasks an accordion component (like I've done with the preservation actions) with each card being an expandable accordion section. The accordion component doesn't work as well for preservation tasks though because many of the tasks won't have additional text, so the accordion section "body" will be empty.  Another issue is layout — the accordion section header and body are two different "boxes" (divs) so there will be a visible line between the first line of the note (header), and the subsequent lines (body).

My second attempt was to make each preservation task card an HTML button. Clicking the button it expands the preservation task to show the additional text. The problem is that it's awkward mixing tasks that are buttons with ones that are just HTML "div" elements (the default).  I think the main issue with mixing HTML element types is accessibility — for instance you can't tab to divs (they aren't interactive), but you can tab to buttons.  Other issues about using divs vs. buttons are described here: https://theadminbar.com/accessibility-weekly/divs-are-not-buttons/.

My third attempt was making every preservation task card a button, whether it had additional text or not. If the task has only one line of output then I disabled the button so it's not clickable.  Disabled buttons don't change color on hover so this strategy works visually, but I worry about the accessibility again. Having a bunch of disabled buttons on the page may be confusing if you are using a screen reader.  This approach does solve the tabbing problem as you can still tab to the disabled buttons (if we use the BS "disabled" class).

### Possible solutions (I can think of)

1. Go with a mix of buttons (expandable cards) and divs (not expandable) and don't worry about not being able to tab to the divs.  I think this is going to get really weird when there are 3 or 4 expandable cards interspersed among a bunch of expandable buttons.
2. Go with all buttons, and try and mitigate possible confusion with aria-labels or other hints.
3. Add an "expand" button only to expandable cards, but don't make it the whole card.  We could experiment with different sizes and layouts for the expand button.

I think using an accordion component is out because it really assumes that every section will have a header and body content unless we always put some sort of additional content into the tasks.

### Call to action

I'm curious if any of you can think of any better solutions than the three that I've come up with.

jraddaoui:
@djjuhasz It does feel rare to have those expandable sections inside expandable sections. I think the accordion border issue could be fixed overwriting that border with classes/css but I don't know how that would work in terms of accessibility, being inside of another accordion. Another option is to use `tabindex="0"` in the divs to make them "tabbables", then make them expandable or not based on the content.

In general, if you ask me, I don't think it's as important to make the whole card clickable as with the preservation actions, I think just the button would be okay and it would make sense to only allow tabbing to those with extra content/button. If you are inside the tasks you are probably looking for something, I think it may be enough to just have an obvious button where is needed, but I'd wait to hear from @fiver-watson about that.

fiver-watson:
Hi @djjuhasz, thanks for such a thoughtful review of the problem and possible solutions. I don't totally understand what the challenge is with the first proposed solution - only being able to tab to cards that can be expanded, rather than all of them, sounds efficient to me, but I think I am missing something. 

In any case, if it is going to help solve implementation issues to not make the entire card clickable, I am open to that - but in that case I think we will need a more obvious button than the current little caret, which I find hard to see. 

Additionally, we may want to wait and see when we have an actual use case, rather than just working with lorem ipsum text, as there may be other things we can do to better indicate the interactivity of the cards?

Given that we can't even include STDOUT stuff from AM anyway right now (and a lot of that is just noise anyway) and all current success messages are pretty short one-liners, right now our only use cases for expanded cards would be: 

* An error, where we want to provide more details and feedback on next steps, or
* A decision point, like whether to approve or deny a deletion request

In both cases, we could potentially style the whole card a bit differently to draw attention to it - a red border for example on an error, etc.... this might make adding a button to expand the card easier to find since we are already drawing the user's eye to it as something different than the other task cards. 

We aren't yet at the point of implementing decision points (even though we know it will come in future iterations), but we could try to limit the next expansion attempt to JUST error cards, and try a) styling them a bit differently so you can see on a scan (and when they are still collapsed) that there is an error and more information available, and b) include a button to expand them to see the details?

If that seems like a good idea, I can try to mock up some options. 

Otherwise, let's generally consider a button approach, but explore ways to make it more obvious.

jraddaoui:
Minor detail ;)

> We aren't yet at the point of implementing decision points (even though we know it will come in future iterations), but we could try to limit the next expansion attempt to JUST error cards, and try a) styling them a bit differently so you can see on a scan (and when they are still collapsed) that there is an error and more information available, and b) include a button to expand them to see the details?

We already have a decision point for the create and review AIP workflow, I think it had an alert and something else in the actions/tasks section. We just hardcoded the auto approve option so we never reach it and it has been long forgotten and untested.

fiver-watson:
Right, I wasn't counting that since that was a made up SAD use case and not an active client feature, but point taken Radda ;)

djjuhasz:
@fiver-watson thank you for your detailed and thoughtful response.  I've quoted a few sections of your response below so I can respond to the points separately.

> I don't totally understand what the challenge is with the first proposed solution - only being able to tab to cards that can be expanded, rather than all of them, sounds efficient to me, but I think I am missing something.

I am probably conflating the UI experience in a visual browser (what I'm seeing) and how a screen reader works (which I have not experienced).  I was thinking that if, for instance, task #15 and task #7 (sorted in descending order) had long notes, and thus were buttons with a tab stop, that when I tab in the browser it jumps straight from task #15 to task #7 skipping the intervening tasks. While skipping tasks is fine when you can see the intervening tasks in the browser my concern was that a screen reader would similarly skip tasks 14 - 8 with no indication that there is content between the 15 and 7.  Now that I think about it, it doesn't make much sense for a screen reader to work this way so it's probably not a valid concern.

> Additionally, we may want to wait and see when we have an actual use case, rather than just working with lorem ipsum text, as there may be other things we can do to better indicate the interactivity of the cards?

We do have a few cases of tasks that can return multi-line error messages in the SFA preprocessing workflow: 

1. Validate SIP structure (one line per missing or unexpected file or directory)
2. Verify manifest (one line per missing file, unexpected file, or mismatched checksum)
3. Validate file formats (one line per forbidden file)

It's a bit more work to trigger the failures than using "Lorem Ipsum" text but I could try to put together a SIP that would output a few multi-line failure messages if it helps with testing and designing the UI.

> we could try to limit the next expansion attempt to JUST error cards, and try a) styling them a bit differently so you can see on a scan (and when they are still collapsed) that there is an error and more information available, and b) include a button to expand them to see the details?

I like the idea of focusing on error cards and how to visually differentiate them and indicate they are expandable.  Just as food for thought I came across this Google Material design guide that has some suggestions for how to implement cards, as well as some situations where cards are *not* the best solution: https://m3.material.io/components/cards/guidelines#8a1d2b9e-2cbe-408b-ba60-c32e8c612e3e

Looking at that guide I'm questioning whether it would be better to simplify the tasks list to use separators instead of full borders. :thinking:

fiver-watson:
@djjuhasz thanks for the links, and the thoughts!

Regarding accessibility and tabbing between elements I did notice this, on the following Material Design page about cards and accessibility - it says that tabbing should move to the next _actionable_ item: 

* https://m3.material.io/components/cards/accessibility#f8d3d4b8-222b-49d2-b044-2d0126216cdb

There are also some further guidelines both above and right below that on proper element labeling to help things like screen readers understand what is important and what to skip etc. Here's a quote from the "Focus" section:

> For non-actionable cards, the card itself is not a tab stop, but every actionable element in the card is a tab stop so they’re all visited before focus navigates to the next card.

And in the "Keyboard" section: 

> Users can Tab to navigate between actionable elements of the cards. If the cards are non-actionable, Tab navigates directly to the actionable buttons or links within the cards.

**On focusing on error messages first:** 

> We do have a few cases of tasks that can return multi-line error messages in the SFA preprocessing workflow:
> 
> 1. Validate SIP structure (one line per missing or unexpected file or directory)
> 2. Verify manifest (one line per missing file, unexpected file, or mismatched checksum)
> 3. Validate file formats (one line per forbidden file)

This is really helpful, and I might generate some of these to see how they currently render and better consider the design. 

Do we currently have _any_ cases at all for successful tasks that could end up being more than one line? Or is it really only errors right now that are actually returning multi-line notes?

> Looking at that guide I'm questioning whether it would be better to simplify the tasks list to use separators instead of full borders. 🤔

Just to clarify this... are you proposing not using cards, but more like one big card or something with section dividers for each task? Just trying to understand.

I would be fine with that in principle. I think the main goals here are: 

* Tasks of a workflow / preservation action are collapsed by default, so users first see the history of workflows/actions on a package view page (done)
* It is clear that workflows / actions can be clicked to expand them and reveal more information about the individual tasks run and their outcomes (done)
* Similarly, we want users to be able to easily scan long lists of tasks at first - only showing what is most important. Users can then choose to expand some tasks for more information when necessary

I suppose you could make the argument that errors are always part of what is "most important" for ingest, so... one option is that we simply park this work, always display multi-line task notes for now, and reconsider it when it feels important or like there's too much noise and we want to simplify the task view. 

Another option is to change the implementation - maybe overflow content appears in a modal, or a separate page like Archivematica does?

Or if you're still game, we can keep at this, trying for something that is much more clearly a button, or other variations.... is there anything we can borrow from this mention of container transforms etc?

* https://m3.material.io/components/cards/guidelines#bf5047a9-e858-435b-812c-7bcb0ea59e72

I'm totally open to your input and also happy  to mock up some possible alternatives if I know which way you are leaning with this work - I don't want you to feel caught up in a bunch of finicky design stuff if it's more frustrating than satisfying, as we have plenty of other important things to focus on!

djjuhasz:
@fiver-watson and I experimented with a couple of variations on the controls to expand and collapse the preservation task cards before settling on displaying ellipses and a "Show more" link on a card that has a note longer than a single line.  

Screenshot from 2025-02-18 16-49-34

When the card is expanded the ellipses are hidden, and the link changes to "Show less".  Clicking "Show less" collapses the card.

Screenshot from 2025-02-18 16-49-53

djjuhasz:
@jraddaoui @mcantelon this is ready for another round of code review.

dashboard/src/components/PreservationTask.vue
```diff
@@ -0,0 +1,111 @@
...
+                {{ card.more }}
+              </p>
+            </Transition>
+            <a
+              :aria-controls="'pt-note-' + index + '-more'"
+              aria-label="Toggle display of additional notes"
+              @click.prevent="card.toggle"
+              href="#"
+            >
+              Show {{ card.isOpen ? "less" : "more" }}
```
jraddaoui: Random thought/nitpick: thinking about translations in the future, I'd consider repeating "Show ".

dashboard/src/components/PreservationTask.vue
```diff
@@ -0,0 +1,111 @@
...
+              Completed:
+              {{ $filters.formatDateTime(task.completedAt) }}
+            </span>
+          </div>
+        </div>
+        <div class="flex-grow-1">
+          <span :id="'pt-note-' + index">{{ card.note }}</span>
+          <span v-if="card.more">
+            <span v-show="!card.isOpen">... </span>
+            <Transition name="fade">
```
jraddaoui, line 79: Didn't know about this, nice!

dashboard/src/components/PreservationTask.vue
```diff
@@ -0,0 +1,111 @@
...
+            </a>
+          </span>
+        </div>
+      </div>
+      <div class="d-flex pt-1">
+        <StatusBadge :status="task.status" />
+      </div>
+    </div>
+  </div>
+</template>
```
jraddaoui, line 105: It would be nice to have some tests for this component.

jraddaoui (approved):
LGTM! Thanks @djjuhasz and @fiver-watson!

dashboard/src/components/PreservationTask.vue
djjuhasz: And I was so proud of my economy in saving five characters! :stuck_out_tongue_winking_eye: Good point about the possible future translation of the string.

dashboard/src/components/PreservationTask.vue
djjuhasz: Ok :sweat:

djjuhasz: Done. :grin:

PR #1108, opened 2025-01-28
Search for packages by name, fixes #1101

### LINKED ISSUE #1101: Feature: search for package by name

If I know a package's name, either in part or in full, I want to be able to find a specific package on the Packages page.

I would like there to be a text box where I can enter a partial or complete package name, hit a search button, and see a list of packages that match my search term.

**Additional context**

djjuhasz:
Add the ability to search for packages by name, including partial matches.  This implementation requires the user to click the search button (looks like a magnifying glass) or press the "Enter" key to perform the search.  It also includes a "reset" button that clears the search field.

Here's how it looks in action:
https://github.com/user-attachments/assets/b3fe047a-c768-42b7-9ae9-74e60ce2e8cc

### DIFF
--- a/dashboard/src/components/Tabs.vue
+++ b/dashboard/src/components/Tabs.vue
@@ -2,7 +2,6 @@
 import type { FunctionalComponent, SVGAttributes } from "vue";
 import type { RouteLocationResolved } from "vue-router/auto";
 import { useRoute } from "vue-router/auto";
-import { isEqual } from "lodash-es";
 
 const route = useRoute();
 
@@ -13,12 +12,15 @@ type Tab = {
   show: boolean;
 };
 
-const { tabs } = defineProps<{
+const { tabs, param } = defineProps<{
   tabs: Tab[];
+  param: string;
 }>();
 
 function isActive(tab: Tab): boolean {
-  return tab.route.path == route.path && isEqual(tab.route.query, route.query);
+  return (
+    tab.route.path == route.path && tab.route.query[param] == route.query[param]
+  );
 }
 </script>
 
--- a/dashboard/src/pages/packages/index.vue
+++ b/dashboard/src/pages/packages/index.vue
@@ -12,12 +12,14 @@ import { useAsyncState } from "@vueuse/core";
 import { useAuthStore } from "@/stores/auth";
 import { useLayoutStore } from "@/stores/layout";
 import { usePackageStore } from "@/stores/package";
-import { useRoute, useRouter } from "vue-router/auto";
-import { ref, watch } from "vue";
+import { useRoute, useRouter, type LocationQueryValue } from "vue-router/auto";
+import { computed, ref, watch } from "vue";
 
 // General icons.
 import IconInfoFill from "~icons/akar-icons/info-fill";
 import IconBundleLine from "~icons/clarity/bundle-line";
+import IconSearch from "~icons/clarity/search-line";
+import IconCloseLine from "~icons/clarity/close-line";
 
 // Pager icons.
 import IconSkipEndFill from "~icons/bi/skip-end-fill";
@@ -36,7 +38,7 @@ const authStore = useAuthStore();
 const layoutStore = useLayoutStore();
 const packageStore = usePackageStore();
 
-const route = useRoute("/packages/");
+const route = useRoute();
 const router = useRouter();
 
 layoutStore.updateBreadcrumb([{ text: "Packages" }]);
@@ -50,17 +52,24 @@ const toggleLegend = () => {
   if (tooltip) tooltip.hide();
 };
 
+const doSearch = () => {
+  router.push({
+    name: "/packages/",
+    query: { ...route.query, name: packageStore.filters.name },
+  });
+};
+
 onMounted(() => {
   if (el.value) tooltip = new Tooltip(el.value);
-  packageStore.filters.status = <PackageListStatusEnum>route.query.status;
 });
 
-const tabs = [
+const tabs = computed(() => [
   {
     icon: RawIconBlocksGroupLine,
     text: "All",
     route: router.resolve({
       name: "/packages/",
+      query: { ...route.query, status: undefined },
     }),
     show: true,
   },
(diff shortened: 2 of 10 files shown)


dashboard/src/pages/packages/index.vue
```diff
@@ -128,9 +151,37 @@ watch(
...
+            doSearch();
+          "
+          type="reset"
+          aria-label="Reset search"
+        >
+          <IconCloseLine />
+        </button>
+        <button class="btn btn-primary" type="submit" aria-label="Do search">
+          <IconSearch />
+        </button>
```
jraddaoui, line 178: What's the point of a submit button if we watch the input?

dashboard/src/pages/packages/index.vue
```diff
@@ -128,9 +151,37 @@ watch(
...
+      </div>
+    </form>
+
+    <Tabs :tabs="tabs" pname="status" />
```
jraddaoui: I thought about changing the tabs/legend order recently, thanks!

internal/persistence/ent/client/filter.go
```diff
@@ -201,6 +201,17 @@ func validPtrValue(ptr any) bool {
 	}
 }
 
+// Contains adds a filter on column containing the given string.
+func (f *Filter[Q, O, P]) Contains(column string, value *string) {
+	if value == nil {
```
jraddaoui: I think we should prevent the filter if it's an empty string too.

dashboard/src/pages/packages/index.vue
```diff
@@ -113,6 +128,14 @@ watch(
     return packageStore.fetchPackages(1);
   },
 );
+
+watch(
+  () => route.query.name,
+  (newSearch) => {
+    packageStore.filters.name = newSearch as string;
+    return packageStore.fetchPackages(1);
```
jraddaoui, line 136: When watching an input and making API requests based on its value, we should consider a few things:

- Debouncing/Throttling: making sure the request doesn't happen each keystroke.
- Request Cancellation: abort previous requests and prevent race conditions.
- Deduplication: about fetching the same query, caching results.

I'm not sure what would be the best way to implement these using our client and Vue but, ideally, we would create something that is reusable in other search inputs. **Doesn't need to happen in this PR.**

Also, it would be nice to provide some "Loading..." visual effect and we should consider pagination too, returning to the first page if the input changes.

dashboard/src/components/Tabs.vue
```diff
@@ -13,12 +12,15 @@ type Tab = {
   show: boolean;
 };
 
-const { tabs } = defineProps<{
+const { tabs, pname } = defineProps<{
   tabs: Tab[];
+  pname: string;
```
jraddaoui: It was hard for me to understand this refers to the parameter name. Not sure if `param` would be clearer, but I wouldn't mind being more specific in this case.

jraddaoui (approved):
Looking good @djjuhasz, thanks!

Some minor things and an important one, but that can happen in another PR.

dashboard/src/pages/packages/index.vue
djjuhasz: I'm not sending the API request to filter the list until the submit button is clicked (or the Enter key is pressed). I originally was sending an API GET request when the search input value changed (i.e. I was watching `packageStore.filters.name` for changes), but now it watching for changes of the `request.query.name` value to trigger sending the request.  I was worried that sending a GET request on each change of the input would lead to performance issues with large datasets, and it was also causing some problems where the full input value was not requested (e.g. the API request might send "sma" when the user typed "small").

internal/persistence/ent/client/filter.go
djjuhasz: Good idea

djjuhasz: Good idea.

djjuhasz: Though that does raise the point of why bother using a string pointer instead of just a string. :thinking:

dashboard/src/pages/packages/index.vue
djjuhasz: Because the API request is fired by submitting the form (well, in a Vue.js kind of SPA version of form submission) I think these issues are probably not as much of a concern.  I was originally using Debounced API requests, but it didn't seem necessary without the "onChange" requests.

`packageStore.fetchPackages(1);` does fetch the first page of the results (that's the 1 parameter).  The pager doesn't remember it's position when you click on a package on the list right now, but that's a problem to address in a different PR I think. :)

The loading is so fast right (the API response time was in the < 10 millisecond range when I tested) now I don't think a "Loading..." effect is necessary, but I think we could add it later if necessary.

jraddaoui: Oh, I see, this watches the route! Why do we need it then? If it's only changed from `doSearch()` it should be enough to fetch there.

Cool about pagination, didn't notice that 1. Loading speeds may change a lot in other environments not running everything locally, but yeah, we can add it later.

dashboard/src/pages/packages/index.vue
jraddaoui: Yeah, that's what I thought, and why I mentioned debouncing/throttling in the other thread.

dashboard/src/pages/packages/index.vue
djjuhasz: There is lots I don't know about how Vue.js does routing, watchers and page flow so the way I've implemented this may not be ideal and I'm always open to other suggestions.  What I do know if I take the route watcher away, when I click the "back" button on the browser the URL in the browser bar updates appropriately, but the packages list doesn't update to reflect the previous search value. :man_shrugging:

jraddaoui: I see that, I would expect that to trigger the `useAsyncState` again but it doesn't seem to do it. Leave it there then, thanks for noticing such details ;)

djjuhasz: Yeah, I would expect the the back button to reload the whole page + app and trigger `useAsyncState` too, but maybe Vue overrides the browser behaviour here to avoid reloading the whole app?

PR #1109, opened 2025-01-28
WIP PoC: Add Grafana Loki and collect logs from k8s pods

jraddaoui:


jraddaoui:
Similar and updated approach in https://github.com/artefactual-sdps/enduro/pull/1684.

PR #1110, opened 2025-01-29
Optionally send unzipped bags to Archivematica

mcantelon:
Added support for sending transfers to Archivematica as unzipped bags.

### DIFF
--- a/enduro.toml
+++ b/enduro.toml
@@ -179,6 +179,10 @@ transferDeadline = "1h"
 # used.
 transferSourcePath = ""
 
+# zipPIP specifies whether or not a PIP should be zipped before being sent from
+# Enduro to Archivematica.
+zipPIP = true
+
 [am.sftp]
 host = "" # The Archivematica Storage Service hostname.
 port = ""
--- a/internal/am/config.go
+++ b/internal/am/config.go
@@ -30,6 +30,10 @@ type Config struct {
 	// "749ef452-fbed-4d50-9072-5f98bc01e52e:sftp_upload").
 	TransferSourcePath string
 
+	// ZipPIP specifies whether or not a PIP should be zipped before being sent
+	// from Enduro to Archivematica.
+	ZipPIP bool
+
 	// Capacity sets the maximum number of worker sessions the worker can
 	// handle at one time (default: 1).
 	Capacity int
--- a/internal/am/start_transfer.go
+++ b/internal/am/start_transfer.go
@@ -19,6 +19,8 @@ type StartTransferActivityParams struct {
 	// Name of the transfer.
 	Name string
 
+	ZipPIP bool
+
 	// RelativePath is the PIP path relative to the Archivematica transfer
 	// source directory.
 	RelativePath string
@@ -55,9 +57,14 @@ func (a *StartTransferActivity) Execute(
 		processingConfig = "automated" // Default value.
 	}
 
+	transferType := "unzipped bag"
+	if opts.ZipPIP {
+		transferType = "zipped bag"
+	}
+
 	payload, resp, err := a.amps.Create(ctx, &amclient.PackageCreateRequest{
 		Name:             opts.Name,
-		Type:             "zipped bag",
+		Type:             transferType,
 		Path:             filepath.Join(a.cfg.TransferSourcePath, opts.RelativePath),
 		ProcessingConfig: processingConfig,
 		AutoApprove:      true,
--- a/internal/am/upload_transfer.go
+++ b/internal/am/upload_transfer.go
@@ -3,6 +3,7 @@ package am
 import (
 	"context"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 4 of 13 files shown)


enduro.toml
```diff
@@ -179,6 +179,9 @@ transferDeadline = "1h"
 # used.
 transferSourcePath = ""
 
+# transferType specifies the transfer type sent from Enduro to Archivematica.
+transferType = "zipped bag"
```
djjuhasz: I would prefer not to use the AM transfer type value as the option flag to enable or disable zipping the PIP that is sent to preservation.

Archivematica accepts six types of transfers (standard, zipfile, unzipped bag, zipped bag, dspace, maildir, TRIM, dataverse), but we've agreed that Enduro should send Bagit Bags to the preservation system for several reasons (see the discussion on #805).  This means the only two transfer types we plan to send to AM are "zipped bag" and "unzipped bag". I think because of this it would be better to use a simple boolean flag to configure whether or not the PIP should be zipped before delivery to the preservation system (e.g. "zipPIP = true / false").  

Using a boolean avoids the question of what values are valid here, and also avoids explicitly tying the functionality to the exact strings used by Archivematica's API transfer type values.  A boolean also simplifies validation as we don't need to check against a list of accepted strings, and deal with capitalization, spacing, etc.

internal/am/start_transfer.go
```diff
@@ -57,7 +60,7 @@ func (a *StartTransferActivity) Execute(
 
 	payload, resp, err := a.amps.Create(ctx, &amclient.PackageCreateRequest{
 		Name:             opts.Name,
-		Type:             "zipped bag",
+		Type:             opts.Type,
```
djjuhasz: If we switch the config for this to a boolean then this string will obviously need to be set in the code to "zipped bag" or "unzipped bag".

internal/am/upload_transfer.go
```diff
@@ -54,33 +54,74 @@ func (a *UploadTransferActivity) Execute(
...
+			switch err.(type) {
+			case *sftp.AuthError:
+				return nil, temporal.NewNonRetryableError(e)
+			default:
+				return nil, e
+			}
+		}
+
+		// Determine total size of files in directory.
+		err := filepath.Walk(params.SourcePath, func(_ string, info os.FileInfo, err error) error {
```
djjuhasz: https://pkg.go.dev/path/filepath#WalkDir is preferred to `filepath.Walk` because it's more efficient.

internal/am/upload_transfer.go
```diff
@@ -54,33 +54,74 @@ func (a *UploadTransferActivity) Execute(
...
+			if err != nil {
+				return err
+			}
 
-		switch err.(type) {
-		case *sftp.AuthError:
-			return nil, temporal.NewNonRetryableError(e)
-		default:
-			return nil, e
+			if !info.IsDir() {
```
djjuhasz: It's a bit confusing to shadow the `info` variable here.  If you switch to `filepath.WalkDir` the default variable names will make this `if !d.IsDir() {` which avoids the problem.

internal/am/upload_transfer.go
```diff
@@ -54,33 +54,74 @@ func (a *UploadTransferActivity) Execute(
...
+		if err != nil {
+			return nil, fmt.Errorf("%s: %v", UploadTransferActivityName, err)
+		}
+		defer src.Close()
+
+		path, upload, err = a.client.UploadFile(ctx, src, filename)
+		if err != nil {
+			e := fmt.Errorf("%s: %v", UploadTransferActivityName, err)
+
+			switch err.(type) {
```
djjuhasz: This code is a duplicated in the directory code branch — it would be nice to refactor this (maybe into a function) to avoid the repetition.

internal/sftp/client.go
```diff
@@ -32,7 +32,8 @@ type Client interface {
 	Delete(ctx context.Context, dest string) error
 	// Upload asynchronously copies data from the src reader to the specified
 	// dest on the SFTP server.
-	Upload(ctx context.Context, src io.Reader, dest string) (remotePath string, upload AsyncUpload, err error)
+	UploadFile(ctx context.Context, src io.Reader, dest string) (remotePath string, upload AsyncUpload, err error)
+	UploadDirectory(ctx context.Context, srcPath string) (remotePath string, upload AsyncUpload, err error)
```
djjuhasz, line 36: I would love to have a single generalized "Upload()" method that could upload a file or directory, but that would require some significant changes and it can wait for a future PR.  One day I'd love to have a second crack at re-writing this whole package now that I know Go better. :laughing:

internal/sftp/goclient.go
```diff
@@ -76,11 +77,78 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (stri
...
 	upload := NewAsyncUpload(conn)
-	go remoteCopy(ctx, &upload, src, remotePath)
+	go remoteCopy(ctx, &upload, src, remotePath, true)
 
 	return remotePath, &upload, nil
 }
 
+// UploadDirectory asynchronously copies a directory to dest over an SFTP connection.
+//
+// When UploaDirectory is called it starts the upload in an asynchronous goroutine,
```
djjuhasz: Typo: "UploaDirectory"

internal/sftp/goclient.go
```diff
@@ -76,11 +77,78 @@ func (c *GoClient) Upload(ctx context.Context, src io.Reader, dest string) (stri
...
+	}
+
+	// Asynchronously upload directory of files.
+	upload := NewAsyncUpload(conn)
+	go uploadDirectory(ctx, conn, srcPath, c.cfg.RemoteDir, &upload)
+
+	return sftp.Join(c.cfg.RemoteDir, transferDir), &upload, nil
+}
+
+func uploadDirectory(ctx context.Context, conn *connection, srcPath, remoteDir string, upload *AsyncUploadImpl) {
```
djjuhasz: `upload` wraps `conn`, so you don't need  to pass `conn` separately here. E.g you can call `upload.conn.Client.Mkdir()`.

internal/sftp/goclient.go
```diff
@@ -107,8 +175,10 @@ func (c *GoClient) dial(ctx context.Context) (*connection, error) {
 // remoteCopy copies data from the src reader to a remote file at dest, and
 // updates upload progress asynchronously. Upload status and progress will be
 // sent to the upload struct via the `upload.Done()` and `upload.Error()` channels.
-func remoteCopy(ctx context.Context, upload *AsyncUploadImpl, src io.Reader, dest string) {
-	defer upload.Close()
+func remoteCopy(ctx context.Context, upload *AsyncUploadImpl, src io.Reader, dest string, singleCopy bool) {
+	if singleCopy {
```
djjuhasz: Rather than passing `singleCopy` and conditionally closing the connection and signaling upload completion I think it would be better to add an `uploadFile()` function (similar to `uploadDirectory()`) that handles those details.

internal/workflow/processing.go
```diff
@@ -908,23 +908,30 @@ func (w *ProcessingWorkflow) transferAM(ctx temporalsdk_workflow.Context, tinfo
...
+		).Get(activityOpts, &zipResult)
+		if err != nil {
+			return err
+		}
 
-	tinfo.SendToFailed.Path = zipResult.Path
-	tinfo.SendToFailed.ActivityName = activities.SendToFailedPIPsName
+		tinfo.SendToFailed.Path = zipResult.Path
+		tinfo.SendToFailed.ActivityName = activities.SendToFailedPIPsName
+		sourcePath = zipResult.Path
```
djjuhasz: Huh, this is totally not because of your changes but I think we should be registering the zipfile path for cleanup here with `cleanup.registerPath()`. :open_mouth: You'll need to pass the cleanup struct into `transferAM()` like we are doing for `transferA3M()`.

djjuhasz: I think you can just reassign `tinfo.TempPath = zipResult.Path` here instead of using `sourcePath`.

djjuhasz (changes requested):
This is looking good @mcantelon, I just had a few change requests.

enduro.toml
mcantelon: Good call!

internal/sftp/client.go
mcantelon: Yeah, that would be a bit cleaner!

mcantelon:
@djjuhasz Thanks for the review!I think I've addressed all the essential things!

enduro.toml
```diff
@@ -179,6 +179,9 @@ transferDeadline = "1h"
 # used.
 transferSourcePath = ""
 
+# transferType specifies the transfer type sent from Enduro to Archivematica.
```
djjuhasz: This comment needs to be updated.

internal/am/config.go
```diff
@@ -30,6 +30,8 @@ type Config struct {
 	// "749ef452-fbed-4d50-9072-5f98bc01e52e:sftp_upload").
 	TransferSourcePath string
 
+	ZipPIP bool
```
djjuhasz, line 35: A comment to describe ZipPIP here would be helpful too.

internal/am/upload_transfer.go
```diff
@@ -113,3 +145,14 @@ func (a *UploadTransferActivity) Heartbeat(ctx context.Context, upload sftp.Asyn
 		}
 	}
 }
+
+func formatUploadError(err error) error {
```
djjuhasz: I think the name "uploadError" (or something similar) would be better because there is a functional difference between returning a `temporal.NewNonRetryableError(e)` and a standard (retryable) error.

djjuhasz (changes requested):
Looking good @mcantelon! I just had a few more minor change requests.

mcantelon:
Thanks @djjuhasz ! I've (hopefully) addressed the feedback. Let me know if anything needs to be tweaked!

djjuhasz (approved):
LGTM @mcantelon! :+1:

mcantelon:
Woot! Thanks @djjuhasz !

Issue #1117, opened 2025-02-10
Problem: it's hard to differentiate between Ingest and Storage concerns


jraddaoui:
After the analysis on issues #1076, #1081 and other discussions outside Github, we realized there is no clear separation between the ingest and storage concerns in Enduro. We tried to provide a unified experience in the user interface but, while doing so, we mixed actions from storage in the ingest application. A clear example is the "move package" operation, check [this comment](https://github.com/artefactual-sdps/enduro/issues/1081#issuecomment-2515571262) to know more.

This issue will cover the solutions we decided to implement as a first step to help us address #1076 and #1081 with a better separation of concerns.

**Dashboard**

- Add Ingest and Storage headings in sidebar.
- Rename Packages to SIPs on Ingest (sidebar, URLs, breadcrumbs, pages).
- Include Locations and AIPs on Storage.
- Rename Packages to AIPs on Storage (sidebar, URLs, breadcrumbs, pages).
- Add Ingest/Storage to breadcrumbs and URLs.
- Have different pages for SIPs and AIPs (browse and index).
  - Show ingest details and link to AIP on SIP page.
  - Show storage details on AIP page, no link back to SIP.
- Wire move operation to Storage endpoints below and only show on AIP page.
- Use new API endpoints detailed below.
- Use UUIDs in URLs.

**API**

Current:

| Method | Endpoint                           | Attributes                      |
| ------ | ---------------------------------- | ------------------------------- |
| GET    | /package                           | `package:list`                  |
| GET    | /package/{id}                      | `package:read`                  |
| POST   | /package/{id}/confirm              | `package:review`                |
| GET    | /package/{id}/move                 | `package:move`                  |
| POST   | /package/{id}/move                 | `package:move`                  |
| GET    | /package/{id}/preservation-actions | `package:listActions`           |
| POST   | /package/{id}/reject               | `package:review`                |
| POST   | /package/upload                    | `package:upload`                |
| GET    | /storage/location                  | `storage:location:list`         |
| POST   | /storage/location                  | `storage:location:create`       |
| GET    | /storage/location/{uuid}           | `storage:location:read`         |
| GET    | /storage/location/{uuid}/packages  | `storage:location:listPackages` |
| POST   | /storage/package                   | `storage:package:create`        |
| GET    | /storage/package/{aip_id}          | `storage:package:read`          |
| GET    | /storage/package/{aip_id}/download | `storage:package:download`      |
[...]

PR #1119, opened 2025-02-10
API redesign for ingest and storage

jraddaoui:
API redesign:

- Rename "package" to "ingest" service.
- Add "sips" prefix to ingest service methods.
- Rename "package" to "aips" in the storage service.
- Use plurals for collection URLs and attributes (sips, aips, locations).
- Rename scopes/attributes to match.
- Rename enums.
- Add name to types and prefix some with SIP.
- Use "uuid" instead of "aip_id" in AIP URLs.

Other changes:

- Generate API Goa code.
- Adapt Go code to work with new Goa generated code.
- Remove references to "package", use SIP or AIP instead.
- Rename "package_" package to ingest.
- Rename .go files with package in their name.
- Generate dashboard API client.
- Update dashboard to work with new API client.
- Update IAC docs API endpoints/attributes.

Refs #1117.


jraddaoui:
@djjuhasz, this is ready for review, sorry I couldn't make this smaller while making everything work. Most of the changes are generated code and re-naming so I've tried to split it in multiple commits. I think it will be easier for you to go commit  by commit for this review, and you could ignore some of them:

- Update API design.
- [IGNORE] Generate API Goa code.
- Adapt Go code to work with new Goa generated code.
- [IGNORE] Rename "package_" package to ingest.
- [IGNORE] Rename .go files with package in their name.
- [IGNORE] Generate dashboard API client.
- Update dashboard to work with new API client.
- Update IAC docs API endpoints/attributes.

The dashboard changes are only the necessary ones to make it work with the new API design. There will be another PR making the re-design there where I'll replace "package" references, rename the store, etc. I think that PR will be the latest of the big ones, then I hope we can continue with smaller chunks.

internal/api/design/package_.go
```diff
@@ -76,16 +76,16 @@ var _ = Service("package", func() {
...
 			Attribute("limit", Int, "Limit number of results to return")
 			Attribute("offset", Int, "Offset from the beginning of the found set")
 
 			Token("token", String)
 		})
-		Result(PackageList)
+		Result(SIPs)
 		HTTP(func() {
-			GET("/")
+			GET("/sip")
```
djjuhasz: I think using plural nouns for collection identifiers is considered best practice, as suggested in these articles:

- https://swagger.io/resources/articles/best-practices-in-api-design/
- https://learn.microsoft.com/en-us/azure/architecture/best-practices/api-design#organize-the-api-design-around-resources
- https://google.aip.dev/122#collection-identifiers

internal/api/design/package_.go
```diff
@@ -99,133 +99,133 @@ var _ = Service("package", func() {
...
 			Response("not_available", StatusConflict)
 		})
 	})
-	Method("preservation_actions", func() {
-		Description("List all preservation actions by ID")
+	Method("list_sip_preservation_actions", func() {
+		Description("List all preservation actions for a SIP")
 		Security(JWTAuth, func() {
-			Scope("package:listActions")
+			Scope("ingest:sip:listActions")
```
djjuhasz: Maybe "ingest:sip:action:list" would be better for future extension?

internal/api/design/package_.go
```diff
@@ -265,47 +265,33 @@ var _ = Service("package", func() {
...
-		Attribute("location_id")
-		Attribute("status")
-		Attribute("workflow_id")
-		Attribute("run_id")
-		Attribute("aip_id")
-		Attribute("created_at")
-		Attribute("started_at")
-		Attribute("completed_at")
+		Attribute("id", UInt, "Identifier of SIP")
+		Attribute("name", String, "Name of the SIP")
```
djjuhasz, line 277: Nice job adding descriptions to all these attributes. :clap:

djjuhasz (changes requested):
Looks good overall @jraddaoui.  I think we should be using plural nouns for collection ids though.

djjuhasz:
> Looks good overall @jraddaoui. I think we should be using plural nouns for collection ids though.

Oops, I've only reviewed the first commit so far.  I didn't mean for this to apply to the whole PR. :blush:

internal/api/design/package_.go
```diff
@@ -249,7 +249,7 @@ var _ = Service("package", func() {
 		Error("internal_error", ErrorResult, "Fault while processing upload.")
 
 		HTTP(func() {
-			POST("/upload")
+			POST("/sip/upload")
```
djjuhasz: I'm not clear on the intended semantics here.  Generally I think the `/sips` endpoint represents data *about* a SIP, rather then the actual SIP.  If the intention is for a single endpoint to create a SIP data record AND upload the SIP then I think `POST /sips` is sufficient.  If the SIP data is created separately from the upload of the actual SIP then I think  `POST /sips/{id}/upload` is more appropriate and follows the API convention of alternating collection nouns with individual resource identifiers.

djjuhasz:
Okay, I've finished reviewing the whole PR.  Looking good @jraddaoui, I just requested those few changes.

internal/api/design/package_.go
jraddaoui: A good moment to change that, thanks!

internal/api/design/package_.go
jraddaoui: I just moved them from the type to the result type ;)

internal/api/design/package_.go
jraddaoui: I've also used them in the ABAC scopes.

internal/api/design/package_.go
jraddaoui: I have my conflicts about this. It's not meant to update an existing SIP, so `POST /sips/{id}/upload` wouldn't work. In fact it doesn't even create a SIP, it just uploads the file to MinIO, and that triggers a workflow that creates the SIP eventually. I think `POST /sips` makes sense to create a SIP (like it is for AIPs and locations) but, without having such an endpoint ATM, it's hard to argue against it, other than reserving it for that reason. Nevertheless, I don't see it wrong like it is, if you have more than one POST operation over the entire collection, and it also makes it clear the purpose of the endpoint.

jraddaoui: Let me know if you feel strong about using `POST /sips`, it should be quick to change. Otherwise, I think it's okay as it is.

jraddaoui:
Thanks @djjuhasz!

djjuhasz (approved):
LGTM! :+1:

PR #1120, opened 2025-02-10
Reorganize dashboard for ingest and storage

jraddaoui:
- Add ingest and storage headings in sidebar.
- Change Packages for SIPs in sidebar ingest section.
- Add AIPs alongside locations in sidebar storage section.
- Modify icons for SIPs and locations.
- Normalize icon usage.
- Consider intermediate text elements in breadcrumb component.
- Add "Ingest" and "Storage" text items to breadcrumbs.
- Update breadcrumb in not found page.
- Add "/ingest" and "/storage" prefixes to routes.
- Rename "/packages" to "/sips".
- Rename "/locations/[id]/packages" to "/locations/[id]/aips".
- Update pages directory structure to match.
- Rename package store to ingest.
- Remove package details card.
- Move download AIP link to package location card.
- Rename PackageLocationCard to AipLocationCard.
- Rename PackageListLegend to SipListLegend.
- Rename PackageListSelectFilters to SipListSelectFilters.
- Rename PackageLocationDialog to LocationDialog.
- Rename PackagePendingAlert to SipPendingAlert.
- Rename PackageReviewAlert to SipReviewAlert.
- Replace other references to package by SIP or AIP.
- Fix ingest websocket proxies.

Refs #1117.


jraddaoui:
@djjuhasz Latest big PR for the re-org. I think it makes more sense to review this PR commit by commit too. A few notes about changes that are not included in it that will come in following ones:

- AIP pages are still missing, AIPs link for the sidebar leads to a not found page.
- SIP location card and location dialog are still a bit confusing, should be better after we move that to the AIP page.
- Location info is still shown in SIPs table, will be removed when we do that in all layers.
- SIP URLs are still using DB ids, another thing that will require changes in all layers.
- Move operation is still the same, it will change when that's moved to only the storage/AIP domain.

dashboard/src/components/Sidebar.vue
```diff
@@ -77,41 +94,50 @@ const closeOffcanvas = () => {
...
-                  <div
-                    class="col-9 d-flex align-items-center"
-                    :class="
-                      layoutStore.sidebarCollapsed
-                        ? 'col-md-12 justify-content-md-center pt-md-2'
-                        : ''
-                    "
-                  >
-                    {{ item.text }}
+          <template v-for="item in menuItems">
```
djjuhasz: What does a nested `<template>` element do inside a Vue.js template?

dashboard/src/components/Breadcrumb.vue
```diff
@@ -9,10 +9,12 @@ const layoutStore = useLayoutStore();
...
       <li
-        v-for="item in layoutStore.breadcrumb"
+        v-for="(item, i) in layoutStore.breadcrumb"
         class="breadcrumb-item"
-        :class="!item.route ? 'active' : ''"
-        :aria-current="!item.route ? 'page' : undefined"
+        :class="i == layoutStore.breadcrumb.length - 1 ? 'active' : ''"
+        :aria-current="
+          i == layoutStore.breadcrumb.length - 1 ? 'page' : undefined
+        "
```
djjuhasz, line 17: I'm seeing a eslint error here: `Elements in iteration expect to have 'v-bind:key' directives.`

dashboard/src/components/PackageLocationCard.vue
```diff
@@ -30,31 +30,31 @@ const choose = async () => {
...
       </div>
-      <div v-if="packageStore.isMoving" class="alert alert-info" role="alert">
+      <div v-if="ingestStore.isMoving" class="alert alert-info" role="alert">
         The package is being moved into a new location.
       </div>
       <h4 class="card-title">Location</h4>
       <p class="card-text">
-        <span v-if="packageStore.isRejected">Package rejected.</span>
-        <span v-else-if="!packageStore.current?.locationId"
+        <span v-if="ingestStore.isRejected">Package rejected.</span>
```
djjuhasz: All of these methods (`isMoving()`, `isRejected()`, `isMovable()`) sound odd in the context of the "ingestStore" because "ingest" isn't a noun.  I think using "sipStore" would make a little more sense in a lot of these cases (e.g. `sipStore.isRejected()`) and maybe we could avoid the changing `packageStore.current` to `ingestStore.currrentSip` and similar.  Of course `sip.isRejected()` would read even better...

dashboard/src/stores/__tests__/package.test.ts
```diff
@@ -2,144 +2,144 @@ import { createPinia, setActivePinia } from "pinia";
 import { beforeEach, describe, expect, it } from "vitest";
```
djjuhasz, line 2: I think this file should be renamed "ingest.test.ts"?  Maybe you do that in a subsequent commit?

dashboard/src/stores/ingest.ts
```diff
@@ -21,21 +21,21 @@ export interface Pager {
   pages: Array<number>;
 }
 
-export const usePackageStore = defineStore("package", {
+export const useIngestStore = defineStore("ingest", {
   state: () => ({
-    // Package currently displayed.
-    current: null as api.EnduroIngestSip | null,
+    // SIP currently displayed.
+    currentSip: null as api.EnduroIngestSip | null,
```
djjuhasz, line 27: Are you using mixed case "Sip" here because the API uses mixed case?  I'd prefer `currentSIP` like we would do in Go.

dashboard/src/components/PackageDetailsCard.vue
```diff
@@ -34,13 +34,11 @@ watch(packageStore.ui.download, () => download());
...
-            v-if="packageStore.current_preservation_actions?.actions"
-            :status="
-              packageStore.current_preservation_actions?.actions[0].status
-            "
+            v-if="ingestStore.currentPreservationActions?.actions"
+            :status="ingestStore.currentPreservationActions?.actions[0].status"
             :note="
               $filters.getPreservationActionLabel(
-                packageStore.current_preservation_actions?.actions[0].type,
+                ingestStore.currentPreservationActions?.actions[0].type,
```
djjuhasz: I like the change from snake case to camel case. :+1:

dashboard/src/components/PackageLocationCard.vue
```diff
@@ -22,6 +23,14 @@ const choose = async () => {
...
   }
 };
+
+const download = () => {
+  if (!ingestStore.currentSip?.aipId) return;
+  const url = storageServiceDownloadURL(ingestStore.currentSip.aipId);
+  window.open(url, "_blank");
+};
+
+watch(ingestStore.ui.download, () => download());
```
djjuhasz: Why do you need a watcher on `ingestStore.ui.download`?

dashboard/src/components/__tests__/PackageLocationCard.test.ts
```diff
@@ -39,7 +40,9 @@ describe("PackageLocationCard.vue", () => {
               <!-- Copy icon. --><span><svg viewBox="0 0 24 24" width="1.2em" height="1.2em" aria-hidden="true"><g fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"><path d="M8 4v12a2 2 0 0 0 2 2h8a2 2 0 0 0 2-2V7.242a2 2 0 0 0-.602-1.43L16.083 2.57A2 2 0 0 0 14.685 2H10a2 2 0 0 0-2 2"></path><path d="M16 18v2a2 2 0 0 1-2 2H6a2 2 0 0 1-2-2V9a2 2 0 0 1 2-2h2"></path></g></svg><span class="visually-hidden">Copy to clipboard</span></span>
             </button>
         </div></span></p>
-        <div class="actions"><button type="button" class="btn btn-primary btn-sm">Choose storage location</button></div>
+        <div class="d-flex flex-wrap gap-2"><button type="button" class="btn btn-primary btn-sm">Choose storage location</button>
```
djjuhasz, line 43: Testing that the exact HTML on the page seems very brittle to me, and not necessary to test the app functionality. :grimacing:

dashboard/src/components/__tests__/AipLocationCard.test.ts
```diff
@@ -4,14 +4,14 @@ import { afterEach, describe, expect, it, vi } from "vitest";
 import { nextTick } from "vue";
 
 import { api } from "@/client";
-import PackageLocationCard from "@/components/PackageLocationCard.vue";
+import AipLocationCard from "@/components/AipLocationCard.vue";
```
djjuhasz, line 7: Again, I'd prefer "AIPLocationCard" (like Go) if possible.

dashboard/src/components/__tests__/SipListLegend.test.ts
```diff
@@ -1,14 +1,14 @@
 import { cleanup, fireEvent, render } from "@testing-library/vue";
 import { afterEach, describe, expect, it } from "vitest";
 
-import PackageListLegend from "@/components/PackageListLegend.vue";
+import SipListLegend from "@/components/SipListLegend.vue";
```
djjuhasz, line 4: And I like "SIPListLegend" better. :wink:

dashboard/src/components/AipLocationCard.vue
```diff
@@ -40,11 +40,11 @@ watch(ingestStore.ui.download, () => download());
         Move operation failed, try again!
       </div>
       <div v-if="ingestStore.isMoving" class="alert alert-info" role="alert">
-        The package is being moved into a new location.
+        The AIP is being moved into a new location.
       </div>
       <h4 class="card-title">Location</h4>
       <p class="card-text">
-        <span v-if="ingestStore.isRejected">Package rejected.</span>
+        <span v-if="ingestStore.isRejected">AIP rejected.</span>
```
djjuhasz, line 47: Oh good, I was hoping you were going to update these messages. :+1:

djjuhasz (approved):
This looks good to me @jraddaoui.  I like the new "Ingest" and "Storage" labels on the navigation menu, and the SIPs icon is pretty slick. :ok_hand: 

I had a few suggestions and a few questions, but I don't think anything is a blocker.

djjuhasz:
Maybe we should change "AIP creation details" to "Ingest details"?
image

dashboard/src/components/Sidebar.vue
jraddaoui: When used with `v-for`, it's just a way to loop without rendering any element:

https://vuejs.org/api/built-in-special-elements#template

dashboard/src/components/Breadcrumb.vue
jraddaoui: Right, I forgot to ask, because the imports I added/modified are not in alphabetical order either. How do I run eslint to fix all that? I tried:

```
$ eslint .

Oops! Something went wrong! :(

ESLint: 6.4.0.

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

    eslint --init

ESLint looked for configuration files in /home/radda/artefactual/enduro/dashboard and its ancestors. If it found none, it then looked in your home directory.

If you think you already have a configuration file or if you need more help, please stop by the ESLint chat room: https://gitter.im/eslint/eslint

$ eslint . -c eslint.config.mjs
YAMLException: Cannot read config file: /home/radda/artefactual/enduro/dashboard/eslint.config.mjs
Error: end of the stream or a document separator is expected (9:10)

  6 | 
  7 |                                          ...
  8 | export default [
  9 |   { files: ["**/*.{js,mjs,cjs,ts,vue}"] },
---------------^
 10 |   {
 11 |     settings: {
    at generateError (/usr/share/nodejs/js-yaml/lib/loader.js:183:10)
    at throwError (/usr/share/nodejs/js-yaml/lib/loader.js:187:9)
    at readDocument (/usr/share/nodejs/js-yaml/lib/loader.js:1645:5)
    at loadDocuments (/usr/share/nodejs/js-yaml/lib/loader.js:1688:5)
    at Object.load (/usr/share/nodejs/js-yaml/lib/loader.js:1714:19)
    at loadLegacyConfigFile (/usr/share/nodejs/eslint/lib/cli-engine/config-array-factory.js:171:21)
    at loadConfigFile (/usr/share/nodejs/eslint/lib/cli-engine/config-array-factory.js:264:20)
    at ConfigArrayFactory._loadConfigData (/usr/share/nodejs/eslint/lib/cli-engine/config-array-factory.js:415:13)
    at ConfigArrayFactory.loadFile (/usr/share/nodejs/eslint/lib/cli-engine/config-array-factory.js:383:18)
    at createCLIConfigArray (/usr/share/nodejs/eslint/lib/cli-engine/cascading-config-array-factory.js:140:35)
```

We should add it to the `package.json` scripts and run it in CI.

dashboard/src/components/PackageLocationCard.vue
jraddaoui: Good point. Because we have the storage store already, I thought about having a general ingest store, in case we add other entities in the future, and I was planning to put both (AIPs and locations) into the storage store.

Now that you say this, I think it makes more sense to have a store per entity (`sips`, `aips` and `locations` stores). Not sure if we should use plurals here too now that they will be dealing with entities/collections, but let me know your thoughts, I'll do this in a following PR when we add AIPs.

dashboard/src/stores/__tests__/package.test.ts
jraddaoui: Ups, no, I just missed this one, thanks!

dashboard/src/stores/ingest.ts
jraddaoui: Yeah, I also had my conflicts about this. I think it is kind of a mess when you work with generated code that uses mixed case, in both sides Go and Typescript code. I even tried to update the API design to generate types following the naming convention, and I got close, but I didn't like how it looked in the generated API spec.

Because we try to follow naming conventions in Go we end up with a mix of `whateverSIP` and `whateverSip` but I don't think there are such conventions in Typescript, and based on the generated client, I though it would be better to use mixed case where possible in the dashboard.

dashboard/src/components/PackageLocationCard.vue
jraddaoui: I just moved it "without looking", but it's used to trigger the download from the review alert component:

https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/components/PackageReviewAlert.vue#L41

dashboard/src/components/__tests__/PackageLocationCard.test.ts
jraddaoui: I agree, just checking how the template is rendered based on initial properties, it feels so much like the requests tapes from AM.

dashboard/src/components/__tests__/AipLocationCard.test.ts
jraddaoui: Let me know if you feel really strong about this after reading my comment above. There will be A LOT of places to change.

dashboard/src/components/__tests__/SipListLegend.test.ts
jraddaoui: Same :(

dashboard/src/components/AipLocationCard.vue
jraddaoui: Yes, still having conflicts about SIP/AIP in this review process. It's part of the ingest workflow for the SIP but the review is supposed to happen over the AIP. I'll revisit when we change the move operation and other interactions between ingest and storage.

jraddaoui:
> Maybe we should change "AIP creation details" to "Ingest details"?

"SIP details" reads in the wireframes, I was actually planning to remove those headings initially, but it looks like Dan kept them in his version of those wireframes. I'll double check and modify this when we add the AIP pages.

dashboard/src/components/Breadcrumb.vue
jraddaoui: Okay, I have enabled it in VSCode and fixed the imports. This `Elements in iteration expect to have 'v-bind:key' directives.` error happens all over the code in many `v-for` elements, alongside other ESLint errors. Better to tackle them all when we include ESLint in the CI process.

dashboard/src/stores/ingest.ts
jraddaoui: If we go that way I'd try to do it all over the generated code (including Go and Typescript), so better do it in another PR.

dashboard/src/components/Breadcrumb.vue
djjuhasz: I run eslint like this:

```
cd dashboard
npx eslint
```

Which outputs many pages of warnings. :(

I'm running eslint 9:
```
❯ npx eslint --version
v9.20.1
```

There is an automatic `--fix` flag for eslint, but the "dry run" output indicates there's only 5 issues that can be automatically fixed. :thinking:

djjuhasz: Bottom of the eslint output :open_mouth: 
```
✖ 1231 problems (1226 errors, 5 warnings)
  0 errors and 5 warnings potentially fixable with the `--fix` option.
```

djjuhasz: Oooh, I think ESLint must be checking the generated code too.  I'll look at excluding the files generated by openapi, and any other code we don't control.

dashboard/src/components/PackageLocationCard.vue
djjuhasz: Having a store per entity sounds good to me.  I like the singular version of the entity names for code libraries (e.g. `sip`, `aip`, `location`) but I'd be okay with plurals too.

dashboard/src/components/__tests__/AipLocationCard.test.ts
djjuhasz: Let's leave it as-is for now.

dashboard/src/components/Sidebar.vue
djjuhasz: Ah, that's good to know about! I've been using `<span>` and `<div>` tags for conditionals but adding unnecessary elements to the DOM is not great.

In this case, what about doing something like this instead?
```<li v-for="item in menuItems.filter((i) => i.show)">```

Here's a test script implementing filter:
https://play.vuejs.org/#eNp9UsFOwzAM/RWTyzZptEJwGt0kQEjAARAgcSAcSud2GWkSJek2qeq/46QaGxPaKbHfc/ye45ZdGZOsGmQTlrnCCuPBoW/MjCtRG209tGCxhA5Kq2sYEHXAFVeFVs6D8Fg7mMIHVwCtymucAGc5Z2NwC72egLcNduM/6NcOLXPpDuHioJirz8tdw9pV1I4EDQd3KKWGd23l/GQw4ipLe/2knAISZmTukSKAbHE2a9tY3HVZSlHMCmUaD6vTWs9RTjkjnDNIe7CR8aSbFMQptSVG8AtC9b6TUkiPdjgUI5jOQCRB9Yiz0CoQkmApNpSifzONj2bpnjg2Zt6Rt1JUydJpRf/QBi4NQtdGSLRPxgvyztkEIhKwnKyvH2IuDClOMNYssPj+J790m5Dj7NmiQ7tCGvIW87mt0Pfw7esjbuj+C9JoGknsI+ALOi2boLGnXTdqTrL3eFHtfdwmoao3d7vxqNzWVBAamP0ecEYbdnPE+k7ueXIR67jqWPcDIxHqUA==

PR #1121, opened 2025-02-10
Replace DB package tables by sip and aip tables

jraddaoui:
Stop using package across the persistence layers. Use SIP on ingest
and AIP on storage renaming DB tables, types and enums. The use of
package still happens on the API design/spec and the dashboard, these
will be modified in following PRs.

- Allow generating migrations for both databases using ariga.io/atlas.
- Add `atlas-hash` Make rule to re-generate sum file in migration dirs.
- Update ingest database, ent schema and client:
  - Rename `package` table to `sip`.
  - Adapt `preservation_action` table relation.
  - Generate `use_atlas` migration to match schema from ent/Atlas.
  - Rename package related functions.
- Update storage database, ent schema and client:
  - Rename `package` table to `aip`.
  - Rename package related functions.
- Rename datatype SIP and PreservationAction.SIPID.
- Rename SIP statuses and types.
- Rename AIP statuses.

Refs #1117.


internal/db/migrations/20250210184720_use_atlas.up.sql
```diff
@@ -0,0 +1,49 @@
+-- drop foreign key constraint from "preservation_action" table
+ALTER TABLE `preservation_action` DROP FOREIGN KEY `preservation_action_ibfk_1`;
+-- modify "sip" table
+ALTER TABLE `sip` COLLATE utf8mb4_bin,
+  MODIFY COLUMN `id` bigint NOT NULL AUTO_INCREMENT,
```
djjuhasz, line 5: Why is it necessary to modify the column types here?

internal/db/migrations/20250210184720_use_atlas.up.sql
```diff
@@ -0,0 +1,49 @@
...
+  ADD INDEX `sip_created_at_idx` (`created_at`),
+  ADD INDEX `sip_location_id_idx` (`location_id`),
+  ADD INDEX `sip_name_idx` (`name` (50)),
+  ADD INDEX `sip_started_at_idx` (`started_at`),
+  ADD INDEX `sip_status_idx` (`status`);
+-- drop foreign key constraint from "preservation_task" table
+ALTER TABLE `preservation_task` DROP FOREIGN KEY `preservation_task_ibfk_1`;
+-- modify "preservation_action" table
+ALTER TABLE `preservation_action` COLLATE utf8mb4_bin,
+  MODIFY COLUMN `id` bigint NOT NULL AUTO_INCREMENT,
```
djjuhasz, line 30: And here?

internal/db/migrations/20250210184720_use_atlas.up.sql
```diff
@@ -0,0 +1,49 @@
...
+  MODIFY COLUMN `workflow_id` varchar(255) NOT NULL,
+  MODIFY COLUMN `started_at` timestamp NULL,
+  MODIFY COLUMN `completed_at` timestamp NULL,
+  MODIFY COLUMN `sip_id` bigint NOT NULL,
+  ADD INDEX `preservation_action_sip_preservation_actions` (`sip_id`),
+  ADD CONSTRAINT `preservation_action_sip_preservation_actions`
+    FOREIGN KEY (`sip_id`) REFERENCES `sip` (`id`) ON UPDATE NO ACTION ON DELETE CASCADE;
+-- modify "preservation_task" table
+ALTER TABLE `preservation_task` COLLATE utf8mb4_bin,
+  MODIFY COLUMN `id` bigint NOT NULL AUTO_INCREMENT,
```
djjuhasz, line 40: And here?

internal/package_/convert.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/timerange"
 )
 
-func packageToGoaPackageCreatedEvent(p *datatypes.Package) *goapackage.PackageCreatedEvent {
+func packageToGoaPackageCreatedEvent(p *datatypes.SIP) *goapackage.PackageCreatedEvent {
```
djjuhasz: I think this function name should change, but we should probably just drop the initial "package"  because it's in the `package_` uh... package. E.g. `package_.toGoaPackageCreatedEvent(sip)` is quite readable and less redundant.

internal/package_/convert.go
```diff
@@ -84,7 +84,7 @@ func preservationTaskToGoa(pt *datatypes.PreservationTask) *goapackage.EnduroPac
 	}
 }
 
-func listPayloadToPackageFilter(payload *goapackage.ListPayload) (*persistence.PackageFilter, error) {
+func listPayloadToPackageFilter(payload *goapackage.ListPayload) (*persistence.SIPFilter, error) {
```
djjuhasz: Rename function to `listPayloadToSIPFilter()`?

internal/persistence/ent/client/preservation_action.go
```diff
@@ -35,7 +35,7 @@ func (c *client) CreatePreservationAction(ctx context.Context, pa *datatypes.Pre
 		SetStatus(int8(pa.Status)). // #nosec G115 -- constrained value.
 		SetNillableStartedAt(startedAt).
 		SetNillableCompletedAt(completedAt).
-		SetPackageID(pa.PackageID)
+		SetSipID(pa.SIPID)
```
djjuhasz, line 38: Should be `SetSIPID()`.

internal/persistence/ent/client/preservation_task_test.go
```diff
@@ -24,17 +24,17 @@ func addDBFixtures(
 ) (*db.PreservationAction, *db.PreservationAction) {
 	t.Helper()
 
-	pkg, err := createPackage(entc, "P1", enums.PackageStatusInProgress)
+	sip, err := createSIP(entc, "S1", enums.SIPStatusInProgress)
 	if err != nil {
-		t.Errorf("create package: %v", err)
+		t.Errorf("create sip: %v", err)
```
djjuhasz: I think this should be "create SIP: %v"

internal/persistence/ent/client/sip.go
```diff
@@ -9,88 +9,88 @@ import (
...
+		q.SetCompletedAt(s.CompletedAt.Time)
 	}
 
 	// Set CreatedAt to the current time
 	q.SetCreatedAt(time.Now())
 
-	// Save the package.
-	p, err := q.Save(ctx)
+	// Save the SIP.
+	ps, err := q.Save(ctx)
```
djjuhasz: Very nitpicky but maybe "ss" for "saved SIP" would be better here?  I'm not sure what "ps" would be.

internal/persistence/persistence.go
```diff
@@ -19,17 +19,17 @@ var (
...
 
 type (
-	PackageUpdater  func(*datatypes.Package) (*datatypes.Package, error)
+	SIPUpdater      func(*datatypes.SIP) (*datatypes.SIP, error)
 	PresTaskUpdater func(*datatypes.PreservationTask) (*datatypes.PreservationTask, error)
 )
 
 type Service interface {
-	// CreatePackage persists the given Package to the data store then updates
+	// CreateSIP persists the given Package to the data store then updates
```
djjuhasz: I think you missed a couple of "Package" instances here, e.g. "CreateSIP persists the given *SIP*..."

internal/storage/persistence/ent/client/client_test.go
```diff
@@ -221,40 +221,40 @@ func TestReadPackage(t *testing.T) {
...
+	t.Run("Returns error when AIP does not exist", func(t *testing.T) {
 		t.Parallel()
 
 		_, c := setUpClient(t)
 
-		l, err := c.ReadPackage(context.Background(), aipID)
+		l, err := c.ReadAIP(context.Background(), aipID)
 		assert.Assert(t, l == nil)
-		assert.ErrorContains(t, err, "package not found")
+		assert.ErrorContains(t, err, "Storage package not found")
```
djjuhasz, line 231: Why not "AIP not found"?

internal/storage/persistence/ent/schema/aip.go
```diff
@@ -14,20 +14,20 @@ import (
...
+func (AIP) Annotations() []schema.Annotation {
 	return []schema.Annotation{
-		entsql.Annotation{Table: "package"},
+		entsql.Annotation{Table: "aip"},
 	}
 }
 
-// Fields of the Pkg.
-func (Pkg) Fields() []ent.Field {
+// Fields of the AIP.
```
djjuhasz, line 29: These comments are not really helpful. :laughing:

internal/storage/persistence/ent/schema/aip.go
```diff
@@ -47,17 +47,17 @@ func (Pkg) Fields() []ent.Field {
...
 			Unique(),
 	}
 }
 
-// Indexes of the Pkg.
-func (Pkg) Indexes() []ent.Index {
+// Indexes of the AIP.
+func (AIP) Indexes() []ent.Index {
 	return []ent.Index{
 		index.Fields("aip_id"),
```
djjuhasz, line 62: Hmm, "aip.aip_id" is kind of confusing.  Maybe we should rename the column "uuid"?

internal/storage/persistence/migrations/20250211024602_rename_package_table.up.sql
```diff
@@ -0,0 +1,4 @@
+-- rename "package" table
+RENAME TABLE `package` TO `aip`;
+-- modify "aip" table
+ALTER TABLE `aip` DROP INDEX `pkg_aip_id`, DROP INDEX `pkg_object_key`, ADD INDEX `aip_aip_id` (`aip_id`), ADD INDEX `aip_location_location` (`location_id`), ADD INDEX `aip_object_key` (`object_key`), DROP FOREIGN KEY `package_location_location`, ADD CONSTRAINT `aip_location_location` FOREIGN KEY (`location_id`) REFERENCES `location` (`id`) ON UPDATE NO ACTION ON DELETE SET NULL;
```
djjuhasz, line 4: If we change the `aip_id` column name to `uuid` then the index name should be "aip_uuid", which makes more sense.

internal/storage/service.go
```diff
@@ -31,7 +31,7 @@ type Service interface {
 	// Used from workflow activities.
 	Location(ctx context.Context, locationID uuid.UUID) (Location, error)
 	ReadPackage(ctx context.Context, aipID uuid.UUID) (*goastorage.Package, error)
-	UpdatePackageStatus(ctx context.Context, aipID uuid.UUID, status types.PackageStatus) error
+	UpdatePackageStatus(ctx context.Context, aipID uuid.UUID, status types.AIPStatus) error
```
djjuhasz, line 34: Can this be renamed to "UpdateAIPStatus" or does that need to wait until the API names are changed?

internal/storage/service.go
```diff
@@ -31,7 +31,7 @@ type Service interface {
...
 	UpdatePackageLocationID(ctx context.Context, aipID, locationID uuid.UUID) error
```
djjuhasz, line 35: Ditto

internal/storage/service_test.go
```diff
@@ -594,10 +594,10 @@ func TestReject(t *testing.T) {
 
 		attrs.persistenceMock.
 			EXPECT().
-			UpdatePackageStatus(
+			UpdateAIPStatus(
```
djjuhasz, line 597: Hmm, shouldn't this be broken because the service method is still called "UpdatePackageStatus()"?

internal/storage/types/aip_status.go
```diff
@@ -0,0 +1,97 @@
+package types
```
djjuhasz, line 1: Huh, are we not using `gen-enums` for this yet?

internal/workflow/local_activities.go
```diff
@@ -17,7 +17,7 @@ import (
 
 type createPackageLocalActivityParams struct {
```
djjuhasz, line 18: This should be changed to `createSIPLocalActivityParams` at some point.

internal/workflow/local_activities.go
```diff
@@ -17,7 +17,7 @@ import (
...
 	Key    string
-	Status enums.PackageStatus
+	Status enums.SIPStatus
 }
 
 func createPackageLocalActivity(
```
djjuhasz, line 23: And this to "createSIPLocalActivity".

djjuhasz (approved):
Whew, finished the review. :sweat_smile: 

It looks great overall @jraddaoui.  

In the database migrations it would be nice to avoid updating all of the column definitions, as I don't think they've actually changed, but that may just be a quirk of Atlas.

I've flagged a number of minor renaming issues, but I don't think there's any serious blockers.

scollazo:
Quick question, are migrations applied by the app each time it starts, or should a migration command be run before the app starts?

internal/db/migrations/20250210184720_use_atlas.up.sql
jraddaoui: As you said, making Atlas happy. Some column types change like the auto-generated `id` column, in others is just the collation.

internal/db/migrations/20250210184720_use_atlas.up.sql
jraddaoui: Same, check the storage DB, is how Atlas/ent generate the `id` column.

internal/db/migrations/20250210184720_use_atlas.up.sql
jraddaoui: ¯\_(ツ)_/¯

But now that you say it, we may want to change the datatypes fields holding those IDs.

internal/package_/convert.go
jraddaoui: This will become the `ingest` package when we tackle the API design changes. This function will probably be `sipToGoaSIP...`. It's also an internal function, working with the package_ package Package :package: :stuck_out_tongue_closed_eyes:

jraddaoui: For now I'll call it `sipToGoaPackageCreatedEvent`.

internal/persistence/ent/client/preservation_action.go
jraddaoui: This is not up to us, dealing with ent generated code.

internal/persistence/ent/client/sip.go
jraddaoui: Persisted/Persistence SIP? I think `dbs` is better, thanks.

internal/storage/persistence/ent/client/client_test.go
jraddaoui: This puzzled me a little too, it looks like we use Goa errors in the persistence layer. So, even if I change the message value:

https://github.com/artefactual-sdps/enduro/blob/dev/issue-1117-db-schema/internal/storage/persistence/ent/client/client.go#L82

The error message in Go is still the API design error:

https://github.com/artefactual-sdps/enduro/blob/dev/issue-1117-db-schema/internal/api/design/storage.go#L75

I'll address this in a following PR, when renaming package to AIP in the API design. And we should consider not using Goa errors in the persistence layer at some point.

internal/storage/persistence/ent/schema/aip.go
jraddaoui: Not at all :grin:

internal/storage/persistence/ent/schema/aip.go
jraddaoui: I agree, I'll do it when I add an "uuid" column to the SIP and I'll probably change "aip_id" to "aip_uuid" in the "sip" table too. This will all be part of the "using UUIDs instead of DBIDs as resource identifiers" PR.

internal/storage/persistence/migrations/20250211024602_rename_package_table.up.sql
jraddaoui: Totally.

internal/storage/service.go
jraddaoui: Since we base our services in the Goa design I was planning to address this when updating that design. Now that you say it, this are service functions not used by Goa, only internally, but I still think it's better to tackle the entire service at the same time.

internal/storage/service.go
jraddaoui: Same.

internal/storage/service_test.go
jraddaoui: This is the storage persistence service mock, not the main storage service.

internal/storage/types/aip_status.go
jraddaoui: Nope :disappointed::

https://github.com/artefactual-sdps/enduro/blob/dev/issue-1117-db-schema/Makefile#L115-L124

internal/workflow/local_activities.go
jraddaoui: Similar, this used the package_ service, which is still based on the API design. I'll fix all package references across the code after doing the API re-design.

internal/workflow/local_activities.go
jraddaoui: Similar, this used the package_ service, which is still based on the API design. I'll fix all package references across the code after doing the API re-design.

jraddaoui:
> Quick question, are migrations applied by the app each time it starts, or should a migration command be run before the app starts?

In this project we use versioned migrations even if they are applied on the application start, so there is no need of an extra command. The `migrate` config value for both databases must be set to true for that to happen though.

internal/db/migrations/20250210184720_use_atlas.up.sql
djjuhasz: Ah okay. I didn't realize there were actual changes to the column definitions.

internal/db/migrations/20250210184720_use_atlas.up.sql
jraddaoui: Thinking about this...

The ent/Atlas auto-generated `id` columns are signed `BIGINT` (9,223,372,036,854,775,807) while the columns we had were unsigned `INT` (4,294,967,295). `BIGINT` uses 8 bytes of storage per row while `INT` uses 4 bytes, and these columns are used for indexes and foreign keys.

TBH, I don't know how big are the performance implications of using one against the other, but I don't think we need such a wide range of IDs in the ingest DB, I'd expect some clean-up after millions of ingested packages, considering they will reach the storage DB. So we may want to keep using `INT` here and `BIGINT` in storage.

Thoughts? @djjuhasz @mcantelon (also @sevein if you have some time).

djjuhasz: I can't imagine we'd ever insert over 4 billion SIPs or AIPs into an Enduro database — if an institution was ingesting a million packages a **day** it would take more than 11 years to reach the 4 billion records limit. :exploding_head:

That said I don't know how easy it is to change the default `id` column size created by entgo / Atlas — if it's a difficult it might be better to just stick with the defaults.  I guess Atlas uses a signed `BIGINT` to align with the Go `int64` type?

jraddaoui: Thanks @djjuhasz, this is what it takes to match the existing schema as much as possible:

https://github.com/artefactual-sdps/enduro/commit/b4cd686814ecbe70ac58a6f81f378e2d1c11e4ee

I'll leave it as it is in this PR.

PR #1129, opened 2025-02-19
Added src to the edit URI

sallain:
This is a guess and I don't have the infrastructure to tell if it works, but maybe this will fix https://github.com/artefactual-sdps/enduro-team-tasks/issues/34?

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -3,7 +3,7 @@ site_url: "https://enduro.readthedocs.io"
 site_description: "Enduro SDPS documentation site."
 site_author: "Artefactual Systems Inc."
 repo_url: "https://github.com/artefactual-sdps/enduro"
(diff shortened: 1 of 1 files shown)


docs/mkdocs.yml
```diff
@@ -3,7 +3,7 @@ site_url: "https://enduro.readthedocs.io"
 site_description: "Enduro SDPS documentation site."
 site_author: "Artefactual Systems Inc."
 repo_url: "https://github.com/artefactual-sdps/enduro"
-edit_uri: "blob/main/docs/"
+edit_uri: "blob/main/docs/src"
```
jraddaoui: I think it will require the final slash.

jraddaoui (changes requested):

jraddaoui (approved):

Issue #1102, opened 2025-01-14
Feature: Filter packages by date


sallain:
**Is your feature request related to a problem? Please describe.**

Being able to call up a list of packages that were ingested  on a specific date, or between two dates, can help me to locate a specific package or to gather rudimentary stats on how many packages are being processed.

**Describe the solution you'd like**

On the Packages page, I'd like to be able to enter a date (or a date range) using some kind of date picker and see all packages that were ingested during that period.

**Describe alternatives you've considered**

**Additional context**

fiver-watson:
This might not be the first implementation, but based on some of the previous SFA feedback, as well as some internal feedback on a few alternative design proposals, I have made some simple updated mockups of how we could add multiple filter options and a simple search to the packages page. 

As always discussion, feedback, alternative proposals, change requests, etc. all welcome! Just giving us a starting point for further team review and planning. 

![SFA_search-and-filter_01](https://github.com/user-attachments/assets/114286b1-2244-43b1-924b-a48fb21e9e51)

In this first image, I have added the search box directly on the packages page, rather than add it to the global header. I think this will make it easier to a) understand that it is currently limited to searching packages, and b) that search results can interact with filters. 

By default, available filter options are hidden. There is an "Add filters" button to the right of the package card results - clicking it will reveal a filter area below the search box and above the results, as shown in the next image. 

Long-term, I hope that we can add some sort options as well - I would imagine that such a button would also be on the right with the "Add filters" button. 

![SFA_seach-and-filter_02](https://github.com/user-attachments/assets/e4d5300a-0bcb-457a-8a7e-f53904a1f70f)

In this image, the "Add filters" option has been clicked, displaying the filter options below. 

For design consistency, I have removed the Status filter tabs in these mockups, and moved them into the proposed filter pills - I would suggest that we re-implement them as checkboxes, so that results can be filtered for multiple statuses at a time (otherwise we might as well just have kept the tabs). 

[...]

PR #1134, opened 2025-02-25
Add custom date range package filters

djjuhasz:
Refs #1102

- Convert pre-defined date range selector to a drop-down widget to reduce size of list
- Add start and end date pickers to a allow selecting a custom times for filtering the package list

Note: Vitest shows a very low coverage for statements in the TimeDrodown component, even though the unit tests should cover most of the code. I've tried to fix the coverage report a number of different ways, but I haven't been able to determine why it's not working correctly.

### DIFF
--- a/dashboard/src/components/TimeDropdown.vue
+++ b/dashboard/src/components/TimeDropdown.vue
@@ -1,7 +1,15 @@
 <script setup lang="ts">
-import { onMounted, ref, useTemplateRef } from "vue";
+import VueDatePicker from "@vuepic/vue-datepicker";
+import type { ModelValue } from "@vuepic/vue-datepicker";
+import "@vuepic/vue-datepicker/dist/main.css";
+import Dropdown from "bootstrap/js/dist/dropdown";
+import { onMounted, ref, watch } from "vue";
 
-const emit = defineEmits<{ change: [field: string, value: string] }>();
+import IconCloseLine from "~icons/clarity/close-line";
+
+const emit = defineEmits<{
+  change: [name: string, start: string, end: string];
+}>();
 
 type option = {
   value: string;
@@ -9,7 +17,7 @@ type option = {
 };
 
 const options: option[] = [
-  { value: "", label: "Any time" },
+  { value: "", label: "Select a time range" },
   { value: "3h", label: "The last 3 hours" },
   { value: "6h", label: "The last 6 hours" },
   { value: "12h", label: "The last 12 hours" },
@@ -19,55 +27,84 @@ const options: option[] = [
 ];
 
 const props = defineProps<{
-  fieldname: string;
+  name: string;
+  label: string;
 }>();
 
-const label = ref("Started");
-const selected = ref(options[0]);
-const dropdown = useTemplateRef("date-filter");
+const el = ref<HTMLElement | null>(null);
+const dropdown = ref<Dropdown | null>(null);
+const defaultLabel = props.label || "Time";
+const btnLabel = ref(defaultLabel);
+const selectedPreset = ref("");
+const startTime = ref<Date | null>(null);
+const endTime = ref<Date | null>(null);
 
 onMounted(() => {
-  if (dropdown.value) {
-    dropdown.value.style.display = "none";
-  }
+  if (el.value) dropdown.value = new Dropdown(el.value);
 });
 
-const toggle = () => {
-  if (dropdown.value) {
-    if (dropdown.value.style.display == "none") {
-      dropdown.value.style.display = "block";
-    } else if (dropdown.value.style.display == "block") {
-      dropdown.value.style.display = "none";
-    }
-  }
-};
-
-const handleChange = (opt: option) => {
-  selected.value = opt;
-
-  if (opt.value === "") {
-    label.value = "Started";
-  } else {
-    label.value = "Started: " + opt.label;
+watch(selectedPreset, async (newValue) => {
+  let sel = options.find((o) => o.value == newValue);
+  if (!sel || sel.value === "") {
+    return;
   }
 
-  toggle();
-  emit("change", props.fieldname, earliestTimeFromOption(opt));
+  btnLabel.value = defaultLabel + ": " + sel.value;
+  startTime.value = earliestTimeFromOption(sel.value);
+  endTime.value = null;
+
+  emitChange();
+});
+
+const emitChange = () => {
+  emit(
+    "change",
+    props.name,
+    formatDate(startTime.value),
+    formatDate(endTime.value),
+  );
 };
 
-const earliestTimeFromOption = (opt: option) => {
-  const formatDate = (date: Date) => {
-    let t = date.toISOString();
-    t = t.split(".")[0] + "Z"; // remove milliseconds.
-    return t;
-  };
+const handleCustomTimeChange = () => {
+  btnLabel.value = defaultLabel + ": Custom";
+  selectedPreset.value = "";
 
+  emitChange();
+};
+
+const handleStartChange = (modelData: ModelValue) => {
+  startTime.value = modelData as Date;
+  handleCustomTimeChange();
+};
+
+const handleEndChange = (modelData: ModelValue) => {
+  endTime.value = modelData as Date;
+  handleCustomTimeChange();
+};
+
+const reset = () => {
+  btnLabel.value = defaultLabel;
+  selectedPreset.value = "";
+  startTime.value = null;
+  endTime.value = null;
+
+  emitChange();
+};
+
+const formatDate = (date: Date | null) => {
+  if (!date) return "";
+  let t = date.toISOString();
+  t = t.split(".")[0] + "Z"; // remove milliseconds.
+  return t;
+};
+
+const earliestTimeFromOption = (value: string) => {
   // convert hours and days to milliseconds.
   const hour = 60 * 60 * 1000;
   const day = 24 * hour;
 
   let start = new Date();
-  switch (opt.value) {
+  switch (value) {
     case "3h":
       start = new Date(Date.now() - 3 * hour);
       break;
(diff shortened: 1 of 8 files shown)


djjuhasz:
Here's how it looks:
image

dashboard/src/components/PreservationActionCollapse.vue
```diff
@@ -70,7 +70,7 @@ function isComplete(task: EnduroPackagePreservationTask) {
         />
         <div
           v-for="(task, index) in action.tasks.slice().reverse()"
-          :key="action.id"
+          :key="task.id"
```
jraddaoui: Now I get the browser warning I was seeing yesterday ;)

dashboard/src/components/TimeDropdown.vue
```diff
@@ -18,56 +25,102 @@ const options: option[] = [
   { value: "7d", label: "The last 7 days" },
 ];
 
+const defaultLabel = "Started";
```
jraddaoui: Should we pass this as a property to the component? Doesn't have to happen in this PR, just a thought.

dashboard/src/components/TimeDropdown.vue
```diff
@@ -18,56 +25,102 @@ const options: option[] = [
...
+  label.value = defaultLabel;
+  selected.value = "";
+  startTime.value = null;
+  endTime.value = null;
+  emit(
+    "change",
+    props.name,
+    formatDate(startTime.value),
+    formatDate(endTime.value),
+  );
```
jraddaoui: Same call for the third time, we could move it to its own function.

dashboard/src/components/TimeDropdown.vue
```diff
@@ -18,56 +25,102 @@ const options: option[] = [
...
+const timeFilter = useTemplateRef("time-filter");
+const selected = ref("");
+const startTime = ref<Date | null>(null);
+const endTime = ref<Date | null>(null);
 
 onMounted(() => {
-  if (dropdown.value) {
-    dropdown.value.style.display = "none";
+  if (timeFilter.value) {
+    timeFilter.value.style.display = "none";
```
jraddaoui: Why is this needed? Does the Bootstrap dropdown doesn't work?

dashboard/src/components/TimeDropdown.vue
```diff
@@ -18,56 +25,102 @@ const options: option[] = [
...
-    } else if (dropdown.value.style.display == "block") {
-      dropdown.value.style.display = "none";
+  if (timeFilter.value) {
+    if (timeFilter.value.style.display == "none") {
+      timeFilter.value.style.display = "block";
+    } else if (timeFilter.value.style.display == "block") {
+      timeFilter.value.style.display = "none";
     }
   }
 };
```
jraddaoui: Same.

dashboard/src/components/TimeDropdown.vue
```diff
@@ -87,30 +140,83 @@ const earliestTimeFromOption = (opt: option) => {
...
-    <button
-      @click="toggle"
-      class="btn btn-primary dropdown-toggle"
-      type="button"
-      data-bs-toggle="dropdown"
-      aria-expanded="false"
+    <div class="input-group">
+      <button
+        :id="'time-filter-' + props.name + '-toggle'"
+        @click="toggle"
```
jraddaoui: I see, it probably doesn't work because the Dropdown component from Bootstrap is not being imported. We should fix this, otherwise the `aria-expanded` attribute won't change and who knows what else.

dashboard/src/components/TimeDropdown.vue
```diff
@@ -87,30 +140,83 @@ const earliestTimeFromOption = (opt: option) => {
...
+        :hidden="startTime === null && endTime === null"
+      >
+        <IconCloseLine />
+      </button>
+    </div>
+    <div
+      :id="'time-filter-' + props.name"
+      class="dropdown-menu p-2"
+      ref="time-filter"
+      style="min-width: 250px"
```
jraddaoui: Why this? Doesn't it fit the content? If really needed I'd move it to a style section after the template, you could target the id with a wildcard.

dashboard/src/components/TimeDropdown.vue
```diff
@@ -87,30 +140,83 @@ const earliestTimeFromOption = (opt: option) => {
...
+          @update:model-value="handleStartChange"
+        />
+        to
+        <VueDatePicker
+          time-picker-inline
+          :id="'time-filter-' + props.name + '-end'"
+          v-model="endTime"
+          placeholder="End time"
+          @update:model-value="handleEndChange"
+        />
```
jraddaoui: Should it read "From" ... "To" ... Are the placeholders transformed to proper labels? Should we add our own? Right now it looks rare to me "Custom range" ... "to" ...

dashboard/src/components/__tests__/TimeDropdown.test.ts
```diff
@@ -1,114 +1,127 @@
...
+
+  it("toggles dropdown visibility when button is clicked", async () => {
+    const dropdown = wrapper.get("#time-filter-createdAt");
+
+    await wrapper.get("#time-filter-createdAt-toggle").trigger("click");
+    expect(dropdown.isVisible()).toBe(true);
+
+    await wrapper.get("#time-filter-createdAt-toggle").trigger("click");
+    expect(dropdown.isVisible()).toBe(false);
+  });
```
jraddaoui: Should help testing the change to use the Dropdown component from Bootstrap, but not a lot of value after that.

dashboard/src/components/__tests__/TimeDropdown.test.ts
```diff
@@ -1,114 +1,127 @@
...
+      "update:model-value",
+      new Date(Date.UTC(2025, 0, 1, 12, 0, 0)),
+    );
+    await datePicker.vm.$nextTick();
+
+    expect(button.text()).toBe("Started: Custom");
+    expect(wrapper.emitted("change")).toEqual([
+      ["createdAt", "2025-01-01T12:00:00Z", ""],
+    ]);
   });
```
jraddaoui, line 91: This test is making a previous one ("handles direct date picker input") unnecessary.

dashboard/tsconfig.vitest.json
```diff
@@ -4,6 +4,6 @@
   "compilerOptions": {
     "composite": true,
     "lib": [],
-    "types": ["node", "jsdom"]
+    "types": ["node", "jsdom", "vitest/globals"]
```
jraddaoui: Why is this needed for?

jraddaoui (changes requested):
Looking good @djjuhasz! Only important feedback would be to use the Dropdown component from Bootstrap.

dashboard/src/components/PreservationActionCollapse.vue
djjuhasz: @jraddaoui are you are getting a new browser warning with this update, or you saw a browser warning yesterday before this change was made?

dashboard/src/components/TimeDropdown.vue
djjuhasz: Yes, I'll have to make it a prop to support the "completed" date filter. :thinking:

dashboard/src/components/PreservationActionCollapse.vue
jraddaoui: Before, this should fix it ;)

dashboard/src/components/TimeDropdown.vue
djjuhasz: I tried getting the Bootstrap Dropdowns working with Vue3 a few months ago, but didn't have any luck.  I believe the problem is due to the Dropdowns dependency on https://popper.js.org/ (which has since been rebranded to https://floating-ui.com/), which isn't included in our current Vue.js/node environment.  

I looked at installing https://bootstrap-vue.org/ for full Bootstrap compatibility but the project seems to have died on Bootstrap 4, with the last release in October 2022. In addition support for Vue3 seems experimental, based on https://www.npmjs.com/package/@vue/compat.

I'm sure it's possible to add Popper / Floating UI to the Dashboard somehow, but I'm don't understand enough about how we are loading Bootstrap to know how to add Popper to the stack.

dashboard/src/components/TimeDropdown.vue
djjuhasz: Yes, without forcing the min-width the date / time values were being truncated in Firefox. :(  I will move this to a style section as you've suggested.

dashboard/src/components/TimeDropdown.vue
djjuhasz: Let's see what @fiver-watson / @sallain or the client think about it.  I like "Custom range" to make it clear that the bottom half of the popup is separate from the "Preset" time ranges, and I think adding a "From" label in addition is unnecessary.

dashboard/tsconfig.vitest.json
djjuhasz: Oh, that was an attempt to get the test coverage working properly — I'll remove it.

dashboard/src/components/TimeDropdown.vue
jraddaoui: Still, they are not proper labels for the related inputs and I'm not sure how `VueDatePicker` works, maybe it actually transforms the placeholders in proper labels. But if it doesn't, this doesn't seem accessible enough.

dashboard/src/components/TimeDropdown.vue
jraddaoui: We should look at this, right now for example, if you click outside the dropdown it doesn't close, and as I said in another comment, I don't know what else may be missing.

djjuhasz: @jraddaoui okay, I've figured out how to get the Bootstrap dropdown working with Vue, using a test component.  I'll update TimeDropdown to use the Bootstrap dropdown.

It's amazing what I've learned from a few weeks of banging my head against Vue. :laughing:

dashboard/src/components/__tests__/TimeDropdown.test.ts
jraddaoui: I kept this one that also checks the button and removed "handles direct date picker input".

dashboard/src/components/__tests__/TimeDropdown.test.ts
jraddaoui: Removed it.

dashboard/src/components/TimeDropdown.vue
jraddaoui: Thanks, I added I minor tweak to the new labels/inputs, so they use the proper `for` and `name` attributes.

jraddaoui:
Thanks @djjuhasz, nice work! I've rebased and added a few minor tweaks to get this squashed and merged before release.

jraddaoui (approved):

Issue #1130, opened 2025-02-20
Problem: It is not clear what is searchable via the SIP browse page search box


fiver-watson:
**Describe the bug**

Currently the SIP browse page includes a table listing per row: 

* ID
* Name
* UUID
* Start time
* Location
* Status

Individual SIP view pages may also contain additional information, such as completed time, total ingest workflow duration, workflow ID, etc. 

Directly above the SIP browse results table is a search box, with a simple placeholder text that says "Search."

However, from this prompt, it is not immediately clear to end users what fields can and cannot be searched using this box. 

**To Reproduce**

1. Log into Enduro
2. Navigate to the Packages / SIPs page
3. Copy a UUID from one of the successful packages and attempt to search using it
4. Copy a "started" timestamp and attempt to search with it
5. Copy a location UUID and attempt to search with it

Outcome: 0 results

**Expected behavior**

The search box placeholder text should help clarify to users that only SIP package name search is currently supported. 

I recommend changing this Placeholder text to: "Search SIP name" (or "Search by SIP name") for now to better clarify this for end users.

jraddaoui:
Using "Search by name". I initially used "Search by package name", planning to replace package by SIP in the re-org, but it seems redundant to me in a browse packages/SIPs page. Also, the column heading is just "Name", so I thought that would make more sense.

sallain:
LGTM!

PR #1135, opened 2025-02-26
Update packages search box placeholder and label

jraddaoui:
Use "Search by name" to clarify what field is queried from that input.

Refs #1130.


djjuhasz (approved):
:+1:

PR #1139, opened 2025-02-27
Pass start and end times to TimeDropdown

djjuhasz:
- Allow passing an initial start and end time to the TimeDropown component
- Pass the "earliest created date" and "latest created date" parameters from the current route to the "Created" date/time filter.

### DIFF
--- a/dashboard/src/components/TimeDropdown.vue
+++ b/dashboard/src/components/TimeDropdown.vue
@@ -28,7 +28,9 @@ const options: option[] = [
 
 const props = defineProps<{
   name: string;
-  label: string;
+  label?: string;
+  start?: Date;
+  end?: Date;
 }>();
 
 const el = ref<HTMLElement | null>(null);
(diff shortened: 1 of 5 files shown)


djjuhasz:
@jraddaoui this PR contains three distinct commits but they are all small. Most importantly for my sanity I finally figured out why the test coverage report was not working properly! :tada:

dashboard/src/components/TimeDropdown.vue
```diff
@@ -41,6 +43,14 @@ const endTime = ref<Date | null>(null);
 
 onMounted(() => {
   if (el.value) dropdown.value = new Dropdown(el.value);
+  if (props.start) {
+    startTime.value = props.start;
+    btnLabel.value = defaultLabel + ": Custom";
```
jraddaoui, line 48: Not a big deal, but this could be misleading if the date comes from a preset value. The times will change for those preset values so it kind of makes sense to make it custom if you want to keep the exact time that was used at the time of filtering, although the end date will vary as it's never set in those presets.

I think it would make more sense to have the preset ranges as part of the dashboard URL `createdAtRange=3h` and then transform them before they are sent to the API to be able to keep showing the last 3 hour SIPs.

Just a note, maybe for the future if we think it deserves the effort, not saying we should change it now as it will require modifying the actual implementation quite a bit.

jraddaoui (approved):
LGTM, just some thoughts for the future, thanks @djjuhasz!

dashboard/src/components/TimeDropdown.vue
djjuhasz: Yes, good points @jraddaoui.  I don't know if Enduro users will want to have URLs that represent a constant time span (in which case we'd need to set the end time as you've mentioned) or would prefer for the URL to always represent the 3 hours before the current time. :thinking:

As you say, I think this is something we can address in the future based on user feedback.

PR #1140, opened 2025-03-03
Fix AIP download URL in dashboard

jraddaoui:
Refs #1117.


djjuhasz:
@jraddaoui looks like the test needs to be updated too.

djjuhasz (approved):
LGTM 👍

PR #1142, opened 2025-03-06
Show SIP list error message on invalid time range

### LINKED ISSUE #1141: Problem: Dashboard package list doesn't show any warnings when the "started" time range is invalid
**Describe the bug**

On the Dashboard "packages" page, if a "started" time filter is set with an end time that comes before the start time, no packages are shown and there is no indication in the Dashboard that the time range is invalid.

**To Reproduce**

1. Go to the Enduro Dashboard in a browser
2. Click on the "Packages" button
3. Click on the "Started" date filter
4. Set a start time in the "Custom range" area
5. Set an end time that comes before the start time

**Expected behavior**

A warning should be shown that the end time must be after the start time.

djjuhasz:
Fixes #1141 

- Return a "400 Bad Request" HTTP error from the Enduro API when invalid package list parameters are sent in the request (it was returning a "500 Internal Server Error")
- Show a Dashboard error message on the SIP list when receiving an error response from the API
- Don't pass empty request parameters (name, status) from the SIP list to the back-end
- Remove some duplicate code for setting package request parameters
- Improve the API error messages for the "ListSIPs" endpoint

### DIFF
--- a/dashboard/src/pages/ingest/sips/index.vue
+++ b/dashboard/src/pages/ingest/sips/index.vue
@@ -99,12 +99,12 @@ const tabs = computed(() => [
   },
 ]);
 
-const doSearch = () => {
+const searchByName = () => {
   let q = { ...route.query };
   if (ingestStore.filters.name === "") {
     delete q.name;
   } else {
-    q.name = ingestStore.filters.name;
+    q.name = <LocationQueryValue>ingestStore.filters.name;
   }
 
   router.push({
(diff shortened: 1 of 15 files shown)


djjuhasz:
Here's a screenshot of what the error message looks like:
image

It's pretty big and scary, we may want to reduce the size at some point.  I would also like to add some validation to the Dashboard date picker that would prevent the user from picking a bad time range (e.g. an end date before a start date) to avoid the big scary warning.

dashboard/src/pages/ingest/sips/index.vue
```diff
@@ -221,7 +212,7 @@ watch(
 
     <div class="d-flex flex-wrap gap-3 mb-3">
       <div>
-        <form id="sipSearch" @submit.prevent="doSearch">
+        <form id="packageSearch" @submit.prevent="searchByName">
```
jraddaoui: I guess the id change is the result of a rebase.

dashboard/src/stores/ingest.ts
```diff
@@ -37,8 +40,8 @@ export const useIngestStore = defineStore("ingest", {
     // A list of SIPs shown during searches.
     sips: [] as Array<api.EnduroIngestSip>,
 
-    // Page is a subset of the total SIP list.
-    page: { limit: 20 } as api.EnduroPage,
+    // Page is a subset of the total package list.
```
jraddaoui: Same?

dashboard/src/stores/ingest.ts
```diff
@@ -151,17 +154,34 @@ export const useIngestStore = defineStore("ingest", {
...
+        if (e instanceof ResponseError) {
+          const body = await e.response.text();
+          const err = api.ModelErrorFromJSON(JSON.parse(body));
+          console.error("API error response", e.response.status, err.message);
+          throw new Error(err.message);
+        } else {
+          console.error("Unknown error", e);
+          throw new Error("Unknown error when fetching packages");
+        }
+      }
```
jraddaoui: Instead of a try/catch, I'd prefer `.then()` and `.catch()` over the response promise.

jraddaoui (approved):
LGTM!

dashboard/src/stores/ingest.ts
djjuhasz: Yep.

dashboard/src/stores/ingest.ts
djjuhasz: Okay, it took me a few tries to get this working, but it's done. :sweat_smile:

jraddaoui: Nice, thanks! Do we need to return?

djjuhasz: Yes, I needed to return the promises, that was my problem. :facepalm:

PR #1143, opened 2025-03-07
Move ent filter and parse range

jraddaoui:
A couple of commits to reduce code duplication when we add the AIP list method to the storage service API.

- Move ent filter to its own package.
- Move parse range function to timerange package.

Refs #1117.


internal/timerange/timerange.go
```diff
@@ -35,3 +35,43 @@ func (r Range) IsZero() bool {
...
 	return r.Start == r.End
 }
+
+// Parse returns a new Range with the given Start and End strings.
+// Parse will return a nil Range if both strings are nil, if only
+// Start is nil it will use an arbitrary time far in the past, and
+// if only End is nil it will use the current time. Parse will return
+// an error if the End time is before the Start time or if the strings
+// cannot be parsed as RFC3339 time format.
+func Parse(start, end *string) (*Range, error) {
```
djjuhasz, line 45: I love the move of time range parsing to the timerange package. :+1:

internal/timerange/timerange.go
```diff
@@ -35,3 +35,43 @@ func (r Range) IsZero() bool {
...
+		s = time.Date(1, 1, 1, 0, 0, 0, 0, time.UTC)
+	} else {
+		s, err = time.Parse(time.RFC3339, *start)
+		if err != nil {
+			return nil, errors.New("time range: cannot parse start time")
+		}
+	}
+
+	if end == nil {
+		e = time.Now()
```
djjuhasz, line 63: I wonder if we should use a time far in the future (e.g. '9999-12-31') instead.  There may be use cases for timerange where the times aren't limited to the past.

djjuhasz (approved):
Love it @jraddaoui. :)  I just had one question/suggestion about how we parse the end date of a time range.

internal/timerange/timerange.go
jraddaoui: Good point, but it's hard for me to see a case where we need to filter by future dates in this type of application. I may be proven wrong pretty quick, but lets do that change when there is a need.

I think it's okay to have a `timerange` without start or end date, and consider that case when applying the filter, better than setting a specific time far in the future or in the past, but that will require to change the current implementation a bit.

PR #1144, opened 2025-03-07
Fix ESLint errors

djjuhasz:
- Exclude dependencies and generated files from ESLint analysis,
- Fix ESLint warnings and errors
- Add ESLint check to GH Actions CI

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -9,8 +9,9 @@
     "build-only": "vite build",
     "type-check": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
     "preview": "vite preview",
-    "lint": "prettier --check .",
+    "lint": "eslint .",
     "format": "prettier --write .",
+    "format-check": "prettier --check .",
     "test": "vitest run",
     "coverage": "vitest run --coverage",
     "deps": "ncu --format=group --interactive"
--- a/.github/workflows/dashboard.yml
+++ b/.github/workflows/dashboard.yml
@@ -12,11 +12,11 @@ jobs:
       - uses: actions/setup-node@v4.0.4
         with:
           node-version: 22
-      - run: npm ci
+      - run: npm clean-install
         working-directory: ./dashboard
       - run: npm run test
         working-directory: ./dashboard
       - run: npm run lint
         working-directory: ./dashboard
-      - run: npm run coverage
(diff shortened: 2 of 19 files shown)


.github/workflows/lint.yml
```diff
@@ -62,3 +62,15 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Install Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: "22"
+      - name: Run eslint
+        working-directory: ./dashboard
+        run: npm install && npm run lint
```
jraddaoui: There is a dashboard workflow that already does this:

https://github.com/artefactual-sdps/enduro/blob/main/.github/workflows/dashboard.yml

dashboard/package.json
```diff
@@ -9,7 +9,7 @@
     "build-only": "vite build",
     "type-check": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
     "preview": "vite preview",
-    "lint": "prettier --check .",
+    "lint": "eslint .",
```
jraddaoui, line 12: Is Prettier ran through ESLint? We want both, right?

jraddaoui (changes requested):
Thanks for looking into this @djjuhasz!

.github/workflows/lint.yml
djjuhasz: Ah, I didn't realize there was a separate dashboard workflow.  I think it was running prettier though, so I'll have to add eslint to the dashboard actions.

dashboard/package.json
djjuhasz: Prettier is called with `npm run format` (the line below lint). I think this is correct - prettier is a code formatter, eslint is a code linter.

jraddaoui: Yeah, sounds good to me. We need to add `format` to the dashboard workflow then as lint is already there. But it won't be the same using `prettier --write .`; maybe add a new `format-check` that runs `prettier --check .`, and run that one in CI.

djjuhasz: I'll add a "format-check" npm script that runs `prettier --check` for use in our CI.

jraddaoui (approved):
Thanks!

PR #1145, opened 2025-03-10
Generate storage enums with abice/go-enum

jraddaoui:
Refs #1117.


djjuhasz:
Adding `internal/storage/types` to the [Codecov ignore list](https://github.com/artefactual-sdps/enduro/blob/main/.codecov.yml#L21) should bring the test coverage up to the minimum threshold (or close).

jraddaoui:
But there is an existing test in that package, I'll create a new enums package to fully ignore it in the tests.

djjuhasz:
@jraddaoui other than the code coverage issue, this looks good to me.

jraddaoui:
Still complaining about the diff, I'll add `[skip-codecov]` before merging.

PR #1150, opened 2025-03-11
Add list AIPs endpoint to storage service API

jraddaoui:
- Add endpoint and scope to API design and docs.
- Generate goa code and dashboard client.
- Add `ListAips` implementation in storage service.
- Update ListAIPs in persistence, to consider the request payload and
  return filtered, ordered and paginated results.
- Update storage service clients and mocks.

Other:
- Ignore coverage and generated API client in ESLint.

Refs #1117.


internal/entfilter/filter.go
```diff
@@ -7,7 +7,8 @@ import (
 	"github.com/google/uuid"
 	"golang.org/x/exp/maps"
 
-	"github.com/artefactual-sdps/enduro/internal/enums"
+	ingest_enums "github.com/artefactual-sdps/enduro/internal/enums"
```
jraddaoui, line 10: At some point I'd like to have a similar directory structure for the ingest and storage packages and sub-packages.

internal/entfilter/filter.go
```diff
@@ -184,9 +185,11 @@ func validPtrValue(ptr any) bool {
 	}
 
 	switch t := ptr.(type) {
-	case *enums.SIPStatus:
+	case *storage_enums.AIPStatus:
 		return t != nil && t.IsValid()
-	case *enums.PreprocessingTaskOutcome:
+	case *ingest_enums.SIPStatus:
+		return t != nil && t.IsValid()
+	case *ingest_enums.PreprocessingTaskOutcome:
```
jraddaoui, line 192: At some point we may want to remove this type of checks from the now general purpose `entfilter` package, I see it as a good candidate for our `go-tools` repository in the future.

internal/storage/service.go
```diff
@@ -238,6 +238,13 @@ func (s *serviceImpl) ListLocations(
 	return s.storagePersistence.ListLocations(ctx)
 }
 
+func (s *serviceImpl) ListAips(
+	ctx context.Context,
+	payload *goastorage.ListAipsPayload,
+) (*goastorage.AIPs, error) {
+	return s.storagePersistence.ListAIPs(ctx, payload)
+}
+
```
jraddaoui, line 247: This service works a bit different than the ingest service. In here, we trust more on the generated Goa types and even return those errors from the persistence layer, so this method just goes directly to the persistence layer.

internal/storage/persistence/ent/client/client.go
```diff
@@ -65,15 +68,59 @@ func (c *Client) CreateAIP(ctx context.Context, goaaip *goastorage.AIP) (*goasto
...
+	qf.Contains(aip.FieldName, payload.Name)
+	qf.Equals(aip.FieldStatus, status)
+	qf.AddDateRange(aip.FieldCreatedAt, createdAt)
+	qf.OrderBy(entfilter.NewSort().AddCol("id", true))
+	qf.Page(ref.DerefZero(payload.Limit), ref.DerefZero(payload.Offset))
+	page, whole := qf.Apply()
+
+	res, err := page.All(ctx)
+	if err != nil {
+		return nil, goastorage.MakeNotAvailable(errors.New("cannot perform operation"))
```
jraddaoui, line 102: Also, it looks like we are returning `409` instead of `500` errors across the entire API, not sure if there was a reason for that, but we may want to consider changing that at some point. For now, I kept the same behavior in this endpoint.

internal/storage/service.go
jraddaoui: Which raises the question, why do we need an ent client and a service when the service could be the ent client at the same time. If we impose that the interactions with storage are done through the service (not directly going to the persistence layer), I think that could simplify the current implementation.

djjuhasz: Yes, I think it's a mistake to tightly couple the API layer to the persistence implementation (ent).

internal/storage/persistence/ent/client/client.go
djjuhasz: My impression from working with the Goa errors for issue #1141 is that the error handling code was "roughed in" but wasn't tested to make sure the HTTP response codes were consistent with the error that produced them.  I think part of the problem is that Goa error handling is not well documented, and there is mixing of different layers of the application --- like in this case where an API error (`goastorage.MakeNotAvailable()`) is being created in persistence layer.

internal/storage/service.go
djjuhasz: Ideally all of the parameters and returns from this service layer should be agnostic to the API implementation (i.e. not accepting `*goastorage.ListAipsPayload` and returning `*goastorage.AIPs`) and the persistence client implementation, like we've done in the ingest service.  Adapters should be used in between layers to convert types.

internal/storage/service_test.go
```diff
@@ -569,6 +569,67 @@ func TestServiceList(t *testing.T) {
...
+				Offset: 0,
+				Total:  1,
+			},
+		}
+
+		attrs.persistenceMock.
+			EXPECT().
+			ListAIPs(ctx, payload).
+			Return(aips, nil).
+			Times(1)
```
djjuhasz: `Times(1)` is the default, it doesn't need to be set explicitly.

internal/workflow/activities/upload_test.go
```diff
@@ -19,6 +19,7 @@ import (
 // StorageService implements goastorage.Service.
 type StorageService struct {
 	JWTAuthHandler          func(ctx context.Context, token string, scheme *security.JWTScheme) (ctx2 context.Context, err error)
+	ListAipsHandler         func(ctx context.Context, req *goastorage.ListAipsPayload) (res *goastorage.AIPs, err error)
```
djjuhasz: Wow, maintaining this mock `StorageService` struct seems like a hassle, why not use mockgen to generate a mock from the `goastorage.Service`?

djjuhasz (approved):
Thanks @jraddaoui.  I've made a few comments on things I think we should improve in the future, but I think for now this is a good iteration.

internal/storage/service.go
jraddaoui: I agree, but I also remember Sevein raising good points about this approach (API first, etc.), I'd like to check with him first before changing the entire service.

internal/storage/persistence/ent/client/client.go
jraddaoui: But there is a `StatusInternalServerError` error type available, that I think fits better in these cases.

internal/workflow/activities/upload_test.go
jraddaoui: I thought the same, and we already have a mock for that service, I'll give it a try ;)

internal/storage/service_test.go
jraddaoui: Copying without looking, there was A LOT of `.Times(1)` in these test :grin:

internal/storage/persistence/ent/client/client.go
djjuhasz: I agree that a 500 HTTP error response is better here, I just want to make sure we aren't returning 500 error codes in cases where it's not appropriate (e.g. when a invalid request parameter value is sent).

internal/storage/service.go
djjuhasz: Oh man, I hope "API first" doesn't imply that we should be tightly coupling our application code to the API implementation. :grimacing:

internal/workflow/activities/upload_test.go
jraddaoui: It was the client and I needed to update the activity to use the `Client` interface, which also required an update to reflect the new method.

PR #1153, opened 2025-03-13
Make AIP move "failed" variable a constant

djjuhasz:


dashboard/src/components/AipLocationCard.vue
```diff
@@ -9,11 +9,9 @@ import { useAuthStore } from "@/stores/auth";
 
 const authStore = useAuthStore();
 const aipStore = useAipStore();
-
-let failed = ref<boolean>(false);
+const failed = ref(false);
 
 const choose = async () => {
-  failed.value = false;
```
jraddaoui, line 16: If it failed once, I think we want to set it to false again before the next try.

djjuhasz: Yeah, good point. It seems a bit odd to me to hide the failure message when the "choose storage location" is clicked, rather then based on the outcome of the move but that's a UI detail we can adjust later if we want to.

jraddaoui (approved):
:ok_hand:

PR #1154, opened 2025-03-18
Rename preservation action/task to workflow/task

jraddaoui:
Replace "Preservation action" by "Workflow" and "Preservation Task"
by only "Task" across all layers of the application.

- Update API design and generated code.
- Update DB schema, generated code and migrations.
- Update services and Ent client.
- Rename data types and enums.
- Rename local activities.
- Update dashboard.

Also, update ingest SIP and Workflow:

- Remove workflow, run and location IDs from SIP.
- Rename workflow ID to Temporal ID in Workflow.

Refs #1117.


jraddaoui:
@sevein, I understand you will be reviewing some PRs for us this week. Sorry you have to look at this pain of PR, there is a lot of code changes involved but very little functional changes. Also, the related issue is quite big and has a lot of background, so I'll try to summarize here what we are trying to do.

In general, we are trying to have a clearer separation between the ingest and storage concerns, this will allow us to have a better distinction between SIP and AIP statuses and implement an AIP delete operation only in the storage domain, even if that will require a lot of duplication between ingest and storage.

In this PR, I'm renaming the preservation actions to workflows and the preservation tasks to only tasks in the ingest domain, before they get duplicated in storage. It also removes/renames some unnecessary/duplicated columns in the SIP and workflow DB schema. Most of the changes are generated code of renaming fields, functions, etc. I think you should take special attention to the API design, DB schema and migrations.

In upcoming PRs, we'll try to do (not exactly in this order):

- Stop using RAW SQL in the ingest domain in favor of the Ent client.
- Use UUIDs as resource identifiers in ingest (as it's for AIPs and Locations in storage)
- Duplicate Workflow and Task in the storage domain.
- Update move operation so it only involves the storage domain.
- Record Workflows/Tasks in the storage domain for the existing AIP upload and move.
- Add AIP deletion request/review and workflow.

internal/ingest/goa.go
```diff
@@ -190,54 +190,57 @@ func (w *goaWrapper) ListSipPreservationActions(
...
 
 func (w *goaWrapper) ConfirmSip(ctx context.Context, payload *goaingest.ConfirmSipPayload) error {
-	goasip, err := w.ShowSip(ctx, &goaingest.ShowSipPayload{ID: payload.ID})
+	goaworkflows, err := w.ListSipWorkflows(ctx, &goaingest.ListSipWorkflowsPayload{ID: payload.ID})
 	if err != nil {
 		return err
 	}
+	if goaworkflows == nil || len(goaworkflows.Workflows) == 0 || len(goaworkflows.Workflows) > 1 {
+		return goaingest.MakeNotAvailable(errors.New("cannot perform operation"))
+	}
```
jraddaoui, line 237: There will only be one workflow per SIP right now, specially being in review. We should improve the errors though, but that's something that needs to happen all over.

internal/ingest/workflow.go
```diff
@@ -5,111 +5,109 @@ import (
 	"fmt"
```
jraddaoui, line 5: This file was called `preservation_action.go`, which I moved to `workflow.go`, but we already had one file called like that, which I moved to `workflows.go`. Sorry about that, the diff is not that much in the end, just what you are seeing all over.

sevein (approved):
I didn't spot anything suspicious. Looks good to me!

PR #1155, opened 2025-03-21
Add workflows and tasks to storage service

jraddaoui:
- Add Workflows and Tasks to storage service:
  - Create enums for Workflow type, Workflow status and Task status.
  - Create local types for Workflow and Task.
  - Create Ent schema for Workflow and Task, and update AIP schema.
  - Generate Ent code and DB migrations.
  - Add methods to create and update Workflows and Tasks.
- Record Workflow and Tasks in the move AIP workflow:
  - Add local activities to create and update Workflows and Tasks.
  - Add helpers to initialize and complete Workflows and Tasks.
  - Create helper to obtain local activity options and reduce duplication.
  - Record move AIP workflow and copy/delete AIP tasks.
  - Update move AIP workflow tests.
- Add list AIP workflows endpoint to storage API:
  - Add new method, types and scope to API design.
  - Generate Goa code.
  - Add endpoint to IAC docs.
  - Add implementation in storage service and Ent client.
  - Add convert methods to transform from Ent to Goa types.
  - Update storage client interface and clients.
- Show AIP workflows and tasks in dashboard:
  - Update API client generated code.
  - Add current workflows to AIP store.
  - Consider AIP workflow types in getWorkflowLabel.
  - Show workflows and tasks in AIP index page.
  - Allow storage AIP workflows and tasks in WorkflowCollapse.
  - Allow storage task in Task component.
  - Allow storage workflow and task statuses in StatusBadge.
  - Restore move button in AIP location card.

Refs #1117.


jraddaoui:
@sevein (if you have some time before David gets back) or @djjuhasz (when you are back), this is ready for an initial code review.

There is a few things left to do:

- Improve errors in the persistence layer, the storage service and the API.
- Add tests for the new functionality in the Ent client, storage service and local activities.
- Improve code comments.
- Remove anything related to move AIP from the ingest domain.

There are also some big topics that I'd like to discuss at some point that relate to these changes:

- Usage of different types in the application (Goa, Ent, custom).
- Usage of local activities.
- Usage of DB IDs vs UUIDs.

I'll add the tests and try to improve the errors in the new changes in this PR, but everything else will probably have to wait for bigger discussions, or at least another PR.

jraddaoui:
Also, it may be easier to review commit by commit, each commit subject and body are  in the PR description, in order.

internal/storage/persistence/persistence.go
```diff
@@ -23,6 +23,7 @@ type Storage interface {
 	AIPDBID(ctx context.Context, aipID uuid.UUID) (int, error)
 	UpdateAIPStatus(ctx context.Context, aipID uuid.UUID, status enums.AIPStatus) error
 	UpdateAIPLocationID(ctx context.Context, aipID, locationID uuid.UUID) error
+	AIPWorkflows(ctx context.Context, aipUUID uuid.UUID) (goastorage.AIPWorkflowCollection, error)
```
sevein, line 25: For consistency, this method looks like it should be named `ReadAIPWorkflows` right?

internal/storage/persistence/persistence.go
```diff
@@ -10,13 +10,20 @@ import (
...
+	WorkflowUpdater func(*types.Workflow) (*types.Workflow, error)
+	TaskUpdater     func(*types.Task) (*types.Task, error)
+)
+
 type Storage interface {
 	// AIP.
 	CreateAIP(ctx context.Context, aip *goastorage.AIP) (*goastorage.AIP, error)
 	ListAIPs(ctx context.Context, payload *goastorage.ListAipsPayload) (*goastorage.AIPs, error)
 	ReadAIP(ctx context.Context, aipID uuid.UUID) (*goastorage.AIP, error)
+	AIPDBID(ctx context.Context, aipID uuid.UUID) (int, error)
```
sevein: Should this method be named `ReadAIPDBID`?

internal/storage/persistence/persistence.go
```diff
@@ -27,4 +34,12 @@ type Storage interface {
 	ListLocations(ctx context.Context) (goastorage.LocationCollection, error)
 	ReadLocation(ctx context.Context, locationID uuid.UUID) (*goastorage.Location, error)
 	LocationAIPs(ctx context.Context, locationID uuid.UUID) (goastorage.AIPCollection, error)
+
+	// Workflow.
+	CreateWorkflow(context.Context, *types.Workflow) error
```
sevein, line 38: I'm not sure I have much to add to the "DB IDs vs UUIDs" discussion, but here's a suggestion. Would it make sense to update this method to look up the AIP's database ID directly from its UUID? That would let you do the insert in a more atomic way (using a transaction or an `INSERT INTO ... SELECT ...` if supported by entgo), avoiding potential race conditions where the AIP record might change or be deleted between the lookup and the insert - perhaps that kind of concurrency is not a concern here? Beyond ensuring data consistency, that would also save you from having to implement `AIPDBID` and the extra roundtrip with the database.

sevein (approved):
LGTM - just a few minor observations.

internal/storage/persistence/persistence.go
jraddaoui: Not sure, we use Read for single resources so I thought List would work better, like it is in the API/Goa service. Then I saw `LocationAIPs` below, so I went with that.

internal/storage/persistence/persistence.go
jraddaoui: Probably, but I'll remove it based on your following comment.

internal/storage/persistence/persistence.go
jraddaoui: Thanks, not sure why I didn't do it that way, seems so obvious now that you say it ;)

internal/storage/local_activities.go
```diff
@@ -45,3 +47,103 @@ func DeleteFromLocationLocalActivity(
...
+	Type       enums.WorkflowType
+}
+
+func CreateWorkflowLocalActivity(
+	ctx context.Context,
+	storagesvc Service,
+	params *CreateWorkflowLocalActivityParams,
+) (int, error) {
+	w := &types.Workflow{
+		UUID:       uuid.New(),
```
djjuhasz, line 63: This non-determinism is going to make unit testing difficult. :(

internal/storage/persistence/ent/client/convert.go
```diff
@@ -0,0 +1,121 @@
...
+	p := &goastorage.AIP{
+		Name:      a.Name,
+		UUID:      a.AipID,
+		Status:    a.Status.String(),
+		ObjectKey: a.ObjectKey,
+		CreatedAt: a.CreatedAt.Format(time.RFC3339),
+	}
+
+	// TODO: should we use UUID as the foreign key?
+	l, err := a.QueryLocation().Only(ctx)
```
djjuhasz, line 69: Seems weird to have an ent query in the adapter. :thinking:

internal/storage/persistence/ent/client/convert.go
```diff
@@ -0,0 +1,121 @@
...
+		w.StartedAt = ref.New(dbw.StartedAt.Format(time.RFC3339))
+	}
+	if !dbw.CompletedAt.IsZero() {
+		w.CompletedAt = ref.New(dbw.StartedAt.Format(time.RFC3339))
+	}
+
+	if len(dbw.Edges.Tasks) > 0 {
+		for _, dbt := range dbw.Edges.Tasks {
+			if dbt != nil {
+				w.Tasks = append(w.Tasks, taskAsGoa(dbt))
```
djjuhasz, line 95: Hmm, this could make for a very large struct, even when the task data isn't relevant.  I'd prefer to see a separate method for getting the workflow task data.

djjuhasz (approved):
Looks good to me @jraddaoui. I made a few comments, but I don't think they need to be addressed in this PR.

internal/storage/local_activities.go
jraddaoui: I usually just ignore the UUID fields, like I've done in some of the tests I just added.

internal/storage/persistence/ent/client/convert.go
jraddaoui: I agree, I'd use the Edges if available, but I just moved this function.

internal/storage/persistence/ent/client/convert.go
jraddaoui: I'm replicating what we have on the ingest domain to be able to reuse the same components without a lot of changes. We ca revisit for both domains in the future.

Issue #1076, opened 2024-11-22
Problem: there is no way to delete an AIP


sallain:
**Is your feature request related to a problem? Please describe.**

Once an AIP has been created, there is no way to delete it through Enduro. It is possible to delete the AIP in Archivematica or the Storage Service, but this means that 1) users have to leave Enduro, and 2) the deletion is not reflected in Enduro.

**Describe the solution you'd like**

While viewing a single AIP in Enduro, a user with sufficient permission should be given the option to delete the AIP through a UI button.

Similar to Archivematica, deletion should be a two-step process:

1. A user clicks a button to request that an AIP be deleted, and is prompted to provide a reason (text field).
2. A user approves or denies the deletion request, and is prompted to provide a reason (perhaps optional for approval but required for denial?)

Once the deletion request is approved, a API call should be sent to the Storage Service to delete the AIP. The Storage Service should be set to auto-approve deletion requests, so that the user doesn't need to repeat the deletion approval there. Enduro should then receive a confirmation that the package has been deleted (NOTE: based on the current SS API, this may require Enduro to poll the Storage Service for the package's status).

Throughout this workflow, all actions - deletion request, deletion approval/denial, deletion confirmation - should be recorded in an audit log that persists indefinitely (or as long as the user needs it to).

See Miro for the [full proposed workflow](https://miro.com/app/board/uXjVMlYiVgs=/?moveToWidget=3458764601125357317&cot=14), including status updates for the package.

**Describe alternatives you've considered**

We've gone through many iterations of this workflow, but this seems to be the simplest flow without compromising auditability of the deletion action.

**Additional context**

![Swiss Federal Archives (SFA) - Proposed DPS deletion workflow](https://github.com/user-attachments/assets/b5e59d01-e41c-44ab-90f4-167eb7cc4709)

Copy of the Miro diagram as of 22 November 2024.

sallain:
As a first iteration of this feature, the team agreed that we would implement the following:

* UI elements for happy path (deletion button, reason entry box, approval button, etc.)
* New status (`Pending`) for AIPs where a deletion request has been made
* API call to Storage Service to delete AIP
* Some way to receive confirmation from the Storage Service that the AIP was deleted
* New status (`Deleted`) for AIPs that have been successfully deleted
* Audit log to capture who/why/when

The first iteration does not need to include:

* Denial button and associated workflow
* Email or UI notifications
* Reports

sallain:
[...]

PR #1157, opened 2025-03-25
Add AIP deletion workflow

jraddaoui:
Add a two steps process to trigger the deletion of an AIP. Create AIP
deletion-request and deletion-review endpoints in the storage service
API. The request will start a Temporal workflow using a deterministic
workflow ID based on the AIP UUID, this ID can only be reused when the
workflow finishes in a failed status (error, canceled, terminated) but
not when a previous workflow succeeds or is currently running. The
workflow will stop and wait for a review signal that is sent from the
API review workflow. The request requires a reason and the review can
not be performed by the requester. Both endpoints require their own
ABAC scopes.

The deletion workflow will look at the AIP and related location info and
it will work with MinIO and AMSS source locations. For the AMSS deletion
there are a few requests involved, to get the origin pipeline from the
AIP details, which is needed to request the deletion, and then it will
poll for status until it changes from "DEL_REQ" to "UPLOADED" or
"DELETED". Approval in the AMSS must be done manually or with a custom
script, since there is no way to approve the deletion in the API yet. 

Add DeletionRequest schema and DB table to hold information about the
request/review and new AIP and Workflow statuses. Add persistence
methods to create and update deletion requests and to get the currently
pending request.

Add delete button, request dialog with reason textarea and review alerts
to the dashboard. Allow task card note updates and render emails in them
as mailto links. Add a poll process after request and review to update
the current status until we implement websockets in the storage service.
Change status filters and time dropdown label in the AIPs browse page
and switch status badge background colors between "queued"
and "processing"/"in progress" statuses. Improve buttons behavior in
the location card and remove move button for now.

Refs #1076.


internal/storage/persistence/ent/schema/deletion_request.go
```diff
@@ -0,0 +1,90 @@
...
+// Fields of the DeletionRequest.
+func (DeletionRequest) Fields() []ent.Field {
+	return []ent.Field{
+		field.UUID("uuid", uuid.UUID{}).
+			Unique(),
+		field.String("requester").
+			Annotations(entsql.Annotation{
+				Size: 1024,
+			}),
+		field.String("requester_iss").
```
djjuhasz, line 39: What's a "requester_iss"?

internal/storage/persistence/ent/schema/deletion_request.go
```diff
@@ -0,0 +1,90 @@
...
+			Annotations(entsql.Annotation{
+				Size: 1024,
+			}),
+		field.String("requester_sub").
```
djjuhasz, line 44: And what's a "requester_sub"?

internal/storage/persistence/migrations/20250328171353_add_status_enums.down.sql
```diff
@@ -0,0 +1,4 @@
+-- reverse: modify "workflow" table
+ALTER TABLE `workflow` MODIFY COLUMN `status` enum('unspecified','in progress','done','error','queued','pending') NOT NULL;
+-- reverse: modify "aip" table
+ALTER TABLE `aip` MODIFY COLUMN `status` enum('unspecified','in_review','rejected','stored','moving') NOT NULL;
```
djjuhasz, line 4: I doubt we'll ever have to reverse this schema migration, but it seems like just removing the three new aip status values from the enum would just break any existing data, no?

internal/api/design/storage.go
```diff
@@ -265,10 +265,8 @@ var _ = Service("storage", func() {
 		Payload(func() {
 			AttributeUUID("uuid", "Identifier of AIP")
 			Token("token", String)
-			Attribute("decision", String, func() {
-				Enum("approve", "reject")
-			})
-			Required("uuid", "decision")
+			Attribute("approved", Boolean)
+			Required("uuid", "approved")
```
djjuhasz, line 269: I think we want to allow including a reason for rejecting the deletion request too?

internal/storage/local_activities.go
```diff
@@ -75,6 +75,24 @@ func CreateWorkflowLocalActivity(
...
+	Status enums.WorkflowStatus
+}
+
+func UpdateWorkflowStatusLocalActivity(
+	ctx context.Context,
+	storagesvc Service,
+	params *UpdateWorkflowStatusLocalActivityParams,
+) error {
+	_, err := storagesvc.UpdateWorkflow(ctx, params.DBID, func(w *types.Workflow) (*types.Workflow, error) {
+		w.Status = params.Status
```
djjuhasz, line 125: I think this will panic if `params` is nil.

internal/storage/local_activities.go
```diff
@@ -75,6 +75,24 @@ func CreateWorkflowLocalActivity(
...
+	DBID   int
+	Status enums.WorkflowStatus
+}
+
+func UpdateWorkflowStatusLocalActivity(
+	ctx context.Context,
+	storagesvc Service,
+	params *UpdateWorkflowStatusLocalActivityParams,
+) error {
+	_, err := storagesvc.UpdateWorkflow(ctx, params.DBID, func(w *types.Workflow) (*types.Workflow, error) {
```
djjuhasz, line 124: Why return `*types.Workflow` when you are mutating the original value through a pointer?

internal/storage/local_activities.go
```diff
@@ -111,7 +127,7 @@ func CreateTaskLocalActivity(
 	t := &types.Task{
 		UUID:         uuid.New(),
 		Name:         params.Name,
```
djjuhasz, line 165: Again, if `params` is nil this is going to panic.

dashboard/src/components/AipDeletionRequestDialog.vue
```diff
@@ -0,0 +1,77 @@
...
+const el = ref<HTMLElement | null>(null);
+const modal = ref<Modal | null>(null);
+
+onMounted(() => {
+  if (!el.value) return;
+  modal.value = new Modal(el.value);
+  modal.value.show();
+});
+
+let reason: string = "";
```
djjuhasz: Doesn't this need to be a `ref()` to use it as a `v-model` value?

dashboard/src/components/AipDeletionReviewAlert.vue
```diff
@@ -0,0 +1,34 @@
...
+
+import { useAipStore } from "@/stores/aip";
+
+const aipStore = useAipStore();
+
+const { note } = defineProps<{
+  note: string;
+}>();
+
+const body = computed(() => note);
```
djjuhasz: Is this necessary? Can't you just call `{{ note }}` directly in the template?

djjuhasz (commented):
Your latest commits LGTM @jraddaoui, I just had a few questions and comments.

internal/storage/persistence/ent/schema/deletion_request.go
jraddaoui: It's the OIDC issuer/provider identifier:

https://openid.net/specs/openid-connect-core-1_0.html#Terminology
https://openid.net/specs/openid-connect-core-1_0.html#IDToken
https://openid.net/specs/openid-connect-core-1_0.html#IssuerIdentifier
https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability

internal/storage/persistence/ent/schema/deletion_request.go
jraddaoui: It's the OIDC subject/user identifier at the issuer (links above).

internal/storage/persistence/migrations/20250328171353_add_status_enums.down.sql
jraddaoui: Good point, and I'd not know what to do in that case, delete the affected rows? What would you do with an AIP in a deleted status. I think we should stop creating/worrying about down migrations and only trust backups for that regard.

internal/api/design/storage.go
jraddaoui: Yeah, I thought the same, but Dan and Sara said it was not necessary when we created the wireframes at sync time. Probably a follow-up after the initial implementation.

internal/storage/local_activities.go
jraddaoui: :thinking: you are right. But we do it in A LOT of activities, we should revisit at some point in general. Why is a pointer needed? What can happen within Temporal?

internal/storage/local_activities.go
jraddaoui: Good point. The way we call it requires to return the mutated value:

https://github.com/artefactual-sdps/enduro/blob/main/internal/storage/persistence/ent/client/workflow.go#L62

But that shouldn't be needed if we do the conversion before calling the updater. I just copied this pattern from other updater functions:

https://github.com/artefactual-sdps/enduro/blob/main/internal/persistence/persistence.go#L22
https://github.com/artefactual-sdps/enduro/blob/main/internal/persistence/ent/client/sip.go#L76

internal/storage/local_activities.go
jraddaoui: Same.

dashboard/src/components/AipDeletionRequestDialog.vue
jraddaoui: Right, I added it without thinking and it worked, didn't know Vue3 transforms it internally, thanks!

dashboard/src/components/AipDeletionReviewAlert.vue
jraddaoui: That's what I tried initially and it din't work, not sure what I did wrong, it works now ;)

jraddaoui:
Thanks @djjuhasz!

This is ready for another (final?) code review, a bunch of commits since you last reviewed. There is some back a forward but I still think it makes more sense to go commit by commit, I'll add comments in some of them where you shouldn't worry about something that changes later.

Check [this comment](https://github.com/artefactual-sdps/enduro/issues/1076#issuecomment-2766523312) first. There are many things left to do, but this is what I could get before the release. I'll address any important feedback you may find and leave the rest to work on it over the week after the release. Alongside three mayor things that are lacking in the PR:

- There is almost zero tests for these changes.
- It requires better validation, error reporting in some interfaces.
- It lacks a lot comments/docs in code.
- Doesn't report request/review errors in the dashboard.

---

It also touches various important topics (like the previous PR) that I think we should discuss as a team and address them as soon as we are able:

-  Types (Goa, Ent, local)

We should normalize on each service and in between services. Currently, it's quite a mess mixing types on the same interface (persistence, service, Goa) and across all of them.

- Duplicated service interfaces:

We should consider stop duplicating those interfaces just to avoid sharing the persistence service.

- Errors (API, persistence):

Based on the decisions for the two points above, we also need to normalize how errors are handled in the API (no logs, no 500, very little info in some of the not available errors).

- Database IDs vs UUID:

Again some inconsistencies, I tried to keep using UUIDs in the storage service (except for internal functions), but the ingest service still uses DB IDs as resource identifiers.

- Usage of local activities:

As you pointed out a while ago, we may want to stop using local activities, even for MySQL interactions. For now, I just followed the existing pattern.

- AMSS location type: 

And the AMSS integration in general. It's so different, specially in the handover between persistence and storage, where Enduro is not involved. Not sure if it's even possible, but we need to find a better way to support the two existing location sources in the same manner.

- AIP integrity:

Now that we have two types of workflows in the storage domain for the AIPs, we need to be more careful about concurrency and integrity. I'm planning to revisit what remains of the move operation in the storage service and delete the unused bits from the ingest service.

- AIP creation and review: 

With the current separation between ingest and storage, we probably want to change how this works right now, avoiding the involvement of the storage service until the reviewed AIP is stored, and removing the "in review" and "rejected" status from the storage domain. 

- AIP/SIP/Workflow/Task statuses:

Part of these changes are implement here but there is a lot more to change in upcoming issues/PRs.

----

I'll create issues for all of them so we can take look when we have the time. Some of them are worth looking at the same time, but they are big topics.

enduro.toml
```diff
@@ -157,6 +157,10 @@ user = "" # Secret: set with env var ENDURO_AM_USER.
 apiKey = "" # Secret: set with env var ENDURO_AM_APIKEY.
 processingConfig = "automated"
 
+# pipelineUUID is the UUID of the AM pipeline registered in the AMSS. Needed
+# for AIP deletion requests sent to the AMSS.
+pipelineUUID = ""
+
```
jraddaoui: I'll remove this commit/changes and get the pipeline UUID from the AMSS API.

internal/storage/workflows/delete.go
```diff
@@ -145,7 +170,70 @@ func (w *StorageDeleteWorkflow) Execute(
...
+	// Complete delete AIP task.
+	taskStatus := enums.TaskStatusDone
+	source := strings.ToUpper(locationInfo.Source.String())
+	taskNote = fmt.Sprintf("AIP deleted from %s source location", source)
+	if !deleted {
+		taskNote = fmt.Sprintf("AIP request rejected in %s source location", source)
+	} else if err != nil {
+		taskStatus = enums.TaskStatusError
+		taskNote = fmt.Sprintf("Failed to delete AIP:\n%v", err)
+	}
```
jraddaoui: This is fixed later, giving priority to the error.

internal/storage/workflows/delete.go
```diff
@@ -291,7 +291,7 @@ func (w *StorageDeleteWorkflow) deleteAIPFromAMSSLocation(
 
 	var re activities.DeleteFromAMSSLocationActivityResult
 	activityOpts := temporalsdk_workflow.WithActivityOptions(ctx, temporalsdk_workflow.ActivityOptions{
-		StartToCloseTimeout: time.Hour * 2,
+		StartToCloseTimeout: time.Minute,
```
jraddaoui: I restored this in a later commit.

internal/storage/local_activities.go
djjuhasz: I wonder if we started passing `params` as a pointer because we were passing `results` as a pointer?  I don't think the workflow and  worker share memory, and I can't imagine ever needing to mutate the params at the calller, so I don't know any reason why we'd need to use a pointer. :thinking:

internal/storage/local_activities.go
djjuhasz: Right, we may want to change the pattern down the road, but it can wait until after this PR.

internal/storage/activities/delete_amss.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	resp, err := a.ssclient.Do(req)
+	if err != nil {
+		return "", err
+	}
+	if resp.StatusCode >= 400 {
+		return "", fmt.Errorf("deletion request to AMSS: response code: %d", resp.StatusCode)
```
jraddaoui: I fixed these errors in the latest commit.

enduro.toml
jraddaoui: Done.

internal/storage/local_activities.go
sevein: Looks like it's a common practice in the Go ecosystem (connect-go, grpc-go, etc), my guess is that they do it to avoid potentially costly copies right?

djjuhasz: @sevein my understanding is the parameters are being sent over GRPC from the workflow server to the worker anyway, so I'm not sure what copy costs it would save?

djjuhasz: I guess there's the copy cost when calling `UpdateWorkflowStatusLocalActivity`.  In this case `UpdateWorkflowStatusLocalActivityParams` is only a string and an enum so I'd expect the copy cost to be trivial.  

I guess the potential drawbacks of using a pointer (accidentally mutating the params, panicking if params is nil) may be unlikely enough that it's not worth the time and mental effort to evaluate if a pointer is worthwhile on a case-by-case basis. :shrug:

internal/storage/activities/delete_amss.go
```diff
@@ -0,0 +1,157 @@
...
+
+	ticker := time.NewTicker(time.Second * 10)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-ticker.C:
+			status, err := a.pollStatus(ctx, params)
```
djjuhasz, line 61: The default timeouts for [cleanhttp.DefaultPooledTransport()](https://github.com/hashicorp/go-cleanhttp/blob/master/cleanhttp.go#L26) are up to 40s for connecting (30s for dial, 10s for TLS handshake) and there is no absolute limit on the total request lifecycle, so I think sending a poll request every 10s has the potential to pile up requests, especially since AMSS can be quite slow to respond under load.  I think it would be better to increase the poll interval to a minute or so to avoid sending a new request before the previous one completes.

Alternatively you could set lower timeouts on the client transport (this could be helpful: https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/#client-timeouts) but realistically I don't think we should expect AMSS to respond within 10s under load.

internal/storage/workflows/delete.go
```diff
@@ -161,3 +249,48 @@ func (w *StorageDeleteWorkflow) waitForReview(ctx temporalsdk_workflow.Context)
...
+	if configValue == nil {
+		return false, fmt.Errorf("missing AMSS config")
+	}
+
+	var re activities.DeleteFromAMSSLocationActivityResult
+	activityOpts := temporalsdk_workflow.WithActivityOptions(ctx, temporalsdk_workflow.ActivityOptions{
+		StartToCloseTimeout: time.Hour * 2,
+		HeartbeatTimeout:    time.Second * 20,
+		RetryPolicy: &temporalsdk_temporal.RetryPolicy{
+			InitialInterval:    time.Second,
```
djjuhasz: I think a one second initial interval is too short here. I would expect a transient connection issue with the AMSS to most likely be to load, or maybe a slow network, and I wouldn't expect either condition to resolve within a second.  I'd suggest an initial interval in the 30 - 60 second range.

internal/storage/workflows/delete.go
```diff
@@ -239,6 +196,72 @@ func (w *StorageDeleteWorkflow) Execute(
...
+	// Wait for user review signal.
+	reviewSignal := w.waitForReview(ctx)
+	logger.Info("Received AIP deletion workflow signal", "signal", reviewSignal)
+
+	// Set AIP status to processing.
+	if err := updateAIPStatus(ctx, w.storagesvc, req.AIPID, enums.AIPStatusProcessing); err != nil {
+		return nil, err
+	}
+
+	// Set Workflow status to pending.
```
djjuhasz: Should be "Set Workflow status to in progress."

internal/storage/workflows/delete.go
```diff
@@ -35,54 +35,96 @@ func (w *StorageDeleteWorkflow) Execute(
...
+	// TODO: Delete the AIP from AMSS or MinIO location.
+	aipStatus = enums.AIPStatusDeleted
+
 	return nil
 }
+
+func (w *StorageDeleteWorkflow) waitForReview(ctx temporalsdk_workflow.Context) *storage.DeletionReviewedSignal {
+	var review storage.DeletionReviewedSignal
+	signalChan := temporalsdk_workflow.GetSignalChannel(ctx, storage.DeletionReviewedSignalName)
+	selector := temporalsdk_workflow.NewSelector(ctx)
```
djjuhasz: I've never used Temporal signals before, but it looks like there's a simpler syntax that doesn't involve selectors if you want to block the workflow until the signal returns, e.g.:

```
workflow.GetSignalChannel(ctx, storage.DeletionReviewedSignalName).Receive(ctx, &review)
return &review
```

See: https://docs.temporal.io/develop/go/message-passing#signals

internal/storage/workflows/delete.go
```diff
@@ -35,54 +35,96 @@ func (w *StorageDeleteWorkflow) Execute(
...
-	if err = errors.Join(err, taskErr); err != nil {
+
+	// Set AIP status to pending.
+	if err := updateAIPStatus(ctx, w.storagesvc, req.AIPID, enums.AIPStatusPending); err != nil {
 		return err
 	}
 
-	// TODO: Delete the AIP from AMSS or MinIO location source.
+	// Wait for user review signal.
+	reviewResult := w.waitForReview(ctx)
```
djjuhasz: I think it's possible for `reviewResult` to be empty here (see my comment below), in which case I think we need to set the workflow status to "error" or something similar instead of "rejected".

internal/storage/workflows/delete.go
```diff
@@ -35,54 +35,96 @@ func (w *StorageDeleteWorkflow) Execute(
...
+
 	return nil
 }
+
+func (w *StorageDeleteWorkflow) waitForReview(ctx temporalsdk_workflow.Context) *storage.DeletionReviewedSignal {
+	var review storage.DeletionReviewedSignal
+	signalChan := temporalsdk_workflow.GetSignalChannel(ctx, storage.DeletionReviewedSignalName)
+	selector := temporalsdk_workflow.NewSelector(ctx)
+	selector.AddReceive(signalChan, func(channel temporalsdk_workflow.ReceiveChannel, more bool) {
+		_ = channel.Receive(ctx, &review)
```
djjuhasz: I think it's possible for `Receive()` to return false without setting `review` if the signal channel is closed before a signal is sent.  I think we need to distinguish between a valid "rejected" signal and some error that prevents the signal from being sent.

dashboard/src/components/AipDeletionRequestDialog.vue
```diff
@@ -41,42 +41,48 @@ const request = () => {
...
+              be reviewed before deletion.
+            </p>
+            <div>
+              <label for="reason" class="form-label">Reason:</label>
+              <textarea
+                class="form-control"
+                id="reason"
+                rows="3"
+                v-model="reason"
+                required
```
djjuhasz, line 64: This is cool, I didn't know they'd added client-side validation to the HTML spec. :+1:

djjuhasz (approved):
@jraddaoui It's hard for me to understand the how all the changes in the bigger commits tie together and I don't fully understand the details of the deletion workflow, but I've done my best to review your changes.  

Overall the code looks good to me, within the limitations and restrictions you've explained on the issue ticket. I agree that we need to figure out the issues you've raised about the application architecture and how the various app layers communicate, but that can come later.

I've raised some questions about some of the HTTP communications and the Temporal signals implementation but I didn't spot any critical problems.

internal/storage/activities/delete_amss.go
jraddaoui: Increased the interval to one minute, thanks!

internal/storage/workflows/delete.go
jraddaoui: Increased initial interval to 30 seconds and reduced backoff coefficient to 1.5.

internal/storage/workflows/delete.go
jraddaoui: Cool, I always saw this implementation in our code, I copied it from the API review signal in the processing workflow, maybe we used the selector to add a future for a time limit at some point. Thanks for pointing that out!

internal/storage/workflows/delete.go
jraddaoui: Done, thanks!

jraddaoui:
Thank you for keeping an eye during the development and the final review @djjuhasz! I think I have addressed all your feedback. I should I have updated the PR description before review, not sure if it would have helped to understand the bigger picture while doing the review day by day, but it's added now, I'll squash with that message and merge.

internal/storage/workflows/delete.go
```diff
@@ -0,0 +1,313 @@
...
+	temporalsdk_workflow "go.temporal.io/sdk/workflow"
+
+	goastorage "github.com/artefactual-sdps/enduro/internal/api/gen/storage"
+	"github.com/artefactual-sdps/enduro/internal/storage"
+	"github.com/artefactual-sdps/enduro/internal/storage/activities"
+	"github.com/artefactual-sdps/enduro/internal/storage/enums"
+	"github.com/artefactual-sdps/enduro/internal/storage/types"
+)
+
+type StorageDeleteWorkflow struct {
```
sevein, line 21: A short description that may help readers understand its purpose:

> StorageDeleteWorkflow handles the controlled deletion of AIPs by ensuring that only stored, valid AIPs can be deleted, and only after explicit user approval. It enforces a review and confirmation step before performing potentially destructive operations.

sevein (commented):
I find the number of state transitions in StorageDeleteWorkflow a bit intimidating. I asked ChatGPT to create a [Mermaid flowchart] to illustrate the different states and while it might not be perfect, it does show the various database states we could see. 

I'm genuinely wondering if the application logic might benefit from more atomic state transitions; for example, should we try, in one single transaction, to look up the status of the AIP, update its status to processing, create the workflow record and the review task? Would that make the application simpler overall by removing the need to handle edge cases, like a workflow record existing without a corresponding review task, and maybe even help prevent future race conditions?

[Mermaid flowchart]: https://mermaid.live/edit#pako:eNp9VNtu2zAM_RVBD0MHpJ0dN25nYBvSXNq0uTXuMGzzHgSbSYw6kiHLvQX590q0nHhFt7cj8og6JI-9pbFIgAZ0mYnHeM2kInf9iBPS_R3RUJlzdzQnfchApYKTH0LeG2pE_xjWhWYtgCVIOhqLmGWkG6v0IVXPHy2nt42oSRdKSEjIB5JpGhaDp7RQxbeI7gyvr2t9zxOmgFR0psqCKEHmUsRQFClf2YoDzexJMMw5yEIXAR7DXhtZQCxkYsnDA3kBDyk8kjtW3NvkpUmyLKtTw5LHRppNX5m02OS6-fduj7ame4x281yKB0jqZq6xsFaVHWR9IncgNyk3Uh5TtSYDKYW0tW4OMnHY1RAaj43rUY_r8Y34Uvxr6BMtbU8MRSljsMqmri7UnYRhQLDz-jVbwN6ftjVrkvLR7P80zyyNF2WeC6n0dqungr9amzWnaOs0GptrqXt_hWVsdr0ss3qSt8aIUNkwrDzxxVapV7zQlKGea5a-NFxw1IclSGO5C-24-3oyIc5RlXL_GCbQ8-T4-Cu5QGMj7KF_NSQRnYqIYvC6EfwJhY3iR9NHOECPIhyiAxFeot8QXqG3EI7QR--9MHr7wg3aBOEYDYFwgtuuuGarljx1G3HcY51oNxKDrIA67lXmwMOsskADewdhM4RzXN17wudvhd_iGhEucF0IQ9qiG_09sDTRv5-tSURUrWGjFQUaJrBkZaYiGvGdprJSifCZxzRQsoQWlaJcrWmwZLqFFi3xv9FP2UqyzT6aM_5LiE19RR9psKVPNHDb3onjt70zx3H88zO_47fosw77pyee77jnrud43rnv-rsWfcEK7ZOO47Y7HcfrnDqO99nRN1bSSLdygCcge6Lkigadzu4VNGOc7Q

**Update!** Just saw this already got merged - looks like I was too slow!

jraddaoui:
Thanks @sevein, I'll consider that in a follow up, at least the status check + update should happen in a transaction. everything else are resources created in the workflow, which I think should be okay.

PR #1162, opened 2025-03-28
Use ISO 8601 format for datepicker dates

### LINKED ISSUE #1156: Problem: SIP and AIP time range widgets display date previews in US format
**Describe the bug**

The Enduro user interface currently has two configurable widgets for limiting results based on a time range: 

* An ingest start time filter on the SIP browse page
* A created or stored time filter on the AIP browse page

Both of these widgets include some preset range options, as well as a set of fields to configure custom ranges. As users click into the start and end time custom range fields to configure them, a calendar select widget appears. While configuring the desired date and time for each field, the calendar select widget displays a date/time preview of the configured value in the bottom left of this widget, as shown in the image below: 

Image

This datetime preview is current using the U.S. `mm/dd/yy` format for the date, rather than the ISO 8601 `YYYY-MM-DD` format. 

## To reproduce

1. Log into Enduro
2. Navigate to the SIPs browse page
3. Open the "Started" filter next to the search box, above the SIP results
4. Configure a date
5. Look at the date preview shown in the widget

## Expected behavior

Date and time is shown as YYYYY-MM-DD hh:mm

## Additional context

Enduro is intended for use internationally. We should conform to existing international standards where possible, to avoid favoring any particular jurisdiction's preferences. 

Note that once configured and confirmed, the calendar widget will disappear, but the date will display in the From and/or To field. This displayed date is also using the U.S. `mm/dd/yyyy` format instead of ISO 8601.

djjuhasz:
Fixes #1156.

### DIFF
--- a/dashboard/src/components/TimeDropdown.vue
+++ b/dashboard/src/components/TimeDropdown.vue
@@ -7,6 +7,8 @@ import { onMounted, ref, watch } from "vue";
 
 import IconClose from "~icons/clarity/close-line";
 
+const dateFormat = "yyyy-MM-dd HH:mm";
+
 const emit = defineEmits<{
   change: [name: string, start: string, end: string];
 }>();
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
:ok_hand:

Issue #1168, opened 2025-04-02
Enhancement: Create a re-usable dashboard pager


djjuhasz:
We have implemented a pager control on the SIPs list page, but the implementation is dependent on the particulars of the SIP list page and data store.   Any Dashboard page that has a potentially large list of entities (e.g. the AIPs page) will need a pager, so it would good to have a re-usable pager implementation to avoid creating a custom pager for each entity list.

**Screenshots**

SIP pager screenshot:
![Image](https://github.com/user-attachments/assets/033ccb6a-7497-4b97-acd5-7c9dac22adad)

**Additional context**

- Vue.js Components are a good fit for creating modular page elements like a pager

PR #1169, opened 2025-04-02
Add a dashboard pager component (#1168)

djjuhasz:
Create a general Vue.js Pager component that can be used for all dashboard lists.

- Create a Vue.js Pager component
- Use the Pager component on the `ingest/sips` and `storage/aips` pages
- Update the page URL when the page changes with the current page
  number, to allow bookmarking a specific page in the results
- Remove pager properties and methods from the AIP and SIP stores

### DIFF
--- a/dashboard/src/components/Pager.vue
+++ b/dashboard/src/components/Pager.vue
@@ -1,0 +1,156 @@
+<script setup lang="ts">
+import { computed } from "vue";
+
+import IconPrev from "~icons/bi/caret-left-fill";
+import IconNext from "~icons/bi/caret-right-fill";
+
+const props = defineProps({
+  offset: {
+    type: Number,
+    default: 0,
+  },
+  limit: {
+    type: Number,
+    default: 20,
+  },
+  total: {
+    type: Number,
+    required: true,
+  },
+  maxVisiblePages: {
+    type: Number,
+    default: 7, // An odd number of pages makes the pager symmetrical.
+  },
+});
+
+const emit = defineEmits<{
+  (e: "page-change", page: number): void;
+}>();
+
+const currentPage = computed(() => {
+  // Calculate current page based on offset and limit.
+  return Math.floor(props.offset / props.limit) + 1;
+});
+
+const totalPages = computed(() => {
+  // Calculate total pages based on total and limit.
+  return Math.ceil(props.total / props.limit);
+});
+
+const visiblePages = computed(() => {
+  if (totalPages.value <= props.maxVisiblePages) {
+    // If total pages is less than max visible, show all pages.
+    return Array.from({ length: totalPages.value }, (_, i) => i + 1);
+  }
+
+  // Calculate range of pages to show.
+  const halfVisible = Math.floor(props.maxVisiblePages / 2);
+  let startPage = Math.max(currentPage.value - halfVisible, 1);
+  let endPage = startPage + props.maxVisiblePages - 1;
+
+  // Adjust if end page exceeds total pages.
+  if (endPage > totalPages.value) {
+    endPage = totalPages.value;
+    startPage = Math.max(endPage - props.maxVisiblePages + 1, 1);
+  }
+
+  return Array.from(
+    { length: endPage - startPage + 1 },
+    (_, i) => startPage + i,
+  );
+});
+
+const goToPage = (page: number) => {
+  if (page < 1 || page > totalPages.value || page === currentPage.value) {
+    return;
+  }
+  emit("page-change", page);
+};
+</script>
+
+<template>
+  <nav role="navigation" aria-label="Pagination navigation">
+    <ul class="pagination justify-content-center">
+      <!-- Previous page -->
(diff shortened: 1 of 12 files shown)


dashboard/src/components/Pager.vue
```diff
@@ -0,0 +1,137 @@
...
+          aria-label="Go to previous page"
+          title="Previous page"
+        >
+          <IconPrev />
+        </a>
+      </li>
+
+      <!-- First page and ellipses-->
+      <template v-if="visiblePages[0] > 1">
+        <li class="page-item" :class="{ disabled: currentPage === 1 }">
```
jraddaoui: I think this will always be enabled, right? Or would it be possible to have the current page outside of the visible pages?

dashboard/src/components/Pager.vue
```diff
@@ -0,0 +1,137 @@
...
+          <a
+            id="first-page"
+            class="page-link"
+            href="#"
+            @click.prevent="goToPage(1)"
+            >1</a
+          >
+        </li>
+        <li class="d-none d-sm-block" aria-hidden="true">
```
jraddaoui: Should we check if the first visible page is 2 and not show this item?

dashboard/src/components/Pager.vue
```diff
@@ -0,0 +1,137 @@
...
+          class="page-link"
+          href="#"
+          @click.prevent="goToPage(page)"
+          >{{ page }}</a
+        >
+      </li>
+
+      <!-- Last page and ellipses-->
+      <template v-if="visiblePages[visiblePages.length - 1] < totalPages">
+        <li class="d-none d-sm-block" aria-hidden="true">
```
jraddaoui: Similar to the comment above but at the end. If `visiblePages[visiblePages.length - 1] === totalPages - 1`, I wouldn't show the ellipsis.

dashboard/src/components/Pager.vue
```diff
@@ -0,0 +1,137 @@
...
+          <a href="#" class="page-link disabled">…</a>
+        </li>
+        <li class="page-item" :class="{ disabled: currentPage === totalPages }">
```
jraddaoui: Same, I think this will always be enabled if `totalPages`  is outside `visiblePages`.

jraddaoui (changes requested):
Looks great to me, thanks @djjuhasz!

It would be nice to remove the existing pager from the AIPs browse page and reuse this one too, I think that should happen in this PR for consistency. Doing so we may realize there is still some duplication around, I wonder if it would be a good idea to pass the fetch function directly to the pager, instead of emitting and doing it outside, but that can happen in a following PR if we think it deserves the effort. Same with the pager related getters from the stores, I think we could move the entire logic into the pager component.

dashboard/src/components/Pager.vue
djjuhasz: Yes, good point. :)

djjuhasz:
@jraddaoui I've made the changes you suggested and switched to the Pager component for paging the `storage/aips` page.  This is ready for another review, thanks.

dashboard/src/components/Pager.vue
```diff
@@ -0,0 +1,156 @@
...
+  (e: "page-change", page: number): void;
+}>();
+
+const currentPage = computed(() => {
+  // Calculate current page based on offset and limit
+  return Math.floor(props.offset / props.limit) + 1;
+});
+
+const totalPages = computed(() => {
+  // Calculate total pages based on offset and limit
```
jraddaoui: Based on total and limit.

dashboard/src/pages/ingest/sips/index.vue
```diff
@@ -204,7 +223,7 @@ watch(
 
     <div class="text-muted mb-3">
       Showing {{ sipStore.page.offset + 1 }} -
-      {{ sipStore.lastResultOnPage }} of
+      {{ sipStore.page.offset + sipStore.page.limit }} of
```
jraddaoui: This may not be true in the final page, where the amount of records doesn't have to match the limit. Also, it currently shows `1 - 0 of 0` if there are no results, I think now it will show `1 to 20 of 0` in that case.

jraddaoui (changes requested):
Thanks @djjuhasz! It would be nice to make those counts a bit more solid.

dashboard/src/pages/ingest/sips/index.vue
djjuhasz: I've added a `ResultCounter` component to show a more readable count of the shown results on the SIPs and AIPs pages.  It now shows:

- "No results found" (0 results)
- "Found 1 result" (1 result)
- "Found 10 results" (more than 1 result, but less than two pages of results)
- "Showing 1 - 20 of 35 results" (two or more pages of results)

djjuhasz: I've made the `ResultCounter` it's own commit for review, but I think I'll squash it with the the dashboard pager commit before merging.

jraddaoui (approved):
Thanks @djjuhasz, really nice!

PR #1172, opened 2025-04-02
Use bine to install dev tools

sevein:
I've created a small tool called [bine](https://github.com/artefactual-labs/bine) to replace the boilerplate we maintain in each project through various `hack/make/dep_*.mk` files used for installing tools like tparse or shfmt. bine introduces a simple configuration file: see ([.bine.json](https://github.com/artefactual-sdps/enduro/blob/dev/use-bine/.bine.json)).

Let me know if you think this would be a good fit for the project - but no worries at all if not! I installed bine using `go get -tool github.com/artefactual-labs/bine` (though I had to upgrade Go to 1.24). I have some ideas for additional features that could be added in the future (see the [wiki](https://github.com/artefactual-labs/bine/wiki)), but those aren't a priority right now.

For the record, I still believe Nix is the best solution for reproducible dev environments - bine is just a stepping stone in that direction.


Makefile
```diff
@@ -159,7 +129,7 @@ gen-mock: $(MOCKGEN)
 
 gosec: # @HELP Run gosec security scanner.
 gosec: GOSEC_VERBOSITY ?= "-terse"
-gosec: $(GOSEC)
+gosec: tool-gosec
 	gosec \
 		$(GOSEC_VERBOSITY) \
 		-exclude-dir=dashboard \
```
jraddaoui, line 135: Missing the $(GOLANGCI_LINT) rules, fmt and lint.

jraddaoui (commented):
This is great @sevein, thanks! I'll take a deeper look to the project soon.

Just a couple of old calls to the golangci-lint tool, and I think you mean Atlas instead of go in the latest commit message.

hack/make/tools.mk
```diff
@@ -0,0 +1,22 @@
...
+    gosec \
+    gotestsum \
+    migrate \
+    mockgen \
+    shfmt \
+    tparse \
+    workflowcheck
+
+# Main target to install all tools.
+tools: $(addprefix tool-, $(TOOLS))
```
djjuhasz, line 18: @sevein I'm curious about the "tool-" prefixing. Did you have namespace problems without the prefix?

.bine.json
```diff
@@ -0,0 +1,78 @@
...
+    {
+      "name": "atlas",
+      "version": "0.32.0",
+      "url": "https://release.ariga.io/atlas",
+      "asset_pattern": "atlas-{goos}-{goarch}-v{version}"
+    },
+    {
+      "name": "ent",
+      "go_package": "entgo.io/ent/cmd/ent",
+      "version": "latest" // 0.14.4 still not compatible with go1.24.
```
djjuhasz: Are you sure that ent 0.14.4 isn't compatible with go 1.24.x? I don't see any errors if I do `make gen-ent` on this branch. :thinking: 

https://github.com/ent/ent/issues/4324 suggests go 1.24.x compatibility was fixed by https://github.com/ent/ent/commit/2b1ad7e4b0128083f50b06f75619c0bea601bdd4 which is included in the ent 0.14.2 ... 1.14.4 tags.

go.mod
```diff
@@ -1,14 +1,14 @@
 module github.com/artefactual-sdps/enduro
 
-go 1.23.6
+go 1.24.2
 
 require (
-	ariga.io/atlas v0.31.0
+	ariga.io/atlas v0.32.0
```
djjuhasz, line 6: Did you have to manually sync this version of the atlas module with the atlas binary used for code generation, or does bine do some magic to sync the versions?

djjuhasz (commented):
This is very nice @sevein. :) I had a couple of questions, but it seems like a big improvement over having a bunch of little make files to manage the installed binaries for the project. :grin:

djjuhasz:
I did test this locally, and it all the `make` commands I tried worked magically in my environment. :tada:

Makefile
sevein: Fixed, thank you!

.bine.json
sevein: You're right! I'm not sure how I came to that conclusion, thanks.

go.mod
sevein: Manually :cry:

sevein: I made a note to remind me of this, it sounds doable.

sevein:
Glad to hear that this is working for you!

I've already rewritten the commits so I'll list the latest changes:
* Rebased againt `dev/down-migrations` (thx radda; atlas update)
* Use `tool-golangci-lint` in `make fmt` and `make lint` (thx radda)
* Bump ent to 0.14.4 for compatibility with go1.24 (thx david)
* Updated `hack/activate/{bash,fish}` to use `go tool bine path` (last minute fix)
* Moved `tools` target to `Makefile` so it's listed by `make help` (last minute fix)

djjuhasz (approved):
LGTM! Thanks @sevein. :+1:

PR #1173, opened 2025-04-02
Remove down migrations

jraddaoui:
Down migrations are hard to maintain and in some cases they could break existing data, see the comments from [this thread](https://github.com/artefactual-sdps/enduro/pull/1157#discussion_r2019482343). This PR will remove the existing down migrations and their generation when running the migrate command. It also upgrades Atlas from 0.31.0 to 0.32.0.


djjuhasz (commented):
@jraddaoui I don't see any changes that would prevent new down migrations from being generated in the future.  Is there a change to the migration generator or call parameters that I'm missing?

jraddaoui:
Yes, the default formatter does the trick: https://github.com/artefactual-sdps/enduro/pull/1173/files#diff-918e5643cec8e9b3393f6301dab40d8e7838fa28b72067549ebbbda244cecd75R102

djjuhasz (approved):
Huh, cool that using the default formatter just creates "up" migrations.  How did you figure that out, I've been searching through the ent and Atlas docs but I can't any find any information about the formatters. :confused:

jraddaoui:
I remembered we didn't have down migrations in `sdps_preprocessing`, and that was the only difference there.

PR #1175, opened 2025-04-07
Add Release action

scollazo:
This will create a .tar.gz with all the binaries and essential configuration files.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+on:
+  push:
+    tags:
+    - "v*"
+name: GoReleaser
+jobs:
+  release:
+    strategy:
+      fail-fast: false
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+    - name: Install Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: go.mod
+    - uses: actions/setup-node@v4.0.4
+      with:
+        node-version: 22
+    - name: Run GoReleaser
+      uses: goreleaser/goreleaser-action@v6
+      with:
+        version: latest
+        args: release
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,78 @@
+---
+version: 2
+project_name: enduro-sdps
+
+before:
+  # Templates for the commands to be ran.
+  hooks:
+  - sh -c 'cd dashboard; npm clean-install; npm run build'
+
+builds:
+- env:
+  - CGO_ENABLED=0
+  id: enduro
+  goos:
+  - linux
+  goarch:
+  - amd64
+  flags:
+  - -trimpath
+  ldflags:
+  - -X "github.com/artefactual-sdps/enduro/internal/version.Long={{ .Version }}-t{{ .ShortCommit }}"
+  - -X "github.com/artefactual-sdps/enduro/internal/version.Short={{ .Version }}"
+  - -X "github.com/artefactual-sdps/enduro/internal/version.GitCommit={{ .Commit }}"
+  main: ./cmd/enduro
+  binary: enduro
+
+- env:
+  - CGO_ENABLED=0
+  id: enduro-am-worker
+  goos:
+  - linux
+  goarch:
+  - amd64
+  flags:
+  - -trimpath
+  ldflags:
+  - -X "github.com/artefactual-sdps/enduro/internal/version.Long={{ .Version }}-t{{ .ShortCommit }}"
+  - -X "github.com/artefactual-sdps/enduro/internal/version.Short={{ .Version }}"
+  - -X "github.com/artefactual-sdps/enduro/internal/version.GitCommit={{ .Commit }}"
+  main: ./cmd/enduro-am-worker
+  binary: enduro-am-worker
+
+- env:
+  - CGO_ENABLED=0
+  id: enduro-a3m-worker
+  goos:
+  - linux
+  goarch:
+  - amd64
+  flags:
+  - -trimpath
+  ldflags:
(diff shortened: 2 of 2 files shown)


.goreleaser.yml
```diff
@@ -0,0 +1,85 @@
...
+  goos:
+  - linux
+  goarch:
+  - amd64
+  flags:
+  - -trimpath
+  ldflags:
+  - -X "main.buildTime={{.Date}}"
+  - -X "main.gitCommit={{.Commit}}"
+  - -X "main.version={{.Version}}"
```
jraddaoui: These are not the flags we currently use here:

https://github.com/artefactual-sdps/enduro/blob/main/hack/build_dist.sh#L38

Also, in this project, should we try to reuse that script instead of using goreleaser.

.goreleaser.yml
```diff
@@ -0,0 +1,85 @@
...
+  sort: asc
+  filters:
+    exclude:
+    - '^docs:'
+    - '^website:'
+
+before:
+  # Templates for the commands to be ran.
+  hooks:
+  - sh -c 'cd dashboard; npm clean-install; npm run build'
```
jraddaoui, line 8: Not sure if we should do this here. The labs version embeds the dashboard in the Go binary if I recall correctly. We may want to look at something like that as an option.

jraddaoui (commented):
Thanks for looking into this @scollazo!

All this was initially set up by @sevein (goreleaser in the other project and build script here), I'd like to get his feedback too.

.goreleaser.yml
scollazo: Not here, as we need to rewrite some js files with the proper values for OIDC auth

sevein (commented):
Sorry I missed your ping - Joel just brought it to my attention.

Like Radda pointed out, this GoReleaser config doesn't yet mirror what our `build_dist.sh` does for version/git stamping - but that's easy to fix. I may be missing some context if this has already been decided, but the main question is whether you want a single "source of truth" for building the binaries or you're ok maintaining two parallel build mechanisms. I see three options:

* Adopt the approach in this pull request: let GoReleaser be an alternate binary builder alongside `build_dist.sh`. You end up with duplicate logic but maybe that's fine for now if you plan to consolidate later. You'd need to make sure that you mirror the build configuration flags.
* Treat `build_dist.sh` as the primary builder and use GoReleaser only for packaging and distribution. You can configure GoReleaser to skip its built-in build and call the script instead - that way GoReleaser handles only packaging, checksums, and releases. I haven't tested it but this looks like it's a [working example](https://src.redpoint.games/redpointgames/kubernetes-autoscaler/-/blob/7858da671604ccd5b718bd64a44a9a776563d231/cluster-autoscaler/cloudprovider/bizflycloud/gobizfly/.goreleaser.yml#L7-14).
* Switch to GoReleaser as the primary builder and drop `build_dist.sh`, i.e. remove `build_dist.sh` entirely and update `build_docker.sh` to use `goreleaser build --id <build_id>`.

If you ask me, I lean towards the third option: all build logic in GoReleaser. It's a project I generally like and will make our build process more manageable. Also important, [go1.24](https://tip.golang.org/doc/go1.24#go-command) now automatically sets module version information in the compiled binary from the VCS tag and/or commit, which makes `build_dist.sh` basically redundant. Of course, you could deal with that across multiple PRs if you decide to go that route.

.github/workflows/release.yml
```diff
@@ -20,3 +20,6 @@ jobs:
       with:
         version: latest
         args: release
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
sevein, line 27: GitHub provides automatic token authentication, see https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication.

I think that you'd need something like this:

```
permissions:
  contents: write
```

sevein: There is an example here: https://github.com/goreleaser/goreleaser-action?tab=readme-ov-file#workflow.

sevein, line 27: Do we need to define custom write permissions?
See https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token.

scollazo:
I updated the flags to match the ones that we use, but now I have a question, is it worth relying on the VERSION.txt file, or should we use the git tag for identifying versions?

If the first approach is the one we want, I'm not sure how to read/use it in goreleaser, any hints?

sevein:
If you choose the first option we talked about, you can safely ignore the `VERSION.txt` file for now so you can merge this sooner. Later, if you switch to GoReleaser as the primary builder, you can look into [setting a custom Git tag](https://goreleaser.com/cookbooks/set-a-custom-git-tag/). At that point you can also reevaluate whether you still need `VERSION.txt`. Its biggest benefit is decoupling versioning from the VCS (for example, in a downloaded ZIP file generated by GitHub) but it may not be worth the extra work.

.goreleaser.yml
```diff
@@ -0,0 +1,85 @@
+---
+version: 2
+project_name: enduro-sdps
+
+changelog:
+  sort: asc
+  filters:
+    exclude:
+    - '^docs:'
+    - '^website:'
```
jraddaoui: Do we really need the changelog here? Is this going to the Github release page? If so, I think we should disable it for now. Otherwise, I don't like it sorting the commits by message and I we don't really use these patterns in the commit messages.

jraddaoui (commented):
Thanks for the input @sevein, that sounds like a plan for me.

@scollazo, I added a comment about the changelog, other than that it looks good to me.

.goreleaser.yml
scollazo: done

jraddaoui (approved):
:ok_hand: Thanks @scollazo!

scollazo:
Squashed and merged. A token has been created with the proper permissions (I hope!), so I'll keep my fingers crossed until we do a new release ;)

PR #1178, opened 2025-04-08
Add "failed" state for workflows and tasks (#1161)

### LINKED ISSUE #1161: Problem: No easy way to separate system errors from validation failures
## PROBLEM

**Context and background**

This issue relates to proposals included in #1160  - separating out one of the implications of the new statuses there into its own issue. 

Talking to SFA staff and to Maaike (who runs some of the legacy Enduro client instances for Enterprise Suport) - it is not easy to sort and separate ingest validation failures from system and network errors. 

For Maaike, Legacy Enduro has a bulk retry option. This is useful when network or disk full errors temporarily interrupt processing, causing packages to error. However, there is no point in retrying packages that failed due to validation issues - they will just fail again. Unfortunately, without clicking through and reviewing the log outputs, there is no way to easily separate those. 

## PROPOSAL

Per the new statuses proposed on #1160,  this issue proposes adding separate workflow and task statuses for content validation failures vs system or network errors. 

**Workflows**

* **FAILED**: The related package has failed to meet policy-defined criteria and ingest was terminated
* **ERROR**:  The workflow has encountered a system error it could not resolve

**Tasks**

* **FAILED**: The related package has failed to meet this task's policy-defined criteria
* **ERROR**:  The task has encountered a system error it could not resolve

This implies that internally the system recognizes the distinction between content validation failures vs system errors and applies the correct status. This also implies updating any related enums, filters, etc.

djjuhasz:
To differentiate workflows that have failed due to validation or policy infractions add a "failed" status for the ingest workflow and its tasks.

- Add a "failed" workflow status
- Add a "failed" task status
- Create a general workflow state struct to efficiently pass the state to activities, and allow them to set the workflow status
- Update the workflow state commentary to explain the purpose of all the fields
- Use the updated "failed" status when an ingest task fails validation


djjuhasz:
It's probably easier to follow through this PR commit by commit due to its size.

internal/enums/preprocessing_task_outcome.go
```diff
@@ -2,10 +2,10 @@ package enums
 
 /*
 ENUM(
-unspecified
-success
-system failure
-validation failure
+unspecified         // Unused!
+success			    // Completed successfully.
```
jraddaoui: Something wrong with this comment alignment.

internal/enums/workflow_type.go
```diff
@@ -2,9 +2,9 @@ package enums
 
 /*
 ENUM(
-unspecified
-create aip
-create and review aip
+unspecified             // Type is indeterminate.
+create aip 	   		    // Create an AIP and store it.
```
jraddaoui: Something wrong with this comment alignment.

internal/workflow/processing.go
```diff
@@ -303,35 +249,45 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
...
 	}
 
 	// Schedule deletion of the original in the watched data source.
 	{
-		if status == enums.SIPStatusDone {
-			if tinfo.req.RetentionPeriod != nil {
-				err := temporalsdk_workflow.NewTimer(ctx, *tinfo.req.RetentionPeriod).Get(ctx, nil)
+		if state.sip.status == enums.SIPStatusDone {
+			if req.RetentionPeriod != nil {
+				err := temporalsdk_workflow.NewTimer(ctx, *req.RetentionPeriod).Get(ctx, nil)
```
jraddaoui, line 257: Not a big deal I guess, but I think it would be better to use the value from the state, just in case. I doubt we'll ever change the retention period in workflow, but more as a general rule to use the status as much as possible after it has been initiated with the request. There are more places below using the request directly.

internal/workflow/state.go
```diff
@@ -0,0 +1,141 @@
...
+		// Initialize the PIP and AIP to empty structs to avoid nil pointer
+		// dereference errors.
+		pip: &pipInfo{},
+		aip: &aipInfo{},
+	}
+}
+
+// tempPath registers a filepath for deletion when a workflow session
+// completes.
+func (s *workflowState) tempPath(path string) {
```
jraddaoui: I'd call it `addTempPath`, or something like that to indicate the addition.

internal/workflow/state.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// path is the path to the PIP. It is populated by the BundleActivity.
+	path string
+
+	// isDir indicates whether the current working copy of the SIP is a
+	// filesystem directory.
+	isDir bool
+
+	// pipType is the type of the PIP.
+	pipType enums.SIPType
```
jraddaoui, line 101: Already inside `pipInfo`, I'd call this just `type`.

jraddaoui (approved):
Neat, thanks @djjuhasz! I love the new state handling, just some nitpicks.

internal/workflow/state.go
djjuhasz: I did initially try `type` but it's a reserve word. I could do something like `_type` or `ptype`, but `pipType` is easier to understand and not much longer, imho.

jraddaoui: Sounds good, I wasn't expecting that being part of a struct, but I always used `Type` when I included that field in one.

internal/workflow/processing.go
djjuhasz: `req` and `state.req` are both pointers to the same underlying struct, so there is no functional difference between the two — if we  mutate `state.req.RetentionPeriod` then `req.RetentionPeriod` will point to the updated value.  It might be a good idea to pass `req` by  value to `Execute` to prevent mutation of the original request, but that would incur some extra resource costs for copying the data.

If we don't change `state.req` to a copy of the `req` then I think we should stick with `req` here because it's shorter than `state.req`.

Issue #1123, opened 2025-02-12
Problem: when transfers are launched concurrently, bag validation fails randomly.


scollazo:
**Describe the bug**

When capacity is set to a number bigger than one, and a bunch of transfers are loaded into minio, there is the following error for some of them in the ui:

```System error: Error attempting to validate the Bag```

And the following error in preprocessing temporal logs:

```
{
  "message": "bagvalidate: runner is busy",
  "source": "GoSDK",
  "stackTrace": "",
  "encodedAttributes": null,
  "cause": null,
  "applicationFailureInfo": {
    "type": "",
    "nonRetryable": false,
    "details": null
  }

```

**To Reproduce**

- Set the enduro capacity to a number bigger than 1 (I used 8), and restart enduro and the enduro-am-worker
- Launch 10-20 transfers (they are the same transfer, renamed):

```
for i in $(seq 1 10); do mc cp enduro/transfers/bag-7537ab2c-4e6b-4820-95bf-bd2c577351c3.zip enduro/transfers/concurrent$i.zip; done
```

**Expected behavior**

The transfers work

**Screenshots**

![Image](https://github.com/user-attachments/assets/b86948d3-3548-465f-a8b2-60e21e329e3d)

![Image](https://github.com/user-attachments/assets/d79ad370-c66b-45f2-8e25-78c654006d74)

**Additional context**

**Comment by scollazo** [comment]
Log from preprocessing:


```
[...]

PR #1184, opened 2025-04-11
Don't use a general bagit-gython validator

jraddaoui:
Refs https://github.com/artefactual-sdps/enduro/issues/1123.

### DIFF
--- a/cmd/enduro-a3m-worker/main.go
+++ b/cmd/enduro-a3m-worker/main.go
@@ -13,7 +13,6 @@ import (
 
 	"ariga.io/sqlcomment"
 	"entgo.io/ent/dialect/sql"
-	bagit_gython "github.com/artefactual-labs/bagit-gython"
 	"github.com/artefactual-sdps/temporal-activities/archiveextract"
 	"github.com/artefactual-sdps/temporal-activities/archivezip"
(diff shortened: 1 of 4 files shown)


cmd/enduro-a3m-worker/main.go
```diff
@@ -262,7 +246,7 @@ func main() {
 			temporalsdk_activity.RegisterOptions{Name: activities.ClassifySIPActivityName},
 		)
 		w.RegisterActivityWithOptions(
-			bagvalidate.New(validator).Execute,
+			bagvalidate.New(nil).Execute,
```
djjuhasz, line 249: @jraddaoui are you sure this is created a new validator on each workflow run? It looks to me like `bagvalidate.New()` will be run once, when the worker is started, and the same bagit_gython instance will be reused. :thinking: 

I would expect that we'd need to create and pass a new bagit_gython instance as an activity parameter in the workflow to get a new instance each run.

jraddaoui: It will reuse the same `gythonValidator`, but the bagit_gython instance is created when Validate is called.

djjuhasz: Ah, I see.

djjuhasz (approved):
LGTM @jraddaoui. :+1:

Issue #1160, opened 2025-03-28
Problem: Packages don't have a status and workflow statuses don't cover all needs


fiver-watson:
## Problem

**Background and context*

This issue will replace issue #1081, which ended up becoming a long conversation about the mixing of ingest and storage domains in Enduro.

We have begun implementing a better separation of concerns via #1117 , and as a consequence, have made changes to the SIPs page, and added a new AIP browse page. These browse pages are supposedly using SIP and AIP status enums on them to display the status entries... but on the SIPs browse page, the status definitions all still describe **workflows**, and on the AIPs browse page, the initial AIP statuses used do not match the page's status filter tabs. 

Additionally, some of the limitations for operators related to discovery via status are outlined in a new enhancement proposal in issue #1159 , which proposes adding a new Workflows page in the Storage domain. Doing so will require clarity around statuses available for SIPs, AIPs, and workflows. 

Finally, Tasks don't currently have a legend to better explain their statuses. We should address this at the same time. 

**To Reproduce**

Examples to illustrate the issue:

1. Log in and navigate to the SIPs browse page in the Ingest domain
2. Click on the (i) info icon in the header above the Status column of the results, to expand the status legend

Resulting error: statuses shown describe workflows not SIPs

1. Log in and navigate to the AIPs browse page in the Storage domain
2. Look at the status filter tabs above the browse results

Resulting error: these statuses relate to the experimental "Create and review AIP" workflow created in early SAD-team Enduro development, and are not immediately useful for filtering AIPs. 

etc...

## Proposal

* Create separate enums for SIPs, AIPs, Workflows, and Tasks
* Add a new AIPs status colum (i) info icon that when clicked, opens an AIP status legend (like on the SIPS page)
* Update the existing definitions of these terms internally, and in the UI
   * SIPs status legend
   *  AIPs status legend
   * Tasks status legend
   * Eventually, via future work on #1159 , Workflows status legend
* Update the status filter tabs on the AIPs browse page
* Update the icons used for all statuses  (SIPs browse page tabs as well, etc)

Proposed enums, definitions and mockups for the legends can be found here: 

* https://miro.com/app/board/uXjVL7s6tus=/?moveToWidget=3458764618433665904&cot=14

Proposed updated icons for each status can be found here:

* https://miro.com/app/board/uXjVL7s6tus=/?moveToWidget=3458764622755430345&cot=14

[...]

PR #1188, opened 2025-04-17
Update AIP statuses (refs #1160)

djjuhasz:
- Remove "in review", "rejected" and "moving" AIP statuses
- Add "queued" AIP status
- Regenerate API, database, and Dashboard models of the AIP status
- Map discontinued status values to one of the new values in the database, application code and front-end
- Add debug logging for database migrations

Status mappings (old -> new):
- "in review" -> "pending
- "rejected" -> "deleted"
- "moving" -> "processing"


internal/db/db.go
```diff
@@ -22,8 +23,23 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/storage/persistence"
 )
 
-// Wait up to five minutes is another process is already on it.
-const lockTimeout = time.Minute * 5
+// Toggle verbose logging for the migrations.
+const verbose = false
```
jraddaoui: I think it would make sense to pass the `debug` configuration value instead of using a constant.

internal/db/db.go
```diff
@@ -22,8 +23,23 @@ import (
...
+type migrateLogger struct {
+	logger  logr.Logger
+	verbose bool
+}
+
+func (l *migrateLogger) Printf(format string, a ...any) {
+	l.logger.V(2).Info(fmt.Sprintf(format, a...))
+}
+
+func (log *migrateLogger) Verbose() bool {
```
jraddaoui: Using `log` confused me a little, I'd use `l *migrateLogger` as the function above.

internal/db/db.go
```diff
@@ -132,31 +148,38 @@ func ConnectSQLite(tp trace.TracerProvider, dsn string) (db *sql.DB, err error)
...
 
-func MigrateEnduroDatabase(db *sql.DB) error {
+func MigrateEnduroDatabase(logger logr.Logger, db *sql.DB) error {
 	src, err := enduroSourceDriver()
 	if err != nil {
 		return fmt.Errorf("error loading migration sources: %v", err)
 	}
 
-	return up(db, src)
+	return up(logger.WithName("enduro-migrate"), db, src)
```
jraddaoui: I think it's better to set the name in main before passing the logger, as we do in many places there. Also, this will name the logger `enduro.enduro-migrate` which seems a bit redundant to me.

I wanted to make it clear that the `enduro` database is the ingest domain DB but I never got to fully address that issue. Not sure if we want to start doing it here calling the loggers `enduro.ingest-migrate` and `enduro.storage-migrate`

internal/db/db.go
```diff
@@ -132,31 +148,38 @@ func ConnectSQLite(tp trace.TracerProvider, dsn string) (db *sql.DB, err error)
...
 
-func MigrateEnduroStorageDatabase(db *sql.DB) error {
+func MigrateEnduroStorageDatabase(logger logr.Logger, db *sql.DB) error {
 	src, err := persistence.SourceDriver()
 	if err != nil {
 		return fmt.Errorf("error loading migration sources: %v", err)
 	}
 
-	return up(db, src)
+	return up(logger.WithName("enduro-storage-migrate"), db, src)
```
jraddaoui: Check comment above.

internal/db/db.go
```diff
@@ -132,31 +148,38 @@ func ConnectSQLite(tp trace.TracerProvider, dsn string) (db *sql.DB, err error)
...
 		return fmt.Errorf("error creating golang-migrate object: %v", err)
 	}
 
+	// Log migration progress.
+	l := &migrateLogger{
+		logger:  logger,
+		verbose: verbose,
+	}
+	m.Log = l
+
```
jraddaoui: You could assign to `m.Log` directly here. And I think it makes more sense to pass the logger and do this in `newMigrate`.

internal/db/db.go
```diff
@@ -177,8 +200,6 @@ func newMigrate(db *sql.DB, src source.Driver) (*migrate.Migrate, error) {
 		return nil, fmt.Errorf("error creating migrate instance: %w", err)
 	}
 
-	m.LockTimeout = lockTimeout
-
```
jraddaoui: Why removing this?

internal/storage/enums/aip_status.go
```diff
@@ -2,14 +2,13 @@ package enums
...
-rejected
-stored
-moving
-pending
-processing
-deleted
+unspecified  // Indeterminate status.
+stored       // Stored in preservation storage.
+pending      // Awaiting a user decision.
+
```
jraddaoui: Why the empty line?

internal/storage/persistence/migrations/20250415223433_update_aip_statuses.up.sql
```diff
@@ -0,0 +1,7 @@
+-- Map "aip" statuses to new values.
+UPDATE `aip` SET `status` = 'pending' WHERE `status` = 'in_review';
+UPDATE `aip` SET `status` = 'processing' WHERE `status` = 'moving';
+UPDATE `aip` SET `status` = 'deleted' WHERE `status` = 'rejected';
+
+-- Update "aip" enum values.
+ALTER TABLE `aip` MODIFY COLUMN `status` enum('unspecified','stored','pending','processing','deleted','queued') NOT NULL;
```
jraddaoui, line 7: After the changes to remove the down migrations, I think the migrations created with the standard formatter won't include `.up` in their name. I'd do the same for consistency, even if all the previous ones have it.

jraddaoui (commented):
Thanks @djjuhasz, looks good to me!

I added a few minor comments but nothing important. There will be a few things to tweak when we revisit the review and move processes as you already noticed, but this is a good increment in that direction too.

internal/db/db.go
djjuhasz: Oops, that was accidental.

internal/storage/enums/aip_status.go
djjuhasz: Also accidental.

internal/storage/persistence/migrations/20250415223433_update_aip_statuses.up.sql
djjuhasz: The migrations don't run if `.up` is not included in their names, at least not in my dev environment. :(  It took me quite a while to figure out why the migration wasn't being applied, because the "migrate.go" command-line script doesn't include the `.up` in the migration file name.

I agree we should drop the ".up" from the migration names, but I don't know how to update the migration scripts to allow that.

jraddaoui: Good catch @djjuhasz, I didn't realize :man_facepalming:, thanks!

I tried with different [formatters](https://github.com/artefactual-sdps/enduro/blob/main/cmd/migrate/main.go#L102) and [migrate methods](https://github.com/artefactual-sdps/enduro/blob/main/internal/db/db.go#L186), but I couldn't make it work without the `.up` suffix either. So now I think we should append `.up` to the [migration name](https://github.com/artefactual-sdps/enduro/blob/main/cmd/migrate/main.go#L108) manually when it's being created, if it's not already included in the name.

We should add a section to the development docs at some point explaining how to create migrations and we should note this requirement there too, alongside [the comments from the command](https://github.com/artefactual-sdps/enduro/blob/main/cmd/migrate/main.go#L1-L23) and instructions to create the empty database.

We can do all this in a different PR ;)

internal/db/db.go
djjuhasz: I don't think we should output the verbose level of logs from the `migrate` library when `cfg.Debug` is true because `cfg.Debug` controls the *encoding* of the log output (STDOUT with colors vs. JSON without colors)  and not the *verbosity* of the log output.  We could try and increase the verbosity of the migration logs based on the `cfg.Verbosity` setting instead, but I think this is overkill. 

The "non-verbose" version of the migration log outputs the name and elapsed time for each migration step that is run, which I think is sufficient for most of our needs.  This default output is still fairly verbose, and only helpful for debugging, so I've set these messages to log level "2".  Because the dev environment has a default `cfg.Verbosity` of 2, we will usually see these basic migration log messages in the dev environment.

I could set the migration logger to output `verbose` logs if `cfg.Verbosity` >= 3 and set the log level for *all* migration messages as level 3 in this case.  I can't see any way to use `migrate.Logger` to implement `logr` style log levels where it will output both level 2 and level 3 log messages.  In `migrate.Logger` the `Verbose()` method returns a boolean, and if it returns false the verbose messages are not emitted at all. I can't see any way to e.g. show the non-verbose  messages at log level 2, and the verbose messages at level 3.

All this is pretty complicated for a level of logging I don't anticipate we will ever need.  My preference would be to keep the current local `verbose` constant, or just hard-code `Verbose()` to always return false.

internal/db/db.go
djjuhasz: I'll switch to `enduro.migrate` and `enduro.storage-migrate`.  I don't want to introduce `enduro.ingest-migrate` at this point because we haven't renamed the `enduro` database yet and I think it will just be confusing.

internal/storage/persistence/migrations/20250415223433_update_aip_statuses.up.sql
djjuhasz: Yes, I agree that some documentation about how to create and apply migrations would be very helpful. :)

internal/db/db.go
jraddaoui: Sounds good!

internal/db/db.go
```diff
@@ -148,48 +147,40 @@ func ConnectSQLite(tp trace.TracerProvider, dsn string) (db *sql.DB, err error)
 	return db, err
 }
 
-func MigrateEnduroDatabase(logger logr.Logger, db *sql.DB) error {
+func MigrateEnduroDatabase(logger logr.Logger, db *sql.DB, debug bool) error {
```
jraddaoui: No need for the debug flag in the end.

internal/db/db.go
```diff
@@ -148,48 +147,40 @@ func ConnectSQLite(tp trace.TracerProvider, dsn string) (db *sql.DB, err error)
...
 	if err != nil {
 		return fmt.Errorf("error loading migration sources: %v", err)
 	}
 
-	return up(logger.WithName("enduro-migrate"), db, src)
+	return up(logger, db, src)
 }
 
-func MigrateEnduroStorageDatabase(logger logr.Logger, db *sql.DB) error {
+func MigrateEnduroStorageDatabase(logger logr.Logger, db *sql.DB, debug bool) error {
```
jraddaoui: Same.

jraddaoui (approved):
Thanks @djjuhasz! LGTM!

internal/db/db.go
djjuhasz: Fixed.

internal/db/db.go
djjuhasz: Fixed.

internal/db/db.go
djjuhasz: I've added some comments to the config for `Debug` and `Verbosity`.

Issue #1189, opened 2025-04-17
Problem: Dashboard error when viewing AIP without permission to list workflows


djjuhasz:
**Describe the bug**

If a Dashboard user has permissions to view an AIP (storage:aips:read) but not to list the AIPs workflows (storage:aips:workflows:list) then a large "Page loading error" appears at the top of the page.

![Image](https://github.com/user-attachments/assets/8a3d9720-8693-45ce-a382-88d97e0d0557)

**To Reproduce**

It's difficult to reproduce this bug because it requires very specific access permissions. I only found the bug because I was testing with the "operator" account in the development environment and added access to list AIPs (storage:aips:list), which allowed me to find and view an individual AIP.

1. Log in to the Enduro dashboard with an account that has permissions to list and view AIPs, but not to list workflows
2. Ingest a SIP and produce an AIP
3. Click "Storage > AIPs" in the left navigation men
4. Click on an AIP in the list

**Expected behavior**

The AIP page should display without an error message.

**Screenshots**

See above.

**Additional context**

None.

sallain:
@jraddaoui please file same issue for SIPs

PR #1190, opened 2025-04-17
Conditionally load AIP workflows (#1189)

djjuhasz:
Only fetch an AIP's workflow list if the authenticated user has the required access permissions (storage:aips:workflows:list).


jraddaoui (approved):
LGTM, thanks @djjuhasz!

PR #1193, opened 2025-04-23
Add database migrations documentation

djjuhasz:


docs/src/dev-manual/db-migrations.md
```diff
@@ -0,0 +1,205 @@
...
+
+Also change the file name in the `internal/storage/persistence/migrations/atlas.sum`
+file.
+
+```text
+20250423225046_add_aip_note_column.up.sql h1:Q7c6YSYzIO30VnKzgGdupGih93N81rkmptb6SYDOkao=
+```
+
+Then regenerate the `atlas.sum` hashes:
+
```
jraddaoui: I'd like to do this in the actual migrate command appending `.up` to the passed name if it's not included already, and just add a note here for migrations created manually altogether.

jraddaoui (commented):
Nice, thanks for looking into this @djjuhasz!

docs/src/dev-manual/db-migrations.md
djjuhasz: I started trying to add the ".up" to the filename in the migrate command, but it's not trivial - you need to find the migration file name (with the prepended timestamp) then update the name of the file, update the `atlas.sum` file and regenerate the file hashes.  I decided to just document the current state as-is, with the intention of improving the migrate command later.

jraddaoui: I mean adding the `.up` suffix before the migration is generated, in https://github.com/artefactual-sdps/enduro/blob/main/cmd/migrate/main.go#L108:

```go
name, _ := p.GetString("name")
if !strings.HasSuffix(name, ".up") {
	name = fmt.Sprintf("%s.up", name)
}
```

This seems to work for me and it avoids all that manual process for generated migrations.

djjuhasz: Ah, that is a simple hack that does the job.  I was overthinking the problem I guess. :laughing: 

I've added this change, and updated the documentation appropriately.

jraddaoui (approved):
Really nice, thanks @djjuhasz!

PR #1195, opened 2025-04-28
Add deletion request tests to storage Ent client

jraddaoui:
Refs #1076.

### DIFF
--- a/internal/storage/persistence/ent/client/deletion_request_test.go
+++ b/internal/storage/persistence/ent/client/deletion_request_test.go
@@ -1,0 +1,284 @@
+package client_test
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/uuid"
+	"gotest.tools/v3/assert"
+
+	"github.com/artefactual-sdps/enduro/internal/storage/enums"
+	"github.com/artefactual-sdps/enduro/internal/storage/persistence"
+	"github.com/artefactual-sdps/enduro/internal/storage/persistence/ent/db"
+	"github.com/artefactual-sdps/enduro/internal/storage/types"
+)
+
+func initialDataForDeletionRequestTests(t *testing.T, ctx context.Context, entc *db.Client) {
+	t.Helper()
+
+	initialDataForTaskTests(t, ctx, entc)
+}
+
+func TestCreateDeletionRequest(t *testing.T) {
+	t.Parallel()
+
+	drUUID := uuid.New()
+	requestedAt := time.Now()
+
+	type test struct {
+		name    string
+		dr      *types.DeletionRequest
+		want    *db.DeletionRequest
+		wantErr string
+	}
+
+	for _, tt := range []test{
+		{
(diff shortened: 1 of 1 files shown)


djjuhasz:
@jraddaoui is it possible for you to break this into multiple PRs? It's a lot of changes to review all at once.

jraddaoui:
Sure, it's all new test files and a full rewrite of another, but I can separate them.

jraddaoui:
I've reused this PR for the first one and created five more:

- https://github.com/artefactual-sdps/enduro/pull/1195 (this one)
- https://github.com/artefactual-sdps/enduro/pull/1197
- https://github.com/artefactual-sdps/enduro/pull/1198
- https://github.com/artefactual-sdps/enduro/pull/1199
- https://github.com/artefactual-sdps/enduro/pull/1200
- https://github.com/artefactual-sdps/enduro/pull/1201

internal/storage/persistence/ent/client/deletion_request_test.go
```diff
@@ -0,0 +1,262 @@
...
+		updater persistence.DeletionRequestUpdater
+		dbID    int
+		want    *types.DeletionRequest
+		wantErr string
+	}
+
+	for _, tt := range []test{
+		{
+			name: "Updates a Deletion Request",
+			updater: func(dr *types.DeletionRequest) (*types.DeletionRequest, error) {
```
djjuhasz, line 121: I'm concerned that not having a standard updater for for updating deletion requests could lead to missed bugs. I've caught issues with updaters in test in the past due to missing fields or unexpected behaviour with nil or empty values.

djjuhasz (changes requested):
See my note about not having a standard updater function.  Other than that, this looks good to me. :)

internal/storage/persistence/ent/client/deletion_request_test.go
jraddaoui: I'm not sure I understand what you mean with a standard updater function, a default function receiving all the possible values to update and checking they are valid? If that's the case, then why do we have this updater functions and don't do that in the actual update function.

I have always had mixed feelings about the updater functions for the reasons you mention, but I think the idea is to give control to the caller to what needs to be updated without worrying about existing values, which as you said could lead to unexpected errors. However, if we do a better job documenting in code the updater types and the update functions, explaining what can be updated and what will be ignored, and if we consider/prevent all possible errors in the update functions, I think it's a nice pattern.

Let me know your thoughts, we could change it for all the updater/update functions in another PR.

---

In this particular case, the update is suited for the only use case we have, which is doing the review, so I have added a new test to show how it will ignore immutable fields from the request in the update.

djjuhasz: Good points about allowing the caller to determine which fields the updater mutates. My main concern is that the updater functions we use are being tested somewhere, and I see now that you are testing the only call I can find to `UpdateDeletionRequest()` in https://github.com/artefactual-sdps/enduro/pull/1199.

Thanks for adding the new test of ignoring immutable fields.

djjuhasz (approved):
:+1:

PR #1197, opened 2025-04-29
Add deletion request tests to storage service

jraddaoui:
Refs #1076.

### DIFF
--- a/internal/storage/service_test.go
+++ b/internal/storage/service_test.go
@@ -1908,6 +1908,34 @@ func TestCreateWorkflow(t *testing.T) {
 		svc := setUpService(t, attrs)
 		ctx := context.Background()
 
+		dbID := 1
+		workflow := &types.Workflow{
+			UUID:       uuid.New(),
+			TemporalID: "temporal-id",
+			Type:       enums.WorkflowTypeMoveAip,
+			Status:     enums.WorkflowStatusInProgress,
+		}
+
+		attrs.persistenceMock.
+			EXPECT().
+			CreateWorkflow(ctx, workflow).
+			DoAndReturn(func(ctx context.Context, w *types.Workflow) error {
+				w.DBID = dbID
+				return nil
+			})
+
+		err := svc.CreateWorkflow(ctx, workflow)
+		assert.NilError(t, err)
+		assert.Equal(t, workflow.DBID, dbID)
+	})
+
+	t.Run("Returns a persistence error", func(t *testing.T) {
+		t.Parallel()
+
+		attrs := &setUpAttrs{}
+		svc := setUpService(t, attrs)
+		ctx := context.Background()
+
 		workflow := &types.Workflow{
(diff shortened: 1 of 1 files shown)


internal/storage/service_test.go
```diff
@@ -2033,3 +2033,126 @@ func TestUpdateTask(t *testing.T) {
 		assert.DeepEqual(t, re, task)
 	})
 }
+
+func TestCreateDeletionRequest(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Creates a DeletionRequest", func(t *testing.T) {
```
djjuhasz, line 2095: I think this test name should be "Returns a persistence error", or something similar.  

It would be nice to have a test that creates a deletion request too, but maybe this got lost when splitting up the PRs?

internal/storage/service_test.go
```diff
@@ -2033,3 +2033,126 @@ func TestUpdateTask(t *testing.T) {
...
+			AIPUUID:     uuid.New(),
+			Reason:      "Updated reason",
+			Status:      enums.DeletionRequestStatusApproved,
+			RequestedAt: time.Now(),
+			ReviewedAt:  time.Now(),
+			Reviewer:    "reviewer@example.com",
+			ReviewerISS: "issuer",
+			ReviewerSub: "subject",
+		}
+		updater := func(dr *types.DeletionRequest) (*types.DeletionRequest, error) { return dr, nil }
```
djjuhasz, line 2173: I'm concerned about defining an ad-hoc updater for testing the deletion request, rather than testing a standard implementation.  I've found errors when testing updaters in the past due missing fields or unexpected empty or nil values.

djjuhasz (changes requested):
I think the successful "CreateDeletionRequest" might have gotten lost in splitting up the original PR.

I'm also concerned that not having a standard updater for the `UpdateDeletionRequest()` method could lead to missing bugs in test.

internal/storage/service_test.go
jraddaoui: See my comments in https://github.com/artefactual-sdps/enduro/pull/1195#discussion_r2067946835. In here, we only pass the updater to the persistence client without doing anything with it or worrying about its output (for that we test the actual definition in the [activity tests](https://github.com/artefactual-sdps/enduro/pull/1199/files#diff-00d6d8ba6d938f78d56ea699fce7ae563cf581cb33d2a07a6c98afba089d353dR291-R325)).

internal/storage/service_test.go
jraddaoui: Good point, it was just like that, quite misleading, but it was the same for workflows and tasks. Since these function are just wrappers to the persistence layer and we use a mock in the tests, I don't think they add a lot of value but I have added actual creation tests for the three entities.

djjuhasz (approved):
LGTM :+1:  Thanks @jraddaoui.

PR #1201, opened 2025-04-29
Add storage delete AIP workflow tests

jraddaoui:
Refs #1076.


djjuhasz (approved):
That's a lot of steps for deleting an AIP. :)  Thanks for adding the tests @jraddaoui.

PR #1204, opened 2025-05-01
Update status badge legends (refs #1160)

djjuhasz:
- Add a "StatusLegend" component and use it for the "ingest/sips" and "storage/aips" page legends
- Add a "WorkflowHelp" component and use it for the "ingest/sips/:id" and "storage/aips/:id" pages
- Add a "Task status legend" to the Workflow Help
- Update the SIP and AIP status legends with the new labels


djjuhasz:
Screenshot of the ingest workflow help + legend:
image

djjuhasz:
SIP list legend (edit: fixed "failed" text):
image

dashboard/src/components/WorkflowHelp.vue
```diff
@@ -0,0 +1,101 @@
+<script setup lang="ts">
+import { api } from "@/client";
+import StatusBadge from "@/components/StatusBadge.vue";
+import IconLink from "~icons/bi/box-arrow-up-right";
+
+const statuses = [
+  {
+    status: api.EnduroIngestSipWorkflowStatusEnum.Done,
```
jraddaoui, line 8: Shouldn't these be task statuses then?

dashboard/src/pages/storage/aips/[id]/index.vue
```diff
@@ -40,47 +56,27 @@ const authStore = useAuthStore();
         authStore.checkAttributes(['storage:aips:workflows:list'])
       "
     >
-      <div class="d-flex">
+      <div>
         <h2 class="mb-0">
-          Workflows
+          Storage workflow details
```
Comment (jraddaoui) [inline dashboard/src/pages/storage/aips/[id]/index.vue:61]: We used "Ingest workflow details" in the SIP page because there can only be one workflow per SIP, but we used "Workflows" here because there can be more than one.

dashboard/src/components/WorkflowHelp.vue
```diff
@@ -0,0 +1,101 @@
...
+              <button
+                id="workflow-help-close"
+                type="button"
+                class="btn-close align-middle"
+                @click="() => emit('update:show', false)"
+                aria-label="Close"
+              ></button>
+            </div>
+          </div>
+          <label for="task-status-legend" class="h5">Task status legend</label>
```
jraddaoui, line 63: Why not a normal heading?

jraddaoui (approved):
LGTM, thanks @djjuhasz!

I'll merge for the release, we can address the minor feedback if needed after that.

dashboard/src/components/WorkflowHelp.vue
djjuhasz: Yes, they should be. Oops.

dashboard/src/pages/storage/aips/[id]/index.vue
Comment (djjuhasz) [reply dashboard/src/pages/storage/aips/[id]/index.vue:61]: Ah, okay. I'll revert this in a new PR.

dashboard/src/components/WorkflowHelp.vue
djjuhasz: Because I was trying to associate a label with the 'legend' table, but it's probably not the right context to use a label element based on the https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label.

jraddaoui: Yeah, and we also shouldn't jump from h2 to h5: https://www.w3.org/WAI/tutorials/page-structure/headings/

> Skipping heading ranks can be confusing and should be avoided where possible: Make sure that a h2 is not followed directly by an h4, for example.

We can still use the h2 element with the h5 class.

djjuhasz: Using an h2 element with an h5 class seems confusing to me too.  Thinking about it, I don't think a heading tag is really appropriate here in any case — this isn't a section heading in a hierarchy, it's a title for the table.  I think I'll just change this to a span element with an "h5" class to avoid issues with the semantic meaning of the tag.

PR #1208, opened 2025-05-02
Fix minor issues with status legend (refs #1160)

djjuhasz:
- Revert AIP "Workflows" heading
- Use SIP task status enum for the workflow task legend
- Use a span element for the workflow task legend title
- Update the workflow and task definitions


jraddaoui (approved):
Thanks @djjuhasz!

Issue #1186, opened 2025-04-15
Problem: Manual package deposit through MinIO doesn't support PKCE


sallain:
**Is your feature request related to a problem? Please describe.**

Some OIDC providers require applications to support PKCE (Proof Key for Code Exchange), which is an extension to the OAuth 2.0 authorization code flow that enhances security. We have discovered that MinIO doesn't support PKCE (and [there doesn't seem to be a plan to implement it](https://github.com/minio/minio/discussions/20239)). Many of the organizations that Enduro is best suited for will be security-conscious; we have already encountered this issue as a blocker for one client.

We currently use MinIO for package deposit, among other things. Users can manually deposit a single package into a MinIO bucket, which then initiates a Redis notification so that Enduro can pick up the package and start processing. We need an alternative solution that provides PKCE support for this process. 

**Describe the solution you'd like**

Implement a solution for manual package deposit that supports PKCE. The solution **must** allow the following:

* Users authenticate using their local OIDC provider and have appropriate roles/permissions to deposit packages in the correct place (different users might deposit in different partitions)
* The ingest will automatically start once the upload is complete
* The solution can handle large packages (e.g. up to 500GB or something)

It would be very nice (though not technically required) if the solution was part of the Enduro UI so that users don't have to switch interfaces.

With these things in mind, the team has suggested implementing either:

* A new page in Enduro (called _New ingest_ or _Deposit_ or something) that contains an uploader, or
* An upload widget at the top of the SIPs page

Exact details TBD

**Describe alternatives you've considered**

We thought about a few different options, explored in a Miro board here: https://miro.com/app/board/uXjVL7s6tus=/?moveToWidget=3458764625178751164&cot=14

**Additional context**

In conjunction

fiver-watson:
Proposal for initial implementation - adding a separate page with a drag and drop uploader. This option we feel adds the most flexibiliy for future changes and enhancements. 

![Image](https://github.com/user-attachments/assets/08c9bb0e-9aa4-44bf-890b-5b250a8bd84c)

djjuhasz:
I've implemented a drag-and-drop uploader using https://uppy.io/.   This is how it looks so far:

https://github.com/user-attachments/assets/40b5407f-5375-4a8d-a2ad-6191e61efbce

djjuhasz:
It's not clear from the video above, but you can add multiple files and you have to click the "Upload" button to start the upload.

jraddaoui:
[...]

Issue #1210, opened 2025-05-06
Problem: it's hard to track the status of an ingested SIP


jraddaoui:
**Is your feature request related to a problem? Please describe.**

The current implementation relies on Redis to pass an event from MinIO when the SIP upload is done, then we have a watcher that listens to those events in Redis and triggers a Temporal workflow. The SIP is not persisted in the database until the workflow starts, so it's not easy to track the status of the SIP and its ingest workflow.

Additionally, the event sent by MinIO doesn't include user information so it's not possible to relate the ingest workflow nor the SIP in Enduro to the user that made the upload in MinIO.

We have a SIP upload endpoint in Enduro's API and we are adding an uploader to the UI in #1186. However, this endpoint only proxies the upload to MinIO and the workflow is still triggered as mentioned above.

**Describe the solution you'd like**

From the upload SIP API endpoint we should be able to directly create the SIP in the database and start the ingest workflow in Temporal related to that SIP, returning the SIP UUID (see #1209) in the API response and being able to relate both the workflow and the SIP with the user that made the upload.

We should be careful about the atomicity of the operation, same as noted in https://github.com/artefactual-sdps/enduro/issues/1076#issuecomment-2756135600 for the AIP deletion workflow. Since we need to create the SIP in the endpoint to return the identifier, I'd suggest to do a manual rollback (deleting the SIP if the workflow can't be started) instead of delegating everything to Temporal as we did in the deletion workflow.

This could be an issue with large uploads where the creation of the SIP in the database fails or the Temporal workflow can't be started, requiring to upload the entire SIP again to retry the process.

**Describe alternatives you've considered**

Split the process in two different endpoints, one for the upload and creation of the SIP in the DB, another to start the workflow. This could be helpful in the future for batch uploads but the same re-upload issue still applies for the first endpoint.

**Additional context**

Discussed a few times during sync time, mentioned by Sevein in https://github.com/artefactual-sdps/enduro/pull/1016#pullrequestreview-2276752114.

The solution mentioned above could remove the need of Redis and the Enduro watcher. In fact, those could get in between triggering two workflows for the same upload if we keep using the same bucket in the upload endpoint and the watcher.

jraddaoui:
This is done. Hard to test from the dashboard, but you should see the UUID populated in the SIP table right after the upload is done. Using cURL in the terminal:

```
$ curl -F "file=@/path/to/sip.zip" http://localhost:9002/ingest/sips/upload

[...]

Issue #1209, opened 2025-05-06
Problem: the ingest domain does not use UUIDs as resource identifiers


jraddaoui:
**Is your feature request related to a problem? Please describe.**

The ingest domain uses database IDs to identify SIPs, Workflows and Tasks.

**Describe the solution you'd like**

Follow the same pattern used in the storage domain with Locations, AIPs, Workflows and Tasks, where we use a UUID as resource identifier.

**Additional context**

Initially part of #1117 and mentioned in other issues and pull requests.

PR #1211, opened 2025-05-06
Add "Upload SIPs" page to Dashboard (#1186)

djjuhasz:
- Return a "202 Created" HTTP success code from API "/ingest/sips/upload" endpoint
- Add an "Upload SIPs" page to the Dashboard using the Uppy file upload library


dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -0,0 +1,54 @@
...
+import { useRouter } from "vue-router/auto";
+
+import { getPath } from "@/client";
+import { useAuthStore } from "@/stores/auth";
+import { useLayoutStore } from "@/stores/layout";
+import IconUpload from "~icons/clarity/plus-circle-line";
+
+import "@uppy/core/dist/style.min.css";
+import "@uppy/dashboard/dist/style.css";
+import "@uppy/progress-bar/dist/style.min.css";
```
jraddaoui: I wouldn't use the minified version, it could help on development and the build process should minify them.

dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -0,0 +1,54 @@
...
+
+layoutStore.updateBreadcrumb([{ text: "Ingest" }, { text: "Upload SIPs" }]);
+
+const uppy = new Uppy({ restrictions: { maxFileSize: maxFileSize } }).use(XHR, {
+  endpoint: getPath() + "/ingest/sips/upload",
+  headers: { Authorization: `Bearer ${authStore.getUserAccessToken}` },
+  allowedMetaFields: false,
+  getResponseData: () => {
+    return { url: "" };
+  },
```
jraddaoui: Is this really needed?

dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -0,0 +1,54 @@
...
+  onAfterResponse: (xhr) => {
+    if (xhr.status === 201) {
+      setTimeout(() => {
+        router.push({
+          path: "/ingest/sips",
+        });
+      }, 500);
+    }
+  },
```
jraddaoui: This will redirect after the first response is received, but why the timeout?

dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -0,0 +1,54 @@
...
+});
```
jraddaoui, line 57: Do we want to do something like what they suggest to refresh the access token if needed?

https://uppy.io/docs/xhr-upload/#how-can-i-refresh-auth-tokens-after-they-expire

We sign out the user directly in other requests:

https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/client.ts#L68-L73

Not sure what would happen here. Would it should an error? Would it continue retrying?

dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -0,0 +1,54 @@
...
+</script>
+
+<template>
+  <div class="container-xxl">
+    <h1><IconUpload class="me-3 text-dark" />Upload SIPs</h1>
```
jraddaoui: Other pages use `class="d-flex"` in the main heading, I think that may align the icon and text better.

internal/api/design/ingest.go
```diff
@@ -214,6 +214,8 @@ var _ = Service("ingest", func() {
 			// direct access to the HTTP request body reader.
 			SkipRequestBodyEncodeDecode()
 
+			Response(StatusCreated)
```
jraddaoui: I'd use `202 Accepted` instead of `201 Created`, there is a lot to happen until the request is completed and the resources created. Even if we are not able yet to return an identifier or URL to track the status, we should be able to do so after #1209 and #1210 are completed.

jraddaoui (commented):
Looking great @djjuhasz!

dashboard/src/pages/ingest/upload/index.vue
djjuhasz: Ah, okay.

dashboard/src/pages/ingest/upload/index.vue
djjuhasz: Yes, Uppy expects a JSON response containing a "URL" field or it shows the upload as "failed". My guess is Uppy wants to provide a link to the uploaded resource.  I think we should replace this with an actual URL after #1210 is implemented.

Uppy has a lot of options, so it's possible that there's a way to turn off the expectation of a URL in the success response, but I haven't found one.

dashboard/src/pages/ingest/upload/index.vue
djjuhasz: I found the immediate redirect after the upload completes to be jarring, and I prefer the brief pause.  We can see how this tests, and adjust or eliminate the timeout if the feedback is negative.

I think once #1210 is implemented we can return a URL for the uploaded SIP and remove this redirect altogether.

dashboard/src/pages/ingest/upload/index.vue
djjuhasz: Yeah, I think this is a good idea -- I've gotten an error a few times trying to upload with an expired token.

internal/api/design/ingest.go
djjuhasz: Yeah, fair enough.  I debated 201 vs. 202 but I think you're right that 202 Accepted is more appropriate right now.

dashboard/src/pages/ingest/upload/index.vue
djjuhasz: I've implemented the `authStore.signinSilent()` method to refresh the access token if needed. This solution works for my tests instead of failing if the token is expired, as it used to do.  It's hard to manually test all the possible scenarios though, and I'm not sure how to do automated testing of how the uploader handles an expired token. :(

I tried to add a `catch()` to the token renewal process in case `authStore.signinSilent()` fails, but I didn't see any difference in the uploader behaviour, and again I'm not sure how to automate testing this functionality.

jraddaoui: Nice, thanks @djjuhasz! Looks like a good approach, I'd like to revisit how we do this in other API interactions, so we can go from this.

> I tried to add a catch() to the token renewal process in case authStore.signinSilent() fails, but I didn't see any difference in the uploader behaviour, and again I'm not sure how to automate testing this functionality.

I think the issue is that we don't return the promise from `authStore.signinSilent()`, I think we should deal with the error from the manager there, no need to do it now.

dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -0,0 +1,72 @@
...
+const maxFileSize = 500 * GiB;
+
+layoutStore.updateBreadcrumb([{ text: "Ingest" }, { text: "Upload SIPs" }]);
+
+const uppy = new Uppy({ restrictions: { maxFileSize: maxFileSize } }).use(XHR, {
+  endpoint: getPath() + "/ingest/sips/upload",
+  allowedMetaFields: false,
+  // Called again for every retry too.
+  async onBeforeRequest(xhr) {
+    if (!authStore.getUserAccessToken) {
```
jraddaoui: I'd use `authStore.isUserValid`, which also checks if the token is expired.

dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -0,0 +1,72 @@
...
+        setTimeout(() => {
+          router.push({
+            path: "/ingest/sips",
+          });
+        }, 500);
+        break;
+      // "401 Unauthorized" is returned when the auth token has expired.
+      case 401:
+        await authStore.signinSilent();
+        break;
```
jraddaoui, line 51: If we do it before request, why do we need it after response?

djjuhasz: That's the way they do it in https://uppy.io/docs/xhr-upload/#how-can-i-refresh-auth-tokens-after-they-expire

I guess theoretically the token could expire between the pre-request check and the request actually reaching the server. :shrug:

jraddaoui: But doing it after response won't fix that issue. If `onBeforeRequest` is called on retries too (as they said), it doesn't make sense to me to do it in `onAfterResponse`.

jraddaoui (approved):
Thanks @djjuhasz!

PR #1213, opened 2025-05-08
Simplify gosec setup

sevein:
This commit removes the `make gosec` target because `make lint` (which runs
`golangci-lint`) already includes gosec checks. By relying on the version of
gosec used internally by golangci-lint, results become more consistent and
reproducible.

The separate `gosec.yml` workflow has been deleted. Instead, a new gosec job
has been added to the existing `nightly.yml` workflow, using the official
`securego/gosec@master` GitHub Action. This ensures our code is still checked
against the latest version of gosec each night.


jraddaoui (approved):
LGTM, thanks @sevein!

Issue #1212, opened 2025-05-08
Problem: intermittent CI test failures due to mismatched timestamps


djjuhasz:
**Describe the bug**

The Enduro unit tests fail intermittently in the Github CI test environment due to mismatched time stamps.

**To Reproduce**

1. Create a Pull Request in Enduro

**Expected behavior**

The unit test should not fail if there is no error in the code or test code.

**Screenshots**

N/A

**Additional context**

- We use "fuzzy" time matching in our tests to allow for small variations in the expected vs. actual time. We may need to increase the allowed variation between times.
- Example failure message:

```
 FAIL: internal/storage/persistence/ent/client TestListAIPs/Returns_all_AIPs (0.03s)
    client_test.go:444: assertion failed: 
        --- aips
        +++ tt.want
          &storage.AIPs{
          	Items: storage.AIPCollection{
          		&{
          			... // 3 identical fields
          			ObjectKey:  {0x7b, 0xa9, 0xa1, 0x18, ...},
          			LocationID: nil,
        - 			CreatedAt:  "2025-05-07T23:56:38Z",
        + 			CreatedAt:  "2025-05-07T23:56:37Z",
          		},
          		&{
          			... // 3 identical fields
          			ObjectKey:  {0x48, 0x8c, 0x64, 0xcc, ...},
          			LocationID: nil,
        - 			CreatedAt:  "2025-05-07T23:56:38Z",
        + 			CreatedAt:  "2025-05-07T23:56:37Z",
          		},
          	},
          	Page: &{Limit: 20, Total: 2},
          }
```

PR #1215, opened 2025-05-08
Use constant times in storage client tests (#1212)

djjuhasz:
Prevent intermittent test failures due to time mismatches by using constant datetime values in the storage persistence client tests.


jraddaoui (approved):
Thanks @djjuhasz!

Issue #1174, opened 2025-04-03
Problem: there is no way for an operator to cancel a deletion request


fiver-watson:
**Describe the bug**

In a high traffic DPS managing many ingests and other ongoing preservation actions, occasionally an operator may accidentally request the deletion of the wrong AIP. Currently the operator cannot cancel their own request, meaning they must wait for an administrator to do so. There may be cases where the adminstrator does not learn of the mistake until **after** approving the deletion - for example, if they are out of office, and don't check their messages before resuming review. This poses a preservation risk, as once approved there is no way to undo a deletion, which will also delete any related replicas. 

**To Reproduce**

1. Log in as an operator and navigate to the view page of a stored AIP
2. Click the Delete button
3. Enter a reason for deletion and submit the request
4. Realize you have the wrong AIP and attempt to cancel the request 

**Expected behavior**

An operator should be able to cancel their own deletion request. In such a case, the behavior should be similar to the deletion operation when a request is rejected - the workflow status should be updated to "Canceled", the AIP status should return to "Stored", and the fact the operator canceled their own request should remain visible in the workflow details. 

**Additional context**

Alternatively, fully removing any sign of the worklfow after it is canceled by the requestor would be a fine outcome as well.

jraddaoui:
Hi @djjuhasz, I thought #1216 was related to #1159 and not to this issue, just checking we are on the same or a similar page.

For this issue, we'll need to know user details about the deletion requester, so I was thinking on adding a new endpoint to check if the user can review or cancel the deletion request for a giving AIP; `/aips/{uuid}/deletion-details` or something along those lines.

I was not sure about returning the user information from that endpoint and do the check in the dashboard or just do the checks directly in the endpoint to avoid returning another user `iss` and `sub` claims. I thought it would be better to do the checks in the API and just return a few boolean fields (deletion_pending, can_review, can_cancel). In the dashboard, we could check the current workflows and make the request conditionally if there is a deletion workflow pending review. I think that will make unnecessary to filter the AIP workflows.

For the actual cancellation I was not sure if we should re-use the review endpoint or add a new `deletion-cancel` endpoint. If we want to record the cancelled status, we'll still need a new deletion request status for it and we need to update the workflow signal to include more than just an `approved` boolean. So I think we could re-use the review endpoint, changing the user check based on the operation (for cancellation it must be the same user as the requester, for approval/rejection it has to be another user) and reusing the status enum to indicate the operation, instead of a boolean.

[...]

PR #1216, opened 2025-05-09
Add workflow list filters for status and type

djjuhasz:
Refs #1174.

Add parameters to the ListAIPWorkflowis API endpoint to allow filtering the results by status and type so I can use a single request to check if an AIP has an "in progress" deletion request workflow.


jraddaoui (approved):
Neat, thanks @djjuhasz!

PR #1217, opened 2025-05-09
Use UUIDs as SIP identifiers

jraddaoui:
- Add `uuid` column to `sip` DB table.
- Use SIP UUID instead of ID in the API design.
- Add UUID to datatypes SIP and SIPUUID to Workflow.
- Update ingest service and Ent client.
- Remove unused SetStatusPending from ingest service.
- Fix conversion and make sure SIPUUID is set on workflow.
- Generate SIP UUID in the ingest workflow request.
- Use SIP UUID in the ingest workflow Temporal ID.
- Use SIP UUID in processing workflow and local activities.
- Dashboard:
  - Use SIP UUID for links.
  - Use SIP UUID instead of AIP UUID in SIPs table.
  - Show SIP UUID in index page.
  - Replace SIP id references.
  - Fix websocket monitor.
- Fix genpkgs script.

Refs #1209.

PS. There will be another PR for Workflows and Tasks.


dashboard/src/monitor.ts
```diff
@@ -59,18 +59,24 @@ function handleWorkflowCreated(data: unknown) {
   const store = useSipStore();
 
   // Ignore event if it does not relate to the current SIP.
-  if (store.current?.id != event.item.sipId) return;
+  console.log(store.current?.uuid);
```
djjuhasz: Are the `console.log()` calls in this file left over from debugging?

internal/ingest/convert.go
```diff
@@ -18,13 +18,8 @@ import (
 )
 
 func sipTogoaingestCreatedEvent(s *datatypes.SIP) *goaingest.SIPCreatedEvent {
```
djjuhasz: Yikes, can we change the name of this function to follow our Go MixedCaps naming conventions?

internal/ingest/ingest.go
```diff
@@ -28,14 +29,13 @@ type Service interface {
 	CreateSIP(context.Context, *datatypes.SIP) error
 	UpdateSIP(
 		ctx context.Context,
-		ID int,
+		id uuid.UUID,
```
djjuhasz, line 32: Why "id" here instead of "uuid" like `goaingest` and `datatypes` fields?

internal/persistence/ent/client/workflow.go
```diff
@@ -42,8 +42,10 @@ func (c *client) CreateWorkflow(ctx context.Context, w *datatypes.Workflow) erro
 		return newDBErrorWithDetails(err, "create workflow")
 	}
 
-	// Update value of task with data from DB (e.g. ID).
+	// Update value of workflow with data from DB (e.g. ID). Maintain SIP UUID.
+	sipUUID := w.SIPUUID
 	*w = *convertWorkflow(r)
+	w.SIPUUID = sipUUID
```
djjuhasz: Hmm, it seems weird to have to manually preserve the SIPUUID like this. :confused: If `w.SIPUUID` doesn't match the DB value of the related `sip.UUID` here, it seems like this could lead to trouble.

internal/workflow/local_activities.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 type createSIPLocalActivityParams struct {
+	UUID   uuid.UUID
```
djjuhasz, line 17: Isn't the `sip.UUID` generated by the DB and immutable?  If so then what is this UUID value used for?

internal/workflow/processing_test.go
```diff
@@ -316,6 +317,7 @@ func (s *ProcessingWorkflowTestSuite) TestConfirmation() {
 			Status:     enums.WorkflowStatusInProgress,
 			StartedAt:  startTime,
 			SIPID:      1,
+			SIPUUID:    sipUUID,
```
djjuhasz, line 319: It seems redundant and potentially contradictory to pass SIPID and SIPUUID here. :confused:

djjuhasz (changes requested):
Looking good @jraddaoui.  I noted a few questions and minor change requests.

dashboard/src/monitor.ts
jraddaoui: Oops, that's right, thanks!

internal/ingest/ingest.go
jraddaoui: I don't like using `uuid` as variable names because it can conflict with the  `github.com/google/uuid` package.

internal/persistence/ent/client/workflow.go
jraddaoui: After the latest commit it should be closer to what we have in the storage domain, using only the SIPUUID and a transaction. It involved changes from this function up to the processing workflow.

internal/workflow/local_activities.go
jraddaoui: The UUID is generated in the request:

https://github.com/artefactual-sdps/enduro/blob/dev/issue-1209-sip-uuid/cmd/enduro/main.go#L403

I thought to do it this way so we can do "the same" later when the workflow is started from the upload endpoint, to be able to return the SIP identifier (re: #1210).

internal/workflow/processing_test.go
jraddaoui: Fixed in the latest commit.

internal/storage/persistence/ent/client/client.go
```diff
@@ -185,7 +185,10 @@ func (c *Client) UpdateAIPLocationID(ctx context.Context, aipID, locationID uuid
 	return nil
 }
 
-func (c *Client) ListWorkflows(ctx context.Context, f *persistence.WorkflowFilter) (goastorage.AIPWorkflowCollection, error) {
+func (c *Client) ListWorkflows(
+	ctx context.Context,
+	f *persistence.WorkflowFilter,
+) (goastorage.AIPWorkflowCollection, error) {
```
jraddaoui: Looks like golangci-lint is not properly failing the CI workflow again after the latest changes.

internal/db/migrations/20250509030046_restore_sip_status_index.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Modify "sip" table
+ALTER TABLE `sip` ADD INDEX `sip_status_idx` (`status`);
```
jraddaoui, line 2: You probably noticed, this was missing from a previous PR, I'll keep it in a separate commit and squash the other two.

jraddaoui:

internal/workflow/local_activities.go
djjuhasz: Ah, okay. I misunderstood the "immutable" type in ent, I thought it was generating a new UUID and then preventing any changes to it.  I see now that the `sip.UUID` can be set when the database row is created.

djjuhasz (approved):
LGTM! Thanks @jraddaoui. :+1:

internal/storage/persistence/ent/client/client.go
djjuhasz: Fixed in 95c9b62 :)

PR #1223, opened 2025-05-13
Use `bine sync`

sevein:
`bine sync` installs all binaries defined in `.bine.json`.

This saves us from having to track the list of tools twice.


Makefile
```diff
@@ -37,8 +37,9 @@ TEST_PACKAGES = $(filter-out $(IGNORED_PACKAGES),$(PACKAGES))
 TEST_IGNORED_PACKAGES = $(filter $(IGNORED_PACKAGES),$(PACKAGES))
 
 # Configure bine.
-include hack/make/tools.mk
 export PATH := $(shell go tool bine path):$(PATH)
+tool-%:
+	@go tool bine get $* 1> /dev/null
```
jraddaoui: I think the targets are sorted alphabetically in this Makefile, it would be nice if we could move this just above `tools`.

jraddaoui (changes requested):
Nice, thanks @sevein!

Makefile
sevein: I can do that! Thanks.

jraddaoui (approved):
Thanks! Shouldn't it be above? ASCIIbetical, `-` before `s` ;)

PR #1224, opened 2025-05-14
Fix usage of golangci-lint

sevein:
Previously we relied on `golangci-lint fmt` in CI, but it only reformats code without ever returning a non‐zero exit code on violations. By switching to `golangci-lint run`, we get both formatter and linter checks in one command (TIL!).

We use `--fix` locally (via make lint) to automatically correct issues during development, but disable it in CI (`--fix=0`) so that the pipeline reports problems without mutating the code.

We can still use `make fmt` (equals to `golangci-lint fmt`) if you need faster formatting, as it does not invoke the linters.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,11 +28,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-golangci-lint-cache-
       - name: Run golangci-lint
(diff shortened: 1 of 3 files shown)


jraddaoui (approved):
Thanks for looking into this @sevein!

PR #1225, opened 2025-05-19
Simplify processing workflow request

jraddaoui:
- Remove configuration values already accessible from the workflow.
- Remove SIPID, trust only SIPUUID and remove unused SIP update.

Refs #1210.

### DIFF
--- a/internal/ingest/workflows.go
+++ b/internal/ingest/workflows.go
@@ -24,10 +24,6 @@ type ReviewPerformedSignal struct {
 }
 
 type ProcessingWorkflowRequest struct {
-	// The zero value represents a new SIP. It can be used to indicate
-	// an existing SIP in retries.
-	SIPID int
-
 	// Unique identifier of the SIP.
 	SIPUUID uuid.UUID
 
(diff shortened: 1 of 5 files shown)


internal/ingest/workflows.go
```diff
@@ -53,29 +49,20 @@ type ProcessingWorkflowRequest struct {
...
-	// TransferDeadline is the maximum time to wait for a transfer to complete.
-	// Set to zero for no deadline.
-	TransferDeadline time.Duration
 }
 
-func InitProcessingWorkflow(ctx context.Context, tc temporalsdk_client.Client, req *ProcessingWorkflowRequest) error {
+func InitProcessingWorkflow(
+	ctx context.Context,
+	tc temporalsdk_client.Client,
+	tq string,
```
djjuhasz: I think it's worth calling this variable "taskQueue" to make its purpose clear — it's part of the function signature and the "string" type doesn't provide any hints about the argument's purpose.

djjuhasz (approved):
LGTM @jraddaoui. I just had one quibble about an argument name. :+1:

Issue #1218, opened 2025-05-09
Enhancement: Add analysis workflow


fiver-watson:
This high-level issue describes a new workflow to be added to Enduro. 

In some cases, content may not have been appraised, yet information professionals may need to establish some basic bit-level preservation and intellectual control over the materials until they can be more thorough reviewed and selected from. 

To support such cases, we propose **adding a new Analysis workflow to Enduro**.  The purpose of an initial analysis workflow is to: 

- Document the state of a given package at receipt
- Provide a high-level overview of the contents of a package that can support appraisal, selection, arrangement, and deduplication
- Generate a report that captures key information about the package contents for later review and selection
- Generate checksums for all submitted content that can then be used to verify integrity until final preservation decisions are made

## Analysis workflow overview

A detailed proposal document has been prepared here: 

* [Analysis workflow proposal overview](https://docs.google.com/document/d/1hX7vPMZfO9gJbVjz4HCwH9gqF4C7K9bh4wSMPjiZMg4/edit?tab=t.0)

What follows is a summary. 

**Enduro operators should be able to:** 

- Upload a package
- Initiate an analysis workflow
- Download a report of the analysis findings

This report can then be used to assist in appraisal and selection prior to preservation ingest. 

Initial analysis workflow activities will include: 

- Receipt
- Package checksum calculation
- Extraction
- File checksum list generation
- File identification via Sigfried
- Duplicate checking among package files
- Directory tree documentation
- Report generation

Depending on development time and priorities during our iterative process, this may further include: 

- Bagging of content
- Storage in a new location

### Workflow initiation

Packages submitted for analysis  must be zippped. The workflow could theoretically be manually launched by an operator either by: 

- Adding a zipped package to a new secondary "analysis" watched minIO bucket, or
- Using the Enduro UI uploader, and selecting the target workflow from a new upload configuration option

In this initial implementation, we will let the sponsoring client determine the preferred method. In the future, it would be possible to support both methods (in addition to others, such as launch via API; watched filesystem directory, etc). 

If they choose the Enduro UI uploader option, here is a simple mockup of the new configuration option we would propose adding: 

![Image](https://github.com/user-attachments/assets/ab9ed4a8-bfe5-4241-b520-1b89cacc5369)

## Development cadence and deliverables

This issue ticket can either be reused, or treated as an epic and linked to other issues as needed. We have proposed organizing this work into a series of deliverable-focused sprints : 

[...]

PR #1226, opened 2025-05-19
Allow setting processing workflow type

jraddaoui:
- Allow setting processing workflow type from watcher configuration.
- Add watcher configuration defaults and validation.
- Remove AutoApproveAIP option from workflow request and pass type.
- Consider workflow type in processing workflow.
- Remove `unspecified` value and update workflow type enum to a string.
- Update DB schema, generate Ent code and add migration.
- Regenerate API and dashboard client code.

Refs #1218.


dashboard/src/main.ts
```diff
@@ -50,9 +50,9 @@ app.config.globalProperties.$filters = {
       | api.EnduroStorageAipWorkflowTypeEnum,
   ) {
     switch (value) {
-      case api.EnduroIngestSipWorkflowTypeEnum.CreateAip:
+      case api.EnduroIngestSipWorkflowTypeEnum.Aip:
         return "Create AIP";
-      case api.EnduroIngestSipWorkflowTypeEnum.CreateAndReviewAip:
+      case api.EnduroIngestSipWorkflowTypeEnum.AndReviewAip:
```
jraddaoui: Don't ask me why ;)

I tried re-generating the client code more than once, but those are the types it chooses.

dashboard/src/pages/ingest/sips/[id]/index.vue
```diff
@@ -23,12 +22,7 @@ const toggleHelp = () => {
...
 
 const createAipWorkflow = computed(
-  () =>
-    sipStore.currentWorkflows?.workflows?.filter(
-      (w) =>
-        w.type === api.EnduroIngestSipWorkflowTypeEnum.CreateAip ||
-        w.type === api.EnduroIngestSipWorkflowTypeEnum.CreateAndReviewAip,
-    )[0],
+  () => sipStore.currentWorkflows?.workflows?.[0],
 );
```
Comment (jraddaoui) [inline dashboard/src/pages/ingest/sips/[id]/index.vue:26]: Kind of unnecessary, it just makes the template a bit cleaner.

internal/db/migrations/atlas.sum
```diff
@@ -9,4 +9,5 @@ h1:4nZSKvUlpsK5RT4Ctn1jUnwvnDQRGaEYafWJxJ7s95w=
 20250402173826_remove_workflow_type.up.sql h1:P0w5tngLlCsfyYm1hW8LqdxDgVAYWIf9QM7oTLu5mJA=
 20250418000742_update_sip_statuses.up.sql h1:FUbEzeO8+t0FVImqnOWApQqKzxulq2GY4urMSsJlbc0=
 20250509030046_restore_sip_status_index.up.sql h1:8iKBW+nC9mgEs6PqQJ4Frq1IGJOjedAyhm3dO27yVZw=
-20250509030521_add_sip_uuid_column.up.sql h1:z5zEsQbzL/nz4q7MgBdkiQa+KKB5rPopYrte0lRrst4=
+20250509030521_add_sip_uuid_column.up.sql h1:rtcC2HuAOJZI+/7ZkWrlrTn+wkrZ34BRqLK5bJRo3h8=
```
jraddaoui, line 12: Looks like the previous hash was not right, I'll think about adding CI to check for missing migrations and things like this.

djjuhasz: I think adding a CI check for a mismatched checksum is a good idea.  Atlas has a [migrate validate command](https://atlasgo.io/cli-reference#atlas-migrate-validate) that we could use.

internal/watcher/config.go
```diff
@@ -13,6 +17,49 @@ type Config struct {
...
+		}
+	}
+	for _, minio := range c.Minio {
+		if minio != nil && minio.WorkflowType == "" {
+			minio.WorkflowType = enums.WorkflowTypeCreateAip
+		}
+	}
+}
+
+func (c Config) Validate() error {
```
djjuhasz, line 36: Nice! I like the config validation. :+1:

dashboard/src/main.ts
djjuhasz: I figured this out: the OpenAPI generator defaults to removing identical prefixes from enum names ("removeEnumValuePrefix").  I created a new branch with the fix so I didn't mess with your PR: https://github.com/artefactual-sdps/enduro/compare/dev/issue-1218-fix-dashboard-workflow-types

Feel free to merge the fix if that works for you.

djjuhasz (approved):
This looks really good @jraddaoui, thanks.

I figured out why the OpenAPI generator was removing the "create" prefix from the ingest workflow type enum keys.  It's not a critical fix, but I do think it makes the dashboard code easier to understand.

dashboard/src/main.ts
jraddaoui: Niiice, thanks for the detective work!

PR #1227, opened 2025-05-20
Add "Cancel deletion" API endpoint (refs #1174)

djjuhasz:
- Add "storage/aips/{uuid}/deletion-cancel" API endpoint
- Add a "test" parameter to the endpoint to allow testing permission to cancel the deletion without actually cancelling it

### DIFF
--- a/internal/storage/workflow.go
+++ b/internal/storage/workflow.go
@@ -8,6 +8,8 @@ import (
 	"github.com/google/uuid"
 	temporalsdk_api_enums "go.temporal.io/api/enums/v1"
 	temporalsdk_client "go.temporal.io/sdk/client"
+
+	"github.com/artefactual-sdps/enduro/internal/storage/enums"
 )
 
 const (
@@ -16,7 +18,7 @@ const (
 	StorageDeleteWorkflowName           = "storage-delete-workflow"
 	StorageUploadWorkflowName           = "storage-upload-workflow"
 	StorageMoveWorkflowName             = "storage-move-workflow"
-	DeletionReviewedSignalName          = "deletion-reviewed-signal"
+	DeletionDecisionSignalName          = "deletion-decision-signal"
 	UploadDoneSignalName                = "upload-done-signal"
 )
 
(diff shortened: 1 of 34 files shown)


internal/storage/workflow.go
```diff
@@ -16,7 +18,7 @@ const (
 	StorageDeleteWorkflowName           = "storage-delete-workflow"
 	StorageUploadWorkflowName           = "storage-upload-workflow"
 	StorageMoveWorkflowName             = "storage-move-workflow"
-	DeletionReviewedSignalName          = "deletion-reviewed-signal"
+	DeletionDecisionSignalName          = "deletion-request-signal"
```
jraddaoui: Why not `deletion-decision-signal`?

internal/storage/workflows/delete.go
```diff
@@ -251,20 +256,19 @@ func (w *StorageDeleteWorkflow) review(
 		return nil, err
 	}
 
-	// Review DeletionRequest.
-	activityOpts = localActivityOptions(ctx)
+	// Update DeletionRequest.
 	err = temporalsdk_workflow.ExecuteLocalActivity(
 		activityOpts,
 		storage.ReviewDeletionRequestLocalActivity,
```
jraddaoui: It would make sense to change the activity name too, not a big deal of course.

jraddaoui (changes requested):
Thanks @djjuhasz, looking great, really nice!

After seeing this, I thought about using a HEAD request, but I think it doesn't really fit here. The only thing I see missing is the Ent generated code and a migration, not sure how the proper status is getting into the DB without them.

internal/storage/workflow.go
djjuhasz: Oops, good catch.  I changed the name twice, and missed updating this to "decision".

djjuhasz:
@jraddaoui what do you think about changing the API param "test" to "check" (as in "check permissions") instead?

jraddaoui:
> @jraddaoui what do you think about changing the API param "test" to "check" (as in "check permissions") instead?

I think that's better, yeah.

jraddaoui:
I think you missed it in the main comment, still missing the Ent generated code and the migration, right? ;)

djjuhasz:
@jraddaoui yes, I missed your comment about the ent generation and DB migrations. Thanks for repeating it. :)

djjuhasz:
@jraddaoui ready for another look. I'll squash all the commits before the final merge, I left them separate for review.

jraddaoui (approved):
Neat!

PR #1232, opened 2025-05-22
Add delete SIP method to persistence

jraddaoui:
Will be used in case of an error starting the ingest workflow from the upload endpoint, to do a manual rollback of the SIP creation.

Refs #1210.

### DIFF
--- a/internal/persistence/ent/client/sip_test.go
+++ b/internal/persistence/ent/client/sip_test.go
@@ -1,7 +1,6 @@
 package entclient_test
 
 import (
-	"context"
 	"database/sql"
 	"fmt"
 	"testing"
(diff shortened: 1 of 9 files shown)


internal/persistence/ent/client/sip_test.go
```diff
@@ -268,6 +268,55 @@ func TestUpdateSIP(t *testing.T) {
...
+			name:    "Fails to delete a missing SIP",
+			id:      12345,
+			wantErr: "not found error: db: sip not found: delete SIP",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			client, svc := setUpClient(t, logr.Discard())
+			ctx := context.Background()
```
djjuhasz: I noticed a note in https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize about using `t.Context()` instead of `context.WithCancel()` in test, and I think it would probably be better for us to use `t.Context()` in our tests.  I think the advantage to `t.Context()` is better shutdown behaviour if the test suite panics or exits suddenly.

djjuhasz (approved):
Looks great @jraddaoui.  I added a note about using `t.Context()` in the tests (which is new to me) but it's optional.

PR #1233, opened 2025-05-23
Update processing workflow request

jraddaoui:
A couple more changes to the workflow request:

- Remove stripTopLevelDir configuration:

No longer used in the processing workflow, remove it from the watcher
configuration and the workflow request.

- Add SIPName to processing workflow request :

Allows to differnetiate between the object key and the SIP name. For now
they are the same value in the existing workflow request, but we'll need
to make the distinction when the workflow is also started from the API.

Refs #1210.


djjuhasz (approved):
LGTM @jraddaoui. Thanks! :+1:

Issue #1231, opened 2025-05-22
Enhancement: add support for Azure storage buckets


joel-simpson:
**Is your feature request related to a problem? Please describe.**
MSF run their environment in the Azure cloud. They already manage their content in Azure buckets. They would like Enduro to be able to connect to their Azure buckets to receive or ingest SIPs. 

**Describe the solution you'd like**

We are currently working on replacing the need for Minio and instead be able to configure S3 buckets for storage. See https://github.com/artefactual-sdps/enduro/issues/1210

When this is done, we'd like to be able to configure Azure buckets in addition to just generic S3 buckets. 

**Describe alternatives you've considered**
none so far. 

**Additional context**

sallain:
Closing unless the client provides feedback.

scollazo:
Reopening, because this is not working with the client authentication settings.

When using an azure blob container for enduro, the upload fails with an HTTP 500 error, and the developer tools show the following response:

```
RESPONSE 403: 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
ERROR CODE: AuthenticationFailed
```


For reference, for using rclone in the client environment, we define the following azure variables:

 - account: <string>
 - tenant: < uuid>
 - client_id: <uuid>
 - client_secret: <string>


But enduro seems to allow only for storageaccount and storagekey .

scollazo:
Authentication works with Shared keys instead of Oauth. Closing (Thanks @jraddaoui ! )

Issue #1202, opened 2025-04-30
Problem: Downloading failed SIPs/PIPs through MinIO doesn't support PKCE


sallain:
**Is your feature request related to a problem? Please describe.**

Some OIDC providers require applications to support PKCE (Proof Key for Code Exchange), which is an extension to the OAuth 2.0 authorization code flow that enhances security. We have discovered that MinIO doesn't support PKCE (and https://github.com/minio/minio/discussions/20239). Many of the organizations that Enduro is best suited for will be security-conscious; we have already encountered this issue as a blocker for one client.

We currently use MinIO as a place to hold packages that have failed - SIPs that have failed an Ingest workflow (in the _failed-sips_ bucket) or PIPs that have failed Archivematica (in the _failed-pips_ bucket). This makes it easy for users to retrieve the packages and inspect them.

**Describe the solution you'd like**

Implement a solution for failed package download that supports PKCE. The solution *must* allow the following:

* Users authenticate using their local OIDC provider and have appropriate roles/permissions to download failed packages
* There is some way of differentiating between a failed SIP and a failed PIP (currently the MinIO bucket is the best indicator of this)
* The solution can handle large packages (e.g. up to 500GB or something)

It would be very nice (though not technically required) if the solution was part of the Enduro UI so that users don't have to switch interfaces.

One initial thought is to add a download link on the SIP detail page (better wireframe to come, promise):

![Image](https://github.com/user-attachments/assets/fd2a1480-3d03-4e63-8469-b7873bd0ed93)

**Describe alternatives you've considered**

**Additional context**

See also #1186

fiver-watson:
Adding some initial mockups after some discussion today (2025-05-01): 

Basically, we decided that in previous mockups (including the rework of the top-page information on issue #955), the previously proposed "workflow summary" widget on the right side of the Summary area on view pages was pretty rendundant - especially since in issue #1166 we propose adding the workflow ID to the workflow header (making most of the rest of the information in the previous "workflow summary" redundant. 

Instead, the right column of SIP pages will get a new widget for **Related packages**. For AIPs, this can be similar to what already exists - an AIP name and a "View" button that will take the user to a view page in the storage domain. Further, if and when we finally address SIP splitting, this widget could potentially list and link to more than 1 AIP. 

For failed packages, depending on the stage they fail at, it can show the failed PIP or SIP, with an option to download it. If the minIO API allows for it, this could also include a DELETE option for failed packages, to remove them from the temp storage. 

[...]

PR #1234, opened 2025-05-23
Create SIP and start workflow from upload

jraddaoui:
- Change the upload bucket to an internal bucket:
  - Update and move configuration.
  - Consider the new bucket in dev env.
- Update upload endpoint:
  - Use SIP prefix plus SIP UUID as object key.
  - Create SIP in persistence, if it fails delete uploaded SIP.
  - Start workflow, if it fails delete uploaded and persistence SIP.
  - Return SIP UUID.
- Update processing workflow:
  - Only persist SIP initially if workflow is started from a watcher.
  - Use temporal-activities/bucketdownload activity to fetch the
    SIP when the request doesn't come from a watcher.

Refs #1210.


internal/api/design/ingest.go
```diff
@@ -193,6 +193,11 @@ var _ = Service("ingest", func() {
 			Token("token", String)
 		})
 
+		Result(func() {
+			AttributeUUID("uuid", "Identifier of uploaded SIP")
+			Required("uuid")
+		})
+
```
jraddaoui, line 200: This is the only change I'm making in the API, to return the SIP UUID. However, there seems to be some changes from the cancel deletion request missing in the dashboard generated client. I guess those are also included in you recent PR, so I'll remove them if we merge that first.

internal/ingest/ingest.go
```diff
@@ -21,6 +22,9 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/persistence"
 )
 
+// Prefix used to store the SIP in the internal bucket after upload.
+const SIPPrefix = "SIP_"
```
jraddaoui, line 26: I'll work in the failed SIPs/PIPs after this. I may make this configurable in that PR, as part of the internal bucket configuration.

internal/workflow/processing.go
```diff
@@ -162,8 +163,8 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
 	// Create the initial workflow state.
 	state := newWorkflowState(req)
 
-	// Persist the SIP as early as possible.
-	{
+	// Persist the SIP as early as possible if the request comes from a watcher.
+	if state.req.WatcherName != "" {
```
jraddaoui, line 167: I thought about adding a boolean flag to the workflow request to indicate if it comes from the upload endpoint and the SIP is placed in the internal bucket, but looking at the watcher name makes sense to me for the current request options. Happy to change that if you think it's worth it already.

jraddaoui:
Ready for code review @djjuhasz. @DanielCosme, I've also added you as reviewer as this is related to #1231 as we spoke today. Also happy to start getting your input in all PRs, so I'll start pinging you both on all of them.

internal/config/config_test.go
```diff
@@ -97,3 +98,50 @@ func TestConfig(t *testing.T) {
...
+				URL:    "s3blob://my-bucket",
+				Bucket: "my-bucket",
+				Region: "planet-earth",
+			},
+		}
+		err := c.Validate()
+		assert.ErrorContains(
+			t,
+			err,
+			"URL and rest of the [internalBucket] configuration options are mutually exclusive",
```
djjuhasz: I'd rewrite this as "the [internalBucket] URL option and the other configuration options are mutually exclusive" or similar.

internal/ingest/ingest_test.go
```diff
@@ -21,24 +21,40 @@ import (
 	persistence_fake "github.com/artefactual-sdps/enduro/internal/persistence/fake"
 )
 
-func testSvc(t *testing.T, b *blob.Bucket, s int64) (ingest.Service, *persistence_fake.MockService) {
+// io.Reader used as number generator for making UUIDs predictable
+type staticRand struct{}
```
djjuhasz: I think we usually use `rand.New(rand.NewSource(1))` (or similar) for a deterministic number generator. Is there a reason to create your own number generator here?

internal/workflow/processing_test.go
```diff
@@ -1631,3 +1637,99 @@ func (s *ProcessingWorkflowTestSuite) TestFailedPIPAM() {
...
+
+func (s *ProcessingWorkflowTestSuite) TestInternalUpload() {
+	cfg := config.Configuration{
+		A3m:          a3m.Config{ShareDir: s.CreateTransferDir()},
+		Preservation: pres.Config{TaskQueue: temporal.A3mWorkerTaskQueue},
+		Storage:      storage.Config{DefaultPermanentLocationID: locationID},
+	}
+	s.SetupWorkflowTest(cfg)
+
+	sipName := "name.zip"
```
djjuhasz, line 1649: Nitpick: I know the actual text is irrelevant but it seems weird that a SIP name would have a "zip" extension. Maybe just "name" would be clearer?

internal/workflow/processing_test.go
```diff
@@ -1631,3 +1637,99 @@ func (s *ProcessingWorkflowTestSuite) TestFailedPIPAM() {
...
+		bucketdownload.Name,
+		sessionCtx,
+		&bucketdownload.Params{
+			Key:     key,
+			DirPath: cfg.Preprocessing.SharedPath,
+		},
+	).Return(&bucketdownload.Result{FilePath: tempPath + "/" + key}, nil)
+
+	// Everything else should be the same as other workflows after
+	// download, so just fail the workflow on extration.
```
djjuhasz, line 1686: This is a nice solution to minimize duplicate test code. :+1:

djjuhasz (approved):
This looks good to me @jraddaoui. I added a few minor notes, but none of them are blockers.

internal/workflow/processing_test.go
jraddaoui: I was thinking about this when I wrote [this comment](https://github.com/artefactual-sdps/enduro/issues/1202#issuecomment-2913412086) and, based on the responses there, we may want to change this in the future. But right now we save the SIP name like that, we use the object key from watched buckets and the uploaded filename from the upload endpoint, both usually contain the file extension.

internal/ingest/ingest_test.go
jraddaoui: Thanks @djjuhasz! I just copied it from storage, I'll replace it ;)

PR #1236, opened 2025-05-23
Allow cancelling a deletion request

### LINKED ISSUE #1174: Problem: there is no way for an operator to cancel a deletion request

djjuhasz:
- Fix a bug with the "List AIP Workflows" API endpoint
- Allow a user to cancel their deletion request


djjuhasz:
Here's what the "Cancel" button looks like on the deletion request dialog:
image

Note that the "Delete" button is hidden in the AIP location "card".  Also note that the deletion request "Approve" and "Reject" buttons are not shown when "Cancel" is shown as the review and cancel actions require mutually exclusive permissions.

(The black "V" icon hovering over the review dialog is from the Vite debug tool, ignore it)

dashboard/src/components/AipDeletionReviewAlert.vue
```diff
@@ -20,12 +34,19 @@ const review = async (approved: boolean) => {
     <p class="line-break" v-html="addEmailLinks(note)"></p>
     <hr />
     <div class="d-flex flex-wrap gap-2">
-      <button class="btn btn-success" type="button" @click="review(true)">
-        Approve
-      </button>
-      <button class="btn btn-danger" type="button" @click="review(false)">
-        Reject
-      </button>
+      <template v-if="canCancel">
```
jraddaoui: The `v-if` could be placed in the button in this case.

dashboard/src/components/AipLocationCard.vue
```diff
@@ -90,11 +90,13 @@ const requestDeletion = async () => {
...
         <button
-          v-if="authStore.checkAttributes(['storage:aips:deletion:request'])"
+          v-if="
+            authStore.checkAttributes(['storage:aips:deletion:request']) &&
+            aipStore.isStored
+          "
           type="button"
           class="btn btn-primary btn-sm"
           @click="requestDeletion"
-          :disabled="!aipStore.isStored"
```
jraddaoui, line 97: I prefer to show/hide based on permissions and disable/enable based on the status, it's an action the user can perform, but not right now.

dashboard/src/pages/storage/aips/[id].vue
```diff
@@ -15,13 +15,12 @@ const router = useRouter();
...
-const { execute, error } = useAsyncState(async function () {
-  aipStore.fetchCurrent(route.params.id.toString()).then(() => {
-    if (authStore.checkAttributes(["storage:aips:workflows:list"])) {
-      aipStore.fetchWorkflows(route.params.id.toString());
-    }
-  });
-}, null);
+const { execute, error } = useAsyncState(
+  aipStore.fetchCurrent(route.params.id.toString()).catch((err) => {
+    error.value = err;
```
Comment (jraddaoui) [inline dashboard/src/pages/storage/aips/[id].vue:20]: I guess this is what populates the error in the alert, nice!

Reading the [PageLoadingAlert.vue](https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/components/PageLoadingAlert.vue#L33-L34) component. Do you think it would make more sense to change the wording to use "resource" instead of "page"?

dashboard/src/stores/aip.ts
```diff
@@ -66,25 +66,53 @@ export const useAipStore = defineStore("aip", {
...
+          ]);
+        })
+        .catch((err) => {
+          console.error("Error fetching AIP", err);
+          this.current = null;
+          this.currentWorkflows = null;
+          layoutStore.updateBreadcrumb([
+            { text: "Storage" },
+            { route: router.resolve("/storage/aips/"), text: "AIPs" },
+          ]);
```
jraddaoui: It would be nice to show something in the breadcrumb that is not the same as the AIPs list page. You are checking the response error so we could even show "Storage / AIPs / Not found" or "Storage / AIPs / Unknown" (or whatever) if we want to make the distinction.

It should be possible to do so in a `.finally()` call over the promise, using an external variable and setting its value from `.then()` and `catch()`.

dashboard/src/stores/aip.ts
```diff
@@ -66,25 +66,53 @@ export const useAipStore = defineStore("aip", {
...
+
+          // Update breadcrumb. TODO: should this be done in the component?
+          layoutStore.updateBreadcrumb([
+            { text: "Storage" },
+            { route: router.resolve("/storage/aips/"), text: "AIPs" },
+            { text: this.current.name },
+          ]);
+        })
+        .catch((err) => {
+          console.error("Error fetching AIP", err);
```
jraddaoui: Is not the response error shown in the console already?

dashboard/src/stores/aip.ts
```diff
@@ -66,25 +66,53 @@ export const useAipStore = defineStore("aip", {
...
-      layoutStore.updateBreadcrumb([
-        { text: "Storage" },
-        { route: router.resolve("/storage/aips/"), text: "AIPs" },
-        { text: this.current.name },
-      ]);
+      return client.storage
+        .storageShowAip({ uuid: id })
+        .then((resp) => {
+          this.current = resp;
+          this.fetchWorkflows(id);
```
jraddaoui: Should we also throw/return the error from `fetchWorkflows` and consider it here?

dashboard/src/stores/aip.ts
```diff
@@ -66,25 +66,53 @@ export const useAipStore = defineStore("aip", {
...
         .then((resp) => {
-          resp.workflows?.reverse();
-          return resp;
+          if (resp && resp.workflows) {
+            resp.workflows.reverse();
+          }
+          this.currentWorkflows = resp;
+        })
+        .catch((err) => {
+          console.error("Error fetching workflows", err);
```
jraddaoui: Same, do we really need to log this? I always get confused with the way browsers log response errors, I just want to avoid their duplication.

dashboard/src/stores/aip.ts
```diff
@@ -201,6 +229,57 @@ export const useAipStore = defineStore("aip", {
...
+        this.fetchCurrent(this.current.uuid).then(() => {
+          if (this.current?.status == api.EnduroStorageAipStatusEnum.Stored) {
+            clearInterval(intID);
+            return;
+          }
+        });
+      }, 1000);
+    },
+    async canCancelDeletion(): Promise<boolean> {
+      if (!this.current) return false;
```
jraddaoui, line 251: Shouldn't this be returning a promise?

dashboard/src/stores/aip.ts
```diff
@@ -201,6 +229,57 @@ export const useAipStore = defineStore("aip", {
...
+              return false;
+            }
+          }
+
+          console.error(
+            "Error checking user authorization to cancel deletion:",
+            err.message,
+          );
+          return false;
+        });
```
jraddaoui: Returning false in both cases. If the console log is still needed (based on the comments above), this could be:

```ts
if (!(err instanceof ResponseError) || err.response.status != 403) {
  console.error(
    "Error checking user authorization to cancel deletion:",
    err.message,
  );
}
return false;
```

Not a big deal and it's probably more readable the way it is.

hack/kube/components/dev/keycloak.yaml
```diff
@@ -122,11 +122,13 @@ data:
               "storage:locations:list",
               "storage:locations:aips:list",
               "storage:locations:read",
+              "storage:aips:deletion:request",
```
jraddaoui, line 125: Thanks! We should give it `storage:aips:deletion:request` permission too.

jraddaoui (commented):
Thanks @djjuhasz, it's looking great! I've added a few minor comments, still need to look at the tests.

dashboard/src/components/AipDeletionReviewAlert.vue
djjuhasz: It's one button "Cancel" vs. two buttons ("Approve", "Reject") in the else branch.

dashboard/src/components/AipLocationCard.vue
djjuhasz: If you can't click a button then what is the purpose of showing it to the user?

I don't like disabled buttons in most cases as it raises the question "why is this disabled?" and the answer is often not clear.  I think that is definitely a problem here as I don't think it's clear that it's the status of AIP that prevents deleting it.

dashboard/src/pages/storage/aips/[id].vue
Comment (djjuhasz) [reply dashboard/src/pages/storage/aips/[id].vue:20]: Hmm, I don't know. To me "Couldn't load resource" seems more technical than "Couldn't load page" and may put off some users.  Maybe we can get away with being more technical in the Enduro context because it's not a public UI, but I'd be curious to hear the opinions of @fiver-watson and @sallain.

Here's what the error looks like now if the "UUID" parameter is not a valid UUID:
image

Here's what it looks like if the UUID is valid, but there is no AIP with that UUID (not found):
image

Comment (djjuhasz) [reply dashboard/src/pages/storage/aips/[id].vue:20]: I think it's also worth noting that in both of these cases the "Retry" button is never going to resolve the problem, so we may want to rethink that.

There is also a special case in the "PageViewAlert.vue" component for a "404 Not found" response error, but it's not triggering here because I'm not passing the `ResponseError` to the component, I'm just passing the message "AIP not found".  I don't think we should be showing the HTTP response error directly to the user because it's usually not helpful to anyone but a developer.

dashboard/src/stores/aip.ts
djjuhasz: Sure, maybe I can just do "Storage / AIPs / Error".

dashboard/src/stores/aip.ts
djjuhasz: A caught error doesn't show in the console by default.  We see a lot of Dashboard errors in the console because we don't do a good job of catching errors. :laughing: 

Right now this is outputting the whole error call stack to the console, which I don't really find useful, so we could stop doing that if it's not useful to you.

image

dashboard/src/stores/aip.ts
djjuhasz: I **think** a `fetchWorkflows` error will propagate up the call chain and be caught by the `catch()` statement on this promise. :thinking:

dashboard/src/stores/aip.ts
djjuhasz: Yeah, I'm all for dropping the response error output to the console, I find them confusing too.  I think "Error fetching workflows" and the HTTP code are probably sufficient for debugging.

dashboard/src/components/AipDeletionReviewAlert.vue
jraddaoui: I don't get this, you can still use a template for the else branch if you want, but you shouldn't need one for the single button.

dashboard/src/components/AipLocationCard.vue
jraddaoui: > If you can't click a button then what is the purpose of showing it to the user?

So the user know there is an action that may become available at some point. I prefer that but it's not a big deal for me, just make sure we follow the same approach for the download button too.

dashboard/src/stores/aip.ts
jraddaoui: It's okay for me, I just recall some response errors being logged always to the console, even if catched. Maybe it's only for 500 errors?

dashboard/src/stores/aip.ts
jraddaoui: I still have to check the tests, but it would be nice to be sure about that.

dashboard/src/stores/aip.ts
jraddaoui: Sounds good.

hack/kube/components/dev/keycloak.yaml
djjuhasz: I think that line does give `storage:aips:deletion:request` permission? :mag:  Maybe you meant `storage:aips:deletion:review` permission?

dashboard/src/stores/aip.ts
djjuhasz: Yeah, I can at least combine the two if statements into a single one. :thinking:

dashboard/src/stores/aip.ts
djjuhasz: I had to look this up, but it looks like line 256 *is* returning a promise:

> Async functions always return a promise. If the return value of an async function is not explicitly a promise, it will be implicitly wrapped in a promise.

-- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#description

dashboard/src/components/AipDeletionReviewAlert.vue
djjuhasz: Oh, I just assumed that the element type of `v-if` and `v-else` had to match, e.g. you couldn't do a `<button v-if="foo">` with a `<template v-else>`.

hack/kube/components/dev/keycloak.yaml
jraddaoui: Yeah! :man_facepalming:

dashboard/src/stores/__tests__/aip.test.ts
```diff
@@ -129,4 +168,140 @@ describe("useAipStore", () => {
...
+    // Third call should return true.
+    result = await store.canCancelDeletion();
+    expect(client.storage.storageCancelAipDeletion).toBeCalledWith({
+      uuid: "aip-uuid-1",
+      cancelAipDeletionRequestBody: {
+        check: true,
+      },
+    });
+    expect(result).toBe(true);
+  });
```
jraddaoui, line 475: Tests look neat, thanks @djjuhasz!

dashboard/src/stores/aip.ts
jraddaoui: Looks like this case is not covered by the tests. It would be nice to be sure, but not a big deal either.

jraddaoui (changes requested):
Thanks @djjuhasz!

I just merged https://github.com/artefactual-sdps/enduro/pull/1234, I guess the generated API client code will be removed from this changes after a rebase.

dashboard/src/pages/storage/aips/[id].vue
Comment (djjuhasz) [reply dashboard/src/pages/storage/aips/[id].vue:20]: I'll create a new issue for discussing the wording and information shown in the error alert component. I've tried to improve the error handling in the aipStore methods in particular, but there are lots of things we could still improve about the dashboard error handling.

dashboard/src/stores/aip.ts
djjuhasz: Done.

dashboard/src/stores/aip.ts
djjuhasz: I've updated most of the aipStore methods to log a console error with a simple description of the action being attempted, and the error response status text.  I think this could be improved on, but it's good enough for now in my opinion.

hack/kube/components/dev/keycloak.yaml
jraddaoui: Is this the only thing left?

djjuhasz:
@jraddaoui ready for another review.  I'll squash all the commits before merging.

hack/kube/components/dev/keycloak.yaml
djjuhasz: I found it helpful for testing that the operator doesn't have `storage:aips:deletion:review` permissions — it caught a bug where the operator could request deletion but then couldn't see the review alert to cancel their request.  That said I'm fine to add it if you still think the operator should be able to review deletion requests in general.

jraddaoui (approved):
Quite a big one in the end, thanks @djjuhasz!

hack/kube/components/dev/keycloak.yaml
jraddaoui: Cool, we can think about this in another PR.

Issue #1136, opened 2025-02-26
Problem: PIPs bigger than 4GB show errors when decompressed with 7zip in Archivematica


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Related issue in the temporal-activities repository:

https://github.com/artefactual-sdps/temporal-activities/issues/37

**Describe the solution you'd like**

As an initial solution, I'd like to optionally send unzipped bags to Archivematica.

**Describe alternatives you've considered**

Make zipped PIPs bigger that 4GB work between Enduro and Archivematica, see the related issue for more info about the actual issue in Go creating those ZIP files.

**Additional context**

Already implemented in https://github.com/artefactual-sdps/enduro/pull/1110 and https://github.com/artefactual-sdps/enduro/pull/1133.

jraddaoui:
After the lates PR this seems to work with compressed ZIPs but it fails deleting the transfer from the SFTP server at the end:

```
2025-02-25T17:15:15.495Z	V(2)	enduro-am-worker.temporal-client	log/with_logger.go:79	error	{"Namespace": "default", "TaskQueue": "am", "WorkerID": "1@enduro-am-77978d78d9-qmprm@", "WorkflowID": "processing-workflow-df7ee783-2ce0-4e8d-87d0-b2a1bf610ca3", "RunID": "05dbbf27-c0f5-42bf-a8bf-854f3f6bf630", "ActivityType": "DeleteTransferActivity", "Attempt": 2, "Error": "delete transfer: path: \"bag-7537ab2c-4e6b-4820-95bf-bd2c577351c3\": SFTP: unable to remove file \"bag-7537ab2c-4e6b-4820-95bf-bd2c577351c3\": failure (SSH_FX_FAILURE)", "error": "Activity error."}
go.temporal.io/sdk/log.(*withLogger).Error
	go.temporal.io/sdk@v1.32.1/log/with_logger.go:79
go.temporal.io/sdk/internal.(*activityTaskHandlerImpl).Execute
	go.temporal.io/sdk@v1.32.1/internal/internal_task_handlers.go:2246
go.temporal.io/sdk/internal.(*activityTaskPoller).ProcessTask
	go.temporal.io/sdk@v1.32.1/internal/internal_task_pollers.go:1080
go.temporal.io/sdk/internal.(*baseWorker).processTaskAsync.func1
	go.temporal.io/sdk@v1.32.1/internal/internal_worker_base.go:446
```

[...]

PR #1239, opened 2025-05-28
Move failed SIP/PIP to internal bucket

jraddaoui:
- Remove failed buckets from configuration and set up.
- Register a single bucketupload activity using the internal bucket.
- Add SIPFailedAs enum with `SIP` and `PIP` values.
- Add hardcoded failed SIP and failed PIP prefixes alongside SIP
  prefix to use in the object key for the failed upload.
- Pass SIP file extension in workflow request or get it  in workflow
  and use it in failed object keys.
- Set workflow state status to done before session ends so it's
  considered in the session cleanup.
- Rename SIP after failure inside internal bucket.
- Modify processing workflow state:
  - Reflect when the SIP has actually been transformed to be able to
  mark it as a failed SIP or as a failed PIP.
  - Remove `pip` and `sendToFailed`.
  - Add `initialPath` and `sip.transformed`.
  - Move `pip.pipType` to `sip.sipType`.

Refs #1202.


internal/enums/sip_failed_as.go
```diff
@@ -0,0 +1,9 @@
+package enums
+
+/*
+ENUM(
+SIP
+PIP
+)
+*/
+type SIPFailedAs string
```
djjuhasz, line 9: I find this enum name kind of awkward.  

Could we just check if the `state.pip.path` has a value and if it does then create the Failed PIP package from that instead of creating a new enum? Removing the `state.sendToFailed.path` would also save having to set both the `pip.path` and `sendToFailed.path` to the same value in a few places.

If you don't want to rely on `state.pip.path` I'd prefer to call the enum "FailedPackageType" or something similar.

internal/workflow/processing.go
```diff
@@ -1177,37 +1182,48 @@ func (w *ProcessingWorkflow) completeTask(
...
+	state *workflowState,
 ) error {
-	if stf.path == "" || stf.activityName == "" {
+	// Missing details.
+	if state.sendToFailed.path == "" || !state.sendToFailed.failedAs.IsValid() {
+		return errors.New("missing failed SIP/PIP path or failed as type")
+	}
+
+	// Failed SIP already in the internal bucket.
+	if state.req.WatcherName == "" && state.sendToFailed.failedAs == enums.SIPFailedAsSIP {
```
djjuhasz: Will the SIP already be named "Failed_SIP_{UUID}" if this is true, or do we need to rename it to be make it clear it failed processing?

internal/workflow/processing.go
```diff
@@ -1177,37 +1182,48 @@ func (w *ProcessingWorkflow) completeTask(
...
 		}
-		stf.path = zipResult.Path
+		state.sendToFailed.path = zipResult.Path
+	}
+
+	prefix := ingest.SIPPrefix
+	if state.sendToFailed.failedAs == enums.SIPFailedAsPIP {
+		prefix = ingest.PIPPrefix
 	}
+	key := fmt.Sprintf("%s%s", prefix, state.sip.uuid.String())
```
djjuhasz: Shouldn't this have a "Failed_" prefix as well?

djjuhasz (changes requested):
Looks good @jraddaoui, I just had a few questions.

internal/enums/sip_failed_as.go
jraddaoui: Yeah, I agree with the enum name, I had a lot of conflicts about this.

The main issue here is that the SIP and PIP are actually the same entity, but we need to make a distinction to know at what point it failed. The idea was to add this enum to the SIP table in a new `failed_as` column, alongside `failed_key` to be able to retrieve the package from the internal bucket. Does it make more sense that way? We may not needed in the workflow state but I thought that works for the DB schema.

About the workflow state, `state.sendToFailed.path` was needed to retain the SIP path at some point in the workflow. For example, we change the `state.sip.path` after extraction, but I want to use the compressed file for the upload, so I keep the previous path in sendToFailed. I think the same doesn't apply to the `state.pip.path` but since we have the `state.sendToFailed.path` already I think it's better to keep using it in both cases.

Now that you mention this, we could probably live without `state.sendToFailed.failedAs` and just check if the `state.pip.id` is populated, but I need to double check when is that set.

internal/workflow/processing.go
jraddaoui: Initial idea was to name them always `SIP_{UUID}` and `PIP_{UUID}` and change that name at download time, I was not even adding the extension.

After today's workshop and all the mentions about data-centric, I'll use the entire filename discussed in https://github.com/artefactual-sdps/enduro/issues/1202#issuecomment-2914119230 as the object key in the internal bucket, to provide more context about those packages outside of Enduro.

internal/workflow/processing.go
jraddaoui: Yeah, see my comment above.

internal/enums/sip_failed_as.go
djjuhasz: Oh, I always get confused about how we handle the SIP/PIP at failure.  If the workflow fails before preservation then I think we move the original SIP to the failed SIP bucket?  But if the workflow fails when we send the PIP to preservation, or later, then we zip* the PIP (with any preprocessing transformations) and move it to the failed PIP bucket, right?

[* I just realized zipping the PIP is going to cause problems if it's bigger than 4GB (#1136) , isn't it?]

Maybe we can simplify the SIP/PIP "sendToFailed" management to make it more intuitive and reduce the number of variables we need to manage?

Here's a proposal:

1. Set `state.sip.path` to the path of the original (zipped) SIP and we never change it
2. Set a `state.working.path` (or something similar) for the current package working directory
3. Set `state.pip.path` if, and only if, preprocessing modifies the original SIP. I think we can assume that if preprocessing is configured to run that it will modify the SIP for now.

If the workflow fails and `state.pip.path` is set, then we zip and upload the PIP to the internal bucket.  We know we have to zip the PIP because preprocessing never returns a zipped PIP anymore.  We don't need a separate `state.sendToFailed.FailedAs` variable because we can just check if `state.pip.path` (or the PIP ID) is set.

If the workflow fails and `state.pip.path` is not set, then we upload the original, zipped SIP to the internal bucket (if it isn't already there).  I think the original SIP will always be zipped right now, won't it?

Because PIPs are never zipped and the SIP is always zipped (for now) we don't need `state.sendToFailed.needsZipping`, and we can ditch `state.sendToFailed` altogether.

jraddaoui: Sorry for the delay on this @djjuhasz, I tried your suggestion but I can't make it work as I think it should. The issue I have is that we treat the SIP as PIP after preprocessing no matter what:

https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/processing.go#L1028-L1034

So, if preprocessing is not enabled and the SIP is a valid Bag, there is no transformation (PIP) before it hits AM, or if the SIP is not a Bag and it fails bagging before sending it to AM; I'd consider that a failed SIP in both cases.

Also, the SIP can theoretically be a directory (I could use the request `isDir` value to check that), and the PIP can already be zipped (`w.cfg.AM.ZipPIP` set to true and failing after).

TBH, I don't like the addition of PIP to the state now that I'm working with that abstraction. We don't have a proper entity for the PIP anywhere and I'd like to consider it a phase/status of the SIP for now, until we really need that entity.

I think this makes it a bit clearer and easier to manage: ad0e943c4e18c31132ab9b6014f4e8f40d4350c0.

internal/workflow/processing.go
jraddaoui: I have renamed the SIP inside the internal bucket in this commit: https://github.com/artefactual-sdps/enduro/pull/1239/commits/088773532bd1d360b72bb485b631582cc4c300bc

It requires to copy and delete the original, so not ideal but at least there is more context this way. I still need to get the original SIP filename and extension in workflow and include it in the keys, that will come in a following commit.

jraddaoui: Adding the SIP name and extension to the failed object keys in https://github.com/artefactual-sdps/enduro/pull/1239/commits/0943d888c91d899eb3493ef068313d4969ba1464.

jraddaoui:
@djjuhasz thanks for the feedback, this is ready for another review. I think it makes more sense to review the new commits one by one and check the comments above. I'll create a new issue to deal with the zipping situation.

internal/enums/sip_failed_as.go
djjuhasz: Thanks @jraddaoui, this does seem clearer and easier than managing the separate "sendToFailed" state variables. It still seems to me like a lot of state variables to manage, but maybe that's just what is required with the current complexities of the system. :shrug:

jraddaoui:
Thanks @djjuhasz! Sorry English is not my first language, appreciated. Some of those comments where already there, but I think I have addressed all of them ;)

internal/workflow/state.go
```diff
@@ -80,26 +78,19 @@ type sipInfo struct {
...
 
-	// isDir indicates whether the current working copy of the SIP is a
-	// filesystem directory.
-	isDir bool
+	// sipType is the type of the SIP.
+	sipType enums.SIPType
 
-	// pipType is the type of the PIP.
-	pipType enums.SIPType
+	// transformed indicates if the SIP should be considered as a PIP.
```
djjuhasz: The "as" is unnecessary: "transformed indicates if the SIP should be considered a PIP."

djjuhasz (approved):
Thanks @jraddaoui, this looks good to me. :+1: 

Sorry for getting picky on the comments, I just noticed that we've gotten a bit inconsistent about using complete sentences as required by our style guide.

jraddaoui:
No worries, thanks!

PR #1240, opened 2025-05-28
Add failed as and failed key to SIPs

jraddaoui:
Review after #1250.

- Add new fields to datatypes SIP.
- Add new nullable columns to the SIP table.
- Add new optional attributes to the SIP Goa design.
- Consider new fields in persistence SIP update and reads.
- Consider new fields in conversions.
- Add new parameters to the update SIP activity and
  consider them in the update.
- Add failed key to workflow state, populate it when
  the failed SIP/PIP is sent to the internal bucket.
- Generate Goa, Ent and dashboard client code.

Refs #1202.


djjuhasz (approved):
LGTM :+1:

PR #1241, opened 2025-05-29
Tweaks for SIP upload

jraddaoui:
- Change object key for uploaded SIPs
  Include the prefix, the file name without extension, the SIP UUID
  and the original extension.
- Update upload max size 
  Use 4 GiB as default in the configuration and the dashboard. Add
  `client_max_body_size 4096M` to the dashboard Dockerfile nginx config.

Refs #1210.

### DIFF
--- a/internal/ingest/upload.go
+++ b/internal/ingest/upload.go
@@ -7,8 +7,10 @@ import (
 	"io"
 	"mime"
 	"mime/multipart"
+	"strings"
 
 	"github.com/google/uuid"
+	"github.com/mholt/archiver/v4"
 	"gocloud.dev/blob"
 
 	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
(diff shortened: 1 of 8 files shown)


internal/ingest/upload.go
```diff
@@ -44,14 +46,24 @@ func (w *goaWrapper) UploadSip(
 		return nil, goaingest.MakeInvalidMultipartRequest(errors.New("invalid multipart request"))
 	}
 
+	// Identify file format to add extension in the object key.
+	// TODO: Use github.com/mholt/archives. We still use the archived github.com/mholt/archiver/v4
+	// in some activities, and using both causes a panic registering the same compressors.
+	format, stream, err := archiver.Identify(part.FileName(), part)
+	if err != nil {
+		return nil, goaingest.MakeInvalidMultipartRequest(errors.New("unable to identify format"))
+	}
```
jraddaoui, line 55: This is actually the same check we have in the extraction activity so, apart from giving us the extension it prevents the error in workflow. It also tries to identify the format from the stream alongside the filename, and detects `tar.xz` and other archive+compression formats.

djjuhasz (approved):
LGTM :+1:

Issue #1237, opened 2025-05-26
Problem: Upload fails when uploading large SIPs via the dashboard uploader


djjuhasz:
**Describe the bug**

Larger SIPs (e.g. a 215 MB SIP) fail to upload when using the Dashboard web uploader.

**To Reproduce**

1. Go to "Upload SIPs"
2. Drag & drop a SIP that is 200 MiB or bigger
3. The uploader reports "upload failed"

**Expected behavior**

The uploader should allow upload of SIPs up to 500 GiB without error.

**Screenshots**

![Image](https://github.com/user-attachments/assets/8250cbe3-d2b5-40f7-88bd-6e9541b8b4e3)

**Additional context**

- This could be a configuration issue, as we have an "[upload.maxSize](https://github.com/artefactual-sdps/enduro/blob/main/enduro.toml#L211)" config option with a default value of 102 MB (97.7 MiB)
- We should synchronize the `upload.maxSize`, the upload limit shown on the uploader page, and the upload limit configured for the Uppy library
- We may want to increase the default upload limit to something closer to the 500 GiB we chose for the uploader

jraddaoui:
Thanks @djjuhasz! One option/suggestion to synchronize the upload max size would be to use the about service/endpoint that already has access to parts of the configuration:

- https://github.com/artefactual-sdps/enduro/blob/main/internal/api/design/about.go
- https://github.com/artefactual-sdps/enduro/blob/main/internal/about/goa.go

djjuhasz:
Fixed by 53e76a2f44ef4dcceeabf2dab3e03ff045347241.

fiver-watson:
Successfully ingested a 227MB SIP during testing. For now that's the biggest I have. Will continue to monitor this and test with bigger packages in the future, but for now considering this addressed.

PR #1245, opened 2025-05-30
Set uploader max filesize from config (#1237)

djjuhasz:
- Set uploader max filesize from the AIP `/about` endpoint "upload_max_size".
- Define a default max upload size of 4 GiB
- Add an "about" store

### DIFF
--- a/dashboard/src/pages/ingest/upload/index.vue
+++ b/dashboard/src/pages/ingest/upload/index.vue
@@ -2,9 +2,11 @@
 import Uppy from "@uppy/core";
 import { Dashboard } from "@uppy/vue";
 import XHR from "@uppy/xhr-upload";
+import { onMounted } from "vue";
 import { useRouter } from "vue-router/auto";
 
 import { getPath } from "@/client";
+import { useAboutStore } from "@/stores/about";
 import { useAuthStore } from "@/stores/auth";
 import { useLayoutStore } from "@/stores/layout";
 import IconUpload from "~icons/clarity/plus-circle-line";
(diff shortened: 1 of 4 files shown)


dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -18,11 +19,34 @@ const authStore = useAuthStore();
 const layoutStore = useLayoutStore();
 
 const GiB = 1024 ** 3; // 1 GiB in bytes
-const maxFileSize = 4 * GiB;
+const uploadMaxDefault = 1 * GiB;
```
jraddaoui: Better if we match the default from the API config, 4 GiB.

dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -18,11 +19,34 @@ const authStore = useAuthStore();
...
+  return client.about
+    .aboutAbout()
+    .then((resp) => {
+      uploadMaxSize.value = resp.uploadMaxSize;
+    })
+    .catch((e) => {
+      console.error("Error fetching upload max size:", e.message);
+      uploadMaxSize.value = uploadMaxDefault;
+    });
+};
```
jraddaoui: What do you think **about** creating a small **about** store? Then use it from the [AboutDialog](https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/components/AboutDialog.vue) too.

jraddaoui (changes requested):
Thanks @djjuhasz! The about store could happen in another PR.

dashboard/src/pages/ingest/upload/index.vue
djjuhasz: Done, see a9c9b98c3d871b9651f28d452666ffdc75343320

dashboard/src/pages/ingest/upload/index.vue
djjuhasz: Fair enough, but it does make testing harder when the default value and the value returned from the API are the same.

jraddaoui (approved):
Thanks @djjuhasz!

PR #1247, opened 2025-06-01
Support macOS/Apple Silicon in dev environment

sevein:
I recently started using a Mac with an M4 chip as my primary driver. To run enduro-sdps, I've had to make a few changes to improve compatibility with macOS and Apple Silicon. These updates also help with broader architecture support, like ARM on major cloud providers.

It's really fast btw. Partly because the virtual machine that hosts Linux and k3s runs natively on the aarch64 cores, i.e. no CPU emulation is required. The setup process is also very easy - using [Colima](https://github.com/abiosoft/colima), which is an open-source, lightweight alternative to Docker Desktop for Mac and is managed from the CLI.


jraddaoui (approved):
Thanks @sevein!!

PR #1249, opened 2025-06-10
Use ent client for SIP updates

jraddaoui:
Slowly moving towards the goal of using Ent in the ingest service. This is related to #1202, where I plan to add new columns to the SIP table to reflect the failure type (SIP or PIP) and the failed object key.

### DIFF
--- a/internal/persistence/ent/client/sip_test.go
+++ b/internal/persistence/ent/client/sip_test.go
@@ -142,6 +142,7 @@ func TestUpdateSIP(t *testing.T) {
 	completed2 := sql.NullTime{Time: started2.Time.Add(time.Second), Valid: true}
 
 	type params struct {
+		sipUUID uuid.UUID
 		sip     *datatypes.SIP
 		updater persistence.SIPUpdater
 	}
@@ -154,6 +155,7 @@ func TestUpdateSIP(t *testing.T) {
 		{
 			name: "Updates all SIP columns",
 			args: params{
+				sipUUID: sipUUID,
 				sip: &datatypes.SIP{
 					UUID:        sipUUID,
 					Name:        "Test SIP",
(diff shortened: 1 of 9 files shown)


internal/persistence/ent/client/sip_test.go
```diff
@@ -188,17 +190,23 @@ func TestUpdateSIP(t *testing.T) {
...
 					AIPID:     aipID,
 					Status:    enums.SIPStatusProcessing,
 					StartedAt: started,
 				},
-				updater: func(p *datatypes.SIP) (*datatypes.SIP, error) {
-					p.Status = enums.SIPStatusIngested
-					p.CompletedAt = completed
-					return p, nil
+				updater: func(s *datatypes.SIP) (*datatypes.SIP, error) {
+					// Inmutable.
```
djjuhasz: Typo: "Immutable"

internal/persistence/ent/client/sip_test.go
```diff
@@ -213,24 +221,54 @@ func TestUpdateSIP(t *testing.T) {
...
+					AIPID:     aipID,
+					Status:    enums.SIPStatusProcessing,
+					StartedAt: started,
+				},
+				updater: func(s *datatypes.SIP) (*datatypes.SIP, error) {
+					// Invalid.
+					s.Status = ""
+
+					// Valid.
+					s.Status = enums.SIPStatusIngested
```
djjuhasz: Is this a copy-paste error? I assume that you wanted to test that trying to set `s.Status = ""` would result in the original value of `enums.SIPStatusProcessing` being retained.

internal/workflow/local_activities.go
```diff
@@ -52,16 +52,30 @@ func updateSIPLocalActivity(
...
-		params.Status,
-		params.CompletedAt,
+		func(s *datatypes.SIP) (*datatypes.SIP, error) {
+			s.Name = params.Name
+			s.Status = params.Status
+
+			if params.AIPUUID != "" {
+				aipUUID, err := uuid.Parse(params.AIPUUID)
+				if err != nil {
+					return nil, err
```
djjuhasz: It might be nice to provide a more contextual error message, e.g `fmt.Errorf("invalid AIPUUID: %s", params.AIPUUID)`.

internal/workflow/local_activities.go
```diff
@@ -52,16 +52,30 @@ func updateSIPLocalActivity(
...
+	_, err := ingestsvc.UpdateSIP(
 		ctx,
 		params.UUID,
-		params.Name,
-		params.AIPUUID,
-		params.Status,
-		params.CompletedAt,
+		func(s *datatypes.SIP) (*datatypes.SIP, error) {
+			s.Name = params.Name
+			s.Status = params.Status
```
djjuhasz, line 61: I think it would be good to validate the status here, and return an error if it's invalid.  Right now I think an invalid status will just result in the previous status being retained, without any feedback to the caller.

djjuhasz (changes requested):
Looks great @jraddaoui, I just made a few minor notes.

internal/persistence/ent/client/sip_test.go
jraddaoui: Good catch, thanks!

internal/workflow/local_activities.go
jraddaoui: You're right, good idea!

jraddaoui:

djjuhasz (approved):
Looks great! Thanks @jraddaoui. :+1:

PR #1250, opened 2025-06-10
Test parameters sent to SIP update activity

jraddaoui:
- Make sure the CompletedAt time is not set while processing.
- Use argument matchers to check that CompletedAt is set.
- Remove `storedAt` from the `state.aip`.

As #1249, this is related to #1202, where I plan to add new columns to the SIP table to add failure details, and I want to test those values are properly set in workflow.

### DIFF
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -97,7 +97,7 @@ func (w *ProcessingWorkflow) cleanup(ctx temporalsdk_workflow.Context, state *wo
 			UUID:        state.sip.uuid,
 			Name:        state.sip.name,
 			AIPUUID:     state.aip.id,
-			CompletedAt: state.aip.storedAt,
+			CompletedAt: temporalsdk_workflow.Now(dctx).UTC(),
 			Status:      state.sip.status,
 		},
 	).Get(activityOpts, nil)
@@ -517,15 +517,14 @@ func (w *ProcessingWorkflow) SessionHandler(
 		}
 	}
 
-	// Persist AIP UUID and storedAt time.
+	// Persist the SIP adding the AIP UUID.
 	{
 		activityOpts := withLocalActivityOpts(sessCtx)
 		_ = temporalsdk_workflow.ExecuteLocalActivity(activityOpts, updateSIPLocalActivity, w.ingestsvc, &updateSIPLocalActivityParams{
-			UUID:        state.sip.uuid,
-			Name:        state.sip.name,
-			AIPUUID:     state.aip.id,
-			CompletedAt: state.aip.storedAt,
-			Status:      enums.SIPStatusProcessing,
+			UUID:    state.sip.uuid,
+			Name:    state.sip.name,
+			AIPUUID: state.aip.id,
+			Status:  enums.SIPStatusProcessing,
 		}).
 			Get(activityOpts, nil)
 	}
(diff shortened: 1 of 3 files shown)


internal/workflow/processing_test.go
```diff
@@ -629,14 +662,30 @@ func (s *ProcessingWorkflowTestSuite) TestAutoApprovedAIP() {
 	s.env.OnActivity(activities.PollMoveToPermanentStorageActivityName, sessionCtx, mock.AnythingOfType("*activities.PollMoveToPermanentStorageActivityParams")).
 		Return(nil, nil).
 		Once()
+
+	s.env.OnActivity(
+		updateSIPLocalActivity,
+		ctx,
+		ingestsvc,
+		mock.MatchedBy(func(params *updateSIPLocalActivityParams) bool {
```
djjuhasz, line 670: This `MatchedBy` function is cool, but time is deterministic in the workflow test context, so you can test the time directly:

```
s.env.OnActivity(
	updateSIPLocalActivity,
	ctx,
	ingestsvc,
	&updateSIPLocalActivityParams{
		UUID:        sipUUID,
		Name:        sipName,
		AIPUUID:     aipUUID.String(),
		Status:      enums.SIPStatusIngested,
		CompletedAt: startTime.Add(time.Second),
	},
).Return(nil, nil).Once()
```

I don't understand why one second is added to the `startTime` in the tests, but `startTime.Add(time.Second)` works consistently.

internal/workflow/processing.go
```diff
@@ -97,7 +97,7 @@ func (w *ProcessingWorkflow) cleanup(ctx temporalsdk_workflow.Context, state *wo
 			UUID:        state.sip.uuid,
 			Name:        state.sip.name,
 			AIPUUID:     state.aip.id,
-			CompletedAt: state.aip.storedAt,
+			CompletedAt: temporalsdk_workflow.Now(dctx).UTC(),
```
djjuhasz, line 100: It looks like we set "CompletedAt" even on error? 

I guess that "StartedAt" and "CompletedAt" are referring to the ingest workflow start and end times? I don't know what it would mean for a SIP to be "completed" in the context of Enduro.  Maybe we should drop these fields from SIP entity at some point in the future.

djjuhasz (approved):
This looks good @jraddaoui.  

I did add a note that we can use explicit time comparisons as `temporalsdk_workflow.Now()` is deterministic in the test environment.  I like explicitly checking the time values better than just checking they are not zero, but I don't feel strongly enough about it to block the PR.

internal/workflow/processing_test.go
jraddaoui: I started doing it that way but, even if it's deterministic, it's influenced by the events that happen in workflow. For example, in the AM workflow I need to add ten seconds instead of one. So I thought, if we modify the workflow and introduce different delays in it, we may need to update all times that are exactly matched in tests.

internal/workflow/processing.go
jraddaoui: > It looks like we set "CompletedAt" even on error?

Yeah, that makes sense to me. It completed, badly, but it completed. 

> I guess that "StartedAt" and "CompletedAt" are referring to the ingest workflow start and end times? I don't know what it would mean for a SIP to be "completed" in the context of Enduro. Maybe we should drop these fields from SIP entity at some point in the future.

Great point, I agree!

Issue #1246, opened 2025-05-31
Enhancement: add ability to see who uploaded a SIP


sallain:
**Is your feature request related to a problem? Please describe.**

There is no way of knowing which user uploaded a SIP. A user may need to track their SIP to ensure that it has completed successfully, or they may want to alert another user that an ingest failed - there are any number of reasons that this data point would be helpful. Having the user who uploaded the SIP listed as part of the SIP data would make this much easier.

**Describe the solution you'd like**

I want to be able to see the username (or email, whatever) of the person who deposited the package in the SIPs table. Eventually, I want to be able to filter the table on the user so that I can see all packages that belong to me (or someone else).

The first step is to associate a SIP with the user that uploaded it. The user display name can be captured the same way that we do for the deletion workflow (see https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/stores/auth.ts#L51-L57), with the same order of preference - _preferred username_ if available, if not then _name_, if not then _email_.

At this time, it is possible to use Enduro unauthenticated or by connecting to an external identity provider. For this iteration, we will only use credentials from the external identity provider. If an unauthenticated user deposits a SIP,  the user can either be listed as "unknown" or left blank (dev's choice!)

The first UI increment will be to show the user in the SIPs table:

![Image](https://github.com/user-attachments/assets/c5ad626c-b341-4454-b05d-f574af09364a)

In order to make space for the user column in the SIPs table, we will remove the UUID column. The UUID is still accessible on the SIP detail page.

The user should also be shown on the SIP detail page, in the left sidebar (mock-up forthcoming)

**Describe alternatives you've considered**

**Additional context**

There are a few aspects to this feature that we are not going to add in this first iteration:

* We are only considering SIPs deposited through the UI. We will deal with automated deposits (triggered from watched buckets) later. In the meantime, SIPs that are started automatically from a watched bucket can either be given an "unknown" user or the field can be left blank.
* The ability to filter by user or to click a "show my packages" button will be added in the next iteration
* Permissions will be added in a later iteration - but, at a minimum, we would like to be able to limit a user to see only their packages 

See [the Enduro product dev Miro board](https://miro.com/app/board/uXjVL7s6tus=/?moveToWidget=3458764627970029292&cot=14) for discussion, details, and mock-ups.

djjuhasz:
@sallain @fiver-watson I've got the SIP uploader showing in the dashboard:

[...]

PR #1252, opened 2025-06-12
Record which user uploads a SIP (refs #1246)

djjuhasz:
- Add a user table to the ingest DB
- Add a one-to-many "uploader" relationship from user to SIP
- Record which user uploads a SIP via the "Upload SIP" API endpoint (when using authentication)


djjuhasz:
@jraddaoui I'm planning to add a user table with a one-to-many relationship to the "sip" table to track who uploaded the SIP, but I thought I'd run it by you before building on it too much.  I thought about making it a many-to-many relationship with a "type" to allow categorizing different user-sip relations (e.g. uploader, creator, deleter) but I think that's more complexity than we need at this point.

djjuhasz:
@jraddaoui it is probably easier to review the commits individually rather than the whole PR at once.

internal/api/auth/claims.go
```diff
@@ -9,6 +9,7 @@ import (
 type Claims struct {
 	Email         string   `json:"email,omitempty"`
 	EmailVerified bool     `json:"email_verified,omitempty"`
+	Name          string   `json:"name,omitempty"`
```
jraddaoui, line 12: In the dashboard we look for the `preferred_username` first ([here](https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/stores/auth.ts#L51-L57)), and it would be nice to follow the same fallback approach in both places.

I'd remove the `preferred_username` check from the dashboard and leave this is as it is for now, we'll see if we want to collect more user info in the future.

internal/datatypes/sip.go
```diff
@@ -12,28 +12,31 @@ import (
...
-	CompletedAt sql.NullTime `db:"completed_at"`
+	CompletedAt sql.NullTime
 
 	// Set if there is a failure in workflow, it can be empty.
-	FailedAs enums.SIPFailedAs `db:"failed_as"`
+	FailedAs enums.SIPFailedAs
 
 	// Object key from the failed SIP/PIP in the internal bucket.
-	FailedKey string `db:"failed_key"`
+	FailedKey string
```
jraddaoui, line 36: I was planning to remove all `db` annotations when we fully remove the raw SQL operations. I think they were only needed for the read, which was recently changed to use Ent, so I think this is okay. I was planning to remove the usage of the `sql.NullTime` and `uuid.NullUUID` in favor of `*time.Time` and `*uuid.UUID` at the same time, but we can do that in another PR.

internal/datatypes/user.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+type User struct {
+	// UUID is a unique identifier for the user that can be used as a public
+	// identifier if no better options (e.g. name, email) are available.
+	UUID uuid.UUID `db:"uuid"`
+
+	// CreatedAt is an optional field. If the value is nil when creating a new
+	// user, CreatedAt will be set to the current time.
+	CreatedAt *time.Time `db:"created_at"`
```
jraddaoui: I think making this a pointer is a bigger hassle than considering the zero value in the user creation.

internal/datatypes/user.go
```diff
@@ -0,0 +1,29 @@
...
+
+	// The Name and Email fields are optional as we are not sure if they will be
+	// provided by all OIDC providers. Because they are nice human-readable
+	// identifiers they should be used for display purposes when available.
+	Email *string `db:"email"`
+	Name  *string `db:"name"`
```
jraddaoui: Not a big deal, but I think it's better to avoid using pointers when the zero value can be considered the same as the nil value.

internal/datatypes/user.go
```diff
@@ -0,0 +1,29 @@
...
+
+	// The JWT Iss and Sub fields are optional SECRET values that uniquely
+	// identify the user in the OIDC provider. They are optional values to allow
+	// for users that are not authenticated via OIDC (e.g. system users).
+	JWTIss *string `db:"jwt_iss"`
+	JWTSub *string `db:"jwt_sub"`
```
jraddaoui: I think `OIDCIss` and `OIDCSub` would be clearer. JWT may be used for other authentication methods in the future.

internal/ingest/upload.go
```diff
@@ -92,11 +101,22 @@ func (w *goaWrapper) initSIP(
 	extension string,
 	wType enums.WorkflowType,
 ) error {
+	user, err := w.findOrCreateUser(ctx, w.perSvc)
```
jraddaoui, line 104: No need to pass `w.perSvc`, the method is part of `w` and should have access to it directly.

internal/ingest/upload.go
```diff
@@ -117,3 +137,61 @@ func (w *goaWrapper) initSIP(
...
+
+func (w *goaWrapper) findOrCreateUser(
+	ctx context.Context,
+	perSvc persistence.Service,
+) (*datatypes.User, error) {
+	claims := auth.UserClaimsFromContext(ctx)
+	if claims == nil {
+		return nil, nil
+	}
+	if claims.ISS == "" {
```
jraddaoui: I just realized it should be `claims.Iss` after seeing your changes, that was my bad. Feel free to leave it or change it now that we are doing it properly in other fields.

internal/ingest/upload.go
```diff
@@ -117,3 +137,61 @@ func (w *goaWrapper) initSIP(
...
+		return nil, fmt.Errorf("invalid user claims: missing ISS")
+	}
+	if claims.Sub == "" {
+		return nil, fmt.Errorf("invalid user claims: missing Sub")
+	}
+
+	user, err := perSvc.ReadJWTUser(ctx, claims.ISS, claims.Sub)
```
jraddaoui: `ReadJWTUser` seems to be like it's going to get another type of user. What about `ReadUserByClaims` and passing the full claims? Or `ReadUserByOIDC` and keeping the same signature?

internal/ingest/upload.go
```diff
@@ -117,3 +137,61 @@ func (w *goaWrapper) initSIP(
...
+	if err != nil {
+		if errors.Is(err, persistence.ErrNotFound) {
+			// User does not exist, create a new one.
+			user, err := w.createUser(ctx, perSvc, claims)
```
jraddaoui, line 160: Same here, `w.createUser` will have access to `w.perSvc`, so no need to pass it.

internal/ingest/upload_test.go
```diff
@@ -145,13 +151,89 @@ func TestUpload(t *testing.T) {
...
+					},
+				).Return(nil, nil)
+			},
+			multipartBody: zipMmultipartBody,
+			contentType:   "multipart/form-data; boundary=foobar",
+			maxUploadSize: 102400000,
+			want:          &goaingest.UploadSipResult{UUID: uuid0.String()},
+		},
+		{
+			name: "Uploads a SIP and creates a user",
```
jraddaoui, line 205: It would be nice to add another success test case, where authentication is disabled and it creates a SIP without uploader.

internal/persistence/ent/client/convert.go
```diff
@@ -69,3 +75,28 @@ func convertTask(task *db.Task) *datatypes.Task {
...
+	}
+	if dbu.Name != "" {
+		u.Name = &dbu.Name
+	}
+	if dbu.JwtIss != "" {
+		u.JWTIss = &dbu.JwtIss
+	}
+	if dbu.JwtSub != "" {
+		u.JWTSub = &dbu.JwtSub
+	}
```
jraddaoui: If the zero value could be considered as nil, this transformation would be straight forward. Also, the other way around, I think there is a flag that allows Ent to return nil fields directly from optional columns, which would also simplify this transformation.

internal/persistence/ent/client/sip.go
```diff
@@ -44,6 +46,16 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
...
+	var uploaderUUID *uuid.UUID
+	if s.UploaderID != nil {
+		u, err := c.ent.User.Query().Where(user.UUID(*s.UploaderID)).Only(ctx)
+		if err != nil {
+			return newDBError(errors.New("create SIP: uploader not found"))
+		}
+		q.SetUser(u)
+		uploaderUUID = &u.UUID
+	}
+
```
jraddaoui: This should happen in a DB transaction, even if we don't have update or delete methods for users, it's an easy thing to do that could prevent issues in the future.

internal/persistence/ent/client/sip.go
```diff
@@ -112,16 +132,35 @@ func (c *client) UpdateSIP(
...
+	var userUUID *uuid.UUID
+	if up.UploaderID != nil {
+		u, err := c.ent.User.Query().Where(user.UUID(*up.UploaderID)).Only(ctx)
+		if err != nil {
+			return nil, rollback(tx, newDBError(errors.New("update SIP: uploader not found")))
+		}
+		q.SetUser(u)
+		userUUID = &u.UUID
+	}
+
```
jraddaoui: I'd not consider the uploader an updatable field, I'd just ignore it in the SIP update.

internal/persistence/ent/client/sip.go
```diff
@@ -135,7 +174,7 @@ func (c *client) DeleteSIP(ctx context.Context, id int) error {
 
 // ReadSIP returns the SIP identified by id.
 func (c *client) ReadSIP(ctx context.Context, id uuid.UUID) (*datatypes.SIP, error) {
-	s, err := c.ent.SIP.Query().Where(sip.UUID(id)).Only(ctx)
+	s, err := c.ent.SIP.Query().Where(sip.UUID(id)).WithUser().Only(ctx)
```
jraddaoui: If we only need the UUID, inside `WithUser()` we can limit the fields returned by the query as an small improvement.

internal/persistence/ent/client/sip.go
```diff
@@ -153,7 +192,7 @@ func (c *client) ListSIPs(ctx context.Context, f *persistence.SIPFilter) (
 		f = &persistence.SIPFilter{}
 	}
 
-	page, whole := filterSIPs(c.ent.SIP.Query(), f)
+	page, whole := filterSIPs(c.ent.SIP.Query().WithUser(), f)
```
jraddaoui: Same here, where it could be more noticeable.

internal/persistence/ent/client/user.go
```diff
@@ -0,0 +1,75 @@
...
+	dbu, err := c.ent.User.Query().Where(user.UUID(id)).Only(ctx)
+	if err != nil {
+		return nil, newDBError(err)
+	}
+
+	return convertUser(dbu), nil
+}
+
+// ReadJWTUser retrieves a user by JWT issuer and subject.
+func (c *client) ReadJWTUser(ctx context.Context, iss, sub string) (*datatypes.User, error) {
```
jraddaoui: Again, I think it's better to refer to OIDC rather than to JWT. Or as I said above, pass the entire claims.

internal/persistence/telemetry_test.go
```diff
@@ -0,0 +1,243 @@
...
+			if tt.wantErr != "" {
+				assert.Error(t, err, tt.wantErr)
+				return
+			}
+
+			assert.NilError(t, err)
+			assert.DeepEqual(t, got, tt.want)
+		})
+	}
+}
```
jraddaoui, line 242: It's nice to get some coverage for these telemetry functions, thanks! Ideally, we would test the span creation, etc. instead of using a noop tracer provider, but that can happen in another PR.

jraddaoui (changes requested):
Really nice @djjuhasz, great work!

I added some comments bellow but most of them are not blockers. The only ones I think we should address before merging:

- Use a transaction on SIP creation.
- Ignore uploader on SIP update.
- Refer to OIDC instead of JWT.

internal/datatypes/sip.go
jraddaoui: Also, I forgot to mention that you added the `db` annotations to the `User` struct, we shouldn't need them, right?

internal/persistence/telemetry_test.go
djjuhasz: I did briefly look for a simple Open Telemetry tracer that we could use in test, but I didn't see anything obvious.  I agree that it would be nice to test the tracer and spans too.

internal/datatypes/sip.go
djjuhasz: Good point, I'll get rid of the `User` db annotations.

djjuhasz:
@jraddaoui this is ready for another review.  I'll squash all of the commits before merging.

internal/persistence/ent/client/sip.go
```diff
@@ -44,26 +52,42 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
 		q.SetCompletedAt(s.CompletedAt.Time)
 	}
 
-	// Set CreatedAt to the current time
-	q.SetCreatedAt(time.Now())
+	var uploaderUUID *uuid.UUID
+	if s.UploaderID != nil {
+		u, err := tx.User.Query().Where(user.UUID(*s.UploaderID)).Only(ctx)
+		if err != nil {
+			return newDBError(errors.New("create SIP: uploader not found"))
```
jraddaoui: This should rollback the transaction.

internal/persistence/ent/client/sip.go
```diff
@@ -44,26 +52,42 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
...
+		}
+		q.SetUser(u)
+		uploaderUUID = &u.UUID
+	}
 
 	// Save the SIP.
 	dbs, err := q.Save(ctx)
 	if err != nil {
 		return newDBErrorWithDetails(err, "create SIP")
```
jraddaoui: Same.

internal/persistence/ent/client/sip.go
```diff
@@ -113,15 +139,30 @@ func (c *client) UpdateSIP(
...
+	if dbs.UploaderID != 0 {
+		u, err := c.ent.User.Query().
+			Where(user.ID(dbs.UploaderID)).
+			Select(user.FieldUUID).
+			Only(ctx)
+		if err != nil {
+			return nil, newDBErrorWithDetails(err, "update SIP: uploader not found")
+		}
+		s.UploaderID = &u.UUID
+	}
```
jraddaoui: Can't we get it in the initial query (L102) and save it as the `dbID` to use it here later?

internal/persistence/ent/client/user.go
```diff
@@ -0,0 +1,79 @@
...
+	// Query the user by UUID.
+	dbu, err := c.ent.User.Query().Where(user.UUID(id)).Only(ctx)
+	if err != nil {
+		return nil, newDBError(err)
+	}
+
+	return convertUser(dbu), nil
+}
+
+// ReadOIDCUser retrieves a user by JWT issuer and subject.
```
jraddaoui: Still reads `...by JWT issuer and subject.`.

internal/persistence/persistence.go
```diff
@@ -39,4 +39,14 @@ type Service interface {
...
 	UpdateTask(ctx context.Context, id int, updater TaskUpdater) (*datatypes.Task, error)
+
+	// CreateUser persists a new user to the data store then updates the user
+	// to add auto-generated data (e.g. ID, CreatedAt).
+	CreateUser(context.Context, *datatypes.User) error
+
+	// ReadUser retrieves a user by UUID.
+	ReadUser(context.Context, uuid.UUID) (*datatypes.User, error)
+
+	// ReadOIDCUser retrieves a user by JWT issuer and subject.
```
jraddaoui: Same.

jraddaoui (changes requested):
Thanks @djjuhasz! Mostly the transaction rollback on error, and a couple of minor details.

internal/persistence/ent/client/sip.go
djjuhasz: Why yes, yes we can. :stuck_out_tongue:

djjuhasz:
@jraddaoui ready for another review. Hopefully the last one. :crossed_fingers: :laughing:

jraddaoui (approved):
Neat :ok_hand:, thanks!

PR #1253, opened 2025-06-12
Add download SIP API endpoints

jraddaoui:
Browsers don't allow to set custom headers (like Authorization) when
downloading files via `window.open()` or `<a href="...">`. Also, using
the fetch API with custom headers requires loading the entire file into
memory, which is not suitable for large files. To work around this, I
implemented a two-step download process, similar to what we do to
set up Websocket connections:

- Request SIP download endpoint:
  - The frontend will call this endpoint with the Authorization header.
  - The backend will validate the token and return a short-lived,
    single-use ticket (via a cookie).

- SIP download endpoint:
  - The frontend will open this endpoint in a new window.
  - The browser will automatically send the ticket cookie.
  - The backend will validate the ticket and stream the file.

Following iterations could allow to download the SIP in a single step
process for out of browser requests to the API.

Refs #1202.

### DIFF
--- a/internal/ingest/download_sip.go
+++ b/internal/ingest/download_sip.go
@@ -1,0 +1,117 @@
+package ingest
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/google/uuid"
+	"gocloud.dev/gcerrors"
+
+	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
+	"github.com/artefactual-sdps/enduro/internal/datatypes"
+	"github.com/artefactual-sdps/enduro/internal/persistence"
+)
+
+func (w *goaWrapper) readSIP(ctx context.Context, id string) (*datatypes.SIP, error) {
+	// Validate the payload UUID.
+	sipUUID, err := uuid.Parse(id)
+	if err != nil {
+		return nil, goaingest.MakeNotValid(errors.New("invalid UUID"))
+	}
+
+	// Read the persisted SIP.
+	sip, err := w.perSvc.ReadSIP(ctx, sipUUID)
+	if err != nil {
+		if errors.Is(err, persistence.ErrNotFound) {
+			return nil, &goaingest.SIPNotFound{UUID: id, Message: "SIP not found"}
+		} else {
+			return nil, goaingest.MakeInternalError(errors.New("error reading SIP"))
+		}
+	}
+
+	return sip, nil
+}
+
+func (w *goaWrapper) DownloadSipRequest(
+	ctx context.Context,
+	payload *goaingest.DownloadSipRequestPayload,
+) (*goaingest.DownloadSipRequestResult, error) {
+	sip, err := w.readSIP(ctx, payload.UUID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check that failed as and failed key values are set.
+	if sip.FailedAs == "" || sip.FailedKey == "" {
+		return nil, goaingest.MakeNotValid(errors.New("SIP has no failed values"))
+	}
+
+	// Check if the failed SIP/PIP exists in the internal bucket.
+	exists, err := w.internalStorage.Exists(ctx, sip.FailedKey)
+	if err != nil {
+		return nil, goaingest.MakeInternalError(errors.New("error checking SIP/PIP file"))
+	}
+
+	if !exists {
+		return nil, &goaingest.SIPNotFound{
+			UUID:    payload.UUID,
+			Message: "Failed SIP/PIP file not found in the internal storage",
+		}
+	}
+
+	// Request a ticket.
+	ticket, err := w.ticketProvider.Request(ctx)
+	if err != nil {
+		return nil, goaingest.MakeInternalError(errors.New("ticket request failed"))
+	}
+
+	// A ticket is not provided when authentication is disabled.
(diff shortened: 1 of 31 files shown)


jraddaoui:
@djjuhasz, I'll fix the AIP download in another issue and PR, with the same approach if this looks good to you.

djjuhasz:
@jraddaoui do you have a source for "Also, using the Fetch API with custom headers requires loading the entire file into memory, which is not suitable for large files."? I've been searching for some mention of this limitation in the Fetch API documentation and via Google search and I can't find any info about it.  It seems weird to me that Fetch won't allow streaming a response when sending a request with an auth header. :confused:

jraddaoui:
How would you do that? You'll need to read the response in the application code. 

https://blog.stephensorensen.com/download-files-using-fetch

djjuhasz:
Wow, I had no idea that proxying a bytestream via a Web App was so hard. There is https://github.com/jimmywarting/StreamSaver.js which allows streaming a file through a WebApp via some magic with service workers, but the solution sounds a bit crazy (using man-in-the-middle to install a service worker in Github static pages??? :exploding_head:)  

It sounds like https://fs.spec.whatwg.org/ might be a solution that doesn't involve deep magic, but I'm not clear if it's production ready, and it sounds like it only works in HTTPS contexts?  (See also: https://developer.mozilla.org/en-US/docs/Web/API/File_System_API)

internal/ingest/download_sip.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/google/uuid"
+	"gocloud.dev/gcerrors"
+
+	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
+	"github.com/artefactual-sdps/enduro/internal/datatypes"
+	"github.com/artefactual-sdps/enduro/internal/persistence"
+)
+
+func (w *goaWrapper) checkSIP(ctx context.Context, id string) (*datatypes.SIP, error) {
```
djjuhasz: Maybe `isSIPDownloadable()` would provide a better hint at what this method is doing?

djjuhasz: I think it's hard to name this method because it's doing two things: loading the SIP and then checking if it's downloadable.  I think splitting up the method into a `loadSIP()` (or something similar) method and a `isFailed(s datatypes.SIP) bool` function would make naming easier and follow the "functions should do one thing" principle, even if it does require a few more lines of code.

internal/ingest/download_sip.go
```diff
@@ -0,0 +1,97 @@
...
+	// Request a ticket.
+	ticket, err := w.ticketProvider.Request(ctx)
+	if err != nil {
+		return nil, goaingest.MakeInternalError(errors.New("failed to request ticket"))
+	}
+
+	// Do not set the cookie unless a ticket is provided.
+	res := &goaingest.DownloadSipRequestResult{}
+	if ticket != "" {
+		res.Ticket = &ticket
```
djjuhasz, line 74: I would expect the API endpoint to return an HTTP error response if it can't return a valid ticket, so the client doesn't proceed to try and download the SIP without a ticket.

internal/ingest/download_sip.go
```diff
@@ -0,0 +1,97 @@
...
+	// Check the SIP status.
+	_, err := w.checkSIP(ctx, payload.UUID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Request a ticket.
+	ticket, err := w.ticketProvider.Request(ctx)
+	if err != nil {
+		return nil, goaingest.MakeInternalError(errors.New("failed to request ticket"))
```
djjuhasz: I think "ticket request failed" or "failed to receive ticket" would be a more accurate error message.

djjuhasz (approved):
Very nice @jraddaoui.  I had a few comments/questions but no blockers.

internal/ingest/download_sip.go
jraddaoui: It is a little abstract IMO, but this is what happens when authentication is disabled. The dashboard will still do the same flow in that case, so we can't return an error.

jraddaoui:
> Wow, I had no idea that proxying a bytestream via a Web App was so hard. There is https://github.com/jimmywarting/StreamSaver.js which allows streaming a file through a WebApp via some magic with service workers, but the solution sounds a bit crazy (using man-in-the-middle to install a service worker in Github static pages??? 🤯)
> 
> It sounds like https://fs.spec.whatwg.org/ might be a solution that doesn't involve deep magic, but I'm not clear if it's production ready, and it sounds like it only works in HTTPS contexts? (See also: https://developer.mozilla.org/en-US/docs/Web/API/File_System_API)

Thanks for looking at that and the review @djjuhasz. Yeah, it is a crazy world :D. I looked at some of those options and even thought about using our own signed URLs, but being only for the browser download from our dashboard and having the ticket store + cookie setup for the websocket, this was a lot more straight forward. In the future, I'd like to see if we can have both authentication mechanisms in the same endpoint (cookie or header) to allow out of browser downloads in a single request.

PS. I added the commit related to the API from the other PR here ;)

internal/ingest/download_sip.go
djjuhasz: Oh, I see. I think a comment to that effect would be helpful.

jraddaoui: Done!

PR #1254, opened 2025-06-12
Pin GitHub Actions and group the Dependabot updates

sevein:
This pull request introduces version pinning across all CI workflows and updates the Dependabot configuration to use groups. By defining a single group for all GitHub Action updates, you'll receive just one pull request (continuously updated by Dependabot) for all updates, significantly reducing noise.

I've tested this before on a different repo, see: https://github.com/artefactual-labs/ccp/pull/83. If this new model works, it could be implemented in other dependency ecosystems like "gomod", "docker" or "npm".

Related: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,8 +3,12 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
-    open-pull-requests-limit: 0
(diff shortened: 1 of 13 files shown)


jraddaoui (approved):
LGTM, thanks @sevein!

PR #1260, opened 2025-06-17
Show user who uploaded a SIP

djjuhasz:
- Return uploader data (if available) in "ingest/sips" API endpoint response
- Replace "UUID" with "Uploaded by" column on Dashboard "SIPs" page
- Add an "Uploaded by" field to the SIP detail page

Depends on https://github.com/artefactual-sdps/enduro/tree/dev/issue-1246-add-ingest-user-table


internal/ingest/goa.go
```diff
@@ -167,7 +168,7 @@ func (w *goaWrapper) ShowSip(
 	}
 
 	s, err := w.perSvc.ReadSIP(ctx, sipUUID)
-	if err == sql.ErrNoRows {
+	if err == persistence.ErrNotFound {
```
jraddaoui, line 171: Ups, thanks!

internal/persistence/ent/client/convert.go
```diff
@@ -13,37 +13,36 @@ import (
...
+		s.AIPID = uuid.NullUUID{UUID: sip.AipID, Valid: true}
 	}
-
-	var uploaderID *uuid.UUID
 	if sip.UploaderID != 0 && sip.Edges.User != nil {
-		uploaderID = &sip.Edges.User.UUID
+		s.Uploader = &datatypes.Uploader{
+			UUID:  sip.Edges.User.UUID,
+			Email: sip.Edges.User.Email,
+			Name:  sip.Edges.User.Name,
```
jraddaoui, line 41: After review in the other PR, we limited the fields returned in the edge, we'll need to add the email and name now too.

internal/datatypes/sip.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/enums"
 )
 
+type Uploader struct {
+	UUID  uuid.UUID
+	Email string
+	Name  string
+}
```
jraddaoui, line 19: Even if we only use some of the fields, could we reuse the existing `User` type?

jraddaoui (changes requested):
Thanks @djjuhasz!

internal/datatypes/sip.go
djjuhasz: I think the disadvantage to re-using the `User` type is that a naive developer could reasonably expect *all* of the `User` fields to be populated in the Uploader struct (e.g. that `CreatedAt` would match the database value).  I would also like to minimize the visibility of the `OIDCIss` and `OIDCSub` fields to avoid accidentally leaking the values.

jraddaoui: Cool, sounds good.

jraddaoui (approved):
Thanks @djjuhasz!

PR #1261, opened 2025-06-17
Remove duplicate API request on SIPs page

djjuhasz:
Remove a duplicate request to the `ingest/sips` endpoint when the SIPs page loads.


jraddaoui (approved):
:ok_hand:

PR #1263, opened 2025-06-18
Check auth before loading SIP workflows (#1238)

### LINKED ISSUE #1238: Problem: Dashboard error when viewing SIP without permission to list workflows
**Describe the bug**

If a Dashboard user has permissions to view a SIP (ingest:sips:read) but not to list the SIPs workflows (ingest:sips:workflows:list) then a large "Page loading error" appears at the top of the page.

**To Reproduce**

It's difficult to reproduce this bug because it requires very specific access permissions.

1. Log in to the Enduro dashboard with an account that has permissions to list and view SIPs, but not to list SIP workflows
2. Ingest a SIP
3. Click "Ingest > SIPs" in the left navigation men
4. Click on a SIP in the list

**Expected behavior**

The SIP page should display without an error message.

**Additional context**

Same issue as #1189 but for SIPs. Check related PR to fix that issue (#1190) and follow-up that changes the implementation a little (#1236).

djjuhasz:


jraddaoui (approved):
LGTM!

PR #1264, opened 2025-06-18
Don't show AIP workflow auth error (#1189)

### LINKED ISSUE #1189: Problem: Dashboard error when viewing AIP without permission to list workflows

djjuhasz:
Fix a regression that was showing an error on the AIP page when the user did not have permissions to view AIP workflows.

### DIFF
--- a/dashboard/src/stores/__tests__/aip.test.ts
+++ b/dashboard/src/stores/__tests__/aip.test.ts
@@ -7,11 +7,11 @@ import { useAipStore } from "@/stores/aip";
 import { useLayoutStore } from "@/stores/layout";
 
 const http500Error = new ResponseError(
-  <Response>{
+  new Response(JSON.stringify({}), {
     status: 500,
     statusText: "Internal Server Error",
-  },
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
LGTM!

PR #1266, opened 2025-06-20
Add failed SIP/PIP download to dashboard

jraddaoui:
- Create a new component to show the related AIP or the failed SIP/PIP
  in the same related packages card, based on the current SIP status.
- Add download action to SIP store. Two steps process to get a cookie
  from the request SIP download endpoint and redirect to a new window
  with the API download URL.
- Collect errors and display them as an alert in the card, replacing
  the download button for 5 seconds.

Refs #1202, depends on #1253.

### DIFF
--- a/dashboard/src/stores/sip.ts
+++ b/dashboard/src/stores/sip.ts
@@ -1,6 +1,6 @@
 import { acceptHMRUpdate, defineStore } from "pinia";
 
-import { api, client } from "@/client";
+import { api, client, getPath } from "@/client";
 import { IngestListSipsStatusEnum, ResponseError } from "@/openapi-generator";
 import router from "@/router";
 import { useLayoutStore } from "@/stores/layout";
@@ -37,6 +37,7 @@ export const useSipStore = defineStore("sip", {
       earliestCreatedTime: undefined as Date | undefined,
       latestCreatedTime: undefined as Date | undefined,
     },
+    downloadError: null as string | null,
   }),
   getters: {
     isPending(): boolean {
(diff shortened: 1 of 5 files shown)


dashboard/src/stores/sip.ts
```diff
@@ -172,6 +173,29 @@ export const useSipStore = defineStore("sip", {
...
+        // Try to parse the error and save it for 5 seconds. It will
+        // replace the download button with an alert including the
+        // error message in the SipRelatedPackages component.
+        if (err instanceof ResponseError) {
+          const body = await err.response.json();
+          if (body.message) {
+            this.downloadError = body.message;
+            setTimeout(() => (this.downloadError = null), 5000);
+          }
+        }
```
djjuhasz, line 196: I think it would be good to show some sort of error message when we catch an error that is not a ResponseError or is a ResponseError but there is no response body.  Something generic like "Download error" would be fine, just so the user gets some acknowledgement that they clicked the button and it something went wrong.

djjuhasz (changes requested):
Looks great @jraddaoui. I just had one request for some extra error handling.

dashboard/src/stores/sip.ts
```diff
@@ -172,6 +173,29 @@ export const useSipStore = defineStore("sip", {
...
+      }
```
jraddaoui: @djjuhasz [I asked you once](https://github.com/artefactual-sdps/enduro/pull/1142#discussion_r1985037383) to use `then/catch` instead of `try/catch` + `async/await`. After working on this, I fully regret that call, sorry for the troubles I put you into, that was my old JS brain.

I initially had this working with then/catch chaining the two promises and, when I got to the tests, I had to use:

```
await store.download();
await vi.runAllTicks();
await Promise.resolve();
await nextTick();
```

Removing any of them made the tests fail :facepalm:. I also think this way makes the code cleaner and more readable and simplifies error handling.

djjuhasz: Haha, I just did some Googling on the difference between the `then().catch().finally()` syntax and the `try/catch` with an await, and it does seem like `await Promise()` is a lot simpler for many scenarios. :)

dashboard/src/stores/sip.ts
jraddaoui: Done in https://github.com/artefactual-sdps/enduro/pull/1266/commits/13099d05380710cc1f237cef3bde9c5d819eb59a.

djjuhasz (approved):
Looks great, thanks @jraddaoui! :+1:

PR #1269, opened 2025-06-23
Fix SIP/PIP download button display

jraddaoui:
Do not show the SIP/PIP download button if the SIP doesn't have failed as or failed key values.

Refs #1202.


jraddaoui:
Sorry @djjuhasz, I just noticed I messed it up while working on the error message alert ;)

djjuhasz (approved):
:+1:

Issue #1262, opened 2025-06-18
Enhancement: add the ability to filter by user who uploaded SIP


sallain:
**Is your feature request related to a problem? Please describe.**

In [#1246](https://github.com/artefactual-sdps/enduro/issues/1246) we added the ability to see who uploaded a SIP. The next step in this work is to add the ability to filter by user. This is the solution for the instigating problem, which was that it can be difficult for users to track the SIPs that they are responsible for processing in systems where there are multiple users. With the filter, users will be able to select themselves

**Describe the solution you'd like**

Here's a mock-up!

![Image](https://github.com/user-attachments/assets/b3c0935a-227f-4f79-98e2-74f28181c041)

It would be nice, if possible, to have the current user floated to the top of the list. It would be annoying to always have to navigate to the bottom of the list if you had a ton of uploaders in your system and your name stared with Z.

**Describe alternatives you've considered**

**Additional context**

djjuhasz:
It could use a glow up, but the "uploaded by" filter is working:

https://github.com/user-attachments/assets/ae0f5876-0063-434f-b2a5-7e71b9abee34

fiver-watson:
Interesting.... TBH I expected it to look more like the "Started" filter - I find it a bit confusing that there is an X next to the filter even when no value has been selected, and with that and the white background, it looks at first glance like a second search box... Finally, it look like the width of the dropdown is already expanded to match the largest input - but that's not how the "started" drop-down works, which overall has a better design. 

If possible, I would suggest that we try to make it more consistent with the Start time filter's appearance and behavior. But still, it's great to see this functionality working!

One last question, maybe this one more for the team: if "Unknown" is a potential outcome for the uploader name (as in the first SIP in your test data, which I assume would come from a watched directory or similar), is it possible to have "Unknown" as an option in the dropdown? Do we want that? They might not all be the _same_ "unknown"... but at the same time, as an Administrator I can imagine wanting to be able to see all ingests without an associated user during a review. So: on the technical side first, is that something we can do?

djjuhasz:
@fiver-watson okay, I've matched the "Uploaded by" dropdown style to the "Started" filter:

https://github.com/user-attachments/assets/8cea6645-71b9-4403-b7a8-e2385d0da749

I agree that we should add an "Unknown" option to the "Uploaded by" list, but that's going to require some work to differentiate between showing all "uploaders" and just showing the SIPs where the uploader is not set.   This is going to have to wait until next iteration.

[...]

PR #1270, opened 2025-06-23
Add an `ingest/users` API endpoint (refs #1262)

djjuhasz:
- Add an `ingest/users` API endpoint with search options
- Add a `ListUsers()` method to the persistence service
- Add a user filter type
- Add an `ingest:users:list` auth attribute


internal/api/design/ingest.go
```diff
@@ -373,6 +409,41 @@ var SIPWorkflows = ResultType("application/vnd.enduro.ingest.sip.workflows", fun
...
+		TypedAttributeUUID("uuid", "Identifier of the user")
+		Attribute("email", String, "Email of the user", func() {
+			Example("nobody@example.com")
+		})
+		Attribute("name", String, "Name of the user", func() {
+			Example("Jane Doe")
+		})
+		Attribute("created_at", String, "Creation date & time of the user", func() {
+			Format(FormatDateTime)
+		})
```
jraddaoui, line 427: Do we care showing this right now?

internal/ingest/convert.go
```diff
@@ -111,6 +111,20 @@ func listSipsPayloadToSIPFilter(payload *goaingest.ListSipsPayload) (*persistenc
 	return &pf, nil
 }
 
+//nolint:unparam  // may return an error in the future
```
jraddaoui: Why is this needed?

internal/ingest/convert.go
```diff
@@ -111,6 +111,20 @@ func listSipsPayloadToSIPFilter(payload *goaingest.ListSipsPayload) (*persistenc
...
+func listUsersPayloadToUserFilter(payload *goaingest.ListUsersPayload) (*persistence.UserFilter, error) {
+	f := persistence.UserFilter{
+		Email: payload.Email,
```
jraddaoui: The usual, should we check `payload` is not `nil`, just saying because I just saw it in the user's `Goa()` method.

internal/persistence/ent/client/user.go
```diff
@@ -77,3 +81,63 @@ func (c *client) ReadOIDCUser(ctx context.Context, iss, sub string) (*datatypes.
...
+		user.FieldCreatedAt: {Name: "CreatedAt"},
+	}
+}
+
+// filterUsers applies the User filter f to query q and return a paginated amd
+// unpaginated query.
+func filterUsers(q *db.UserQuery, f *persistence.UserFilter) (page, whole *db.UserQuery) {
+	qf := entfilter.NewFilter(q, sortableFields())
+
+	qf.AddDateRange(sip.FieldCreatedAt, f.CreatedAt)
```
jraddaoui, line 132: I see, we have it in the filter and in the sortable fields, but we can't set any of those in the API request. No need to remove it in any case, we may need it at some point.

jraddaoui (approved):
Looks great @djjuhasz, thanks!

internal/ingest/convert.go
djjuhasz: Unparam was flagging that the error return is always nil.  I could drop the error return value, but I am trying to match the signature of `listSipsPayloadToSIPFilter()`.

internal/api/design/ingest.go
djjuhasz: No, but it seems like the kind of thing we might want to show in the future, it's easy to add now, and it's not sensitive data.

internal/ingest/convert.go
jraddaoui: Right! I'd not return the error then, I don't think we need to match `listSipsPayloadToSIPFilter`, and not having to worry about the error when calling it is always nice.

```go
r, pg, err := w.perSvc.ListUsers(ctx, listUsersPayloadToUserFilter(payload))
```

djjuhasz: I added simple string length validation for the email and name parameters, so this can return an error legitimately now.

internal/ingest/convert.go
djjuhasz: Done. This also made me realize that this code wasn't being tested, so I fixed that too. :)

Issue #1265, opened 2025-06-20
Problem: AIP download endpoint bypasses authentication and access control checks


jraddaoui:
**Describe the bug**

The storage service API download AIP endpoint allows to download AIPs as an unauthenticated user when authentication is enabled in the configuration. It's also possible to download them as an authenticated user without the `storage:aips:download` attribute in the claims.  

**To Reproduce**

1. Enable authentication and ABAC in the API configuration.
2. Try to download an AIP directly through the API as an unauthenticated user.
3. Or as an authenticated user without the `storage:aips:download` attribute.
4. AIP is downloaded.

**Expected behavior**

The AIP should not be downloaded, a `401 Unauthorized` or `403 Forbiden` response should be returned.

**Additional context**

The Goa authentication and access control setup is bypassed when the Goa handler is replaced by a custom one to set the write timeout:

https://github.com/artefactual-sdps/enduro/blob/main/internal/api/api.go#L65

Additionally, browsers don't allow to set custom headers (like Authorization) when downloading files via `window.open()` or `<a href="...">`. And using the fetch API with custom headers requires loading the entire file into memory, which is not suitable for large files. To work around this, check the two-step download process implemented for failed SIP/PIP downloads in https://github.com/artefactual-sdps/enduro/pull/1253.

Following iterations could allow to download the AIP in a single step process for out of browser requests to the API.

fiver-watson:
Seems to be fixed?

PR #1271, opened 2025-06-24
Fix AIP download authn/authz checks 

jraddaoui:
Follow the implementation for SIP download made in #1253 and #1266.
Do not overwrite the download endpoint to set the wirte timeout, create
request download and actual download endpoints following the Goa server
stubs. Update dashboard code to request in two steps using a cookie.

Refs #1265.

### DIFF
--- a/dashboard/src/components/SipReviewAlert.vue
+++ b/dashboard/src/components/SipReviewAlert.vue
@@ -16,6 +16,10 @@ const emit = defineEmits<{
 const aipStore = useAipStore();
 const sipStore = useSipStore();
 
+if (sipStore.current?.aipId) {
+  aipStore.fetchCurrent(sipStore.current.aipId);
+}
+
 const confirm = async () => {
   const locationId = await openDialog(LocationDialog);
   if (!locationId) return;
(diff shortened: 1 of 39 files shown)


jraddaoui:
@djjuhasz, this is ready for code review if you have some time before leaving.

dashboard/src/components/SipReviewAlert.vue
```diff
@@ -13,6 +13,7 @@ const emit = defineEmits<{
   (e: "update:expandCounter", value: number): void;
 }>();
 
+// TODO: Load current AIP in aipStore to make download work.
```
djjuhasz: Is download not working with this PR then?

djjuhasz (approved):
LGTM @jraddaoui.  There was just one surprising TODO that suggests AIP download isn't working.

dashboard/src/components/SipReviewAlert.vue
jraddaoui: There is an AIP download link in the review alert from the create and review AIP workflow, it's just there where it won't  work for now.

jraddaoui: I fixed this before merging ;)

PR #1275, opened 2025-06-24
Add an "Uploaded by" filter to the SIPs page

### LINKED ISSUE #1262: Enhancement: add the ability to filter by user who uploaded SIP

djjuhasz:
Fixes #1262.

Add an "Uploaded by" filter to the Dashboard SIPs page to allow filtering SIPs by the uploading user.

### DIFF
--- a/dashboard/src/pages/ingest/sips/index.vue
+++ b/dashboard/src/pages/ingest/sips/index.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { useAsyncState } from "@vueuse/core";
+import Dropdown from "bootstrap/js/dist/dropdown";
 import Tooltip from "bootstrap/js/dist/tooltip";
 import { computed, onMounted, ref, watch } from "vue";
 import { useRoute, useRouter } from "vue-router/auto";
@@ -18,6 +19,7 @@ import type { IngestListSipsStatusEnum } from "@/openapi-generator";
 import { useAuthStore } from "@/stores/auth";
 import { useLayoutStore } from "@/stores/layout";
 import { useSipStore } from "@/stores/sip";
+import { useUserStore } from "@/stores/user";
 import IconInfo from "~icons/akar-icons/info-fill";
 import IconAll from "~icons/clarity/blocks-group-line?raw&font-size=20px";
 import IconClose from "~icons/clarity/close-line";
@@ -33,6 +35,11 @@ import IconSIPs from "~icons/octicon/package-dependencies-24";
 const authStore = useAuthStore();
 const layoutStore = useLayoutStore();
 const sipStore = useSipStore();
+const userStore = useUserStore();
+
+const uploaderEl = ref<HTMLElement | null>(null);
+const uploaderDD = ref<Dropdown | null>(null);
+const uploaderDDLabel = ref("Uploaded by");
 
 const route = useRoute();
 const router = useRouter();
@@ -223,6 +230,25 @@ const updateDateFilter = (
   });
 };
 
+const updateUploaderFilter = () => {
+  let q = { ...route.query };
+
+  if (sipStore.filters.uploaderId) {
+    q.uploaderId = sipStore.filters.uploaderId as LocationQueryValue;
+  } else {
+    delete q.uploaderId;
+  }
+
+  // Reset the page number because the found results may reduce the total number
+  // of pages.
+  delete q.page;
+
+  router.push({
+    name: "/ingest/sips/",
+    query: q,
+  });
+};
+
 const { execute, error } = useAsyncState(() => {
   if (route.query.name) {
     sipStore.filters.name = <string>route.query.name;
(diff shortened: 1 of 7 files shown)


jraddaoui:
@djjuhasz ping me when is fully ready, it has told me to refresh the diff a couple of times since I started the review.

dashboard/src/stores/user.ts
```diff
@@ -0,0 +1,81 @@
...
+  constructor(user: api.EnduroIngestUser) {
+    this.createdAt = user.createdAt || new Date();
+    this.email = user.email || "";
+    this.name = user.name || "";
+    this.uuid = user.uuid || "";
+  }
+
+  toString(): string {
+    return this.name || this.email || this.uuid || "Unknown";
+  }
```
jraddaoui: In this context the users will always have UUID, right? It's not like the uploader that may not exist.

dashboard/src/stores/user.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+  },
+  actions: {
+    async fetchUsers() {
+      try {
+        const response = await client.ingest.ingestListUsers({
+          email: this.filters.email,
+          name: this.filters.name,
+          limit: this.page.limit,
+          offset: this.page.offset,
```
jraddaoui, line 36: I know it's a bit unmanageable to have a really long select and I guess you are planning to add filtering to the actual select values. In the meantime I'd remove pagination, I'd prefer to show all users even if there are more than 20 to not be able to filter by a given one.

dashboard/src/stores/user.ts
```diff
@@ -0,0 +1,81 @@
...
+// User is a wrapper around api.EnduroIngestUser that provides additional
+// functionality, such as a formatted string representation.
+export class User implements api.EnduroIngestUser {
+  createdAt: Date;
+  email: string;
+  name: string;
+  uuid: string;
+
+  constructor(user: api.EnduroIngestUser) {
+    this.createdAt = user.createdAt || new Date();
```
jraddaoui: Similar here, the created at date is a required field in the API responses. Having a default value here feels wrong.

dashboard/src/stores/user.ts
```diff
@@ -0,0 +1,81 @@
...
+    this.email = user.email || "";
+    this.name = user.name || "";
+    this.uuid = user.uuid || "";
+  }
+
+  toString(): string {
+    return this.name || this.email || this.uuid || "Unknown";
+  }
+}
```
jraddaoui: In general I think this wrapper adds very little value, I think an action function in the store to get the display name/email with an API user as parameter would be more than enough.

jraddaoui (changes requested):
Looks neat @djjuhasz, really nice incremental work for the entire feature. Just some minor details for this one.

dashboard/src/stores/user.ts
djjuhasz: Yes, I think we'll need to add some sort of search / autocomplete functionality to the uploader filter in the future.

The API doesn't support "removing" pagination, but I've bumped the limit up to 100 for this query — I think the dropdown will be unmanageable after that point.

dashboard/src/stores/user.ts
djjuhasz: This User class is gone now.

dashboard/src/stores/user.ts
djjuhasz: Yes, true.

djjuhasz:
@jraddaoui this is ready for another look.  I'll squash the commits before merging.

dashboard/src/pages/ingest/sips/index.vue
```diff
@@ -33,6 +35,14 @@ import IconSIPs from "~icons/octicon/package-dependencies-24";
...
 const layoutStore = useLayoutStore();
 const sipStore = useSipStore();
+const userStore = useUserStore();
+
+const uploaderEl = ref<HTMLElement | null>(null);
+const uploaderDD = ref<Dropdown | null>(null);
+const uploaderDDLabel = ref("Uploaded by");
+
+// Increase the default limit for the uploader dropdown to 100 users.
+userStore.page.limit = 100;
```
jraddaoui: What's the point of doing it from here? This will change it permanently after visiting this page.

jraddaoui (approved):
Thanks @djjuhasz, LGTM!

dashboard/src/pages/ingest/sips/index.vue
djjuhasz: I don't think it actually matters at this point because this is the only page using the userStore, and there is no way to adjust the limit from the UI at this point.   In any case, I've changed the default limit in the user store instead.

Issue #1276, opened 2025-06-25
Problem: the deletion of the original SIP fails for internal uploads if preprocessing is enabled


jraddaoui:
**Describe the bug**

When the SIP deposit is done through Enduro's UI/API and a preprocessing child workflow is configured, the download is done directly into the preprocessing shared directory. We use different activities to download from the internal bucket or from watched bucket and, if preprocessing is enabled we set the destination path to the shared folder, so this is the current destination for the different options (`/tmp` will be the system's temporary directory, `1234` will be a random string and `/sharedpath` is the configured preprocessing shared path):

- Internal bucket without preprocesing: `/tmp/bucketdownload1234/sip.zip`
- Internal bucket with preprocessing: `/sharedpath/sip.zip`
- Watched bucket without preprocesing: `/tmp/enduro1234/sip.zip`
- Watched bucket with preprocessing: `/sharedpath/enduro1234/sip.zip`

After download, we mark the parent directory of the downloaded SIP for deletion at the end of the workflow:

- Internal bucket without preprocesing: `/tmp/bucketdownload1234`
- Internal bucket with preprocessing: `/sharedpath`
- Watched bucket without preprocesing: `/tmp/enduro1234`
- Watched bucket with preprocessing: `/sharedpath/enduro1234`

Deleting the entire shared path is not desired. Also, downloading the SIP to the same path could cause collisions for SIPs with the same filename.

**To Reproduce**

1. Enable and configure a preprocessing child workflow.
2. Upload a SIP through Enduro's UI/API.
3. Check Temporal logs for the remove paths activity at the end.

**Expected behavior**

The downloaded SIP is deleted at the end of the workflow without failure and without deleting the entire shared folder.

**Additional context**

Initially reported by @scollazo as part of #1272. To avoid using the default temporary directory for processing we have #926.

sallain:
LGTM!

PR #1277, opened 2025-06-25
Fix SIP download from internal bucket (#1276)

jraddaoui:
When preprocessing is enabled and a shared path configured, add a parent directory with the SIP UUID to avoid collisions and normalize the deletion of the original SIP from the parent directory.

Refs #1276.


djjuhasz (approved):
LGTM, thanks @jraddaoui. :+1:

PR #1283, opened 2025-07-09
Use UUIDs as identifiers for workflows and tasks

jraddaoui:
- Add UUID columns to ingest workflows and tasks.
- Use workflow UUID and a transaction to create tasks.
- Replace workflow and task IDs by UUIDs in API design.
- Update dashboard API client and use UUIDs instead of IDs.

Refs #1209.

### DIFF
--- a/internal/db/migrations/20250709054510_add_uuid_columns.up.sql
+++ b/internal/db/migrations/20250709054510_add_uuid_columns.up.sql
@@ -1,0 +1,10 @@
+-- Rename "task_id" column to "uuid" in "task" table and add unique index
+ALTER TABLE `task` CHANGE COLUMN `task_id` `uuid` CHAR(36) NOT NULL, ADD UNIQUE INDEX `uuid` (`uuid`);
+-- Add nullable "uuid" column to "workflow" table
+ALTER TABLE `workflow` ADD COLUMN `uuid` CHAR(36) NULL;
+-- Add "uuid" values
+UPDATE `workflow` SET `uuid` = BIN_TO_UUID(
+    RANDOM_BYTES(16) & 0xffffffffffff0fff3fffffffffffffff | 0x00000000000040008000000000000000
+) WHERE uuid IS NULL;
+-- Make "uuid" column not nullable and add unique index
+ALTER TABLE `workflow` MODIFY COLUMN `uuid` CHAR(36) NOT NULL, ADD UNIQUE INDEX `uuid` (`uuid`);
--- a/dashboard/src/components/__tests__/Task.test.ts
+++ b/dashboard/src/components/__tests__/Task.test.ts
@@ -12,12 +12,11 @@ describe("Task.vue", () => {
       props: {
         index: 1,
         task: {
-          id: 1,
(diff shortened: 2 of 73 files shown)


internal/db/migrations/20250709054510_add_uuid_columns.up.sql
```diff
@@ -0,0 +1,10 @@
+-- Rename "task_id" column to "uuid" in "task" table and add unique index
+ALTER TABLE `task` CHANGE COLUMN `task_id` `uuid` char(36) NOT NULL, ADD UNIQUE INDEX `uuid` (`uuid`);
```
sevein: For consistency, `CHAR(36)`.

internal/db/migrations/20250709054510_add_uuid_columns.up.sql
```diff
@@ -0,0 +1,10 @@
...
+-- Add nullable "uuid" column to "workflow" table
+ALTER TABLE `workflow` ADD COLUMN `uuid` CHAR(36) NULL;
```
sevein, line 4: Have we ever tried the native UUID type in MySQL 8? Looks like Temporal [prefers](https://github.com/temporalio/temporal/blob/main/schema/mysql/v8/temporal/schema.sql) `BINARY(16)`.

sevein (approved):
LGTM!

One potential consideration for the future could be adopting a naming convention where field names aren't tied to the storage type. I think this is frequently described in various Go style guides ([this one] from Dave Cheney off the top of my head). I think the main idea is that `WorkflowID` would still make sense even though the type changes to `uuid.UUID`.

Asked o3 to word the convention for me and it gave me this:
> Use plain `ID` for the record's own primary key, prefix foreign keys with the referenced entity as in `WorkflowID` or `UserID`, fall back to `<Entity>PK` for the rare composite or legacy key that is not a simple surrogate, and reserve `<Entity>Ref` or `<Entity>ExternalID` for handles coming from outside systems.

Some advantages:
* Less refactoring if the column type changes again.
* Predictable grepping, e.g. `git grep WorkflowID`.
* No accidental breaking in encoding formats, e.g. JSON etc.

[this one]: https://dave.cheney.net/2019/01/29/you-shouldnt-name-your-variables-after-their-types-for-the-same-reason-you-wouldnt-name-your-pets-dog-or-cat

internal/db/migrations/20250709054510_add_uuid_columns.up.sql
jraddaoui: Yeah, I saw that when I was looking for ways to generate the UUIDs, not sure why Ent/Atlas decide to use char.

jraddaoui:
Thanks @sevein!

Good point, we may have left the use of `id` in the API but we use `{uuid}` in the URL patterns so I thought that would be clearer (not so clear for workflows and tasks as they don't have specific URLs, but it's the same for other resources). Internally, we have a little mess about using `id` or `uuid` as we have both in the database and both are used for now, so I thought it would be better to make that distinction.

PR #1287, opened 2025-07-15
Add a script to get an API access token

jraddaoui:


Tiltfile
```diff
@@ -99,8 +99,18 @@ if os.environ.get('TRIGGER_MODE_AUTO', '').lower() in true:
...
+k8s_resource(
+  "enduro",
+  labels=["Enduro"],
+  port_forwards=["9000:9000", "9002:9002"],
+  trigger_mode=trigger_mode
+)
+k8s_resource(
+  "enduro-dashboard",
+  port_forwards="8080:80",
+  labels=["Enduro"],
```
djjuhasz, line 110: Nitpick: Let's use a consistent order for the resource properties please (labels before port_forwards) :sweat_smile:

hack/auth/main.go
```diff
@@ -0,0 +1,116 @@
...
+		ClientID:    client,
+		RedirectURL: "urn:ietf:wg:oauth:2.0:oob",
+		Scopes:      scopes,
+		Endpoint: oauth2.Endpoint{
+			AuthURL:  fmt.Sprintf("%s/protocol/openid-connect/auth", host),
+			TokenURL: fmt.Sprintf("%s/protocol/openid-connect/token", host),
+		},
+	}
+
+	codeVerifier, err := codeVerifier()
```
djjuhasz: I think it's a bit confusing to shadow the function name with the `codeVerifier` var, and it could potentially lead to errors if this code is ever extended.

hack/auth/main.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		fmt.Printf("Unable to generate code verifier: %v\n", err)
+		os.Exit(1)
+	}
+	codeChallenge := codeChallenge(codeVerifier)
```
djjuhasz: shadowing again.

djjuhasz (approved):
Seems okay @jraddaoui.  I had to do a `tilt down` and `tilt up` to get the OOB redirect URL to work, but it worked fine after that.

I noted a couple of minor nitpicks.

Tiltfile
jraddaoui: That required to change a few more resources, but done.

jraddaoui:
Thanks @djjuhasz! Restarting the Keycloak resource would have been enough, sorry I forgot to mention that.

Issue #1221, opened 2025-05-13
Problem: dashboard automatic updates for the ingest domain require all permissions


jraddaoui:
**Is your feature request related to a problem? Please describe.**

When ABAC authorization is enabled in the API/dashboard a user must have all the ingest domain attributes to be able to establish a websocket connection between the dashboard and the API to get automatic updates:

https://github.com/artefactual-sdps/enduro/blob/main/internal/api/design/ingest.go#L21-L27

**Describe the solution you'd like**

Ideally, the user should be able to see updates based in their actual permissions. For example, if they can see SIPs but not their workflows, they should be able to get SIP created/updated events, but not workflow/task created/updated events.

**Describe alternatives you've considered**

More than an alternative, I'd suggest a **two steps approach** as passing the event based on the user attributes may not be trivial:

**First**, require only all read/list permissions:

- "ingest:sips:list"
- "ingest:sips:read"
- "ingest:sips:workflows:list"

Removing the two that shouldn't affect any user to get updates, only to perform the action:

- "ingest:sips:review"
- "ingest:sips:upload"

**Second**, consider options to be more granular and only send events to each connection based on the user attributes.

Issue #1057, opened 2024-10-28
Problem: Regularly get "write EPIPE" error from Dashboard in Tilt dev env


djjuhasz:
**Describe the bug**

In the Enduro Tilt dev environment when I load the package list in the Dashboard the page often fails to load a displays a totally blank white screen.  The Dashboard log has the following error messages:

```
VITE v5.4.8  ready in 705 ms

  ➜  Local:   [http://localhost:80/](http://localhost/)
  ➜  Network: use --host to expose
3:32:41 PM [vite] ✨ new dependencies optimized: @vueuse/core, bootstrap/js/dist/tooltip, @vue/reactivity
3:32:41 PM [vite] ✨ optimized dependencies changed. reloading
Reconnecting... Error port-forwarding enduro-dashboard (8080 -> 80): creating stream: Timeout occurred
3:33:19 PM [vite] ws proxy socket error:
Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:161:15)
    at writeGeneric (node:internal/stream_base_commons:152:3)
    at Socket._writeGeneric (node:net:954:11)
    at Socket._write (node:net:966:8)
    at writeOrBuffer (node:internal/streams/writable:570:12)
    at _write (node:internal/streams/writable:499:10)
    at Writable.write (node:internal/streams/writable:508:10)
    at Socket.ondata (node:internal/streams/readable:1007:22)
    at Socket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:559:12)
```

**To Reproduce**

1. Start the Tilt dev environment with `tilt up`
2. Add a one or more SIPs to the the Minio "sips" bucket to start processing
3. Open the Enduro Dashboard in a browser window (http://localhost:8080/packages)
4. Click on the "Packages" link on the left side of the screen
5. Page fails to load, and after some time (30s?) the above errors are written the Tilt log for the "enduro-dashboard" resource

**Expected behavior**

The package list should load without errors

**Screenshots**

![image](https://github.com/user-attachments/assets/893f7e73-d84f-4735-9a66-4910d948e3f0)

**Additional context**

If I wait a couple of minutes then try to reload http://localhost:8080/packages it usually loads fine.

PR #1288, opened 2025-07-15
Send websocket events based on user attributes

jraddaoui:
- Allow saving key/value in the auth ticket store
- Use constants for API auth attributes/scopes
- Add auth.TicketProvider interface
- Send websocket events based on user attributes

Refs #1221.


jraddaoui:
@djjuhasz, this is ready for an initial code review, just lacking tests for the monitor endpoints. It'll make more sense if you review commit by commit, thanks ;)!

internal/api/auth/ticket_test.go
```diff
@@ -44,14 +44,15 @@ func TestTicketProviderRequest(t *testing.T) {
...
 
 		store.EXPECT().
-			SetEx(gomock.Any(), ticket, auth.TicketTTL).
+			SetEx(gomock.Any(), ticket, value, auth.TicketTTL).
 			Return(nil)
 
 		provider := auth.NewTicketProvider(ctx, store, rander)
 
-		ticket, err := provider.Request(ctx)
+		ticket, err := provider.Request(ctx, value)
```
djjuhasz: `ticket` here is shadowing the value of `ticket` from the parent scope.

internal/api/auth/ticket_test.go
```diff
@@ -44,14 +44,15 @@ func TestTicketProviderRequest(t *testing.T) {
...
 		assert.NilError(t, err)
 		assert.Equal(t, ticket, ticket)
```
djjuhasz: This is always true due to the shadowing of `ticket`.

internal/api/auth/ticket_test.go
```diff
@@ -67,7 +68,7 @@ func TestTicketProviderRequest(t *testing.T) {
 		rander := iotest.ErrReader(errors.New("rand source error"))
 		provider := auth.NewTicketProvider(ctx, store, rander)
 
-		ticket, err := provider.Request(ctx)
+		ticket, err := provider.Request(ctx, nil)
 		assert.Error(t, err, "error creating ticket: rand source error")
 		assert.Assert(t, ticket == "")
```
djjuhasz: I'd expect `assert.Equal(t, ticket, "")` here and I think it will provide better reporting if the assertion fails.

internal/api/auth/ticket_test.go
```diff
@@ -81,13 +82,13 @@ func TestTicketProviderRequest(t *testing.T) {
...
+			SetEx(gomock.Any(), ticket, nil, auth.TicketTTL).
 			Return(errors.New("fake error"))
 
 		rander := rand.New(rand.NewSource(1)) //#nosec
 		provider := auth.NewTicketProvider(ctx, store, rander)
 
-		ticket, err := provider.Request(ctx)
+		ticket, err := provider.Request(ctx, nil)
 		assert.Error(t, err, "error storing ticket: fake error")
 		assert.Assert(t, ticket == "")
```
djjuhasz: Again, I think `assert.Equal()` is better here.

internal/api/auth/ticket_test.go
```diff
@@ -154,7 +161,7 @@ func TestTicketProviderCheck(t *testing.T) {
 		rander := rand.New(rand.NewSource(1)) //#nosec
 		provider := auth.NewTicketProvider(ctx, nil, rander)
 
-		err := provider.Check(ctx, nil)
+		err := provider.Check(ctx, nil, nil)
```
djjuhasz, line 164: What happens if you pass a ticket here?  Does the check always succeed if no store is defined?

internal/api/auth/attributes.go
```diff
@@ -0,0 +1,25 @@
+package auth
+
+const (
+	IngestSipsDownloadAttr         = "ingest:sips:download"
```
djjuhasz: According to our style guide "SIP" and "AIP" should be all caps: https://google.github.io/styleguide/go/decisions.html#initialisms

jraddaoui:
@djjuhasz, there is a "small" back and forward between https://github.com/artefactual-sdps/enduro/pull/1288/commits/e5df4344853e87f00f4f033e4427cf867dc417a0 and https://github.com/artefactual-sdps/enduro/pull/1288/commits/0e9b2aac6449187fc63a620fc9daac5dfa6e2476 that I realized after adding the tests.

This makes me wonder if we should just log the errors instead of breaking that loop (except context done).

jraddaoui:
Maybe that's the issue in https://github.com/artefactual-sdps/enduro/issues/1057, I'll take a look.

I'll squash the latest three commits ;)

internal/ingest/monitor_test.go
```diff
@@ -0,0 +1,247 @@
...
+				&goaingest.SIPUpdatedEvent{UUID: testUUID},
+				&goaingest.SIPStatusUpdatedEvent{UUID: testUUID},
+				&goaingest.SIPWorkflowCreatedEvent{UUID: testUUID},
+				&goaingest.SIPWorkflowUpdatedEvent{UUID: testUUID},
+				&goaingest.SIPTaskCreatedEvent{UUID: testUUID},
+				&goaingest.SIPTaskUpdatedEvent{UUID: testUUID},
+			},
+		},
+		{
+			name: "Filters all events for a user wihtout permissions",
```
djjuhasz: Typo: "wihtout"

djjuhasz (changes requested):
LGTM @jraddaoui, I just flagged some minor issues.

internal/api/auth/attributes.go
jraddaoui: `IngestSIPsDownloadAttr` or `IngestSIPSDownloadAttr` :thinking:

jraddaoui: I'll go with `IngestSIPSDownloadAttr`.

internal/api/auth/ticket_test.go
jraddaoui: Yes: https://github.com/artefactual-sdps/enduro/blob/main/internal/api/auth/ticket.go#L67

TBH, I never liked how we deal with API authentication being disabled, it's all kind of hidden in the token verifier and the ticket provider.

internal/api/auth/ticket_test.go
jraddaoui: :man_facepalming:

jraddaoui:

djjuhasz (approved):
Looks great, thanks @jraddaoui! :+1:

jraddaoui:
Thanks @djjuhasz! I've squashed everything you already reviewed into the first four commits. But, while testing all possible permutations, I found a couple of issues that required small fixes. Please, take a look to the latest three commits when you have some time.

djjuhasz (approved):
The new changes look fine to me @jraddaoui.  It's not clear to me from the commits why you added a "noattributes" user, but it seems fine.

jraddaoui:
Thanks @djjuhasz, just a user that can authenticate but can't do anything, only for testing.

PR #1296, opened 2025-07-22
Add a "list source items" API endpoint

djjuhasz:
- Add a "ingest/sources/{uuid}/items" API endpoint to list SIP source items
- Add configuration for a single SIP source
- Add a "BucketSource" SIP Source implementation based on https://pkg.go.dev/gocloud.dev/blob buckets


enduro.toml
```diff
@@ -227,6 +227,22 @@ bucket = "internal"
 storageAccount = ""
 storageKey = ""
 
+# sipsource configures a SIP Source location where SIPs can be put and later
+# ingested by Enduro via the Dashboard.
```
jraddaoui: Via the Dashboard or the API.

internal/api/design/ingest.go
```diff
@@ -326,6 +326,36 @@ var _ = Service("ingest", func() {
 			})
 		})
 	})
+	Method("list_source_items", func() {
+		Description("List SIP source items")
+		Security(JWTAuth, func() {
+			Scope("ingest:sources:items:list")
```
jraddaoui: Use the constant.

internal/sipsource/config.go
```diff
@@ -0,0 +1,53 @@
...
+	ErrMissingID = errors.New("SIP source: missing ID")
+	// ErrMissingName is returned when the SIP source name is empty.
+	ErrMissingName = errors.New("SIP source: missing name")
+	// ErrMissingBucket is returned when no bucket is configured.
+	ErrMissingBucket = errors.New("SIP source: missing bucket")
+)
+
+type Config struct {
+	// ID is the UUID of the SIP source.
+	ID uuid.UUID
```
jraddaoui, line 24: What happens when an invalid UUID is used?

enduro.toml
```diff
@@ -227,6 +227,22 @@ bucket = "internal"
...
+[sipsource]
+id = "e6ddb29a-66d1-480e-82eb-fcfef1c825c5"
```
jraddaoui, line 234: We should document here this must be an UUID4.

internal/api/design/ingest.go
```diff
@@ -326,6 +326,36 @@ var _ = Service("ingest", func() {
...
+			Token("token", String)
+			Required("uuid")
+		})
+		Result(SourceItems)
+		Error("not_implemented")
+		Error("not_found")
+		Error("not_valid")
+		Error("internal_error")
+		HTTP(func() {
+			GET("/sources/{uuid}/items")
```
jraddaoui: They are named "sources" here and "sipsources" in the configuration, it would be nice to match both.

internal/api/design/ingest.go
```diff
@@ -326,6 +326,36 @@ var _ = Service("ingest", func() {
...
+			Response(StatusOK)
+			Response("not_implemented", StatusNotImplemented)
```
jraddaoui: Already implemented ;)

internal/api/design/ingest.go
```diff
@@ -505,3 +535,31 @@ var SIPTask = ResultType("application/vnd.enduro.ingest.sip.task", func() {
 	})
 	Required("uuid", "name", "status", "started_at", "workflow_uuid")
 })
+
+var SourceItem = ResultType("application/vnd.enduro.ingest.source.item", func() {
+	Description("SourceItem describes an item in a data source location.")
```
jraddaoui: SIP source?

internal/api/design/ingest.go
```diff
@@ -505,3 +535,31 @@ var SIPTask = ResultType("application/vnd.enduro.ingest.sip.task", func() {
...
+		})
+		Attribute("size", Int64, "Size of the item in bytes")
+		Attribute("is_dir", Boolean, "True if the item is a directory, false if it is a file")
+	})
+	View("default", func() {
+		Attribute("key")
+		Attribute("mod_time")
+		Attribute("size")
+		Attribute("is_dir")
+	})
```
jraddaoui: No need for a view here.

internal/config/config.go
```diff
@@ -82,8 +84,9 @@ func (c *Configuration) Validate() error {
 		c.API.Auth.Validate(),
 		c.BagIt.Validate(),
 		c.Preprocessing.Validate(),
-		c.ValidatePREMIS.Validate(),
+		c.SIPSource.Validate(),
 		c.Watcher.Validate(),
+		c.ValidatePREMIS.Validate(),
```
jraddaoui, line 88: If you were looking for alphabetical order, it should go before the watcher.

internal/ingest/goa.go
```diff
@@ -233,3 +235,43 @@ func (w *goaWrapper) ListUsers(ctx context.Context, payload *goaingest.ListUsers
...
+	if err != nil {
+		if errors.Is(err, sipsource.ErrMissingBucket) {
+			return nil, goaingest.MakeNotFound(errors.New("SIP source not found"))
+		}
+
+		w.logger.Error(err, "Error listing source items")
+		return nil, goaingest.MakeInternalError(errors.New("internal error"))
+	}
+	if page == nil || len(page.Items) == 0 {
+		return nil, goaingest.MakeNotFound(errors.New("no items found"))
```
jraddaoui: Returning not found may be confusing, I'd just return an empty list like we do in other list endpoints.

internal/ingest/goa_test.go
```diff
@@ -575,3 +578,165 @@ func TestListUsers(t *testing.T) {
...
+			},
+			wantErr: "SIP source not found",
+		},
+		{
+			name: "Returns an internal error",
+			mockRecorder: func(mr *sipsource_fake.MockSIPSourceMockRecorder) {
+				mr.ListItems(
+					mockutil.Context(),
+					nil,
+					0, // No limit specified
```
jraddaoui: Should we end inline comments with a dot too? I guess so, right? In any case, I don't think it adds a lot of value.

internal/ingest/goa_test.go
```diff
@@ -575,3 +578,165 @@ func TestListUsers(t *testing.T) {
...
+			name: "Returns an error when no items found",
+			mockRecorder: func(mr *sipsource_fake.MockSIPSourceMockRecorder) {
+				mr.ListItems(
+					mockutil.Context(),
+					nil,
+					0, // No limit specified
+				).Return(
+					&sipsource.Page{
+						Items: []*sipsource.Item{},
+						Limit: 100, // Default limit
```
jraddaoui: Same for these two.

internal/ingest/ingest_test.go
```diff
@@ -43,6 +43,7 @@ func testSvc(t *testing.T, b *blob.Bucket, s int64) (
 		b,
 		s,
 		rand.New(rand.NewSource(1)), // #nosec: G404
+		nil,                         // sipSource
```
jraddaoui: Same.

internal/sipsource/bucket_source.go
```diff
@@ -0,0 +1,101 @@
+package sipsource
```
jraddaoui, line 1: Similar, I'm getting dizzy about `sipsource` vs `source`.

internal/sipsource/bucket_source.go
```diff
@@ -0,0 +1,101 @@
...
+
+var _ SIPSource = (*SIPBucketSource)(nil)
+
+// NewBucketSource creates a new SIPSource from the provided configuration.
+func NewBucketSource(ctx context.Context, cfg *Config) (*SIPBucketSource, error) {
+	if cfg.IsEmpty() {
+		// Return an empty SIP source if the configuration is empty.
+		return &SIPBucketSource{}, nil
+	}
+	if err := cfg.Validate(); err != nil {
```
jraddaoui: Why validating again here? It's validated within the entire config already.

internal/sipsource/bucket_source.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+
+	bucket, err := bucket.NewWithConfig(ctx, cfg.Bucket)
```
jraddaoui, line 34: We should support Azure buckets here, I'll follow up about that in another comment.

internal/sipsource/bucket_source.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// ListItems returns a paged list of items in the SIP source bucket.
+//
+// If the token parameter is nil, the first page of items will be returned.
+// Subsequent calls to ListItems should provide the NextToken from the previous
+// response to retrieve the next page.
+//
+// The limit parameter specifies the maximum number of items to return. If limit
+// is less than or equal to zero, the page limit will be set to defaultLimit.
```
jraddaoui, line 64: I'd say `100` instead of `defaultLimit`.

internal/sipsource/config.go
```diff
@@ -0,0 +1,53 @@
...
+	var errs error
+	if c.ID == uuid.Nil {
+		errs = errors.Join(errs, ErrMissingID)
+	}
+	if c.Name == "" {
+		errs = errors.Join(errs, ErrMissingName)
+	}
+	if c.Bucket == nil {
+		errs = errors.Join(errs, ErrMissingBucket)
+	}
```
jraddaoui, line 49: We should validate the bucket config too, like we do with the the watched and the internal bucket.

jraddaoui (changes requested):
Looking great @djjuhasz! Some comments but nothing really big below.

jraddaoui:
As I said in one comment, it would be nice to add the existing support we have for the internal storage bucket to these sources. Daniel implemented that support in:

https://github.com/artefactual-sdps/enduro/blob/main/internal/config/internal_storage.go

But it would be great to move that to:

https://github.com/artefactual-labs/gotools/tree/main/bucket

Also, Azure has different authentication methods, and it would be nice to add support for OAuth2 alongside Shared Keys. Check this chat in Slack:

https://artefactual.slack.com/archives/C08B447USUF/p1753211678182809

It doesn't need to happen in this PR, we can create another issue for it.

jraddaoui:
Just realized, the new endpoint and attribute are missing in the required attributes table: https://github.com/artefactual-sdps/enduro/blob/main/docs/src/admin-manual/iac.md#required-attributes

internal/sipsource/bucket_source.go
djjuhasz: If we use `100` here and the `defaultLimit` changes in the future then it's easy to forgot to update the comment — you can't do a general search and replace on `100`.

internal/sipsource/bucket_source.go
djjuhasz: I agree with your comment that Azure bucket support should be implemented in https://github.com/artefactual-labs/gotools/tree/main/bucket, but I think that should be a separate issue and PR.

internal/sipsource/bucket_source.go
djjuhasz: Hmm, yeah I think I was originally using `cfg.Validate()` to make sure a bucket was configured, but now I'm doing the explicit check for a bucket.  I'll remove this.

internal/sipsource/bucket_source.go
jraddaoui: Thinking on someone reading the function docs may not be able to read defaultLimit, so it didn't add a lot of value like that. But it's not like we are publishing this package, so no worries.

internal/sipsource/config.go
djjuhasz: It returns a `400 Bad Request` response with a JSON payload:
```
{
    "name": "invalid_format",
    "id": "Neoweii1",
    "message": "uuid must be formatted as a uuid but got value \"adfe\", uuid: adfe: invalid UUID length: 4",
    "temporary": false,
    "timeout": false,
    "fault": false
}
```

jraddaoui: I meant in the configuration, using "invalid-uuid" in the TOML file.

djjuhasz: Using a nil UUID returns an error about the UUID format instead:
```
{
    "name": "invalid_format",
    "id": "Ixfgr-dZ",
    "message": "uuid must be formatted as a uuid but got value \"00000000-0000-0000-0000-000000000000\", uuid: expected RFC4122 format, but got Reserved",
    "temporary": false,
    "timeout": false,
    "fault": false
}
```
:thinking:

jraddaoui: Oh, that's the error from the configuration I guess, It looked like one of those HTTP response errors, sorry.

internal/api/design/ingest.go
djjuhasz: Yeah, I waffled back and forth between "sipsource" and "source".  I wanted to leave open the possibility of other types of (non-SIP) "source" locations, but I have don't know what those types may be.  I'll switch to "sipsource" throughout.

internal/sipsource/config.go
djjuhasz: Both of my examples were HTTP response errors. I'll try an invalid UUID in the config now.

jraddaoui: No idea what the Reserved format is, but makes sense to me, it has the UUID structure but it's not UUID4.

djjuhasz: If the UUID value in the config is not a valid UUID then the worker panics on startup with an error:
```
Failed to read configuration: failed to unmarshal configuration: 1 error(s) decoding:

* error decoding 'SIPSource.ID': invalid UUID length: 7
```

jraddaoui: Meh, not ideal, but not the worst either. I defeats the idea of returning all config errors, but I guess that's what you get when you want to use an UUID type directly in the config struct. I'd consider making that a string in the struct and parse on validation to make sure that string is a valid UUID.

jraddaoui: And thanks for looking, I just remembered that we have the decoder:

https://github.com/artefactual-sdps/enduro/blob/main/internal/config/config.go#L168-L178

And that this issue already exists:

https://github.com/artefactual-sdps/enduro/blob/main/internal/storage/config.go#L8

So up to you, not a huge deal.

internal/ingest/ingest_test.go
djjuhasz: Considering that the `ingest.NewService()` constructor has 12 parameters, I find it helpful to provide some hint as to which one is being set to `nil`.  I think we should probably refactor the function to either take a data structure (e.g. `ingest.ServiceParams`) as a param or use the `ingest.NewService().WithOption()` pattern to make the constructor easier to understand and call.

In this case I've updated the `ingest.NewService()` call to use variables with descriptive names where the parameter value doesn't make it clear what is being set. In the case of the SIP source parameter I'm passing an empty `&sipsource.BucketSource{}` value, to provide context.

internal/api/design/ingest.go
```diff
@@ -326,27 +326,25 @@ var _ = Service("ingest", func() {
...
 		})
-		Result(SourceItems)
-		Error("not_implemented")
+		Result(SIPSourceObjects)
 		Error("not_found")
 		Error("not_valid")
 		Error("internal_error")
 		HTTP(func() {
-			GET("/sources/{uuid}/items")
+			GET("/sip_sources/{uuid}/objects")
```
jraddaoui: We have been using hyphens as separator in the URLs.

hack/kube/components/dev/keycloak.yaml
```diff
@@ -120,6 +120,7 @@ data:
               "ingest:sips:read",
               "ingest:sips:upload",
               "ingest:sips:workflows:list",
+              "ingest:sources:items:list",
```
jraddaoui: Forgot this one: `ingest:sipsources:objects:list`

jraddaoui (approved):
Thanks @djjuhasz! Looks great!

Issue #1222, opened 2025-05-13
Problem: there is no automatic updates in the dashboard for the storage concern


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Enduro's API is split in two different services, ingest and storage.

The API service for the ingest domain has a websocket implementation and a couple of endpoint that can be used to establish a permanent connection and get events from the dashboard to perform automatic updates.

The storage service lacks such implementation in both sides, there is no monitor endpoints and the API doesn't send events when data is added/changed, and the dashboard is not ready to accept those events en reflect them.

**Describe the solution you'd like**

Follow the same implementation we have in the ingest service for the storage service.

**Describe alternatives you've considered**

None.

**Additional context**

#1221 will make some changes to the ingest service implementation, this issue should be addressed after that one to follow the same patterns regarding authorization.

PR #1298, opened 2025-07-24
Add related UUID to storage workflows and tasks

jraddaoui:
Add AIP UUID to Workflows and Workflow UUID to tasks in the API design,
they will be needed to look for the related resource in the dashboard
when the Websocket events are received.

Also, normalize usage of UUID vs ID in API design. Now I understand better
why @sevein said: https://github.com/artefactual-sdps/enduro/pull/1283#pullrequestreview-3004118843.

Let me know your thoughts @djjuhasz, there are many places still using
`LocationID` or `AIPID` (variables, params, etc), but I thought it would be
better to be consistent in the API. Going the other way will require more
changes, but we could still do that if you think is better.

Refs #1222.


djjuhasz (commented):
@jraddaoui thanks for your work and for opening a discussion on how to label UUIDs in Enduro. 

I think @sevein (and ChatGPT :laughing:) make some good points about preferring "WorkflowID" to "WorkflowUUID", especially in the API and front-end where we never want to expose database identifiers.  My preference would be to use "ID" for UUIDs in the API and front-end. 

I think in the database layer (+ ORM) we need to have both a primary key (int) and an UUID column based on previous research that showed that using a UUID as a primary key in MySQL + Ent wasn't really practical.  I think it's fine to use "id" for the primary key, and "uuid" for the UUID column in the database. Switching from a UUID different type of identifier at the database level would require adding a new column with a new datatype I think, so the arguments about obscuring the underlying implementation don't apply IMHO.

For the middle layer of the application (the service layer?) I think we would ideally be using UUIDs exclusively for identifying and retrieving data objects, in which case using the "ID" naming convention is unambiguous, concise and doesn't imply a particular implementation.  That said, I don't think it's going to be quite a bit of work to change from using database keys to UUIDs throughout the service layer, so I don't think you should try and do that in this PR.  For now, I think it's fine to use e.g. "UUID" and "WorkflowUUID" in the service layer to distinguish UUIDs from database keys.

All that said, I don't want to make you do a bunch of extra work here, so I'm fine if you want to go ahead and merge this PR as is.  I did review the code and didn't see any issues.

jraddaoui:
Cool, yeah, I kind of agree. Lets merge this then, so I can continue with the Websocket updates PRs, at least it's normalized in the design using `uuid` for URL patterns a field names. I'll create another issue to change that to `id` across the entire API, and we'll get to normalizing the backend like that at some point; ideally removing raw SQL altogether and making a decision about the types we use (Goa vs Ent vs custom).

PR #1299, opened 2025-07-24
Create transformKeys helper function

jraddaoui:
Parses a JSON string and recursively transforms all object keys to snake_case, returning the transformed object. Needed to transform the Websocket events to match the HTTP schema.

Refs #1222.


djjuhasz (approved):
LGTM @jraddaoui. Thanks!

PR #1301, opened 2025-07-25
Add storage events and monitor endpoints design

jraddaoui:
Add expected WebSocket events and the monitor request and monitor endpoints to the storage service API design. Update naming of ingest service events to avoid collisions and improve the generated schema. This changes somehow added the Location representation to the schema which updated the type name for some enums.

Refs #1222.


internal/event/redis.go
```diff
@@ -111,7 +111,7 @@ func (s *SubscriptionRedisImpl) loop() {
 				s.logger.Error(err, "Monitor event is invalid.")
 				continue
 			}
-			s.c <- client.NewMonitorEventOK(&payload)
+			s.c <- client.NewMonitorIngestEventOK(&payload)
```
djjuhasz, line 114: Should we remove "Monitor" from `client.NewMonitorIngestEventOK` too?

djjuhasz (approved):
Looks good @jraddaoui.  I only had one question.

internal/event/redis.go
jraddaoui: This is generated code, based on the endpoint name.

PR #1302, opened 2025-07-26
Update event package for different event types

jraddaoui:
- Add generic Service[T] interface supporting type-safe event handling.
- Add IngestEventService and StorageEventService with type aliases.
- Add Serializer interface for ingest/storage events serialization.
- Support separate Redis channels for ingest and storage via config.
- Send Redis events without blocking and consider stop channel.
- Skip events when the buffer is fills up.
- Increase buffer size to 256.
- Add storage publish helper.

Refs #1222, depends on #1301.

### DIFF
--- a/internal/event/inmem.go
+++ b/internal/event/inmem.go
@@ -5,28 +5,23 @@ import (
 	"sync"
 
 	"github.com/google/uuid"
-
-	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
 )
 
-// EventServiceInMemImpl represents a service for managing events in the system.
-type EventServiceInMemImpl struct {
+// serviceInMemImpl represents a generic service for managing events in memory.
+type serviceInMemImpl[T any] struct {
 	mu   sync.Mutex
-	subs map[uuid.UUID]*SubscriptionInMemImpl
+	subs map[uuid.UUID]*subscriptionInMemImpl[T]
 }
 
-// NewEventServiceInMemImpl returns a new instance of EventService.
-func NewEventServiceInMemImpl() *EventServiceInMemImpl {
-	return &EventServiceInMemImpl{
-		subs: map[uuid.UUID]*SubscriptionInMemImpl{},
+// newServiceInMem returns a new instance of a generic event service.
+func newServiceInMem[T any]() *serviceInMemImpl[T] {
+	return &serviceInMemImpl[T]{
+		subs: map[uuid.UUID]*subscriptionInMemImpl[T]{},
 	}
 }
 
-// PublishEvent publishes event to all of a user's subscriptions.
-//
-// If user's channel is full then the user is disconnected. This is to prevent
-// slow users from blocking progress.
-func (s *EventServiceInMemImpl) PublishEvent(ctx context.Context, event *goaingest.IngestEvent) {
+// PublishEvent publishes event to all subscriptions.
+func (s *serviceInMemImpl[T]) PublishEvent(ctx context.Context, event T) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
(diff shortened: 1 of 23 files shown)


internal/event/inmem.go
```diff
@@ -78,23 +73,25 @@ func (s *EventServiceInMemImpl) unsubscribe(sub *SubscriptionInMemImpl) {
...
-	c       chan *goaingest.IngestEvent // channel of events
-	once    sync.Once                   // ensures c only closed once
-	id      uuid.UUID                   // subscription identifier
+// subscriptionInMemImpl represents a stream of events.
+type subscriptionInMemImpl[T any] struct {
+	// Service the subscription was created from.
+	service *serviceInMemImpl[T]
+	// Channel of events.
+	c chan T
+	// Ensures c only closed once.
```
djjuhasz: Grammar: "Ensures c *is* only closed once."

djjuhasz (commented):
Do we really want to call `s.unsubscribe(sub)` any time that receiving on `sub.c` blocks?  If the `sub.c` buffer fills up (maybe from a lot of publish events in a short time) then the subscriber will immediately be dropped, won't it? :confused:

djjuhasz:
Sorry, I didn't mean to submit that code review. :(

internal/event/redis.go
```diff
@@ -39,93 +41,91 @@ func NewEventServiceRedis(logger logr.Logger, tp trace.TracerProvider, cfg *Conf
...
-		c:      make(chan *goaingest.IngestEvent, EventBufferSize),
-		stopCh: make(chan struct{}),
+	sub := subscriptionRedisImpl[T]{
+		logger:     logger,
+		pubsub:     pubsub,
+		c:          make(chan T, EventBufferSize),
+		stopCh:     make(chan struct{}),
+		serializer: serializer,
 	}
 	go sub.loop()
```
djjuhasz: Seems to me that we should pass `ctx` to `loop()`, and exit the loop if the context is cancelled. :thinking:

djjuhasz (approved):
This is a lot to take in @jraddaoui with the generics, and the whole pub/sub channel system.  I don't really understand how it all works together, but the generics you added look okay to me. :shrug:  

I raised some issues with the event pub/sub system, but that predates this PR so feel free to leave things as they are for now.

sevein (approved):
The new generic type-safe implementation of the event bus is so nice. For a future iteration, what do you think about making the event package fully domain-agnostic, similar to go-cloud's design? My thinking is that right now, our low-level event package imports types from high-level domains like ingest (my fault), which inverts the proper dependency flow and creates a conceptual circular dependency. If we move the domain-specific serializers into their respective packages, we can correct this flow and the event package would become a core reusable utility without knowledge of the domains it serves. No need to address it in this PR, just a thought for the next iteration.

internal/event/redis.go
sevein: Well spotted, I think this is a problem in my original implementation. The monitor handler is calling the Close method, but we could honour the ctx in the loop to ensure that the child goroutine is terminated if the monitor is stuck for some reason, which can be rare but it could happen. This can probably wait, but it'd be nice to do it at some point.

jraddaoui: Done!

jraddaoui:
Thanks @djjuhasz!

> Do we really want to call s.unsubscribe(sub) any time that receiving on sub.c blocks? If the sub.c buffer fills up (maybe from a lot of publish events in a short time) then the subscriber will immediately be dropped, won't it? 😕

I have changed that to skip the events and also increased the buffer size to 256. Added similar changes in the Redis implementation, where I made the stop channel buffered and checked it in the event send loop to ensure responsive shutdown without blocking.

---

Thanks @sevein!

> The new generic type-safe implementation of the event bus is so nice. For a future iteration, what do you think about making the event package fully domain-agnostic, similar to go-cloud's design? My thinking is that right now, our low-level event package imports types from high-level domains like ingest (my fault), which inverts the proper dependency flow and creates a conceptual circular dependency. If we move the domain-specific serializers into their respective packages, we can correct this flow and the event package would become a core reusable utility without knowledge of the domains it serves. No need to address it in this PR, just a thought for the next iteration.

I think that's great and I have it almost done, will have a PR for you on Monday. That also made me realize that it may be better to split the config for both domains too. Please, take a look to the changes I mention to David above (sorry I amended the commit to merge it), if you spot something I missed I can address it in that PR.

PR #1312, opened 2025-08-04
Make the event package domain-agnostic

jraddaoui:
Move serializers and publish functions to the respective ingest and storage packages, remove helpers and any other domain specific code from the event package and update tests. Update configuration for a better separation between domains.

Refs #1222.


internal/ingest/event.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil, err
+	}
+	if err := ingestclient.ValidateMonitorResponseBody(&payload); err != nil {
+		return nil, err
+	}
+	return ingestclient.NewMonitorIngestEventOK(&payload), nil
+}
+
+// PublishEvent publishes an ingest event with type safety.
+func PublishEvent(ctx context.Context, svc event.Service[*goaingest.IngestEvent], event any) {
```
sevein: I wouldn't say that it's type-safe because of the runtime type check that could lead to a panic. Here's an example using a compile-time contraint that you may want to consider: https://gist.github.com/sevein/0cd28e149ab9f4c1317531d9c3da0979.

I've never understood why Goa doesn't export an interface that can be used for cases like this. Maybe there's something that could be contributed to Goa in that regards? Perhaps at another time.

sevein (approved):
Thanks for doing this, it looks great! Just one minor suggestion.

internal/ingest/event.go
jraddaoui: Thanks @sevein, that looks parfait :ok_hand:!

Issue #1291, opened 2025-07-15
Enhancement: Support ability to select SIP from a staging location


sallain:
**Is your feature request related to a problem? Please describe.**

For many reasons, a user may need to upload a SIP from a remote server (or "staging") location, like an S3 bucket. In one instance, we know of users who must deposit their SIPs into a security scanning application before uploading to Enduro, and they cannot download the SIPs locally once the security scan is completed. In other circumstances, SIPs that are already in a staging location may be too large to download locally to use the existing uploader.

**Describe the solution you'd like**

Implement the ability to connect Enduro to a server that can act as a staging location. For the user, it should work like this:

* User (or system) uploads package to staging location (primary use case is S3 bucket, but MinIO bucket and Azure bucket should be supported as well)
* User logs into Enduro, navigates **Upload SIPs** and then **Select from staging** tab
* User searches staging location via Enduro UI and selects package
* User starts ingest

It might look like this (see notes below)

<img width="1548" height="1046" alt="Image" src="https://github.com/user-attachments/assets/5d8df7c9-0d01-498a-9576-877af19f95b0" />

For a first iteration, we could limit this feature to single-select (multi-select is shown in the mock-up above), and to only being able to connect to one staging location. It might also make sense to only target S3 buckets (this may include MinIO as well, since it uses S3) and leave Azure for later.

**Describe alternatives you've considered**

We looked at several options with the instigating client, but this solution seemed the most universal and generally desirable for the product. More info on our discovery process can be provided by request.

fiver-watson:
@djjuhasz in the docs i am drafting up now, the placeholder header I have left this I have tentatively labeled as a "source location" instead of a staging directory. Thoughts?

I'm thinking the tab could then be either "Select from source" or "Select from location" or just "watched location upload." Thoughts?

CC @jraddaoui @sallain ... [context: "staging" has a different meaning for sysadmins in terms of server deployment and DJ suggested we don't use that term. Trying to define the language for how we talk about this upload menthod and the locations you select from, and distinguish these from a watched location where import is automatic, like minIO used to be]

fiver-watson:
Radda has told me that David is already using SIP source in the code. @djjuhasz with that in mind, why don't we label that second tab as: "Select from source" ?

I talked to @jraddaoui and @sallain who are in agreement with this suggestion, FYI

djjuhasz:
@fiver-watson "Select from source" sounds good to me. :)

PR #1313, opened 2025-08-05
Add SIP source upload

jraddaoui:
Refs #1291.

---

**Start ingest workflow from SIP source** 

Add API endpoint to create a SIP from a SIP source and init the ingest
workflow. Add `SIPSourceID` field to the workflow request and register
two `bucketdownload` activities, one using the internal bucket and the
other using the SIP source bucket. Based on the `SIPSourceID` field run
one activity or the other.

---

**Allow SIP source upload from the dashboard**

Split the upload SIPs page into local and cloud upload using tabs. Move
the existing Uppy uploader to a component and create a new one for the
SIP source upload. In the new component, fetch a list of objects from a
hardcoded SIP source using scroll load and display a list-group with
check-boxes and multi-selection. Checking the objects and submitting
the upload will group and make all requests to the API to start an ingest
workflow for each SIP.

---

**Fix user creation on concurrent add SIP requests**

Add unique index over the user OIDC columns combined to prevent
the creation of multiple users with the same OIDC iss and sub claims.
To allow setting that index without prefix I reduced both column sizes
to 255 chars.

Additionally, when multiple requests are sent at the same time by a new
user, most of them would fail due to the new unique index. To allow that
scenario I decided to use Ent's upsert feature. This will create a new
user or update the email and name of an existing one if there is a
conflict in the OIDC columns.


djjuhasz (commented):
Wooow! It works (and looks) great so far @jraddaoui. Well done! :clap: 

I reviewed all of your commits so far and everything looked good to me, though I was mostly looking at the big picture and didn't focus on the little details.  I'll do a more thorough review when you've finished up the PR and the commits are squashed just so I'm not commenting on minor issues that may be fixed in subsequent commit.

jraddaoui:
Thanks @djjuhasz, a lot of that is your work and it was looking great already!

I squashed all that was done into a couple of commits and based this PR against the user creation fixes. I'll add a few more commits that we can fix-up later into the existing ones.

jraddaoui:
Hi @djjuhasz, this is ready for a full look.

The idea would be to merge as three commits as outlined in the PR description. The first two and the latest one, I'll squash all the others into the dashboard commit. I thought it would be nice to leave them like that in case you want to look at some of them individually, but I could easily squash them before review if you prefer.

The latest commit relates to the user + SIP creation and the transaction PR from the other day. Sorry about the back an forward on this, in the end the `ReadOIDCUser` and `CreateUser` functions are unused for now, but I thought they could become useful in the near future.

It's still lacking some test in both ends, but that will need to happen in another PR if we want to include this in the release. BTW, I just noticed we are not getting coverage reports from Codecov, I think for quite a while already.

dashboard/src/components/SIPUploadCloud.vue
```diff
@@ -0,0 +1,147 @@
+<script setup lang="ts">
```
djjuhasz: Maybe we should change the name of this component to `SIPSourceUpload.vue` to match the back-end naming and the label suggested by @fiver-watson on https://github.com/artefactual-sdps/enduro/issues/1291#issuecomment-3133525369?

If we do make  that change then I think we should change the local upload to `SIPLocalUpload` or `SIPFilesystemUpload` or something similar.

dashboard/src/components/SIPUploadCloud.vue
```diff
@@ -0,0 +1,147 @@
...
+    } catch (error) {
+      console.error("Failed to load SIPs:", error);
+      errorMessage.value = "Failed to load SIPs.";
+      items.value = [];
+    }
+  },
+  null,
+);
+
+useInfiniteScroll(
```
djjuhasz: Infinite scroll, very cool! 😍

dashboard/src/components/SIPUploadCloud.vue
```diff
@@ -0,0 +1,153 @@
...
+      @click="errorMessage = null"
+      aria-label="Close"
+    ></button>
+  </div>
+
+  <h2 class="mb-3">1. Select SIPs to Ingest</h2>
+  <div class="mb-3">
+    <div class="form-text d-flex gap-2 justify-content-end">
+      <span>Selected SIPs: {{ checkedItems.length }}</span>
+      <span>•</span>
```
djjuhasz: It's not a blocker for this PR but I think for accessibility it would be better to use a `<menu>` tag and make the "Select all" and "Clear selections" links menu items and use CSS to add the separators between the items.

dashboard/src/pages/ingest/upload/index.vue
```diff
@@ -1,87 +1,64 @@
...
+    show: true,
   },
-  getResponseData: () => {
-    return { url: "" };
+  {
+    icon: IconCloudUpload,
+    text: "Select from source",
+    route: router.resolve({
+      name: "/ingest/upload/",
+      query: { ...route.query, source: "cloud" },
```
djjuhasz: Maybe "cloud" should be "source" or "sip-source"?

internal/persistence/ent/client/user.go
```diff
@@ -97,25 +98,55 @@ func readOIDCUser(ctx context.Context, entc *db.Client, iss, sub string) (int, *
...
+		return 0, newRequiredFieldError("OIDCSub")
+	}
+
+	// If CreatedAt is not set, use the current time.
+	if u.CreatedAt.IsZero() {
+		u.CreatedAt = time.Now()
+	}
+
+	// Create or update the user. There is no save method for
+	// upsert queries so we cant get the DB user directly.
```
djjuhasz: Typo: "can't"

internal/persistence/ent/schema/user.go
```diff
@@ -44,12 +44,12 @@ func (User) Fields() []ent.Field {
 			Optional(),
 		field.String("oidc_iss").
 			Annotations(entsql.Annotation{
-				Size: 1024,
+				Size: 255,
```
djjuhasz, line 47: I'm a bit worried there could be edge cases where the issuer URL is longer than 255 chars. 🤔

djjuhasz (approved):
This looks great to me @jraddaoui.  I suggested a few minor changes but none of them are blockers for this PR.

dashboard/src/pages/ingest/upload/index.vue
jraddaoui: Yeah, using "source", I also changed the param name to "from" to avoid confusion.

internal/persistence/ent/schema/user.go
jraddaoui: Hopefully 255 is fine, you never know, but here an example provider URL using 255 chars generated with AI:

```
https://login.example-identity-provider.com/tenant/region/eu-west/application/authentication/v2/organizations/org-42ff83b1-91b3-4a9e-81ab-9d15fcb67a61/projects/project-4be21d8d-59a0-4d5a-8df6-3d0c9b4ac2fd/environments/production
```

We already applied that migration in the dev environment, lets leave it like that for now.

dashboard/src/components/SIPUploadCloud.vue
jraddaoui: I used SIPUploadSource and SIPUploadLocal.

jraddaoui:
Thanks @djjuhasz! I addressed most of your feedback and, one extra thing I noticed, the route protection and all was not considering the new attributes that could allow visiting the /sips/upload page. We needed to allow OR conditions now in the route conditions so I modified the route permissions check to allow that. Also updated the sidebar check and the tabs/component checks option. Another minor detail, `onMounted` was not enough if somehow you click the upload link again while being already in that page, it removed everything from the page, we are using `watch` now ;)

PR #1315, opened 2025-08-06
Fix uploader creation on SIP upload

jraddaoui:
Use a single transaction to find or create the user and create the SIP. Reuse the User datatype for the SIP Uploader to be able to pass the OIDC fields on SIP creation. Also, check user claims before streaming the upload and rename SIP edge in the DB schema from to uploader.

Refs #1291.

### DIFF
--- a/internal/datatypes/sip.go
+++ b/internal/datatypes/sip.go
@@ -12,12 +12,6 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/enums"
 )
 
-type Uploader struct {
-	UUID  uuid.UUID
-	Email string
-	Name  string
-}
-
 // SIP represents a SIP.
 type SIP struct {
 	ID     int
@@ -42,7 +36,7 @@ type SIP struct {
 	FailedKey string
 
 	// Uploader is the user that uploaded the SIP.
-	Uploader *Uploader
+	Uploader *User
 }
 
 // Goa returns the API representation of the SIP.
--- a/internal/persistence/ent/client/sip.go
+++ b/internal/persistence/ent/client/sip.go
@@ -28,6 +28,17 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
 	if s.Name == "" {
 		return newRequiredFieldError("Name")
 	}
+	if s.Uploader != nil {
+		if s.Uploader.UUID == uuid.Nil {
+			return newRequiredFieldError("Uploader.UUID")
+		}
+		if s.Uploader.OIDCIss == "" {
+			return newRequiredFieldError("Uploader.OIDCIss")
+		}
+		if s.Uploader.OIDCSub == "" {
+			return newRequiredFieldError("Uploader.OIDCSub")
+		}
+	}
 
 	tx, err := c.ent.BeginTx(ctx, nil)
 	if err != nil {
@@ -51,20 +62,14 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
 		q.SetCompletedAt(s.CompletedAt.Time)
 	}
 
-	// If Uploader.UUID is set, find the user and link it to the SIP.
-	var uploader *datatypes.Uploader
-	if s.Uploader != nil && s.Uploader.UUID != uuid.Nil {
-		u, err := tx.User.Query().Where(user.UUID(s.Uploader.UUID)).Only(ctx)
+	// If Uploader is set, find or create the user and link it to the SIP.
+	if s.Uploader != nil {
+		uID, err := findOrCreateUser(ctx, tx.Client(), s.Uploader)
 		if err != nil {
-			return rollback(tx, newDBErrorWithDetails(err, "create SIP"))
+			return rollback(tx, err)
 		}
-		q.SetUser(u)
 
-		uploader = &datatypes.Uploader{
-			UUID:  u.UUID,
-			Email: u.Email,
-			Name:  u.Name,
-		}
+		q.SetUploaderID(uID)
 	}
 
 	// Save the SIP.
@@ -76,14 +81,11 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
 		return rollback(tx, newDBError(err))
 	}
 
-	// Update SIP with DB data, to get generated values (e.g. ID).
+	// Update SIP with DB data, to get generated values (e.g. ID). Manually set
+	// the uploader data because the dbs result doesn't include the user edge.
+	uploader := s.Uploader
 	*s = *convertSIP(dbs)
-
-	// Manually set the uploader data because the dbs result doesn't include the
-	// user edge.
-	if uploader != nil {
-		s.Uploader = uploader
-	}
+	s.Uploader = uploader
 
 	return nil
 }
@@ -106,11 +108,7 @@ func (c *client) UpdateSIP(
 	// Get the current SIP data from the database.
 	dbs, err := tx.SIP.Query().
 		Where(sip.UUID(id)).
-		WithUser(func(q *db.UserQuery) {
-			q.Select(user.FieldUUID)
-			q.Select(user.FieldEmail)
-			q.Select(user.FieldName)
-		}).
+		WithUploader().
 		Only(ctx)
 	if err != nil {
 		return nil, rollback(tx, newDBError(err))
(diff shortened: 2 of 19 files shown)


internal/persistence/ent/client/sip.go
```diff
@@ -51,20 +52,33 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
 		q.SetCompletedAt(s.CompletedAt.Time)
 	}
 
-	// If Uploader.UUID is set, find the user and link it to the SIP.
-	var uploader *datatypes.Uploader
-	if s.Uploader != nil && s.Uploader.UUID != uuid.Nil {
-		u, err := tx.User.Query().Where(user.UUID(s.Uploader.UUID)).Only(ctx)
+	// If Uploader is set, find or create the user and link it to the SIP.
+	if s.Uploader != nil {
+		// Try to find the user by OIDC Iss and Sub.
```
djjuhasz: This is a fairly long conditional block with nested conditionals — I think the code would be easier to read if the logic to get or create the uploader was moved to its own function.

internal/persistence/ent/client/sip.go
```diff
@@ -51,20 +52,33 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
...
+		u, err := readOIDCUser(ctx, tx.Client(), s.Uploader.OIDCIss, s.Uploader.OIDCSub)
 		if err != nil {
-			return rollback(tx, newDBErrorWithDetails(err, "create SIP"))
+			if !errors.Is(err, persistence.ErrNotFound) {
+				return rollback(tx, err)
+			}
+
+			// User does not exist, create a new one. Allow the UUID to be
```
djjuhasz: Nitpick: "The user does not exist, so create a new one" or similar to be a proper sentence.

internal/persistence/ent/client/sip.go
```diff
@@ -51,20 +52,33 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
...
+			// set by the caller, or generate a new one if not provided.
+			if s.Uploader.UUID == uuid.Nil {
+				s.Uploader.UUID = uuid.New()
```
djjuhasz: I'm not a fan of generating a UUID here — it's non-deterministic and not tested. I'd prefer to generate the UUID in the `goaWrapper` caller where we have an injected RNG that can be deterministic for tests.

internal/persistence/ent/client/sip.go
```diff
@@ -51,20 +52,33 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
...
+			s.Uploader = u
 		}
-		q.SetUser(u)
 
-		uploader = &datatypes.Uploader{
-			UUID:  u.UUID,
-			Email: u.Email,
-			Name:  u.Name,
+		// Get and set the uploader ID on the SIP.
+		uID, err := tx.User.Query().Where(user.UUID(s.Uploader.UUID)).OnlyID(ctx)
```
djjuhasz: I think it would be better to return the user ID (PK) from `readOIDCUser()` then to do a second database query for the same user.  Alternately you could use a direct ent query to check if the user exists in the first place.

internal/persistence/ent/client/sip.go
```diff
@@ -76,14 +90,11 @@ func (c *client) CreateSIP(ctx context.Context, s *datatypes.SIP) error {
 		return rollback(tx, newDBError(err))
 	}
 
-	// Update SIP with DB data, to get generated values (e.g. ID).
+	// Update SIP with DB data, to get generated values (e.g. ID). Manually set
+	// the uploader data because the dbs result doesn't include the user edge.
+	uploader := s.Uploader
```
djjuhasz, line 86: `s.Uploader` is not getting updated from the database when you create a new user, so I think `uploader` will be missing the `CreatedAt` value, and could have the wrong `Email` and `Name` if the `s.Uploader` values are different than the database values.

internal/persistence/ent/client/user.go
```diff
@@ -15,12 +15,16 @@ import (
 
 // CreateUser creates and persists a new user.
 func (c *client) CreateUser(ctx context.Context, u *datatypes.User) error {
+	return createUser(ctx, c.ent, u)
+}
+
+func createUser(ctx context.Context, entc *db.Client, u *datatypes.User) error {
```
djjuhasz: It wasn't obvious to me that the separate `createUser` and `readUser` functions were to allow passing in a database transaction client — I think it would be helpful to add a comment to explain this.

internal/persistence/ent/schema/sip.go
```diff
@@ -66,7 +66,7 @@ func (SIP) Edges() []ent.Edge {
 	return []ent.Edge{
 		edge.To("workflows", Workflow.Type).
 			Annotations(entsql.OnDelete(entsql.Cascade)),
-		edge.From("user", User.Type).
+		edge.From("uploader", User.Type).
```
djjuhasz, line 69: This is a nice little change for readability in the code. :+1:

internal/datatypes/sip.go
```diff
@@ -42,7 +36,7 @@ type SIP struct {
 	FailedKey string
 
 	// Uploader is the user that uploaded the SIP.
-	Uploader *Uploader
+	Uploader *User
```
djjuhasz, line 39: Dropping the `Uploader` type works better than I expected. :+1:

djjuhasz (changes requested):
Looking good @jraddaoui, I just noticed a few potential problems when linking an uploader to a SIP.

internal/persistence/ent/client/sip.go
jraddaoui: I think it's updated. If we get the user from `readOIDCUser` it's transformed [here](https://github.com/artefactual-sdps/enduro/blob/dev/issue-1291-create-user-tx/internal/persistence/ent/client/user.go#L90) and assigned to the `s.Uploader` [here](https://github.com/artefactual-sdps/enduro/blob/dev/issue-1291-create-user-tx/internal/persistence/ent/client/sip.go#L73), if we create a new user it's transformed [here](https://github.com/artefactual-sdps/enduro/blob/dev/issue-1291-create-user-tx/internal/persistence/ent/client/user.go#L46) where the passed `s.Uploader` is also being updated.

djjuhasz: Oh, you're right.  I forgot that `createUser()` mutates the passed user. :+1:

internal/persistence/ent/client/sip.go
jraddaoui: We recently talk about using UUIDs in the internal service interfaces too, and the `User` struct from the datatypes package doesn't even include the DB ID, something that I was planning to do in the other types when the service interfaces is fully changed. And we don't get the DB ID from the user creation either.

I was trying to use the existing functions/process as much as possible but, now that you say this, I could return the DB ID from the private functions as another var and just ignore it from the public functions. I will do that ;)

internal/persistence/ent/client/sip.go
jraddaoui: Okay, I made it a required field if the uploader is set. And I did the same for the OIDC fields, as they are required in readOIDCUser.

internal/persistence/ent/client/sip.go
jraddaoui: Done!

jraddaoui:

internal/persistence/ent/client/user.go
```diff
@@ -15,13 +16,16 @@ import (
...
 // CreateUser creates and persists a new user.
 func (c *client) CreateUser(ctx context.Context, u *datatypes.User) error {
-	return createUser(ctx, c.ent, u)
+	_, err := createUser(ctx, c.ent, u)
+	return err
 }
 
-func createUser(ctx context.Context, entc *db.Client, u *datatypes.User) error {
+// createUser creates and persists a new user.
+// It allows to pass an Ent Client that binds to a running transaction.
```
djjuhasz: Nitpick: "It allows passing an Ent Client..."

internal/persistence/ent/client/user.go
```diff
@@ -66,16 +70,19 @@ func (c *client) ReadUser(ctx context.Context, id uuid.UUID) (*datatypes.User, e
...
 // ReadOIDCUser retrieves a user by OIDC issuer and subject.
 func (c *client) ReadOIDCUser(ctx context.Context, iss, sub string) (*datatypes.User, error) {
-	return readOIDCUser(ctx, c.ent, iss, sub)
+	_, u, err := readOIDCUser(ctx, c.ent, iss, sub)
+	return u, err
 }
 
-func readOIDCUser(ctx context.Context, entc *db.Client, iss, sub string) (*datatypes.User, error) {
+// readOIDCUser retrieves a user by OIDC issuer and subject.
+// It allows to pass an Ent Client that binds to a running transaction.
```
djjuhasz: Here too.

djjuhasz (approved):
LGTM! Just one minor grammar issue x 2. :)

Issue #1282, opened 2025-07-08
Enhancement: Ensure that Enduro emits consistent audit events for key actions


fiver-watson:
**Is your feature request related to a problem? Please describe.**

For both security and chain of custody information over time, many institutions have firm requirements around the ability to audit key actions performed in the preservation environment. These typically include, at minimum: 

* SIP deposits to begin ingest or other workflows
* AIP deletion requests and approvals
* AIP move operations
* AIP downloads

**Describe the solution you'd like**

The application should capture key information whenever the above events occur, including: 

* The event name, using a consistent controlled vocabulary 
* The agent or user who initiated the action
* Timestamp of when the action occurred
* Related package UUID

Possibly also: related workflow name and instance ID?

The application should use a consistent format for capture, conforming to existing de facto standards, specifications, and/or best practices for audit logging whenever relevant. The goal is to ensure that third-party logging applications can easily ingest and manage the audit log events without excessive processing or transformation. 

**Additional context**

In Enduro-STE, I recall us making an internal distinction between a user action and a system action - something similar may be useful here?

First draft brainstorm for a proposed event vocabulary: 

```
* SIP.upload
* AIP.deletion.request
* AIP.deletion.approval
* AIP.deletion.rejection
* AIP.move
* AIP.download
```

sallain:
This is an additional thought that comes from an RFP I'm working on - in that RFP, the client specifies that the audit logs should be immutable (the phrase they use is "tamper-proof"). I'm not sure if that's standard for audit logs so just thought it was best to note it here!

djjuhasz:
I've done some research into audit logs and found the following recommendations are considered best practice according to the sources cited below.

Almost all of the articles I read assume that the audit logs are written to a file on disk.  Writing audit logs to disk also fits well with the "Data Centric" principle of Artefactual's Product Conceptual Model. There are a few examples of audit logs being written to a database, but this only seems to be common for DBMS that are logging audit data for database events.  The articles also assume that the raw audit logs will be aggregated and analyzed using a log management application — which isn't surprising as many of the articles are written by companies that sell a log management product.  Many of the articles also warn that a log rotation and retention policy must be implemented to limit the disk space required by the logs — the audit logs can grow very quickly for heavily used applications.

[...]

PR #1320, opened 2025-08-19
Add an ingest event audit logger

djjuhasz:
Refs #1282.

- Add an ingest audit logger that subscribes to the ingest event service and writes "SIP deposited" events to an audit log file
- Configure the development environment to do audit logging

### DIFF
--- a/cmd/enduro/main.go
+++ b/cmd/enduro/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/about"
 	"github.com/artefactual-sdps/enduro/internal/api"
 	"github.com/artefactual-sdps/enduro/internal/api/auth"
+	"github.com/artefactual-sdps/enduro/internal/auditlog"
 	"github.com/artefactual-sdps/enduro/internal/config"
 	"github.com/artefactual-sdps/enduro/internal/db"
 	"github.com/artefactual-sdps/enduro/internal/event"
@@ -162,6 +163,20 @@ func main() {
 	if err != nil {
 		logger.Error(err, "Error creating Ingest Event service.")
 		os.Exit(1)
+	}
+
+	// Set up the ingest audit logger, if one is configured.
+	if cfg.AuditLog.Filepath != "" {
+		al := auditlog.NewFromConfig(cfg.AuditLog, ingest.HandleAuditEvent)
+		err := al.Listen(ctx, ingestEventSvc)
+		if err != nil {
+			logger.Error(err, "Error starting ingest audit logger.")
+			os.Exit(1)
+		}
+		defer al.Close()
+		logger.V(1).Info("Starting audit logger", "path", cfg.AuditLog.Filepath)
+	} else {
+		logger.V(2).Info("Audit logging is disabled.")
 	}
 
 	// Set up the storage event service.
--- a/enduro.toml
+++ b/enduro.toml
@@ -80,6 +80,23 @@ migrate = true
 redisAddress = "redis://redis.enduro-sdps:6379"
 redisChannel = "enduro-ingest-events"
 
+[auditlog]
+# filePath is the absolute path to the ingest audit log. The enduro user will
+# need write access to the given directory. Rotated logs will use the same
+# directory and filename, but with a timestamp appended. If filePath is not set,
+# or set to an empty string, audit logging will be disabled.
+filePath = "/home/enduro/logs/enduro_ingest_audit.log"
+# maxSize is the maximum size, in megabytes, of the audit log file. When the
+# log file reaches maxSize or greater it will be rotated (default: 100 MB).
+maxSize = 100
+# compress determines if the rotated log files are compressed using gzip
+# (default: false).
+compress = false
+# verbosity determines the minimum level threshold for a log message to be
+# recorded. Currently supported levels are: -4 = DEBUG, 0 = INFO (default),
+# 4 = WARN, 8 = ERROR.
+verbosity = 0
+
 [extractActivity]
 dirMode = "0o700"
 fileMode = "0o600"
@@ -143,26 +160,26 @@ shareDir = "/home/a3m/.local/share/a3m/share"
 capacity = 1
 
 [a3m.processing]
-AssignUuidsToDirectories                     = true
-ExamineContents                              = true
-GenerateTransferStructureReport              = true
-DocumentEmptyDirectories                     = true
-ExtractPackages                              = true
-DeletePackagesAfterExtraction                = true
-IdentifyTransfer                             = true
-IdentifySubmissionAndMetadata                = true
-IdentifyBeforeNormalization                  = true
-Normalize                                    = true
-TranscribeFiles                              = true
-PerformPolicyChecksOnOriginals               = true
+AssignUuidsToDirectories = true
+ExamineContents = true
+GenerateTransferStructureReport = true
+DocumentEmptyDirectories = true
+ExtractPackages = true
+DeletePackagesAfterExtraction = true
+IdentifyTransfer = true
+IdentifySubmissionAndMetadata = true
+IdentifyBeforeNormalization = true
+Normalize = true
+TranscribeFiles = true
+PerformPolicyChecksOnOriginals = true
 PerformPolicyChecksOnPreservationDerivatives = true
-AipCompressionLevel                          = 1
-AipCompressionAlgorithm                      = 6
+AipCompressionLevel = 1
+AipCompressionAlgorithm = 6
 
 [am]
 address = ""
-user = "" # Secret: set with env var ENDURO_AM_USER.
-apiKey = "" # Secret: set with env var ENDURO_AM_APIKEY.
+user = ""                      # Secret: set with env var ENDURO_AM_USER.
+apiKey = ""                    # Secret: set with env var ENDURO_AM_APIKEY.
 processingConfig = "automated"
 
 # capacity limits the number of transfers a worker can process at one time
--- a/internal/auditlog/config.go
+++ b/internal/auditlog/config.go
@@ -1,0 +1,57 @@
+package auditlog
+
+import (
+	"io"
+	"log/slog"
+
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+type Config struct {
+	// Filepath specifies the location of the audit log file. If Filepath is
+	// not set, audit logging will be disabled.
+	Filepath string
+
+	// MaxSize sets the maximum size of the audit log file in megabytes before
+	// it is rotated (default: 100 MB).
+	MaxSize int
+
+	// Compress determines if the rotated log files are compressed using gzip
+	// (default: false).
+	Compress bool
+
+	// Verbosity sets the minimum log level that will be logged. Valid levels
+	// are: -4 = DEBUG, 0 = INFO (default), 4 = WARN, 8 = ERROR.
+	Verbosity int
+}
+
+// NewFromConfig creates a new Auditlog instance from the given config, using
+// lumberjack for log rotation.  If the Filepath in the config is not set,
+// a nil Auditlog will be returned.
+func NewFromConfig[T any](cfg Config, h EventHandler[T]) *Auditlog[T] {
+	if cfg.Filepath == "" {
+		return nil
+	}
+
+	logger, w := loggerFromConfig(cfg)
+	return &Auditlog[T]{
+		logger:  logger,
+		w:       w,
+		handler: h,
+		stopCh:  make(chan struct{}),
+	}
+}
+
+func loggerFromConfig(cfg Config) (*slog.Logger, io.WriteCloser) {
+	w := &lumberjack.Logger{
+		Filename: cfg.Filepath,
+		MaxSize:  cfg.MaxSize,
+		Compress: cfg.Compress,
+	}
+
+	logger := slog.New(slog.NewJSONHandler(w, &slog.HandlerOptions{
+		Level: slog.Level(cfg.Verbosity),
+	}))
+
+	return logger, w
+}
--- a/internal/auditlog/event.go
+++ b/internal/auditlog/event.go
@@ -1,0 +1,19 @@
+package auditlog
+
+import "log/slog"
+
+type Event struct {
+	Level    slog.Level
+	Msg      string
+	Type     string
+	ObjectID string
+	UserID   string
+}
+
+func (e Event) Args() []any {
+	return []any{
+		"type", e.Type,
+		"objectID", e.ObjectID,
+		"userID", e.UserID,
+	}
+}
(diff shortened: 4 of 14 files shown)


djjuhasz:
Here's what the audit log output looks like right now:
```
{"time":"2025-08-18T20:37:20.622422365Z","level":"INFO","msg":"SIP deposited","type":"SIP.deposit","objectID":"bc70d029-8fa6-4455-9523-739f93fbe8a1","userID":"44675334-8d13-474f-8954-4db06a393d65"}
{"time":"2025-08-18T21:51:28.115995153Z","level":"INFO","msg":"SIP deposited","type":"SIP.deposit","objectID":"ed225ceb-e7a5-47ce-a2e9-871b32ac2879","userID":"44675334-8d13-474f-8954-4db06a393d65"}
{"time":"2025-08-18T21:51:28.150393933Z","level":"INFO","msg":"SIP deposited","type":"SIP.deposit","objectID":"1778a8b2-beb0-4cdd-8b7c-3aee703776b9","userID":"44675334-8d13-474f-8954-4db06a393d65"}
```

djjuhasz:
@sevein if you have time to review this PR sometime this week, I'd appreciate it.  If not then @jraddaoui can review it next week when he's back from vacation.

sevein (commented):
In regards to the changes to the Dockerfile, it looks like a significant amount of effort has gone into creating a secure prod-ready image based on distroless, and this change feels like undoing that work for the sake of developer ergonomics.

Perhaps we could explore other alternatives? We could either maintain a separate, development-focused Dockerfile build stage that produces an image with a shell (feels undesirable), or perhaps we could address the `mkdir` and shell access concerns in a way that doesn't require changing the base image. For instance, the log directory can be created in a previous multi-stage build and then copied into the final distroless image. For shell access, we can adopt the use of ephemeral containers for debugging purposes - I remember we considered this option before and it had some drawbacks but seemed good enough to preserve the benefits of the distroless image while providing a good dev experience?

sevein (commented):
Here is some feedback on the new audit log implementation:

* The `ingest.Service` interface is being polluted by the `WithAuditLogger` method because is not a core responsibility of the ingest service but a setup concern. I would encourage instead to continue using the pattern of passing dependencies to the `ingest.NewService` constructor. You could adopt the "functional options" pattern (e.g. pass WithAuditLogger to the constructor) but it doesn't look like a good use case to me.
* The is no feedback if the audit logger fails to subscribe to the event bus on startup. We could alter the signature of `ingest.NewService` constructor to return such error, allowing the app to fail if the audit system can't be initialized.
* I'm wondering if the audit log's subscription logic should be moved instead to the dedicated `auditlog` package, i.e. a package where we address all aspects of audit logging including subscription details? Not entirely sure.
* It feels like the audit logging system should have a Shutdown method because you'd want to have a mechanism to wait for the logging goroutine to complete successfully during shutdown. Perhaps something that can be implemented with a `sync.WaitGroup` ensuring that the main application does not exit until all the pending audit events have been processed.

Overall I think that the auditlog pkg should contain the audit logging service, not just its config. The subscription and graceful shutdown mechanisms seem audit-related operational concerns as opposed to core ingest business logic.

djjuhasz:
@sevein we previously discussed using ephemeral containers and switching from a distroless container in this PR: https://github.com/artefactual-sdps/enduro/pull/1005.

I think at this point we should consider using a distroless distribution for our Docker build a failed experiment and discontinue its use because its disadvantages outweigh its advantages.  We switched to the current distroless Linux distrubtion from an Alpine Linux distribution on a suggestion from you after we introduced the https://github.com/artefactual-labs/bagit-gython dependency, which won't run in musl based Linux distribution.  I chose to use distroless without any knowledge of the drawbacks for development, and without any thorough analysis of the tradeoffs.  

Now that I've tried developing with a distroless container I've found it a significant obstacle to quick and iterative development due to its lack of standard binaries we depend on like libxml and ssh/sftp, and its lack of a shell for inspecting the running container and allowing file access.  Furthermore we don't use this Docker container for any production deployments of Enduro so the security benefits of using a distroless container are negligible in my opinion.  If we do want to create a secure, production-ready, Enduro image I think we should have a dedicated Dockerfile or build system for doing so, and take a more holistic approach to security for the entire Enduro system (e.g. I've used shell access to an Archivematica installation to get access to a PIP created by Enduro).

In the current case the distroless container caused several issues that made it harder to develop this audit logging feature:
1. I could not easily create a new directory for the logs and give the Enduro user write access to the directory to create a log file, due to the lack of standard Linux commands like `mkdir` and `chmod` in the image.
2. I could not easily access the log file written to the container to check that the logger was functioning properly

In the first case, it looks like we could create an empty `/logs` directory and copy it to the distroless image, which is only slightly more cumbersome than just creating the directory directly but it wasn't immediately obvious to me as a solution.  In the second case, we have had no success in our several attempts to access files inside a distroless container using ephemeral containers or other methods.  Ephemeral containers are not easy to understand or use, and thus present an significant barrier to development even if we can get them to work for this purpose.

For all these reasons, I think we should switch back to a standard Linux distribution with shell access for our Enduro Docker file.

djjuhasz:
@sevein I've tried several times to create a pure "auditlog" package, but the problem with creating an event listener is the events emitted from an event subscription are typed, and the events we publish have no common interface that would allow me to go from at typed event like `*goaingest.SIPCreatedEvent` to a `[]string` or struct that I could pass to `slog.Info(msg, args...)`.  I'm trying to wrap my head around how to create an adapter to get the typed subscription like `Subscription[T *goaingest.IngestEvent]` to emit a general `auditlog.Event` event, but I haven't been able to come up with a solution yet. :(

sevein:
> For all these reasons, I think we should switch back to a standard Linux distribution with shell access for our Enduro Docker file.

Makes sense, thank you!

djjuhasz:
@sevein and @jraddaoui I've completely restructured the audit logger based on @sevein's feedback to decouple it from the ingest service altogether. I've used a event handler function to allow general conversion of an event service event to an `auditlog.Event` that can then be logged by slog.  

@sevein I don't know if you want to continue code review on this, or if you want to pass it over to @jraddaoui as he's back on Monday.  If you do continue, then please note that I overwrote the previous commit history because this was such a major change — sorry if that makes it harder to review again.

djjuhasz:
Oh, I've also incorporated log rotation into this PR because it's a relatively small amount of code, and to avoid having to do a big gnarly rebase of https://github.com/artefactual-sdps/enduro/pull/1324.

cmd/enduro/main.go
```diff
@@ -164,6 +165,20 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Set up the ingest audit logger, if one is configured.
+	if cfg.AuditLog.Filepath != "" {
```
jraddaoui, line 169: We are not really consistent with this, but in other configurations we have an "Enabled" boolean. Not a big deal but I kind of like that pattern.

cmd/enduro/main.go
```diff
@@ -164,6 +165,20 @@ func main() {
...
+		al := auditlog.NewFromConfig(cfg.AuditLog, ingest.HandleAuditEvent)
+		err := al.Listen(ctx, ingestEventSvc)
```
jraddaoui, line 171: I don't think we should tie the audit logs with the current event services used for the WebSocket updates. Those are meant to send updates to the dashboard so they may not include some details, or we may be missing some of the events needed. For example, the SIP creation event is published once from the processing workflow without user details, and for AIP download there is no event as that doesn't require updates in the dashboard.

I'd consider setting up a different event.Service for the new auditlog Event type and publishing different events based on the audit log needs.

enduro.toml
```diff
@@ -80,6 +80,23 @@ migrate = true
...
+# maxSize is the maximum size, in megabytes, of the audit log file. When the
+# log file reaches maxSize or greater it will be rotated (default: 100 MB).
+maxSize = 100
+# compress determines if the rotated log files are compressed using gzip
+# (default: false).
+compress = false
+# verbosity determines the minimum level threshold for a log message to be
+# recorded. Currently supported levels are: -4 = DEBUG, 0 = INFO (default),
+# 4 = WARN, 8 = ERROR.
+verbosity = 0
```
jraddaoui, line 98: Do we care about verbosity in an audit log? I don't think so, we want to log all audit events, I'd remove it from the config.

internal/auditlog/config.go
```diff
@@ -0,0 +1,57 @@
...
+
+func loggerFromConfig(cfg Config) (*slog.Logger, io.WriteCloser) {
+	w := &lumberjack.Logger{
+		Filename: cfg.Filepath,
+		MaxSize:  cfg.MaxSize,
+		Compress: cfg.Compress,
+	}
+
+	logger := slog.New(slog.NewJSONHandler(w, &slog.HandlerOptions{
+		Level: slog.Level(cfg.Verbosity),
```
jraddaoui, line 53: Whatever is the default level, use it in all events. And no need to have it in the config or the event type.

internal/auditlog/event.go
```diff
@@ -0,0 +1,19 @@
+package auditlog
+
+import "log/slog"
+
+type Event struct {
+	Level    slog.Level
```
jraddaoui, line 6: Again, I don't think we care about this here.

internal/auditlog/event.go
```diff
@@ -0,0 +1,19 @@
...
+	Msg      string
+	Type     string
+	ObjectID string
+	UserID   string
```
jraddaoui, line 10: I think the user id is not enough in this logs, at least I'd like to see the user email in the audit events. Otherwise it doesn't provide a lot of info as a log, still requires to go to the DB to get the actual user details.

internal/ingest/convert.go
```diff
@@ -177,3 +178,22 @@ func sipSourceObjectsToGoa(objects []*sipsource.Object) goaingest.SIPSourceObjec
...
+			Msg:      "SIP created",
+			Type:     "sip.created",
+			ObjectID: e.UUID.String(),
+			UserID:   userID,
+		}
+	default:
+		// Ignore unsupported event types.
+		return nil
+	}
+}
```
jraddaoui, line 199: Why do we have both `Type` and `Msg` fields? I think the more human readable version would be enough. To be clear, I'd use the `Msg` value as `Type` and remove the `Msg` field.

internal/auditlog/event.go
```diff
@@ -0,0 +1,19 @@
+package auditlog
+
+import "log/slog"
+
+type Event struct {
+	Level    slog.Level
+	Msg      string
+	Type     string
+	ObjectID string
```
jraddaoui, line 9: That looks so AtoM and OOP :smile:, maybe "ResourceID"? Not a big deal of course.

cmd/enduro/main.go
jraddaoui: Given this comment and the current implementation logging to a file per service, I'd also consider just passing the logger directly to the API service.

jraddaoui (commented):
Thanks @djjuhasz! See some comments below.

About the `defer` issue we were talking in the meeting and using the group actors. Wouldn't it be enough to use the main/run pattern we use in other projects? Calling `run` from `main` and if there is an error calling `os.Exit(1)`, `run` would execute the deferred functions before returning the error.

Also, what's the plan for storage? A different or the same file? We don't have internal user IDs there, and even if we add them, they will be different from ingest.

cmd/enduro/main.go
djjuhasz: I've been avoiding an "enabled" boolean on the principle of avoiding an unnecessary variables — the audit log won't work without a valid file path so toggling the feature based on the presence of a file path seems alternative.  I'm get that an explicit "enabled" boolean needs less explanation though, so I'm fine to add it if you want to go that way.

enduro.toml
djjuhasz: Why would you assume that everyone using Enduro would want to log all audit events?

internal/auditlog/config.go
djjuhasz: In my research on audit logs it seemed like common practice to include error levels and to log events like failed logins, which would be a WARN or ERROR level event. I know we are only using INFO right now, but I think we should build on the assumption that we will have different event levels in the future.

internal/auditlog/event.go
djjuhasz: Yeah, I agree we should log user emails, especially since we don't have a user management UI.

internal/ingest/convert.go
djjuhasz: @fiver-watson suggested a dot separated type vocabulary (e.g. `SIP.uploaded`) for the audit log events in the original description of #1282. I think a more discursive event message is better for human readers. I'm fine with eliminating the `Type`, but I'd be more reluctant to removing the human readable `Msg`.

internal/auditlog/config.go
jraddaoui: I'd think of that as the outcome of the action, which we are not recording yet. I think it's quite confusing this way, we'd need to represent somehow the levels of each event. What I read is that we should make a distinction between blocking events (those that MUST be logged) and non-blocking events (those not that relevant that may be skipped based on the system load). I think we shouldn't worry about that with the current implementation, and I think the same about the verbosity.

For now I think it adds some confusion and it's something that we don't need to make configurable, I may be proven wrong, but we could easily add it later if that's the case.

cmd/enduro/main.go
jraddaoui: Not a big deal, it's okay like it is.

enduro.toml
jraddaoui: Yeah, you are right, I'm not sure if verbosity is the way to filter them, but I can't think of a better option. For now I think it is unnecessary though.

internal/ingest/convert.go
jraddaoui: I see, and based on our discussion we could have different messages based on the outcome of the action, while the type would be the same, so I'm cool with having both. One minor detail, should it be "sip.creation" as the type then?

djjuhasz: Yes, changing the type to "SIP.creation" is a good idea. :+1:

fiver-watson: Also curious about "SIP creation" as an audit log event in general.... perhaps from the perspective of the Enduro DB we are "creating" the SIP when first registering it, but otherwise, I'm not sure how we end up with a SIP creation event in Enduro when the application doesn't support SIP creation, and producers generally do this before content ever reaches Enduro...?

If we are thinking of the staging locations (where ingest doesn't proceed immediately for example), I'd say "SIP deposited" would be a better event, and then perhaps an "Ingest started" or "SIP ingest started" etc event when things are launched? Created doesn't feel quite right... but I also might not be understanding the context here, so apologies in advance!

djjuhasz:
We implemented a simpler audit logger implementation in commit 6dba5179ed6f2b0c571b842d9b17092c43165b35 that doesn't require subscribing to the event service and emitting events to the Monitor API that are not required for the Dashboard.

PR #1324, opened 2025-08-20
Add audit log rotation (refs #1282)

djjuhasz:
Add an audit log rotator that will rotate the log file when it reaches the configured "maxSize". The rotator also supports compressing the rotated log files with gzip.


sevein (approved):
LGTM

djjuhasz:
I've added log rotation to https://github.com/artefactual-sdps/enduro/pull/1320.

PR #1325, opened 2025-08-21
Add an AIP deletion request creation event

djjuhasz:
Refs #1282.

- Add an AIP deletion request creation event type to the Goa API
- Publish an event to the storage event service when an AIP deletion request is created

### DIFF
--- a/docs/src/dev-manual/api/openapi3.json
+++ b/docs/src/dev-manual/api/openapi3.json
@@ -45,6 +45,34 @@
         ],
         "type": "object"
       },
+      "AIPDeletionRequestCreatedEvent": {
+        "example": {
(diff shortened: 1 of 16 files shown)


sevein (approved):
LGTM

sevein (approved):

jraddaoui:
Check my comments in https://github.com/artefactual-sdps/enduro/pull/1320#discussion_r2299407029, I'd consider removing this event based on the direction we take there, I think we should not mix audit with WebSocket events. This event will be emitted to the dashboard but ignored there.

jraddaoui:
@djjuhasz not sure if you were already planning to do so, but I think we should revert this changes.

PR #1328, opened 2025-08-25
Fix code coverage CI action

djjuhasz:
Update codecov action to change "file" argument to "files" — the "file" argument has been deprecated as of version 5 of https://github.com/codecov/codecov-action.


jraddaoui (approved):
:ok_hand: Thanks!!

PR #1329, opened 2025-08-25
Add user IDs to deletion request creation event

djjuhasz:
- Add requester and reviewer fields to the AIP deletion request creation event
- Regenerate Goa API files and dashboard API client


djjuhasz:
This is no longer required with the simpler audit logger we've implemented in commit 6dba5179ed6f2b0c571b842d9b17092c43165b35.

PR #1330, opened 2025-08-26
Add a deletion request updated event to API

djjuhasz:
Refs #1282.


jraddaoui:
Check my comments in https://github.com/artefactual-sdps/enduro/pull/1320#discussion_r2299407029, I'd consider not adding this event based on the direction we take there, I think we should not mix audit with WebSocket events. This event will be emitted to the dashboard but ignored there.

djjuhasz:
This is no longer required, as we've chosen to implement the audit logger without subscribing to the ingest event service.

Issue #1321, opened 2025-08-19
Problem: user filter looks like it's been cleared if you leave and return to page, but hasn't actually been cleared


sallain:
**Describe the bug**

If you apply a user filter, then navigate away and return to the SIP browse page, the filter will still be applied, but the name of the user is no longer shown on the filter, making it seem at first glance that the filter has been cleared.

<img width="1340" height="497" alt="Image" src="https://github.com/user-attachments/assets/2fe6efae-df2c-4818-894c-abdeba0d80f2" />

Clicking the "X" on the filter removes the X, but doesn't actually clear the filter, unless the user forces a browser refresh.

**To Reproduce**

1. Log in and navigate to the SIPs browse page
2. Use the new "Uploaded by" filter to filter the SIPs to those uploaded by "operator@example.com"
3. Navigate away, to the Locations page for example
4. Return to the SIPs browse page

**Expected behavior**

The filter should either:

- Be cleared when returning to the page, or
- Continue to show the user it is filtering results to (i.e. operator@example.com)

**Screenshots**

**Additional context**

Originally described on #1275 by @fiver-watson

sallain:
LGTM!

PR #1332, opened 2025-08-26
Clear uploader filter on load

jraddaoui:
Refs #1321.

### DIFF
--- a/dashboard/src/stores/sip.ts
+++ b/dashboard/src/stores/sip.ts
@@ -27,7 +27,7 @@ export const useSipStore = defineStore("sip", {
       status: "" as IngestListSipsStatusEnum | undefined,
       earliestCreatedTime: undefined as Date | undefined,
       latestCreatedTime: undefined as Date | undefined,
-      uploaderId: "" as string,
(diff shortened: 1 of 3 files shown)


dashboard/src/stores/sip.ts
```diff
@@ -27,7 +27,7 @@ export const useSipStore = defineStore("sip", {
       status: "" as IngestListSipsStatusEnum | undefined,
       earliestCreatedTime: undefined as Date | undefined,
       latestCreatedTime: undefined as Date | undefined,
-      uploaderId: "" as string,
+      uploaderId: "" as string | undefined,
```
djjuhasz: Might as well make the default value of `uploaderId` `undefined` I think.

djjuhasz (changes requested):
Sorry about that SFTP test, it's been a hassle.  Thanks for fixing it.

Just one comment on the uploader filter change.

dashboard/src/stores/sip.ts
jraddaoui: Done!

djjuhasz (approved):
:+1:

PR #1333, opened 2025-08-27
Add audit logging for SIP ingest events

djjuhasz:
- Pass a parameter struct to the ingest service constructor to avoid passing a long list of positional arguments
- Add an audit logging package
- Log system startup and shutdown and SIP ingest events to the audit log
- Switch to a Debian (slim) Docker image for the enduro workers to simplify debugging of the running containers

### DIFF
--- a/cmd/enduro/main.go
+++ b/cmd/enduro/main.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"net/http/pprof"
 	"os"
@@ -31,6 +32,7 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/about"
 	"github.com/artefactual-sdps/enduro/internal/api"
 	"github.com/artefactual-sdps/enduro/internal/api/auth"
+	"github.com/artefactual-sdps/enduro/internal/auditlog"
 	"github.com/artefactual-sdps/enduro/internal/config"
 	"github.com/artefactual-sdps/enduro/internal/db"
 	"github.com/artefactual-sdps/enduro/internal/event"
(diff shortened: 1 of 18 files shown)


internal/ingest/download_sip_test.go
```diff
@@ -153,20 +153,12 @@ func TestDownloadSipRequest(t *testing.T) {
...
-				0,
-				nil,
-				nil,
-			)
+			svc := ingest.NewService(ingest.ServiceParams{
+				Logger:             logr.Discard(),
+				PersistenceService: psvcMock,
+				TicketProvider:     ticketProvider,
+				InternalStorage:    bucket,
+			})
```
jraddaoui, line 161: :heart_eyes:

jraddaoui (approved):
Thanks @djjuhasz, looks great!

cmd/enduro/main.go
```diff
@@ -94,6 +96,29 @@ func main() {
...
+	// close. Deferred functions run in "first in, last out" order.
+	auditLogger := auditlog.NewFromConfig(cfg.Auditlog)
+	defer func() {
+		auditLogger.Log(ctx, &auditlog.Event{
+			Level: slog.LevelInfo,
+			Msg:   "Enduro stopping",
+			Type:  "system",
+		})
+		auditLogger.Close()
+	}()
```
jraddaoui: As you said, this may not get logged on `os.Exit()` calls. We may want to give it a try to the main/run suggestion from the other PR. It doesn't need to happen here, there are more defer calls that should be affected by that issue already.

djjuhasz: Yeah, I think down the line it's a good idea to either add the logger shutdown to the rungroup actor shutdown, or as you said to on our call the other day move the app runner to a `Run()` function that returns an error, while calling `os.Exit()` exclusively in main.

jraddaoui: Also, I get it's good to start the logger earlier so the defer is the last but, should we log those system events later, to avoid some noise in failed starts?

djjuhasz: Okay, I've moved the startup and shutdown audit log messages close to the the start of the workflow and activity worker.

PR #1334, opened 2025-08-28
Log deletion request audit events

djjuhasz:
Refs #1282.

- Log deletion request creation, cancelation, approval and rejection
- Add auditlog log level constants and use them in preference to `slog` log levels to decouple `auditlog.Log()` calls from the `slog` package
- Add the AIP UUID to the deletion request returned from the persistence service after an update, so it's available to the audit logger


internal/storage/persistence/ent/client/deletion_request.go
```diff
@@ -61,18 +61,23 @@ func (c *Client) UpdateDeletionRequest(
...
+	// Get the UUID of the related AIP.
+	aipUUID := dr.Edges.Aip.AipID
+
 	up, err := upd(convertDeletionRequest(dr))
 	if err != nil {
 		return nil, rollback(tx, fmt.Errorf("update deletion request: %v", err))
 	}
 
 	q := tx.DeletionRequest.UpdateOneID(id).
-		SetUUID(up.UUID).
```
jraddaoui, line 75: Good catch!

jraddaoui (approved):
LGTM, thanks @djjuhasz!

PR #1342, opened 2025-09-02
Add ability to configure an institution logo

### LINKED ISSUE #1309: Enhancement: add ability to configure institution logo

In a consortia model, it's nice for users to have a quick, visual way of identifying which site a user is on.

Add a small logo to the persistent top bar of the site, and make it configurable so that an admin can add a small logo. Like this!

Image

We thought about a few different locations but decided that the upper right made the most sense.

**Additional context**

We have an existing issue to move the login button to the top right as well, and with the (i) icon there it all might get a bit crowded. However, that's a low priority so we will proceed with this change and then redesign the area when the time comes.

djjuhasz:
Fixes #1309.

Add the ability to display an institution logo in the Dashboard header by setting the required values in a Tilt `.env` file. The logo configuration values are documented in
`docs/admin-manual/dashboard-config.md`.


djjuhasz:
The institution logo is hidden at small screen widths to avoid display problems with the header wrapping to two lines.

[Screencast from 2025-09-02 17-13-51.webm](https://github.com/user-attachments/assets/de7f58a3-8ee5-4dd9-ac2c-f01607118f64)

dashboard/.env
```diff
@@ -10,3 +10,6 @@ VITE_OIDC_ABAC_CLAIM_PATH_SEPARATOR=\$VITE_OIDC_ABAC_CLAIM_PATH_SEPARATOR
 VITE_OIDC_ABAC_CLAIM_VALUE_PREFIX=\$VITE_OIDC_ABAC_CLAIM_VALUE_PREFIX
 VITE_OIDC_ABAC_USE_ROLES=\$VITE_OIDC_ABAC_USE_ROLES
 VITE_OIDC_ABAC_ROLES_MAPPING=\$VITE_OIDC_ABAC_ROLES_MAPPING
+
+VITE_INSTITUTION_LOGO=\$VITE_INSTITUTION_LOGO
+VITE_INSTITUTION_NAME=\$VITE_INSTITUTION_NAME
```
jraddaoui, line 15: I think we need to add all env. vars. here so they are considered by the replacement script.

dashboard/.env.development
```diff
@@ -10,3 +10,7 @@ VITE_OIDC_ABAC_CLAIM_PATH_SEPARATOR=
 VITE_OIDC_ABAC_CLAIM_VALUE_PREFIX=
 VITE_OIDC_ABAC_USE_ROLES=false
 VITE_OIDC_ABAC_ROLES_MAPPING=
+
+VITE_INSTITUTION_LOGO=http://localhost:8080/artefactual-logo.png
```
jraddaoui, line 14: :thinking: This is going to require making the CSP meta configurable too.

dashboard/public/artefactual-logo.png
jraddaoui, line 1: Can we use a smaller logo that doesn't take that much space in the header, I think just the Artefactual "a" would be great.

docs/src/admin-manual/dashboard-config.md
```diff
@@ -0,0 +1,55 @@
...
+
+When running an Enduro Dashboard Docker container, such as when using the Tilt
+development environment, the Vite environment variables can be set via the
+`dashboard/.env` (production), `dashboard/.env.development` (development), or
+`dashboard/.env.test` (test) environment files.
+
+If the Enduro Dashboard is not being run via Docker, but the code is being built
+with [Vite] then the institution logo can be configured in the production
+environment file (`dashboard/.env`) before building the distribution files.
+
```
jraddaoui, line 16: So nice that you added this! It would be great to mention the replacement script we use to be able to set those env. vars. after build (needed when using RPM packages for example).

https://github.com/artefactual-sdps/enduro/blob/main/dashboard/hack/inject-vite-envs.sh

We already mention this script in the IAC page, where we show the other dashboard env. vars. for OIDC configuration. It would be nice to organize this a bit better, not sure exactly how and it doesn't need to happen here.

jraddaoui (changes requested):
Looking great @djjuhasz, nice work!

docs/src/admin-manual/dashboard-config.md
fiver-watson: Another idea would be to add this documentation to the general configuration docs I am already working on. As it is, this PR's changes to the README toc will conflict with my PR, and even if kept apart, both pages should be updated to point to each other as "see also" for other configuration file stuff.

Just a thought!

dashboard/.env.development
djjuhasz: Ah, right.

dashboard/public/artefactual-logo.png
djjuhasz: We could, but this logo is closer in size to the SFA logo and I think other institutional logos will tend to have a similar aspect ratio.

docs/src/admin-manual/dashboard-config.md
djjuhasz: @jraddaoui I think the variable injection script is only necessary for changing application variables after the Vite distribution files are built.  I think for most institutions changing the logo variables in the config files then building the distribution files will be fine because the logo values shouldn't change often. 

I can still add the note about the injection script if you feel it's necessary but it's fairly complex and I'd rather avoid that complexity if possible.

djjuhasz: @fiver-watson I'd rather keep the Dashboard config information separate because it's a different config file (well several config files) and a separate application. 

I agree that linking between the two config files is a good idea though! :)

dashboard/public/artefactual-logo.png
jraddaoui: A wide logo makes the header breadcrumb issue more noticeable:

image

It was an existing issue and in the end it looks smaller than I was expecting, so it's okay to leave it like that. We may want to fix the breadcrumb issue in another PR, at least adding ellipsis and avoiding wrapping.

docs/src/admin-manual/dashboard-config.md
jraddaoui: Yeah, I'd like to make its own section for the injection script, and linking it from both places. The releases we make include the built dashboard and also the RPM packages, so I think it's a good idea to mention all options.

dashboard/src/components/Header.vue
```diff
@@ -3,76 +3,92 @@ import { openDialog } from "vue3-promise-dialog";
...
-        aria-controls="menu-offcanvas"
-        aria-label="Open navigation"
-      >
-        <IconMenu
-          class="text-dark mx-1"
-          style="font-size: 1.5em"
-          aria-hidden="true"
-        />
-      </button>
+      <div class="container-fluid">
```
jraddaoui: The `container-fluid` adds some padding to the sides we don't want. We could use the specific flex classes we want, or just add `px-0` to remove it.

dashboard/src/components/InstitutionLogo.vue
```diff
@@ -0,0 +1,28 @@
...
+  name: string;
+  url: string;
+}>();
+
+const imgTag = `<img
+    id="institution-logo"
+    src="${props.logo}"
+    alt="${props.name}"
+    height="30"
+    class="d-none d-sm-block mx-3"
```
jraddaoui: This classes should be added to the link or div containers, otherwise we "show" the link without content. Also, why a constant with v-html instead of doing it in the template directly?

jraddaoui (changes requested):
Thanks @djjuhasz! Just a couple of extra details I noticed while giving it a try locally.

dashboard/src/components/InstitutionLogo.vue
djjuhasz: I used a variable and v-html to avoid having to repeat the HTML for the `` twice, like so:

```
<a
    v-if="props.logo && props.url"
    :href="props.url"
    target="_blank"
    rel="external"
  >
    <img
      id="institution-logo"
      src="${props.logo}"
      alt="${props.name}"
      height="30"
      class="d-none d-sm-block mx-3"
    />
  </a>
  <div v-else-if="props.logo">
    <img
      id="institution-logo"
      src="${props.logo}"
      alt="${props.name}"
      height="30"
      class="d-none d-sm-block mx-3"
    />
  </div>
```

I haven't figured out a really good way to  with Vue's templating system to conditionally add a link that doesn't require repeating any nested elements. :(

dashboard/src/components/Header.vue
djjuhasz: I added the "container-fluid" class based on the examples at https://getbootstrap.com/docs/5.3/components/navbar/.  I'm fine to remove it, but I wish our use of Bootstrap classes didn't feel so arbitrary, I really have a hard time figuring out what classes to use. :(

dashboard/src/components/InstitutionLogo.vue
jraddaoui: Right, good point!

dashboard/src/components/Header.vue
jraddaoui: In this case I just want to remove the extra padding, that's breaking the alignment of the Enduro heading and the sidebar.

djjuhasz: I've removed the div altogether in the latest commit.

jraddaoui (approved):
Thanks @djjuhasz! Just the docs comment left, which can happen in another PR if you prefer.

docs/src/admin-manual/dashboard-config.md
djjuhasz: Okay, let's document the injection script in another PR.

PR #1343, opened 2025-09-03
Add CSP meta to dashboard

jraddaoui:
Implemet CSP meta in the dashboard using `vite-plugin-csp-guard`. Allow
to disable it at build/serve time using an ENDURO_DISABLE_CSP_META env.
var. Make sure the environment files are read during VITE build/serve.
Add CSP header and comments in Dockerfile NGINX config.

Also:

- Consider `index.html` and remove not replaced placeholders in the `inject-vite-envs.sh` script.
- Add institution vars to ImportMetaEnv.

Refs #1352.

### DIFF
--- a/dashboard/Dockerfile
+++ b/dashboard/Dockerfile
@@ -30,6 +30,24 @@ server {
     listen 80;
     root /usr/share/nginx/html;
     absolute_redirect off;
+
+    # Clickjacking protection, not supported via CSP meta.
+    add_header Content-Security-Policy "frame-ancestors 'none'" always;
+
+    # Full CSP via header, if not using the meta tag.
+    # Replace or remove the OIDC provider and institution logo URLs as needed.
+    # add_header Content-Security-Policy "default-src 'self'; script-src-attr 'none'; img-src 'self' data: http://institution.com/logo.png; frame-src 'none'; object-src 'none'; base-uri 'self'; form-action 'self'; connect-src 'self' http://oidcprovider.com/realms/example/; frame-ancestors 'none'" always;
+
+    # When serving over HTTPS:
+    # (Append to CSP) ... ; upgrade-insecure-requests
+
+    # Reporting:
+    # (Append to CSP) ... ; report-uri https://example.com/csp
+
+    # Reporting (not broadly supported yet):
+    # add_header Reporting-Endpoints 'csp="https://example.com/csp"' always;
+    # (Append to CSP) ... ; report-to csp
+
     location /api/ingest/monitor {
         proxy_pass http://backend/ingest/monitor;
         proxy_http_version 1.1;
--- a/dashboard/hack/inject-vite-envs.sh
+++ b/dashboard/hack/inject-vite-envs.sh
@@ -11,12 +11,18 @@ cp -r $ENDURO_DASHBOARD_DIST/* $ENDURO_DASHBOARD_ROOT/
 VITE_ENVS=$(printenv | awk -F= '$1 ~ /^VITE/ {print $1}' | sed 's/^/\$/g' | paste -sd,);
 echo "Vite envs: ${VITE_ENVS}"
 
-# Inject environment variables into distribution files
+# Inject environment variables into distribution files and remove
+# placeholders that were not replaced (env. vars. not set).
 for file in $ENDURO_DASHBOARD_ROOT/assets/*.js;
 do
-    echo "Inject VITE environment variables into $(basename $file)"
+    echo "Inject VITE environment variables into assets/$(basename $file)"
     envsubst $VITE_ENVS < $file > $TMP_DIR/$(basename $file)
+    sed -E -i 's/\$VITE_[A-Z0-9_]+//g' $TMP_DIR/$(basename $file)
     cp $TMP_DIR/$(basename $file) $file
 done
+echo "Inject VITE environment variables into index.html"
+envsubst $VITE_ENVS < $ENDURO_DASHBOARD_ROOT/index.html > $TMP_DIR/index.html
+sed -E -i 's/\$VITE_[A-Z0-9_]+//g' $TMP_DIR/index.html
+cp $TMP_DIR/index.html $ENDURO_DASHBOARD_ROOT/index.html
 
 rm -rf $TMP_DIR
--- a/dashboard/vite.config.ts
+++ b/dashboard/vite.config.ts
@@ -3,8 +3,38 @@ import { URL, fileURLToPath } from "node:url";
 import vue from "@vitejs/plugin-vue";
 import Icons from "unplugin-icons/vite";
 import VueRouter from "unplugin-vue-router/vite";
-import { defineConfig } from "vite";
+import { defineConfig, loadEnv } from "vite";
+import csp from "vite-plugin-csp-guard";
 import vueDevTools from "vite-plugin-vue-devtools";
+
+// Load environment variables from .env* files and the current process.
+// The variables from the current process take precedence.
+const fileEnv = loadEnv(process.env.NODE_ENV || "", process.cwd(), "");
+const env = { ...fileEnv, ...process.env };
+
+const cspPolicy: Record<string, string[]> = {
+  "default-src": ["'self'"],
+  "script-src-attr": ["'none'"],
+  "img-src": ["'self'", "data:"],
+  "frame-src": ["'none'"],
+  "object-src": ["'none'"],
+  "base-uri": ["'self'"],
+  "form-action": ["'self'"],
+};
+
+if (env.VITE_OIDC_AUTHORITY) {
+  cspPolicy["connect-src"] = ["'self'", env.VITE_OIDC_AUTHORITY];
+}
+
+if (env.VITE_INSTITUTION_LOGO) {
+  cspPolicy["img-src"].push(env.VITE_INSTITUTION_LOGO);
+}
+
+if (env.NODE_ENV == "development") {
+  cspPolicy["style-src-elem"] = ["'self'", "'unsafe-inline'"];
+  cspPolicy["script-src-elem"] = ["'self'"]; // Needed to avoid SRI in dev.
+  cspPolicy["frame-src"] = ["'self'"];
+}
 
 // https://vitejs.dev/config/
 export default defineConfig({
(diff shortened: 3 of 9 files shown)


jraddaoui:
@sbreker, @djjuhasz, this is ready for an initial review. Please, check some additional context I added in the related issue. And the documentation for the plugin we are using can be found [here](https://vite-csp.tsotne.co.uk/).

I tried to add [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) but in some cases we update the final assets after build to replace env. vars. in the scripts/html, invalidating the hashes. In the end it would only protect against tampered JS/CSS file on disk as everything is same-origin for now. I'd still like to investigate other options to the environment substitution at some point, but that won't be part of this PR.

I was also worried about not being able to generate a nonce in a SPA, but it looks like we can live without `unsafe-inline` in production for now.

dashboard/hack/inject-vite-envs.sh
```diff
@@ -11,12 +11,18 @@ cp -r $ENDURO_DASHBOARD_DIST/* $ENDURO_DASHBOARD_ROOT/
...
+# placeholders that were not replaced (env. vars. not set).
 for file in $ENDURO_DASHBOARD_ROOT/assets/*.js;
 do
-    echo "Inject VITE environment variables into $(basename $file)"
+    echo "Inject VITE environment variables into assets/$(basename $file)"
     envsubst $VITE_ENVS < $file > $TMP_DIR/$(basename $file)
+    sed -E -i 's/\$VITE_[A-Z0-9_]+//g' $TMP_DIR/$(basename $file)
     cp $TMP_DIR/$(basename $file) $file
 done
+echo "Inject VITE environment variables into index.html"
```
djjuhasz, line 23: Why do the environment vars need to be injected into index.html?

dashboard/Dockerfile
```diff
@@ -30,6 +30,21 @@ server {
     listen 80;
     root /usr/share/nginx/html;
     absolute_redirect off;
+
+    # Clickjacking protection, not supported via CSP meta.
+    add_header Content-Security-Policy "frame-ancestors 'none'" always;
+
+    # Full CSP via header, if not using the meta tag.
```
djjuhasz, line 37: What does "if not using the meta tag" mean?

dashboard/Dockerfile
```diff
@@ -30,6 +30,21 @@ server {
...
+    # Replace or remove the OIDC provider and institution logo URLs as needed.
+    # add_header Content-Security-Policy "default-src 'self'; script-src-attr 'none'; img-src 'self' data: http://institution.com/logo.png; frame-src 'none'; object-src 'none'; base-uri 'self'; form-action 'self'; connect-src 'self' http://oidcprovider.com/realms/example/; frame-ancestors 'none'" always;
+
+    # When serving over HTTPS:
```
djjuhasz, line 41: If using the "meta tag" and HTTPS does "upgrade-insecure-requests" still need to be added manually or will the CSP plugin handle that?

dashboard/Dockerfile
```diff
@@ -30,6 +30,21 @@ server {
...
+    # (Append to CSP) ... ; upgrade-insecure-requests
+
+    # Reporting:
+    # add_header Reporting-Endpoints 'csp="https://example.com/csp"' always;
+    # (Append to CSP) ... ; report-to csp
```
djjuhasz, line 49: It might be helpful to add "report-uri" to this example too, until "report-to" is universally supported: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/report-to#usage_notes

djjuhasz (approved):
Looks good to me @jraddaoui.  I had a few questions and one suggestion for docs, but no blockers.

dashboard/vite.config.ts
```diff
@@ -3,9 +3,39 @@ import { URL, fileURLToPath } from "node:url";
...
+}
+
+if (env.VITE_INSTITUTION_LOGO) {
+  cspPolicy["img-src"].push(env.VITE_INSTITUTION_LOGO);
+}
+
+if (env.NODE_ENV == "development") {
+  cspPolicy["style-src-elem"] = ["'self'", "'unsafe-inline'"];
+  cspPolicy["script-src-elem"] = ["'self'", "'unsafe-eval'", "'unsafe-inline'"];
+  cspPolicy["frame-src"] = ["'self'"];
```
sbreker, line 36: In dev mode it would be better if we could use `Content-Security-Policy-Report-Only` and use the prod directives instead of flagging with unsafe-* because this will hide CSP violations from devs.

When using 'report-only', devs can check the browser console for any CSP violations caused by current work. CSP 'report-only' mode will not block the offending scripts so the application will continue to work properly.

The `Content-Security-Policy-Report-Only` response header allows developers to test CSP directives by reporting violations without actually enforcing them.

dashboard/Dockerfile
sbreker: `style-src` is not specified, so I think the `default-src 'self'` will be used for styles. Is this intentional?

sbreker (commented):
Hi @jraddaoui! Thanks for this! I added a couple comments about adding report-only mode, and the default for `style-src`.

dashboard/hack/inject-vite-envs.sh
jraddaoui: We now need to replace those env. vars. in the meta tag value too, which is added to index.html.

dashboard/Dockerfile
jraddaoui: If you disable the CSP at build, this is what would be in the meta tag. Just a full example in case you prefer the header to the meta.

dashboard/Dockerfile
jraddaoui: > If using the "meta tag" and HTTPS does "upgrade-insecure-requests" still need to be added manually or will the CSP plugin handle that?

No, this will never be added to the meta right now. You may not know how the app will be served at build time, like it happens with the builds we make (release/RPM).

> style-src is not specified, so I think the default-src 'self' will be used for styles. Is this intentional?

Yes, that's intentional. It's okay for dev and prod, while only the more granular `style-src-elem` is updated for dev.

dashboard/vite.config.ts
jraddaoui: Thanks @sbreker, this is a really good point and something that I had a hard time making a decision.

My first point against report only is that a violation that can be easily missed if you don't pay close attention to the console. I kind of prefer to break the app so you notice the issue right away. Second, which adds to the first one, is that the HMR (hot reload) from Vite needs unsafe-eval and unsafe-inline, if we forbid them in report-only we'll get a lot of noise in the console.

In the build, all inline JS/CSS will be extracted, even if we add inline content it will be removed, so I think there is no need to report those in dev. However, for unsafe-eval not all the violations may be removed, the ones from Vite/Vue won't be present but a dev can still introduce them directly or via third party library.

However, I still don't want to report those given the issues mentioned above. So what I'm thinking now is that we could add lint rules to prevent the addition in our code and, even if they are not common in modern libs we could scan the build for possible violations in CI. One suggestion from ChatGPT, after build:

```bash
rg -n "eval\\s*\\(|new\\s+Function\\s*\\(|set(Time|Inter)val\\s*\\(\\s*['\`\"]" dist/ || true
```

Thoughts?

dashboard/Dockerfile
djjuhasz: Ah, I didn't realize a CSP could be specified by header or meta tag, or that the CSP Guard plugin uses meta tags for its CSP.  I think it would be helpful to note here that CSP headers are preferred to using meta tags and that some CSP tags are not supported when using meta tags.

dashboard/vite.config.ts
```diff
@@ -15,30 +45,43 @@ export default defineConfig({
     vue({}),
     vueDevTools(),
     Icons({ compiler: "vue3" }),
+    // Include CSP plugin if not explicitly disabled.
+    ...(env.ENDURO_DISABLE_CSP?.trim().toLowerCase() === "true"
```
djjuhasz: It would be nice to document the `ENDURO_DISABLE_CSP` variable somewhere other than this config file, either by adding it to the `.env` file (set to false) or adding note to the dashboard config doc page.

djjuhasz: Also maybe naming the var `ENDURO_DISABLE_META_CSP` would make the distinction about it using meta tags a little clearer?

dashboard/vite.config.ts
djjuhasz: @jraddaoui do you have a reference for the assertion that "HMR (hot reload) from Vite needs unsafe-eval and unsafe-inline"?  I've been Googling around for an explanation but haven't found anything that clearly states these requirements.

jraddaoui: Thanks for making me look at this again @djjuhasz!

Initial research suggested both are needed, I removed and added them while testing as I was seeing them appear in the console, I think it was the way the plugin reloads the meta and handles SRI what made me think they were needed as I removed them, but all the errors were from unsafe-inline in styles. So it looks like we don't need neither of them for scripts, just unsafe-inline in style-src-elem [due to Vite HMR](https://vite.dev/guide/features.html#css) and self in frame-src for the Vue dev tools.

Thoughts about report vs fail?

dashboard/Dockerfile
jraddaoui: I agree, and I think that should go in the documentation in a build/serve the dashboard section, where we can place the replacement script docs, etc. I'll do that in another PR.

dashboard/vite.config.ts
sbreker: Thanks for the explanation @jraddaoui! I agree - the noise in the console log would make `report-only` not an effective way to detect violations. If scanning in CI could be accomplished, this would be an effective workaround to ensure new commits to not break the app when CSP is enforcing the prod rules.

sbreker (approved):
Thanks @jraddaoui!  :+1:

dashboard/vite.config.ts
jraddaoui: Thanks @sbreker! In the end unsafe-eval is not needed in dev either, so we can make it fail.

PR #1346, opened 2025-09-04
Upgrade Keycloak from 24.0.4 to 26.3

djjuhasz:
Bump the version of Keycloak installed by the Kube helm chart from 24.0.4 to the latest version of 26.3 (currently 26.3.3).


jraddaoui (approved):
Thanks!

PR #1348, opened 2025-09-05
Add `sub` claim to Keycloak enduro client

djjuhasz:
Version 25 of Keycloak removed the OIDC `sub` claim from the default access token claims, which breaks several checks of the `sub` claim in Enduro. This commit explicitly adds the `sub` claim back to the Keycloak enduro client.


jraddaoui (approved):
Thanks @djjuhasz!

Issue #1229, opened 2025-05-22
Problem: AIP Deletion, approval, and rejection buttons are shown to users without permission to use those methods


fiver-watson:
**Describe the bug**

Currently all users will see the option to initiate an AIP deletion, as well as approve or reject a deletion request. This is despite the fact that some users may not have permissions to take these actions - for example, basic users approving their own deletion requests. 

**To Reproduce**

1. Log in
2. Navigate to Storage > Browse AIPs
3. Select AIP from the browse list and navigate to its View page
4. Initiate an AIP deletion request
5. Refresh the page 

Resulting error: user is still shown review buttons (i.e. Approve and Reject), though clicking them does nothing. Consequently, it is not clear to users whether they have permissions to do this but Enduro isn't working, or if in fact they cannot do this. 

**Expected behavior**

Buttons that cannot be used by specific users based on permissions should either: 

* Present in a manner that makes it clear they are deactivated, or
* Not be shown to users who do not have the permissions required to use them

djjuhasz:
After initiating a deletion request the requestor no longer sees the "Approve" or "Reject" buttons for the request. The Approve/Delete buttons are also not shown to users that do not have [storage:aips:deletion:review] or [storage:aips:deletion:request] permissions.  A user that has permissions to request a deletion but does *not* have permissions to review a deletion may still be able to see the Approve and Delete buttons for a deletion request where they are not the requestor, even though they shouldn't.

The "Delete" button to initiate a deletion request is not shown to a user if they don't have [storage:aips:deletion:request] permissions.  The "Delete" button is also not shown for an AIP that already has a deletion request pending, even if the user has permissions to request deletion.

djjuhasz:
Commit 27136ca hides the deletion request "Approve" and "Reject" buttons if the user doesn't have `[storage:aips:deletion:review]` permissions. I think this change fixes the last instance where the AIP deletion buttons were being shown when the user has insufficient permissions to perform the related action.

fiver-watson:
Interestingly, when I first logged in as operator, I saw the approve and delete buttons on the first Pending AIP I opened... but then when I checked them all again, they were no longer shown. Cache issue maybe?

Also of note: the "Cancel" button is not working with the operator account for this SIP: 

* https://sfa-dev.archivematica.net:8844/storage/aips/f45d2275-72c0-47c6-bcaf-f612b7438776

Is this a bug, or some issue w how we have handled the test site and redeployments etc?

[...]

PR #1351, opened 2025-09-08
Check auth for deletion review buttons

djjuhasz:
Refs #1229.

Don't show "Approve" and "Reject" buttons when user isn't authorized to review AIP deletion requests.


jraddaoui (approved):
LGTM!

PR #1353, opened 2025-09-10
Link config docs

fiver-watson:
Add links between the two pages.


djjuhasz (approved):
Thanks @fiver-watson. :+1:

PR #1357, opened 2025-09-13
Provide early feedback in Dashboard "Create AIP" workflow

djjuhasz:
Add a "Copy SIP" task to the beginning of the "Create AIP" workflow to provide feedback to the operator at the beginning of the workflow on the workflow progress.

### DIFF
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -3,7 +3,6 @@
 package workflow
 
 import (
-	"database/sql"
 	"errors"
 	"fmt"
 	"io"
@@ -17,7 +16,6 @@ import (
 	"github.com/artefactual-sdps/temporal-activities/bagvalidate"
 	"github.com/artefactual-sdps/temporal-activities/bucketcopy"
 	"github.com/artefactual-sdps/temporal-activities/bucketdelete"
-	"github.com/artefactual-sdps/temporal-activities/bucketdownload"
 	"github.com/artefactual-sdps/temporal-activities/bucketupload"
 	"github.com/artefactual-sdps/temporal-activities/removepaths"
 	"github.com/artefactual-sdps/temporal-activities/xmlvalidate"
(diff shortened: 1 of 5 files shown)


internal/workflow/processing.go
```diff
@@ -315,91 +313,38 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+		// TODO: make activities idempotent.
+		var res createWorkflowLocalActivityResult
 		ctx := withLocalActivityOpts(sessCtx)
 		err := temporalsdk_workflow.ExecuteLocalActivity(
 			ctx,
 			createWorkflowLocalActivity,
 			w.ingestsvc,
 			&createWorkflowLocalActivityParams{
-				UUID:       state.workflowUUID,
+				RNG:        w.rng,
```
jraddaoui, line 326: Is this really solving the deterministic UUIDs issue? I thought we would need to use [side effects](https://docs.temporal.io/develop/go/side-effects) and that's why I preferred to pass the UUIDs from workflow instead of generating them in the activities.

But now I see:

> An Activity or a Local Activity may also be used instead of a Side effect, as its result is also persisted in Workflow Execution History.

But we also said that activities may be executed more than once and that's why we need to make them idempotent.

jraddaoui (commented):
Thanks @djjuhasz! I added a comment about the UUID changes, everything else looks good to me.

internal/workflow/processing.go
djjuhasz: @jraddaoui I actually think the logic behind retrying the the session activities manually is misguided, I think we should set up Temporal retry policies for the workflow and let Temporal manage retries instead of trying to do it ourselves.  I'm guessing the manual retry logic is the result of some sort of failures that were happening early in Enduro's history, but I wonder if it might have been due to a Temporal bug that has since been fixed.  I think the current implementation of the retry has some issues around data persistence and the Temporal retry logic — for instance we are writing the workflow execution ID to the database, and the execution ID will change if the workflow is executed a second time, so I'm not sure how that would work if there was a mixture of Temporal retries and our internal session retries.

In any case, I think generating the UUIDs in an activity is an improvement over generating them in the workflow without using side effects, which is the current state.  And while this change doesn't make the activity idempotent I still think it is better than the status quo.

jraddaoui: I was thinking about full workflow retries, something that we did in another project and required to make the activities idempotent. I think those retires get a different run id but the same workflow id. Anyway, nothing to worry in this PR.

jraddaoui (approved):
LGTM, thanks @djjuhasz!

internal/workflow/processing.go
djjuhasz: @jraddaoui I think you are right about new workflow executions using the same workflow ID but a different run ID.  I assumed `temporalsdk_workflow.GetInfo(ctx).WorkflowExecution.ID` was the *execution* ID, but it looks like it's actually the workflow ID based on https://github.com/temporalio/sdk-go/blob/master/internal/workflow.go#L331

Issue #1289, opened 2025-07-15
Problem: Enduro UI does not give user an indication that a workflow is starting when ingesting large SIPs


fiver-watson:
Up until recently, we have been primarily using smaller test packages for QA for convenience, which typically means that ingest both starts and finishes very quickly. Handy for testing, but not good for fully understanding the user experience. Now that we are testing with larger packages, we are seeing some of the gaps. Namely: 

When a large SIP is first ingested, Enduro does not record any initial tasks. Consequently, the user is shown an empty SIP view page with no indication of what workflow is running, or that the package has been received. 

**To Reproduce**

1. Use one of the test packages in [this Drive directory](https://drive.google.com/drive/folders/1GWVX0vmq_ls5nXGPVlwW9nkfMGNKGKhy?usp=drive_link), 6GB or larger
2. Once the upload is complete, look at the SIP view page of the package

**Outcome**

See this 15GB import as an example: https://sfa-dev.archivematica.net:8844/ingest/sips/69752d6f-b5f4-4402-bbe3-e4780bdbdf66

This page did not show the workflow name or any activities at all for several minutes, until many activities appeared all at once, up to the Bag SIP activity  (if I recall correctly). This despite that fact that: 

* there is an almost 6 minute gap between when the workflow started and when the first "Extract SIP" activity completed, and 
* in inspecting the individual activity time stamps, there is also a gap between the "Validate SIP name" and "verify SIP manifest activities" where an update could have been provided (37 seconds - which feels long if a user is staring at the screen expecting an update)

Finally, note that at the time this issue was filed (11:40am Eastern, 2025-07-15, almost 23 hours after the ingest started), this is how things were stalled out - with no feedback from Archivematica on status or whether the SIP was still processing: 

<img width="1339" height="925" alt="Image" src="https://github.com/user-attachments/assets/23510feb-f57e-4c80-bbe4-b8c3746c85f8" />

(However, this last issue is likely large and may require changes in AM - it is likely to be addressed in a separate ticket. For now it's just a bit of an FYI on the user experience when working with larger packages.)

**Expected behavior**

* We know what workflow is going to be run when an ingest is initiated - we should show the workflow header with the name, start time, and status of the workflow in the ingest workflow details area (at the bottom of the SIP view page) as soon as possible - or, if queued, then without the start time. 
[...]

PR #1362, opened 2025-09-17
Add spinner to "in progress" workflow status badge

djjuhasz:
Refs #1289.

- Add a spinner to the workflow "in progress" badge
- Automatically expand the task list for "in progress" workflows

### DIFF
--- a/dashboard/src/components/StatusBadge.vue
+++ b/dashboard/src/components/StatusBadge.vue
@@ -15,10 +15,9 @@ export type StatusEnum = packageEnum | workflowEnum | taskEnum;
 <script setup lang="ts">
 import { computed } from "vue";
 
-import type { api } from "@/client";
+import { api } from "@/client";
 
 type badgeType = "package" | "workflow";
-
 type badgeStyle = string[];
 
 const props = defineProps<{
@@ -78,8 +77,14 @@ const colorClass = computed(() => {
 
 <template>
   <span>
-    <span :class="['badge', colorClass]">
+    <span :class="['badge', colorClass]" id="status-badge">
       {{ props.status.toUpperCase() }}
+      <div
+        v-if="props.status == api.EnduroIngestSipWorkflowStatusEnum.InProgress"
+        class="spinner-border spinner-border-sm text-black"
+        role="progress"
+        aria-hidden="true"
+      ></div>
     </span>
     <span v-if="props.note" class="badge text-dark fw-normal"
       >({{ props.note }})</span
--- a/dashboard/src/components/WorkflowCollapse.vue
+++ b/dashboard/src/components/WorkflowCollapse.vue
@@ -23,42 +23,78 @@ const tasks = computed<api.EnduroIngestSipTask[] | api.EnduroStorageAipTask[]>(
 const props = defineProps<{
   workflow: api.EnduroIngestSipWorkflow | api.EnduroStorageAipWorkflow;
   index: number;
+  of: number;
 }>();
 
 const { workflow, index } = toRefs(props);
 
 let expandCounter = ref<number>(0);
 
-const showSipReviewAlert = (
-  workflow: api.EnduroIngestSipWorkflow | api.EnduroStorageAipWorkflow,
-) => {
+const showSipReviewAlert = () => {
   return (
-    workflow.type == api.EnduroIngestSipWorkflowTypeEnum.CreateAndReviewAip &&
-    workflow.status == api.EnduroIngestSipWorkflowStatusEnum.Pending
+    workflow.value.type ==
+      api.EnduroIngestSipWorkflowTypeEnum.CreateAndReviewAip &&
+    workflow.value.status == api.EnduroIngestSipWorkflowStatusEnum.Pending
   );
 };
 
-const showAipDeletionReviewAlert = (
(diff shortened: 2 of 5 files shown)


djjuhasz:
Spinner and auto-expand together:
[Screencast from 2025-09-17 09-33-27.webm](https://github.com/user-attachments/assets/2e0ce830-13c1-4271-9111-4b0b9a1a95be)

dashboard/src/components/WorkflowCollapse.vue
```diff
@@ -46,12 +46,21 @@ const showAipDeletionReviewAlert = (
...
   );
 };
+
+// Auto-expand the workflow task list if the workflow is "in progress".
+const wfBtn = ref<HTMLElement | null>(null);
+onMounted(() => {
+  if (workflow.value.status === "in progress") {
+    wfBtn.value?.click();
+  }
+});
```
jraddaoui: Feels like a hack to do this. Why not checking in the attributes directly? I think it only requires to remove the collapsed class and set aria-expanded to true in the button, and add show to the content classes. Like the first item in the first example:

https://getbootstrap.com/docs/5.3/components/accordion/

jraddaoui (changes requested):
Thanks @djjuhasz!

dashboard/src/components/WorkflowCollapse.vue
```diff
@@ -46,12 +46,21 @@ const showAipDeletionReviewAlert = (
     workflow.status == api.EnduroStorageAipWorkflowStatusEnum.Pending
   );
 };
+
+// Auto-expand the workflow task list if the workflow is "in progress".
+const wfBtn = ref<HTMLElement | null>(null);
+onMounted(() => {
+  if (workflow.value.status === "in progress") {
```
jraddaoui: Also, it would be nice to use the constant that comes in the API client code instead of the "in progress" string.

dashboard/src/components/StatusBadge.vue
```diff
@@ -78,8 +78,14 @@ const colorClass = computed(() => {
 
 <template>
   <span>
-    <span :class="['badge', colorClass]">
+    <span :class="['badge', colorClass]" id="status-badge">
       {{ props.status.toUpperCase() }}
+      <div
+        v-if="props.status === 'in progress'"
```
jraddaoui: Same here.

dashboard/src/components/WorkflowCollapse.vue
djjuhasz: Oh you mean use `api.EnduroIngestSipWorkflowStatusEnum.InProgress` instead of `"in progress"`?  :stuck_out_tongue_winking_eye: Sometimes I think us programmers are our own worst enemies.

dashboard/src/components/WorkflowCollapse.vue
djjuhasz: Fine, I'll do it the "right" way. :roll_eyes: :stuck_out_tongue_closed_eyes:

djjuhasz:
This is ready for another look @jraddaoui.

dashboard/src/components/WorkflowCollapse.vue
```diff
@@ -50,15 +50,27 @@ const showAipDeletionReviewAlert = (
...
+          },
+        ]"
         type="button"
         data-bs-toggle="collapse"
-        :data-bs-target="'#w-body-' + index"
-        aria-expanded="false"
-        :aria-controls="'w-body-' + index"
+        :data-bs-target="'#wf' + index + '-tasks'"
+        :aria-expanded="
+          workflow.status == api.EnduroIngestSipWorkflowStatusEnum.InProgress
```
jraddaoui: Do we need to look for the AIP workflow status too? I know is really ugly looking at those constants but it gives some safety. Maybe moving it to a computed value or a helper function makes it a little less ugly ;D

dashboard/src/components/StatusBadge.vue
jraddaoui: Missed this one?

jraddaoui (commented):
Thanks @djjuhasz!

dashboard/src/components/WorkflowCollapse.vue
djjuhasz: I don't know why but 
```
workflow.status == api.EnduroIngestSipWorkflowStatusEnum.InProgress ||
workflow.status == api.EnduroStorageAipWorkflowStatusEnum.InProgress
```

Causes a compile time error:
```
This comparison appears to be unintentional because the types '"done" | "unspecified" | "error" | "queued" | "pending" | "failed" | "canceled"' and '"in progress"' have no overlap.
```

Using `api.EnduroIngestSipWorkflowStatusEnum.InProgress` or `workflow.status == api.EnduroStorageAipWorkflowStatusEnum.InProgress` works, but I can't use both in the same conditional. :shrug:

jraddaoui: Typescript typing mate, I don't know :man_shrugging:. If you decide to use a computed value or helper function, I guess two conditionals won't hurt that much.

djjuhasz:
I'm going to break this into multiple PRs.

PR #1311, opened 2025-08-01
Add AIP view and manage docs

fiver-watson:
Adds a new page for "Viewing and managing AIPs" to the Storage section of the user manual.

PR #1363, opened 2025-09-18
Auto-expand workflow task lists

### LINKED ISSUE #1131: Change Request: Ensure that Preservation action tasks are visible by default when only one Action is listed on a package view page

Throughout development and iteration on the Enduro user interface, we have received feedback (some observed, some reported) about the discoverability of the tasks listed under a Preservation Action shown on a package page.  We made some changes in #986 , but are still receiving various reports from users, such as this latest feedback from  a client: 

Image

RE: The first feedback about the label "Preservation actions" - this will be addressed for SIPs as part of the SIP view page redesign currently proposed in #955. For now we will keep the label "Preservation actions" on the AIP view page, as there is a much higher chance of having multiple preservation workflows run there, and the feedback acknowledged that the info icon was useful in explaining the domain term. So: out of scope for this issue

RE: the second piece of feedback on having the task list open by default, we propose making a change to help address this. 

Add a conditional rule for the visibility of tasks lists (i.e. expanded or collapsed by default on page load): 

* If there is only 1 preservation action, then expand the task list by default on page load
* If there is more than 1 preservation action listed, then keep all collapsed by default on page load
* If a workflow is underway and is paused with a "Pending" status awaiting a user decision, expand the task list by default so that the task awaiting the user decision is more easily discoverable. 

* Ignoring the feedback and leaving things as-is
* Always expanding all Preservation actions to show tasks 
* Always expanding the FIRST preservation action to show tasks
* Moving Preservation actions to its own tab on the package details page

Of these options, we feel the selected / preferred option is the best because: 

* Doing nothing ignores useful user feedback
* Previous testing found that users often stopped scrolling before reaching the bottom of a long task list
* This means if there are multiple preservation actions, earlier ones may not be readily discovered by users
* With the incoming redesign, we generally expect there to be only 1 workflow / preservation action for SIPs, describing the ingest / Create AIP workflow. Having this expanded by default will address the feedback
* For AIPs, it is likely we will over time have more as more functionality is added - Moves, Fixity checks, Deletions, Reingests, DIP creation, Manual replication requests, etc. We want to ensure these are all discoverable to users on an AIP view page
* Implementing the logic conditionally means the same code can be used on SIP and AIP view pages
* Moving everything to a dedicated "Preservation actions" tab does not directly address the feedback on its own, and may in fact reduce the discoverability of the task details overall

djjuhasz:
Refs #1311.

- Make a workflow task list expanded by default if the workflow is currently "in progress", "pending" a user decision or is the only workflow on the page
- Update HTML `id` attributes for workflow accordion elements

### DIFF
--- a/dashboard/src/components/WorkflowCollapse.vue
+++ b/dashboard/src/components/WorkflowCollapse.vue
@@ -23,42 +23,78 @@ const tasks = computed<api.EnduroIngestSipTask[] | api.EnduroStorageAipTask[]>(
 const props = defineProps<{
   workflow: api.EnduroIngestSipWorkflow | api.EnduroStorageAipWorkflow;
   index: number;
+  of: number;
 }>();
 
 const { workflow, index } = toRefs(props);
 
 let expandCounter = ref<number>(0);
 
-const showSipReviewAlert = (
-  workflow: api.EnduroIngestSipWorkflow | api.EnduroStorageAipWorkflow,
-) => {
+const showSipReviewAlert = () => {
   return (
-    workflow.type == api.EnduroIngestSipWorkflowTypeEnum.CreateAndReviewAip &&
-    workflow.status == api.EnduroIngestSipWorkflowStatusEnum.Pending
+    workflow.value.type ==
+      api.EnduroIngestSipWorkflowTypeEnum.CreateAndReviewAip &&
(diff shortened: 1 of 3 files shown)


djjuhasz:
The logic to expand an "in progress" workflows was originally part of  PR #1362

jraddaoui (approved):
Neat! Thanks @djjuhasz!

PR #1364, opened 2025-09-18
Add spinner to workflow "in progress" badge

djjuhasz:
Refs #1289.

Add an animated spinner to the workflow "in progress" badge as a visual cue that the workflow is still running.

### DIFF
--- a/dashboard/src/components/StatusBadge.vue
+++ b/dashboard/src/components/StatusBadge.vue
@@ -15,10 +15,9 @@ export type StatusEnum = packageEnum | workflowEnum | taskEnum;
 <script setup lang="ts">
 import { computed } from "vue";
 
-import type { api } from "@/client";
+import { api } from "@/client";
 
 type badgeType = "package" | "workflow";
-
 type badgeStyle = string[];
 
 const props = defineProps<{
@@ -80,6 +79,12 @@ const colorClass = computed(() => {
   <span>
     <span :class="['badge', colorClass]">
       {{ props.status.toUpperCase() }}
+      <div
+        v-if="props.status == api.EnduroIngestSipWorkflowStatusEnum.InProgress"
+        class="spinner-border spinner-border-sm text-black"
+        role="progress"
+        aria-hidden="true"
+      ></div>
     </span>
     <span v-if="props.note" class="badge text-dark fw-normal"
       >({{ props.note }})</span


djjuhasz:
This was original part of PR #1362

dashboard/src/components/StatusBadge.vue
```diff
@@ -80,6 +79,14 @@ const colorClass = computed(() => {
   <span>
     <span :class="['badge', colorClass]">
       {{ props.status.toUpperCase() }}
+      <div
+        v-if="props.status == api.EnduroIngestSipWorkflowStatusEnum.InProgress"
```
jraddaoui, line 83: Do we want to do this for AIP workflows too?

dashboard/src/components/StatusBadge.vue
```diff
@@ -80,6 +79,14 @@ const colorClass = computed(() => {
...
+        class="spinner-border spinner-border-sm text-black"
+        role="progress"
+        aria-hidden="true"
+      >
+        <span class="visually-hidden">Running...</span>
```
jraddaoui: I think this span could be removed as it will read "In progress running", just "In progress" should be enough for screen readers.

jraddaoui (changes requested):

dashboard/src/components/StatusBadge.vue
djjuhasz: Adding `api.EnduroStorageAipWorkflowStatusEnum.InProgress` in the comparison causes a compile error, and I don't think it's possible to do the type filtering I did in #1363 because we aren't passing a workflow object to this page, just a "status" string.  At the end of the day we are comparing the string "in progress" to another string here, despite all the fancy Typescript type magic going on — the SIP and AIP "in progress" string is exactly the same in both cases.

dashboard/src/components/StatusBadge.vue
djjuhasz: The examples in Bootstrap all have a `<span class="visually-hidden">Loading...</span>` so I'm trying to follow that pattern.  I don't know enough about screen readers to guess how they will render an empty `<div>` tag. :thinking:

djjuhasz: The Bootstrap examples also don't have an `aria-hidden="true"` attribute on the spinner `<div>`, so I should probably remove that if adding the "Running..." span. :thinking:

dashboard/src/components/StatusBadge.vue
djjuhasz: One distinction with the generated enums that took me a bit to figure out is that the generated code declares a `EnduroIngestSipWorkflowStatusEnum` object *and* a `EnduroIngestSipWorkflowStatusEnum` type.  The object has key/value pairs whereas the type is just a list of allowed string literals (e.g. "in progress", "done", "unspecified").  In this StatusBadge component the `props.status` type is a union list of all the possible status string literals (type `StatusEnum`) but we aren't actually passing a `EnduroIngestSipWorkflowStatusEnum` or `EnduroStorageAipWorkflowStatusEnum` *object* to the component.

jraddaoui: But we are not passing just a string, we have all the types above in that script and use them for the property.

https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/components/StatusBadge.vue#L1-L12

I don't like it either, but at least it gives us some safety. If this needs to work the same for AIPs, which was my only initial question, I'd like to notice if we change that enum at some point.

dashboard/src/components/StatusBadge.vue
jraddaoui: Right, this is not the usual icon + read text, the span is inside the actual div and the icon is added through CSS only. So you are right, that aria-hidden is probably hiding the span already.

In this case we are adding a visual reinforcement to an existing text, I'm not sure we need the textual reinforcement for screen readers, but I guess it won't hurt either. If the spinner works just with the div I'd leave like that, an empty div shouldn't be an issue.

jraddaoui (approved):
Thanks @djjuhasz, feel free to ignore my comments ;)

dashboard/src/components/StatusBadge.vue
djjuhasz: @jraddaoui if you look at the defnition:

```
export const EnduroIngestSipWorkflowStatusEnum = {
    Unspecified: 'unspecified',
    InProgress: 'in progress',
    Done: 'done',
    Error: 'error',
    Queued: 'queued',
    Pending: 'pending',
    Failed: 'failed'
} as const;
```

`EnduroIngestSipWorkflowStatusEnum.InProgress` is not explicitly typed as `StatusEnum` or any other type, so it's inferred to be plain string type, and the same for `EnduroStorageAipWorkflowStatusEnum.InProgress`.  One of the weird things (imho) about these generated types is that the `EnduroIngestSipWorkflowStatusEnum` object doesn't use the `EnduroIngestSipWorkflowStatusEnum` type at all. :confused: 

So what https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/components/StatusBadge.vue#L1-L12 guarantees is that `props.status.value` is of type `StatusEnum`, but  I believe the value is converted to a plain string so it can be compared to `EnduroStorageAipWorkflowStatusEnum.InProgress` which is a string type.

jraddaoui: I understand what you say @djjuhasz. Moreover, the union type creates an overlap with both values that makes it harder to compare. Nevertheless:

```ts
const isInProgressWorkflow = computed(() => {
  return (
    (props.status as string) ===
      api.EnduroIngestSipWorkflowStatusEnum.InProgress ||
    (props.status as string) ===
      api.EnduroStorageAipWorkflowStatusEnum.InProgress
  );
});
```

This would reference both enum values directly so we get the type safety for both and compile time errors if any of them change in the future. The string casting prevents the union type issue.

Issue #1352, opened 2025-09-09
Problem: The dashboard does not implement CSP headers/meta


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Content Security Policy (CSP) is an added layer of security that helps to detect and mitigate certain types of attacks, including Cross Site Scripting (XSS) and data injection attacks. These attacks are used for everything from data theft to site defacement or distribution of malware. CSP provides a set of standard HTTP headers that allow website owners to declare approved sources of content that browsers should be allowed to load on that page — covered types are JavaScript, CSS, HTML frames, fonts, images and embeddable objects such as Java applets, ActiveX, audio and video files.

**Describe the solution you'd like**

Add CSP headers and/or meta tag to the Enduro dashboard. Ideally, this would run in development in a similar manner as in production, to be able to test new code and identify changes needed in the CSP.

**Describe alternatives you've considered**

Just use NGINX headers.

**Additional context**

The dashboard requires to connect to a configurable OIDC provider that is configured with environment variables. If the dashboard is built with that environment configured that value could be added to the CSP `connect-src` directive. However, we also provide a built version (Docker image, RPM package, release assets) where those environment variables can be replaced/injected after build (using [this script](https://github.com/artefactual-sdps/enduro/blob/main/dashboard/hack/inject-vite-envs.sh) or similar). The same applies to the configurable institution logo recently added, in this case that image URL needs to be added to the `img-src` directive.

jraddaoui:
This was initially implemented in #1343. I decided to use the [vite-plugin-csp-guard](https://vite-csp.tsotne.co.uk/) plugin for a couple of reasons:

- **Development integration.** An easy way to have the same CSP for production and development, being added by Vite while building and serving. Making it easier to find CSP issues while developing.
- **SRI support.** Adding [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) allows to identify modified assets after build.

However, after working with this implementation and trying to deploy it with Ansible, I found the following issues:

- **Poor performance in dev.** [The claim the plugin makes](https://vite-csp.tsotne.co.uk/api-docs#dev-properties) about the performance implications is more noticeable than I was expecting, causing multiple reloads of the same page and routing issues.
- **SRI is not possible.** The way we use the environment to configure the dashboard and the env replacement/injection script modifies the assets after build.
[...]

PR #1372, opened 2025-09-24
Use CSP header instead of meta

jraddaoui:
Refs https://github.com/artefactual-sdps/enduro/issues/1352. See https://github.com/artefactual-sdps/enduro/issues/1352#issuecomment-3329906827 for more info.


djjuhasz (approved):
LGTM @jraddaoui. It's too bad the Vite CSP plugin didn't work as expected, but this solution is easier to understand and if it improves performance in dev then I'm all for it — the constant page reloads were really frustrating and time-consuming.

PR #1375, opened 2025-09-26
Reconnect websockets when they close unexpectedly

### LINKED ISSUE #1057: Problem: Regularly get "write EPIPE" error from Dashboard in Tilt dev env

djjuhasz:
Fixes #1057.

Attempt to reconnect the ingest and storage monitor websockets when the connection is unexpected closed at the server end (e.g. on server restart). My tests so far indicate that this change fixes the problem I've been having with the Dashboard hanging intermittently and throwing Vite "write EPIPE" errors.

### DIFF
--- a/dashboard/src/client.ts
+++ b/dashboard/src/client.ts
@@ -1,5 +1,3 @@
-import { handleIngestEvent } from "./monitor-ingest";
-import { handleStorageEvent } from "./monitor-storage";
 import * as api from "./openapi-generator";
 import * as runtime from "./openapi-generator/runtime";
 
@@ -10,8 +8,6 @@ export interface Client {
   about: api.AboutApi;
   ingest: api.IngestApi;
   storage: api.StorageApi;
-  connectIngestMonitor: () => void;
-  connectStorageMonitor: () => void;
 }
 
 function getPath(): string {
@@ -24,42 +20,6 @@ function getPath(): string {
     "/api";
 
   return path.replace(/\/$/, "");
-}
-
-function getWebSocketURL(): string {
-  let url = getPath();
-
-  if (url.startsWith("https")) {
-    url = "wss" + url.slice("https".length);
-  } else if (url.startsWith("http")) {
-    url = "ws" + url.slice("http".length);
-  }
-
-  return url;
-}
-
-function connectIngestMonitor() {
-  const url = getWebSocketURL() + "/ingest/monitor";
-  const socket = new WebSocket(url);
-  socket.onmessage = (event: MessageEvent) => {
-    const body = JSON.parse(event.data);
-    const data = api.IngestEventFromJSON(body);
-    if (data.ingestValue) {
-      handleIngestEvent(data.ingestValue);
-    }
-  };
-}
-
-function connectStorageMonitor() {
-  const url = getWebSocketURL() + "/storage/monitor";
-  const socket = new WebSocket(url);
-  socket.onmessage = (event: MessageEvent) => {
-    const body = JSON.parse(event.data);
-    const data = api.StorageEventFromJSON(body);
-    if (data.storageValue) {
-      handleStorageEvent(data.storageValue);
-    }
-  };
 }
 
 function createClient(): Client {
@@ -84,8 +44,6 @@ function createClient(): Client {
     about: new api.AboutApi(config),
     ingest: new api.IngestApi(config),
     storage: new api.StorageApi(config),
-    connectIngestMonitor: connectIngestMonitor,
-    connectStorageMonitor: connectStorageMonitor,
   };
 }
 
--- a/dashboard/src/monitor.ts
+++ b/dashboard/src/monitor.ts
@@ -1,0 +1,186 @@
+import { api, client } from "@/client";
+import { handleIngestEvent } from "@/monitor-ingest";
+import { handleStorageEvent } from "@/monitor-storage";
+
+export type RetryOptions = {
+  initialDelay: number;
+  maxDelay: number;
+  backoff: number;
+  maxAttempts: number;
+  jitterFn: () => number;
+};
+
+abstract class MonitorConnection {
+  type: "ingest" | "storage";
+  url: string;
+  socket: WebSocket | null = null;
+  isConnected: boolean = false;
+  retry: RetryOptions;
+
+  constructor(
+    type: "ingest" | "storage",
+    baseUrl: string,
+    retry?: RetryOptions,
+  ) {
+    this.type = type;
+    this.url = this.getWebSocketURL(baseUrl + "/" + this.type + "/monitor");
+    this.retry = retry || {
+      initialDelay: 1000, // 1 second.
+      maxDelay: 30000, // 30 seconds.
+      backoff: 2,
+      maxAttempts: 10,
+      jitterFn: () => Math.random() * 500, // add up to 500ms of jitter.
+    };
+  }
+
+  abstract dial(): Promise<void>;
+
+  getWebSocketURL(url: string): string {
+    if (url.startsWith("https")) {
+      url = "wss" + url.slice("https".length);
+    } else if (url.startsWith("http")) {
+      url = "ws" + url.slice("http".length);
+    }
+
+    return url;
+  }
+
+  async retryBackoff(fn: () => Promise<void>) {
+    for (let attempt = 0; attempt < this.retry.maxAttempts; attempt++) {
+      try {
+        await fn();
+        return;
+      } catch {
+        if (attempt === this.retry.maxAttempts - 1) {
+          throw new Error("Max attempts reached");
+        }
+
+        // Exponential backoff with jitter.
+        const delay =
+          Math.min(
+            this.retry.initialDelay * this.retry.backoff ** attempt,
+            this.retry.maxDelay,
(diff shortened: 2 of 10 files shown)


dashboard/src/client.ts
```diff
@@ -48,6 +48,18 @@ function connectIngestMonitor() {
       handleIngestEvent(data.ingestValue);
     }
   };
+  socket.onclose = () => {
+    console.error("Ingest monitor socket closed");
+    setTimeout(() => {
+      console.log("Reconnecting ingest monitor socket...");
+      client.ingest.ingestMonitorRequest().then(() => {
```
jraddaoui: I'd consider putting this initial request inside the `connectIngestMonitor` function directly, and the same for storage. It would reduce some duplication, leaving the calls in App.vue just with `client.connectIngestMonitor();` and `client.connectStorageMonitor();` inside the if, and similar here on close.

https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/App.vue#L19-L24

jraddaoui (approved):
LGTM @djjuhasz, thanks for looking into this! Just a minor suggestion to reduce duplication.

dashboard/src/client.ts
djjuhasz: @jraddaoui I don't understand what you mean.  The `connectIngestMonitor()` method is autogenerated, I'm not sure how we would modify the function put the `ingestMonitorRequest()` call "inside" it. :confused:

djjuhasz: Oh, I see. `connectIngestMonitor()` isn't auto-generated, `ingestMonitorRequest()` is auto-generated.

djjuhasz:
@jraddaoui I totally rewrote the monitor reconnection logic to include exponential backoff with jitter, and an attempt limit.  I've also moved the ingest and monitor state to Pinia to preserve the WebSocket connections through page updates.  I've also added tests for all of the reconnection logic.

dashboard/src/monitor.ts
```diff
@@ -0,0 +1,180 @@
...
+};
+
+export interface MonitorConnector {
+  dial(): Promise<void>;
+  isConnected(): boolean;
+  getWebSocketURL(url: string): string;
+  close(): void;
+}
+
+export class MonitorConnection {
```
jraddaoui: I'd make this an abstract class with an abstract dial method and implement `MonitorConnector` directly. That will force all subclasses to implement `MonitorConnector` even if it's forgotten and it will reduce some minor duplication (see below).

```ts
export abstract class MonitorConnection implements MonitorConnector {
  // Existing code.
  
  abstract dial(): Promise<void>;
}
```

dashboard/src/monitor.ts
```diff
@@ -0,0 +1,180 @@
...
+  async connectSocket(url: string): Promise<void> {
+    this.socket = new WebSocket(url);
+
+    this.socket.onopen = () => {
+      console.log(`${this.type} monitor socket connected`);
+    };
+    this.socket.onerror = () => {
+      console.error(`${this.type} monitor socket error`);
+    };
+  }
```
jraddaoui: With the abstract dial method you could move the `onclose` set up here too:

```ts
connectSocket(url: string): void {
  this.socket = new WebSocket(url);

  this.socket.onopen = () => {
    console.log(`${this.type} monitor socket connected`);
  };
  this.socket.onerror = () => {
    console.error(`${this.type} monitor socket error`);
  };

  // Reconnect on close.
  this.socket.onclose = () => {
    console.error(`${this.type} monitor socket closed, reconnecting...`);
    this.dial();
  };
}
```

dashboard/src/monitor.ts
```diff
@@ -0,0 +1,180 @@
...
+}
+
+export class IngestMonitorConnection
+  extends MonitorConnection
+  implements MonitorConnector
+{
```
jraddaoui: This will only need to extend MonitorConnection:

```ts
export class IngestMonitorConnection extends MonitorConnection {}
```

dashboard/src/monitor.ts
```diff
@@ -0,0 +1,180 @@
...
+      return client.ingest
+        .ingestMonitorRequest()
+        .then(async () => {
+          return this.connectSocket(this.url).then(() => {
+            if (this.socket) {
+              // Reconnect on close.
+              this.socket.onclose = () => {
+                console.error("ingest monitor socket closed, reconnecting...");
+                this.dial();
+              };
```
jraddaoui: And this could be removed.

dashboard/src/monitor.ts
```diff
@@ -0,0 +1,180 @@
...
+          throw err;
+        });
+    });
+  }
+}
+
+export class StorageMonitorConnection
+  extends MonitorConnection
+  implements MonitorConnector
+{
```
jraddaoui: Same.

dashboard/src/monitor.ts
```diff
@@ -0,0 +1,180 @@
...
+      return client.storage
+        .storageMonitorRequest()
+        .then(async () => {
+          return this.connectSocket(this.url).then(() => {
+            if (this.socket) {
+              // Reconnect on close.
+              this.socket.onclose = () => {
+                console.error("storage monitor socket closed, reconnecting...");
+                this.dial();
+              };
```
jraddaoui: Same.

dashboard/src/monitor.ts
```diff
@@ -0,0 +1,180 @@
...
+  maxAttempts: number;
+  jitterFn: () => number;
+};
+
+export interface MonitorConnector {
+  dial(): Promise<void>;
+  isConnected(): boolean;
+  getWebSocketURL(url: string): string;
+  close(): void;
+}
```
jraddaoui: Actually, with the abstract class I think we could remove this interface entirely, the `MonitorConnection` class will define the contract.

dashboard/src/monitor.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  close(): void {
+    if (this.socket) {
+      this.socket.onclose = null; // Disable reconnection on close.
+      this.socket.close();
+      this.socket = null;
+    }
+  }
+
+  async connectSocket(url: string): Promise<void> {
```
jraddaoui: Why is this method async and returning a promise?

```ts
connectSocket(url: string): void {}
```

That would work, right? Then we'll need to update the dial calls:

```ts
this.connectSocket(this.url);
if (this.socket) {}
```

jraddaoui (changes requested):
Thanks @djjuhasz, looks great to me, nice work!

We could simplify a little the monitor classes by using an abstract class, but it's not a huge deal.

dashboard/src/monitor.ts
djjuhasz: I assumed `new Websocket()` would connect asynchronously and return a promise, but it looks like creating the socket is blocking (and pretty much instantaneous), then the `onopen` handler fires asynchronously when the connection is established. :thinking:

When looking this up I found a great example of a web socket client class: https://websocket.org/guides/languages/javascript/.  I wish I'd found that before trying to write this whole class myself. :facepalm:

djjuhasz:
@jraddaoui I think I've addressed all your feedback and this is ready for another look.  I also stopped exporting the "MonitorConnect" class, and changed the `isConnected()` method to an `isConnected` property, following the example of: https://websocket.org/guides/languages/javascript/

jraddaoui (approved):
Thanks @djjuhasz!

Issue #1355, opened 2025-09-11
Feature: automatically remove SIP once AIP has been created


sallain:
**Is your feature request related to a problem? Please describe.**

It is cumbersome to manage transfer storage, especially in high-volume instances. In order to effectively manage both archival workflows and storage, a user must manually monitor that a transfer package has successfully passed through the workflow and then delete it manually.

**Describe the solution you'd like**

We need to do analysis! 

**Describe alternatives you've considered**

**Additional context**

Note that, at this time, there is a one-to-one relationship between a SIP and an AIP. It's likely that this will change in the future - we may, for example, someday implement batch management that results in one SIP splitting into multiple AIPs (and this is already the case for other users we know of). If it's possible to design this work to accomodate such a scenario, then that's great; however, it shouldn't be a blocker.

jraddaoui:
There are some pieces from the previous Enduro version that relate to this, retention period and completed directory. Not saying that we must follow this approach, but I'll try to explain what we have already in code so we can consider it in the analysis.

Although it's not included in the [default config](https://github.com/artefactual-sdps/enduro/blob/main/enduro.toml#L118-L134), you can currently configure a retention period in the [watched buckets](https://github.com/artefactual-sdps/enduro/blob/main/internal/watcher/config.go#L121). Also, we have not tested filesystem watchers in this version of Enduro, but you can set a retention period alongside a completed directory in [their config](https://github.com/artefactual-sdps/enduro/blob/main/internal/watcher/config.go#L80-L96).

The workflow request made from the watchers [will include both values](https://github.com/artefactual-sdps/enduro/blob/main/cmd/enduro/main.go#L443-L444) making them available in workflow. Then, at the end of the workflow between the session and the final clean up, and only if the workflow status is done, [they are considered](https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/processing.go#L251-L278). If both are set the retention period will take precedence, making the workflow wait for that time before deleting the original SIP. If only the completed directory is set, it will move the original SIP to that path without waiting.

[...]

PR #1376, opened 2025-09-29
Add retention period to all SIP origin configs 

jraddaoui:
Allow setting a retention period to the internal uploads and the SIP
sources, and update the current behavior in watched buckets for
consistency. The retention period will default to "0" (automatic 
deletion) and has to be set to a negative value to disable the deletion
of the original SIP after a successful ingest. Record a workflow task to
show details in the API/dashboard.

Refs #1355.


jraddaoui:
Still working on the tests and the documentation, but this is ready for an initial review when you have some time, @djjuhasz!

cmd/enduro-a3m-worker/main.go
```diff
@@ -247,7 +247,8 @@ func main() {
 		)
 		// TODO: At some point there may be multiple SIP sources, this activity should
 		// work similar to the watched bucket download activity and use the source ID
-		// to determine which bucket to use.
+		// to determine which bucket to use. Alternatively, we could register multiple
+		// copies of this activity, one per source.
```
djjuhasz, line 251: How would registering the activity multiple times work? Is it enough to register the activity with a different `Name` each time? :thinking:

enduro.toml
```diff
@@ -132,6 +132,11 @@ bucket = "sips"
 # the later only works properly when the preservation system is "a3m".
 # Default: "create aip".
 workflowType = "create aip"
+# retentionPeriod is the duration to retain SIPs before they are deleted after a
+# successful ingest. Do not set to disable automatic deletion of SIPs. Set to "0"
+# to delete uploaded SIPs immediately after they have been ingested.
+# It must be a string format compatible with https://pkg.go.dev/time#ParseDuration.
+retentionPeriod = "0"
```
djjuhasz, line 139: I think it would be better to use a negative duration for indefinite storage so a zero or positive value in this config file can be overridden by an environment variable that disables deletion (e.g. `ENDURO_WATCHER_EMBEDDED_RETENTIONPERIOD=-1s`).

internal/workflow/delete_original.go
```diff
@@ -0,0 +1,92 @@
...
+	if state.req.RetentionPeriod == nil {
+		return nil
+	}
+
+	// Create a "delete original SIP" task.
+	id, err := w.createTask(
+		ctx,
+		&datatypes.Task{
+			Name:         "Delete original SIP",
+			Note:         fmt.Sprintf("The original SIP will be deleted in %s", state.req.RetentionPeriod.String()),
```
djjuhasz, line 27: It might be more useful to report the time at which the SIP will be deleted, rather than the duration.

internal/workflow/delete_original.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Set the default (successful) delete original SIP task completion values.
+	task := datatypes.Task{
+		ID:     id,
+		Status: enums.TaskStatusDone,
+		Note:   "SIP successfully deleted",
+	}
+
+	// Set a timer for the retention period.
+	if err := temporalsdk_workflow.NewTimer(ctx, *state.req.RetentionPeriod).Get(ctx, nil); err != nil {
```
djjuhasz: I think you can use `temporalsdk_workflow.Sleep(ctx, *state.req.RetentionPeriod)` to block until the timer is complete, `NewTimer()` looks to me like it's meant more for asynchronous use.

internal/workflow/processing.go
```diff
@@ -246,31 +246,23 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
...
-					activities.DisposeOriginalActivityName,
-					req.WatcherName,
-					req.CompletedDir,
-					req.Key,
-				).Get(activityOpts, nil)
+	// Schedule deletion or disposal of the original SIP.
+	if state.status == enums.WorkflowStatusDone {
+		if req.RetentionPeriod != nil {
+			if err := w.deleteOriginalSIP(ctx, state); err != nil {
+				w.logger.Error("Failed to delete original SIP", "err", err.Error())
```
djjuhasz: The error returned from `w.deleteOriginalSIP()` includes "delete original SIP" in the message too, I'd recommend returning the deletion error without the prefix text from `w.deleteOriginalSIP()`.

internal/workflow/processing.go
```diff
@@ -246,31 +246,23 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
...
-				activityOpts := withActivityOptsForLocalAction(ctx)
-				_ = temporalsdk_workflow.ExecuteActivity(
-					activityOpts,
-					activities.DisposeOriginalActivityName,
-					req.WatcherName,
-					req.CompletedDir,
-					req.Key,
-				).Get(activityOpts, nil)
+	// Schedule deletion or disposal of the original SIP.
+	if state.status == enums.WorkflowStatusDone {
```
djjuhasz: This multi-nested conditional is hard to understand.  Can we just return here if the workflow status is not done? e.g. 

```
if state.status != enums.WorkflowStatusDone {
    return nil
}
```

The only thing we'd be skipping that runs now is the final "Workflow completed successfully!" log message, but I don't think that makes sense for a status other than "done".  I'm not sure we can even reach this point in the workflow without `state.status == enums.WorkflowStatusDone`. :thinking:

internal/workflow/processing.go
```diff
@@ -246,31 +246,23 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
...
+		if req.RetentionPeriod != nil {
+			if err := w.deleteOriginalSIP(ctx, state); err != nil {
+				w.logger.Error("Failed to delete original SIP", "err", err.Error())
+			}
+		} else if req.CompletedDir != "" {
```
djjuhasz: Is `req.CompletedDir` really mutually exclusive from having a retention period? There's no note in the `ProcessingWorkflowRequest` field comments to indicate that you can't set both `RetentionPeriod` and `CompletedDir`.  Maybe the SIP should be moved to the completed dir (if set), and then the retention period should be applied regardless of whether the SIP is in the completed dir or the original dir?

djjuhasz: Ah, I've traced the `ProcessingWorkflowRequest.CompletedDir` back to https://github.com/artefactual-sdps/enduro/blob/main/internal/watcher/config.go#L83 and it is checked for mutually exclusivity with `RetentionPeriod` at https://github.com/artefactual-sdps/enduro/blob/main/internal/watcher/filesystem.go#L57.

It's really hard to trace the path from `req.CompletedDir` back to `watcher.FilesystemConfig` though, and I think it's a mistake to be passing `req.CompletedDir` when it's an filesystem watcher implementation specific field that we aren't using anywhere else.  Maybe it was necessary to pass `CompletedDir` in the past because we didn't have access to the watcher config directly, but now we have access to it both through the watcher service (https://github.com/artefactual-sdps/enduro/blob/main/internal/watcher/watcher.go#L43) AND the workflow config (https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/processing.go#L46). :(

enduro.toml
```diff
@@ -243,6 +248,11 @@ passphrase = ""
 # maxSize is the maximum upload size allowed by the server in bytes.
 # Default: 4294967296 (4 GiB).
 maxSize = 4294967296
+# retentionPeriod is the duration to retain SIPs before they are deleted after a
+# successful ingest. Do not set to disable automatic deletion of SIPs. Set to "0"
+# to delete uploaded SIPs immediately after they have been ingested.
+# It must be a string format compatible with https://pkg.go.dev/time#ParseDuration.
+retentionPeriod = "0"
```
djjuhasz, line 255: I don't think it's necessary or even desirable to have a retention period for uploaded SIPs — I don't think there is any way for a system operator to access the SIP in the internal storage location, so I don't think there's a reason to keep it after processing is completed.  Because the SIPs in internal storage can't be seen through the UI, any SIPs stored in internal storage end up being "hidden" SIPs that just take up space.

internal/ingest/goa.go
```diff
@@ -96,11 +96,12 @@ func (w *goaWrapper) AddSip(ctx context.Context, payload *goaingest.AddSipPayloa
...
-		SIPSourceID: sourceID,
-		SIPName:     s.Name,
-		Type:        enums.WorkflowTypeCreateAip,
-		Key:         payload.Key,
+		SIPUUID:         s.UUID,
+		SIPSourceID:     sourceID,
+		SIPName:         s.Name,
+		Type:            enums.WorkflowTypeCreateAip,
+		Key:             payload.Key,
+		RetentionPeriod: w.sipSource.RetentionPeriod(),
```
djjuhasz, line 104: It seems very awkward to me how we are passing the `RetentionPeriod` to `InitProcessingWorkflow` even though we already have access to the configured value through the `ProcessingWorkflow.cfg`.  I think this is a result of how watchers are handled, where each watcher calls `InitProcessingWorkflow` with it's specific `RetentionPeriod`, but it feels like we are duplicating the information unnecessarily.

I don't think it's feasible in this PR to change how we manage and share configuration data from all the various places a SIP can be ingested, but I think it's something we should consider improving in the future.

djjuhasz (commented):
@jraddaoui looks good, but I had a few questions, a few recommendations and I am just finding it confusing how the relationships between the various SIP ingest locations and how we are starting the ingest workflow are managed.

cmd/enduro-a3m-worker/main.go
jraddaoui: Yeah, we already do that:

https://github.com/artefactual-sdps/enduro/blob/main/cmd/enduro-am-worker/main.go#L258-L265

It's the same activity from temporal-activities, we just register them with different names and buckets.

internal/workflow/delete_original.go
jraddaoui: But being part of the note field it's not going to work across different timezones. I thought we displayed the start time in the tasks while they are in progress, but I don't see that now in the dashboard. I think that may be a better solution.

internal/workflow/delete_original.go
jraddaoui: Looks like Sleep uses the timer internally, so just a little cleaner for us.

https://github.com/temporalio/sdk-go/blob/master/internal/workflow.go#L1556-L1574

internal/workflow/processing.go
jraddaoui: Okay, done.

enduro.toml
jraddaoui: It was already like this and that's quite an edge case, but we can change it.

enduro.toml
jraddaoui: Good point. Also, the behavior for the failed SIPs is different already for internal uploads, instead of copying the SIP as failed SIP in the internal bucket we "rename" it to avoid duplication. Following what @djjuhasz said, we could also delete the original in case of failure. However, we never experienced with workflow retries where it may be good to keep a copy to avoid the re-upload.

In any case, I was just following the [issue notes](https://github.com/artefactual-sdps/enduro/issues/1355#issuecomment-3309310033), so I'd love to hear from @fiver-watson before changing this.

internal/ingest/goa.go
jraddaoui: Sounds good, I think we didn't pass the entire config to the workflow initially. And I think in some cases it may be better to just pass the value than looking for it in workflow.

enduro.toml
jraddaoui: I like the idea of being unset means no deletion, in case somebody forgets it. We could keep both options if you want ("Do not set or use a negative value to disable automatic deletion of SIPs.") but I think it's more confusing just to support environment overrides. I'd keep it as it is unless you feel strongly about it.

enduro.toml
fiver-watson: **Team discussion outcome, 2025-10-01**: 

* Aim for consistency - leave retention periods for all upload methods
* Set default behavior when no retention period is configured to be same as "0" - i.e. delete immediately upon success
* In our default config, keep retention period for internal bucket to 0

In the docs, we will add a big warning about setting a retention period for the internal upload space, mentioning that it is also the processing space, and if full, nothing can be ingested.

jraddaoui:
@djjuhasz, I think I have addressed all you feedback, including the default value change we discussed with the team. Still lacks documentation, and the workflow tests I added are based on https://github.com/artefactual-sdps/enduro/pull/1381, so we'll to merge that one first. But this is ready for another look when you have some time.

enduro.toml
djjuhasz: Defaulting to keeping SIPs forever makes sense to me.  We might want to change the comment wording a bit to make it very clear that the `retentionPeriod` setting should be commented out or removed entirely to disable automatic deletion.

internal/workflow/delete_original.go
djjuhasz: Ah, good point about the timezone problems with displaying the deletion time. I agree that it would be helpful to show the start time for a task in progress.

enduro.toml
djjuhasz: Oh okay, I'd forgotten we decided to make the default behaviour deleting the SIP immediately.  Ignore the above.

djjuhasz (approved):
LGTM 👍

jraddaoui:
Thanks @djjuhasz! But we need to merge https://github.com/artefactual-sdps/enduro/pull/1381 first and update the tests here to use the changes we made there (`defaultParams()`, etc). That one is still waiting for another look ;)

jraddaoui:
Thanks again @djjuhasz, I'll merge this one and add the documentation in another PR.

PR #1381, opened 2025-10-03
Refactor processing workflow tests

jraddaoui:
Split large test file into three and introduce a reusable expectation system to reduce code duplication, improve reusability, and enable easier extension. The expectation functions provide parameterized mock setups for common workflow activities.

- Extract expectation functions for reusable activity mocks.
- Centralize test constants and variables in expectations file.
- Separate test setup logic into dedicated setup file.
- Add ExecuteAndValidateWorkflow helper.

### DIFF
--- a/internal/workflow/processing_expectations_test.go
+++ b/internal/workflow/processing_expectations_test.go
@@ -1,0 +1,479 @@
+package workflow
+
+import (
+	"path/filepath"
+	"time"
+
+	"github.com/artefactual-sdps/temporal-activities/archiveextract"
+	"github.com/artefactual-sdps/temporal-activities/archivezip"
+	"github.com/artefactual-sdps/temporal-activities/bagcreate"
+	"github.com/artefactual-sdps/temporal-activities/bagvalidate"
+	"github.com/artefactual-sdps/temporal-activities/bucketupload"
+	"github.com/artefactual-sdps/temporal-activities/removepaths"
+	"github.com/artefactual-sdps/temporal-activities/xmlvalidate"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/mock"
+	"go.artefactual.dev/tools/ref"
+
+	"github.com/artefactual-sdps/enduro/internal/a3m"
+	"github.com/artefactual-sdps/enduro/internal/datatypes"
+	"github.com/artefactual-sdps/enduro/internal/enums"
+	"github.com/artefactual-sdps/enduro/internal/workflow/activities"
+)
+
+const (
+	tempPath         = "/tmp/enduro123456"
+	extractPath      = "/tmp/enduro123456/extract"
+	transferPath     = "/home/a3m/.local/share/a3m/share/enduro2985726865"
+	prepSharedPath   = "/home/enduro/preprocessing/"
+	prepDownloadPath = "/home/enduro/preprocessing/enduro123456"
+	prepExtractPath  = "/home/enduro/preprocessing/enduro123456/extract"
+	failedSIPKey     = "Failed_SIP_name-e2ace0da-8697-453d-9ea1-4c9b62309e54.zip"
+	failedPIPKey     = "Failed_PIP_name-e2ace0da-8697-453d-9ea1-4c9b62309e54.zip"
+	sipID            = 1
+	workflowID       = 1
+	copySIPTaskID    = 100
+	valBagTaskID     = 101
+	valPREMISTaskID  = 102
+	uploadTaskID     = 103
+	reviewAIPTaskID  = 104
+	moveAIPTaskID    = 105
+	sipName          = "name.zip"
+	key              = "transfer.zip"
+	watcherName      = "watcher"
+)
+
+var (
+	ctx             = mock.AnythingOfType("*context.valueCtx")
+	sessionCtx      = mock.AnythingOfType("*context.timerCtx")
+	internalCtx     = mock.AnythingOfType("*internal.valueCtx")
+	sipUUID         = uuid.MustParse("e2ace0da-8697-453d-9ea1-4c9b62309e54")
+	locationID      = uuid.MustParse("f2cc963f-c14d-4eaa-b950-bd207189a1f1")
+	amssLocationID  = uuid.MustParse("e0ed8b2a-8ae2-4546-b5d8-f0090919df04")
+	transferID      = uuid.MustParse("65233405-771e-4f7e-b2d9-b08439570ba2")
+	aipUUID         = uuid.MustParse("9e8161cc-2815-4d6f-8a75-f003c41b257b")
+	workflowUUID    = uuid.MustParse("8fdfaea1-06ed-4cf6-8bdf-d15d80420f35")
+	startTime       = time.Date(2024, 7, 9, 16, 55, 13, 50, time.UTC)
+	retentionPeriod = ref.New(time.Second)
+)
+
+// expectationParams holds configurable parameters for setting up activity expectations.
+type expectationParams struct {
+	// SIP type used for classification.
+	sipType enums.SIPType
+
+	// SIP status for status updates.
+	sipStatus enums.SIPStatus
+
+	// Workflow type for workflow creation.
+	workflowType enums.WorkflowType
+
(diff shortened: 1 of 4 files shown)


jraddaoui:
Wait for it, I thought it would be better to ask you from some feedback at this point, but there are only a few remaining things to do, so I'll finish and explain it all at once. Sorry for the noise.

jraddaoui:
Ready for review when you have some time @djjuhasz. I think it makes no sense looking at the diff, I'd check the three test files individually. In the setup file I have only moved the existing functions and added a helper to execute and validate the workflow. In the other two is where most of the changes are happening, preprocessing_test now only contains the test cases, and preprocessing_expectations_test includes the new expectations and variables.

I've moved all the variables we had duplicated on each test alongside the other global vars/const and added expectation functions for those that were used more than once across all tests. I've used Copilot with Claude Sonnet 4 in agent mode in a few point in this changes, it was really helpful identifying those duplicated vars/expectations and adding comments but not so helpful with the parameterization and other aspects.

This was the length of the current test cases:

- TestConfirmation: 244 lines
- TestAutoApprovedAIP: 253 lines
- TestAMWorkflow: 249 lines
- TestRejection: 176 lines
- TestChildWorkflows: 343 lines
- TestFailedSIP: 178 lines
- TestFailedPIPA3m: 214 lines
- TestFailedPIPAM: 168 lines
- TestInternalUpload: 146 lines

And a summary:

- Shortest test: TestInternalUpload (146 lines)
- Longest test: TestChildWorkflows (343 lines)
- Average length: ~219 lines
- Total test lines: ~1,971 lines (excluding helper functions and imports)

Compared with these changes:

- TestConfirmation: 73 lines
- TestRejection: 67 lines
- TestAutoApprovedAIP: 58 lines
- TestAMWorkflow: 88 lines
- TestChildWorkflows: 141 lines
- TestFailedSIP: 56 lines
- TestFailedPIPA3m: 61 lines
- TestFailedPIPAM: 45 lines
- TestInternalUpload: 68 lines

Summary

- Shortest test: TestFailedPIPAM (45 lines)
- Longest test: TestChildWorkflows (141 lines)
- Average length: ~73 lines
- Total test lines: 657 lines (excluding helper functions and imports)

Hopefully this makes adding new tests a little less overwhelming.

internal/workflow/processing_expectations_test.go
```diff
@@ -0,0 +1,501 @@
...
+// expectationFunc represents a function that sets up activity expectations for testing.
+// Functions configure mock activities using s.env.OnActivity() calls with the provided
+// test suite and parameters.
+type expectationFunc func(s *ProcessingWorkflowTestSuite, params expectationParams)
+
+// expectations is a map of activity names to their expectation functions.
+// Each key represents a logical activity name that corresponds to a Temporal activity.
+// Functions use global constants and variables for consistent test data, with customization
+// through expectationParams.
+var expectations = map[string]expectationFunc{
```
djjuhasz, line 142: I think it would be better to create a copy of the `expectations` map in each test to allow mutating the expectations for a specific test without changing the outcome of other test and to avoid contention problems when mutating the map.

jraddaoui:
Added an extra commit to include a couple of expectation wrappers for a3m and added comments to the tests to specify what's being tested. A couple of new tests added using this expectations can be seen here: https://github.com/artefactual-sdps/enduro/pull/1376/commits/0c6990c54009ce75a073882520169af4c6a83087.

jraddaoui:
With the latest changes, including the tests mentioned above:

- TestConfirmation: 59 lines
- TestRejection: 52 lines
- TestAutoApprovedAIP: 29 lines
- TestAMWorkflow: 96 lines
- TestChildWorkflows: 126 lines
- TestFailedSIP: 63 lines
- TestFailedPIPA3m: 77 lines
- TestFailedPIPAM: 50 lines
- TestInternalUpload: 65 lines
- TestInternalUploadError: 74 lines
- TestSIPSourceUpload: 59 lines

Summary:

- Shortest test: TestAutoApprovedAIP (29 lines)
- Longest test: TestChildWorkflows (126 lines)
- Average length: ~68 lines
- Total test lines: ~750 lines (excluding helper functions and imports)

internal/workflow/processing_expectations_test.go
```diff
@@ -0,0 +1,501 @@
...
+
+	// Workflow status for workflow status updates.
+	// Used in "setWorkflowStatus" expectation.
+	// Default: enums.WorkflowStatusPending.
+	workflowStatus enums.WorkflowStatus
+
+	// Task identifier for task operations.
+	// Used in "createTask" and "completeTask" expectations.
+	// Required.
+	taskID int
```
djjuhasz, line 75: Having a parameter that is required but only for a couple of expectations seems like it could lead to errors and be tricky to maintain (e.g. changing which expectations use the param will require updates in `preprocessing_test.go` and the comment here).  I think it would be better to use a constructor of some sort that sets a default value for `taskID` and other params where the "zero" value doesn't make sense as a default.

internal/workflow/processing_expectations_test.go
djjuhasz: Huh, it looks like testify doesn't support running tests in parallel (https://github.com/stretchr/testify?tab=readme-ov-file#suite-package), so I guess there's no risk of a simultaneous read and write to the map. :thinking:

djjuhasz (changes requested):
Very nice @jraddaoui! I made a couple of notes, but my main request is to make the expectations map a local variable instead of a global.

internal/workflow/processing_expectations_test.go
jraddaoui: But why would you want to mutate those expectations in test? The idea was to use them as they are with some minor modification through parameters, if you need a different expectation you can use it as before in the tests. For example for the download, I added the most common download activity call, other downloads just use their own expectation in test:

https://github.com/artefactual-sdps/enduro/blob/dev/processing-tests/internal/workflow/processing_expectations_test.go#L259-L277
https://github.com/artefactual-sdps/enduro/blob/dev/processing-tests/internal/workflow/processing_test.go#L609-L616

internal/workflow/processing_expectations_test.go
jraddaoui: This particular field could even be removed and use always 0 as the task ID because it's used as return param and then as the input, but this is closer to reality where all tasks would get a different ID. But I'm not sure I understand your suggestion, each call to an expectation uses a new params instance, where would you add a constructor?

> (e.g. changing which expectations use the param will require updates in preprocessing_test.go and the comment here)

I think this will happen more than once and I don't see an issue with it. We'll probably add more expectations and parameters at some point.

jraddaoui:
Thanks @djjuhasz!

internal/workflow/processing_expectations_test.go
djjuhasz: I was thinking a dev would modify the expectations to change the expected outcome of an activity - to make an activity fail for instance.  You're right that you can just write out the full activity mock like we did previously to change the expectations but it was not clear to me from the expectation structure or comments that it's meant to be a read-only data structure. There's also no protection in place to discourage a dev from mutating the map.

internal/workflow/processing_expectations_test.go
djjuhasz: Because the `expectationParams` is a single struct I would expect to instantiate it once with all of the parameters needed for a test, and then reuse it for each activity.  If the parameters are meant to be specific to an activity, then I would expect each activity to have it's own parameters struct.

jraddaoui: What would you prefer?

djjuhasz: I'd prefer it being a single struct (as now) but having a `defaultParams()` constructor that returns a new instance with defaults for all the fields.

jraddaoui: Lets give it a try ;)

jraddaoui: Thanks @djjuhasz! I like the idea, let me know how it looks to you.

djjuhasz (approved):
LGTM :+1:

Issue #1373, opened 2025-09-25
Problem: Source location does not provide enough context to easily select correct SIP


fiver-watson:
**Describe the bug**

Currently the SIP source location UI in Enduro displays only the name of the files / SIPs included there, and displays the contents of the related source location in a seemingly random order. This makes it harder for operators to find the right package when there is a common naming convention in place. 

**To Reproduce**

1. Log in
2. Navigate to Upload SIPs and click on the source location tab
3. View the included SIPs

**Expected behavior**

Ideally: 

* Include the size of the SIP/file
* Include the date of upload/deposit
* Sort the results from most recent to oldest deposit, to make it easier for users to immediately find a SIP they just deposited

**Screenshots**

An initial mockup of the additional metadata fields was provided by David while investigating: 

<img width="1358" height="518" alt="Image" src="https://github.com/user-attachments/assets/b5b1fa76-39dd-405a-a31a-7afd6b414575" />

djjuhasz:
@fiver-watson it turns out that the underlying [Bucket.ListPage()](https://pkg.go.dev/gocloud.dev/blob#example-Bucket.ListPage) method only returns the list of bucket objects in "lexicographical order of UTF-8 encoded keys" (the key being the "name" of the file).  

Because the object list is paged, to sort on modification date/time we'd have to page through all of the objects in a bucket and then re-sort the list by date before displaying the list.  Loading the entire object list and then re-sorting it would be okay for a bucket containing a moderate number of objects (e.g. thousands, tens of thousands) but could use an unacceptable amount of CPU time and RAM to sort a very large number of of objects.

fiver-watson:
Interesting...

I'd be curious to hear from other team members on this (@jraddaoui @sallain @joel-simpson ). 

Personally, I suspect that thousands of packages in a source location is unlikely, particularly after we add the retention period? Like, if your volume is going to be high enough to have thousands of packages coming in over a short period of time, you are more likely to use a watched location to automate the ingest process and use a watched location for deposit instead, so there is no need to manually start the ingest.

If we believe that thousands  of SIPs wouldn't cause too big a perfomance hit, and that even tens of thousands might be okay, then I personally think this is a fine threshold, and we could proceed and put a warning in the docs about potential performance hits for any number of SIPs exceeding those rough amounts. 

[...]

PR #1385, opened 2025-10-08
Add file size and deposit date to SIP source list

djjuhasz:
Refs #1373.

Add the SIP file size and deposit timestamp to the Dashboard SIP source file list.

### DIFF
--- a/dashboard/src/composables/format.ts
+++ b/dashboard/src/composables/format.ts
@@ -16,3 +16,14 @@ export function FormatDuration(from: Date, to: Date) {
   const diff = moment(to).diff(from);
   return moment.duration(diff).humanize();
 }
+
+export function humanFileSize(bytes: number, precision = 0): string {
+  const base = 1000;
+  const units = ["bytes", "KB", "MB", "GB", "TB"];
+  let i = 0;
+  while (bytes >= base && i < units.length - 1) {
+    bytes /= base;
+    i++;
+  }
+  return `${bytes.toFixed(precision)} ${units[i]}`;
+}
--- a/dashboard/src/components/SIPUploadSource.vue
+++ b/dashboard/src/components/SIPUploadSource.vue
@@ -4,6 +4,7 @@ import { ref } from "vue";
 import { useRouter } from "vue-router/auto";
 
 import { api, client } from "@/client";
+import { humanFileSize } from "@/composables/format";
 import IconBundle from "~icons/clarity/bundle-line";
 
 const router = useRouter();
@@ -12,13 +13,13 @@ const router = useRouter();
 // TODO: Fetch SIP sources from the backend.
 const sourceId = "e6ddb29a-66d1-480e-82eb-fcfef1c825c5";
 const items = ref<api.EnduroIngestSipsourceObject[]>([]);
-const checkedItems = ref<string[]>([]);
+const selectedSips = ref<string[]>([]);
 const nextCursor = ref<string | undefined>(undefined);
 const listContainer = ref<HTMLElement>();
 const errorMessage = ref<string | null>(null);
 
 const { execute, isLoading } = useAsyncState(
-  async (cursor: string | undefined) => {
+  async (cursor?: string | undefined) => {
     // Clear previous error.
     errorMessage.value = null;
     try {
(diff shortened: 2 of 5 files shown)


djjuhasz:
Screenshot: image

djjuhasz:
Sorting by deposit date (most recent first) will be added in a subsequent PR.

jraddaoui (commented):
Thanks @djjuhasz! A few suggestions to improve the layout:

- If the deposited date has a fixed width, we could move the size column to sit just before the date. That should keep sizes aligned vertically.
- Responsive tweaks:
  - On small screens, split each row into two lines.
  - Add text ellipsis to the name to prevent overlaps or layout shifts before the split.

djjuhasz:
Here's how it looks with the filesize right justified:
image

I don't think it looks very good with the filesize "column" having no label, and the repetition of the "Deposited" label seems excessive.  I think if we can make an HTML table with proper rows and columns (with headers) that is responsive that could look better and use the space more effectively.  

There's a few sites suggesting ways to make responsive tables, e.g. 
- https://getbootstrap.com/docs/5.3/content/tables/#responsive-tables
- https://css-tricks.com/responsive-data-tables/
- https://code-boxx.com/css-responsive-table/

djjuhasz:
Here's what it looks like as a responsive table:
[Screencast from 2025-10-14 16-05-11.webm](https://github.com/user-attachments/assets/1bc39852-6748-421a-affa-f6d504396f03)

Note that I'm hiding the "size" and "deposited" (date) columns at the smallest Bootstrap screen width breakpoint (<576 pixels wide).

djjuhasz:
Here's a screenshot of the responsive table version, for easy reference:
image

djjuhasz:
More compact version of the table (using `table-sm`):
image

dashboard/src/composables/format.ts
```diff
@@ -0,0 +1,34 @@
+import moment from "moment";
+
+export function FormatDateTime(value: Date | undefined) {
```
sevein, line 3: I think that all these functions that you're exporting should use camelCase.
PascalCase is used for classes, components, types...

dashboard/src/composables/format.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (!value || Number.isNaN(value.getTime())) {
+    return "";
+  }
+  return moment(value).format("YYYY-MM-DD HH:mm:ss");
+}
```
sevein: Here's a polished version:

```suggestion
export function formatDateTime(value?: Date): string {
  if (!(value instanceof Date) || isNaN(value.getTime())) {
    return "";
  }
  return moment(value).format("YYYY-MM-DD HH:mm:ss");
}
```

dashboard/src/composables/format.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function FormatDateTimeString(value: string) {
+  const date = new Date(value);
+  return moment(date).format("YYYY-MM-DD HH:mm:ss");
+}
+
+export function FormatDuration(from: Date, to: Date) {
```
sevein: You could also define the return type here (string).

dashboard/src/composables/format.ts
```diff
@@ -0,0 +1,34 @@
+import moment from "moment";
+
+export function FormatDateTime(value: Date | undefined) {
+  if (!value || Number.isNaN(value.getTime())) {
+    return "";
+  }
+  return moment(value).format("YYYY-MM-DD HH:mm:ss");
```
sevein: This could be a const.

dashboard/src/composables/format.ts
```diff
@@ -0,0 +1,34 @@
...
+  } else if (bytes >= base ** 3) {
+    return `${(bytes / base ** 3).toFixed(precision)} GB`;
+  } else if (bytes >= base ** 2) {
+    return `${(bytes / base ** 2).toFixed(precision)} MB`;
+  } else if (bytes >= base) {
+    return `${(bytes / base).toFixed(precision)} KB`;
+  } else {
+    return `${bytes} bytes`;
+  }
+}
```
sevein: Here's perhaps a more readable alternative:

```suggestion
export function humanFileSize(bytes: number, precision = 0): string {
  const base = 1000;
  const units = ["bytes", "KB", "MB", "GB", "TB"];
  let i = 0;

  while (bytes >= base && i < units.length - 1) {
    bytes /= base;
    i++;
  }

  return `${bytes.toFixed(precision)} ${units[i]}`;
}
```

sevein (approved):
👍

sevein (commented):
When rows are highlighted, the striped background pattern becomes visually confusing, resulting in up to four different row colors on screen. Highlighted rows should probably not inherit the striped style.

dashboard/src/composables/format.ts
djjuhasz: @sevein are you referencing a particular Typescript style guide? I don't think we have any defined style conventions for our Typescript / Vue code.

djjuhasz:
I just removed the table striping rather than trying to change the Bootstrap default styling for highlighted + striped rows.

Issue #1383, opened 2025-10-06
Feature: make homepage text customizable


sallain:
**Is your feature request related to a problem? Please describe.**

The homepage currently contains some generic text about Enduro.

<img width="1239" height="366" alt="Image" src="https://github.com/user-attachments/assets/f9f43a74-fee2-44bf-a302-74df08f3c8b8" />

It would be nice if a system admin could customize this text to be institution specific. For example, they might want to include links to internal procedures or policies or contact info for troubleshooting.

**Describe the solution you'd like**

Make that text customizable per deployment!

**Describe alternatives you've considered**

More analysis to come. 

**Additional context**

jraddaoui:
I have been looking at options to extend Enduro's dashboard with custom HTML/components. I still need to test a few ideas to allow better customization options, but I think we can address this issue with a basic implementation that we can extend later.

This initial implementation would follow what @djjuhasz did for the institution logo. Using an environment variable to set a path URL to an HTML file that will leave in the same host. If unset it will load the default homepage, if set it will be fetched, sanitized and placed instead of the default content. This will allow really basic customization using HTML only (no styles or JS). This env. var. and the HTML file could be configured per deployment.

If that is enough for this issue, it's mostly done in https://github.com/artefactual-sdps/enduro/pull/1386.

I'll create another issue to discuss following iterations.

sallain:
Thanks @jraddaoui!

PR #1386, opened 2025-10-09
Add custom home page HTML support to dashboard 

jraddaoui:
Add VITE_CUSTOM_HOME_URL environment variable to allow configuring
custom HTML content for the dashboard home page. Content is fetched
dynamically and sanitized with DOMPurify.

Refs #1383.


dashboard/src/pages/index.vue
```diff
@@ -1,42 +1,88 @@
...
-    </p>
-    <p>
-      See the
-      <a href="https://enduro.readthedocs.io/" target="_blank"
-        >Enduro documentation</a
-      >
-      for more information!
-    </p>
+    <!-- Custom HTML content -->
+    <div v-if="url">
```
djjuhasz: You could use `<template v-if="url">` here to avoid the extra `<div>` in the rendered HTML.

dashboard/src/pages/index.vue
```diff
@@ -1,42 +1,88 @@
...
+        </div>
+      </div>
+      <div v-else-if="error" class="alert alert-warning" role="alert">
+        {{ error }}
+      </div>
+      <div v-else-if="content" v-html="content"></div>
+    </div>
+
+    <!-- Default content -->
+    <div v-if="!url || error">
```
djjuhasz, line 51: `<template>` could be used here too.

dashboard/src/pages/index.vue
```diff
@@ -1,42 +1,88 @@
...
+    <div v-if="url">
+      <div v-if="loading" class="text-center p-3">
+        <div class="spinner-border text-muted" role="status">
+          <span class="visually-hidden">Loading...</span>
+        </div>
+      </div>
+      <div v-else-if="error" class="alert alert-warning" role="alert">
+        {{ error }}
+      </div>
+      <div v-else-if="content" v-html="content"></div>
```
djjuhasz, line 47: What will happen if `url` is defined, `loading` is false, `error` is null, and `content` is empty?

djjuhasz: I'm not sure if this is what is happening on the dev environment, but I'm seeing a blank home page:

Image

djjuhasz (changes requested):
@jraddaoui looks good overall, but I'm getting a blank homepage using the Tilt dev environment (see screenshot above)

dashboard/src/pages/index.vue
djjuhasz: It looks like `http://localhost:8080/$VITE_CUSTOM_HOME_URL` is being sent as the custom home page request.
image

jraddaoui: > What will happen if url is defined, loading is false, error is null, and content is empty?

It will show an empty page. It will fetch the URL without failing, so I guess that may only happen if serving an empty HTML page or if the content is sanitized entirely. I'll show an error and fallback to the default page, but that won't catch if the HTML is just tags without real content.

> It looks like http://localhost:8080/$VITE_CUSTOM_HOME_URL is being sent as the custom home page request.

Right, thanks for testing it! I removed the example but forgot to leave the new variable unset in the dev environment, and there is no injection script there. It should be fixed now.

jraddaoui:
Thanks @djjuhasz!

djjuhasz (approved):
LGTM, thanks @jraddaoui!

PR #1387, opened 2025-10-10
Sort `ListSIPSourceObjects` API results by date

djjuhasz:
Refs #1373

- Sort the objects returned from the `ListSIPSourceObjects" API endpoint by last modified date, in descending order (most recent first).
- Add a `sipsource.ListOptions` struct and use it the `sipsource.List()` calls.
- Add sort options to the `sipsource.List()` method.
- Return an "invalid cursor" error when a bad token is provided

### DIFF
--- a/internal/sipsource/bucket_source.go
+++ b/internal/sipsource/bucket_source.go
@@ -3,14 +3,14 @@ package sipsource
 import (
 	"context"
 	"fmt"
+	"io"
+	"slices"
 	"time"
 
 	"github.com/google/uuid"
 	"go.artefactual.dev/tools/bucket"
 	"gocloud.dev/blob"
 )
-
-const defaultLimit = 100
 
 // BucketSource represents a SIP source that uses a cloud storage bucket
 // to store SIPs. It implements the SIPSource interface.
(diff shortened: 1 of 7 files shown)


internal/sipsource/sipsource.go
```diff
@@ -45,3 +57,71 @@ type Object struct {
...
+// SortByKey returns a Sort that sorts by the object key.
+func SortByKey() *Sort {
+	return &Sort{attr: "key"}
+}
+
+// SortByModTime returns a Sort that sorts by the object modification time.
+func SortByModTime() *Sort {
+	return &Sort{attr: "modtime"}
+}
+
```
jraddaoui, line 91: Why not `SortBySize()`?

internal/sipsource/sipsource.go
```diff
@@ -45,3 +57,71 @@ type Object struct {
...
+func (s *Sort) ByDesc() *Sort {
+	s.desc = true
+	return s
+}
+
+// ByAsc sets the sort order to ascending (A-Z).
+func (s *Sort) ByAsc() *Sort {
+	s.desc = false
+	return s
+}
```
jraddaoui: Just `Asc()` and `Desc()` sounds better to me.

internal/sipsource/sipsource.go
```diff
@@ -45,3 +57,71 @@ type Object struct {
...
+		if s.desc {
+			return b.ModTime.Compare(a.ModTime)
+		}
+		return a.ModTime.Compare(b.ModTime)
+	case "key":
+		if s.desc {
+			return strings.Compare(b.Key, a.Key)
+		}
+		return strings.Compare(a.Key, b.Key)
+	}
```
jraddaoui, line 120: Instead of checking `s.desc` on each case to change how the values are compared, we could compare always `a` against `b` and negate the result after the switch:

```go
var re int
switch s.attr {
case "modtime":
	re = a.ModTime.Compare(b.ModTime)
case "key":
	re = strings.Compare(a.Key, b.Key)
default:
	return 0
}

if s.desc {
	return -re
}

return re
```

internal/sipsource/bucket_source.go
```diff
@@ -59,47 +59,84 @@ func (s *BucketSource) Close() error {
...
+		if err != nil {
+			return nil, fmt.Errorf("SIP bucket source: list objects: %w", err)
+		}
+		objects = append(objects, &Object{
+			Key:     i.Key,
+			ModTime: i.ModTime,
+			Size:    i.Size,
+			IsDir:   i.IsDir,
+		})
 	}
```
jraddaoui, line 101: Just thinking about it for a next iteration, what do you think about caching? Not sure how that would work with thousands of objects, we may need to check if Redis is a good option, but if we are planning to consume the entire list it may be a good idea caching the full objects list for a little while.

jraddaoui (approved):
LGTM @djjuhasz!

internal/sipsource/sipsource.go
djjuhasz: I didn't add `SortBySize()` because we don't need it yet. We're only using `SortByModTime()` right now for the API results — I added `SortByKey()` because `SortByModTime()` wasn't reliable for the tests.

Maybe I should just add all three sort fields to the API endpoint though, and just make `SortByModTime().Desc()` the default.

internal/sipsource/sipsource.go
djjuhasz: Yeah, I thought about that but I thought we might want to add `Asc()` and `Desc()` as getters in the future.

internal/sipsource/bucket_source.go
djjuhasz: I think cache invalidation would be challenging to get right.  I don't think there's a quick check to find out if a file has been added to the SIP source bucket, in which case we would be doing purely-time based cache expiration.  

I don't know what would make sense for a cache expiration time, a short expiration time (<1 second) would result in mostly cache misses unless a lot of users were using Enduro at the same time, but a long expiration time (>30s) could be a pain when trying to add a file to the SIP source and then upload it. I don't know how effective an expiration time in the middle would be, but I suspect we'd have a lot more cache misses than hits.

internal/sipsource/sipsource.go
djjuhasz: Adding sorting to the API endpoint is going to require a significant amount of extra work and changes so I think I'll just leave this as-is for now.  I do think having both `SortByKey()` and `SortByModTime()` is helpful because it establishes the pattern for multiple sort attributes.

internal/sipsource/sipsource.go
djjuhasz: I guess we could implement `IsDesc()` to get the sort order rather than `Desc()`.

internal/sipsource/bucket_source.go
jraddaoui: I was thinking on a short expiration time, if we are planning to consume all pages from the dashboard at "once".

djjuhasz: Ah, I see. Yes, for buckets with a lot of objects I think it would make sense to cache the bucket object list between paging requests to avoid having to repeatedly fetch and re-sort the list from the bucket on each request.  I think timing the cache expiration could be tricky if the subsequent pages are retrieved on user action (e.g. scrolling to the bottom of the list, pressing a button) but I'm sure we could work out the details.  I also think we are going to have to make some significant changes to the current SIP Source UI to make it usable with hundreds or thousands of bucket objects — e.g.  expandable "folders", search, filtering.

Issue #1401, opened 2025-11-04
Problem: AIP deletion in Archivematica Storage Service (AMSS) locations requires approval in AMSS


jraddaoui:
**Is your feature request related to a problem? Please describe.**

AIP deletions follow a request/review process in Enduro and, when the AIP is stored in an AMSS location, they also require a request/review process in AMSS. To avoid this manual step, we set up a management script that automatically approves deletion requests in AMSS on a schedule. However, that script could also pick up other deletion requests in AMSS that did not originate from Enduro.

This process also requires Enduro to poll AMSS for status updates to know when a request has been reviewed. We configure a timeout of 2 hours for that activity, assuming the script runs every hour. If the script is not configured correctly, or if manual approval does not occur within that period, the AIP may end up with a status mismatch between Enduro and AMSS, preventing any further deletion requests from being sent to AMSS.

**Describe the solution you'd like**

In AMSS v0.25.0 it will be possible to review AIP deletion requests through the API (re: [issue](https://github.com/archivematica/Issues/issues/1696) and [PR](https://github.com/artefactual/archivematica-storage-service/pull/813)). I'd like to fully automate the request/review in AMSS after the approval in Enduro. 

**Describe alternatives you've considered**

Leave it as it is.

**Additional context**

Ideally, this would be added with a configuration variable to maintain backwards compatibility with AMSS v.0.24.0 and lower.

PR #1402, opened 2025-11-04
Add configurable approval for AMSS AIP deletions

jraddaoui:
Introduce a new `aipDeletion.approveAMSS` boolean option to the
storage service configuration that enables automatic approval of
AIP deletion requests in Archivematica Storage Service (AMSS)
locations. This requires AMSS v0.25.0 or later.

Refs https://github.com/artefactual-sdps/enduro/issues/1401.

### DIFF
--- a/enduro.toml
+++ b/enduro.toml
@@ -138,32 +138,6 @@ workflowType = "create aip"
 # It must be a string format compatible with https://pkg.go.dev/time#ParseDuration.
 retentionPeriod = "0"
 
-[storage]
-enduroAddress = "enduro.enduro-sdps:9002"
-
-# defaultPermanentLocationId is the UUID of the storage location used for
-# permanent AIP storage in the "auto-approve" processing workflow. The value of
-# "f2cc963f-c14d-4eaa-b950-bd207189a1f1" represents the first permanent location
-# defined in the mysql-create-locations-job.yaml Kubernetes manifest.
-defaultPermanentLocationId = "f2cc963f-c14d-4eaa-b950-bd207189a1f1"
-
-[storage.database]
-driver = "mysql"
-dsn = "enduro:enduro123@tcp(mysql.enduro-sdps:3306)/enduro_storage"
-migrate = true
-
-[storage.internal]
-endpoint = "http://minio.enduro-sdps:9000"
-pathStyle = true
-key = "minio"
-secret = "minio123"
-region = "us-west-1"
-bucket = "aips"
-
-[storage.event]
-redisAddress = "redis://redis.enduro-sdps:6379"
-redisChannel = "enduro-storage-events"
-
 # Change the taskqueue setting to your prefered preservation system, by default
 # it is a3m. Use 'am' to configure Archivematica instead.
 [preservation]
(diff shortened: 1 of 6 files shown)


internal/storage/activities/delete_amss.go
```diff
@@ -121,7 +147,7 @@ func (a *DeleteFromAMSSLocationActivity) requestDeletion(
 	ctx context.Context,
 	params *DeleteFromAMSSLocationActivityParams,
 	pipelineUUID string,
-) error {
+) (int64, error) {
```
djjuhasz: I think it would be helpful to name the return variables here to provide a hint about what the `int64` return represents (e.g. `id int64, err error`).

internal/storage/activities/delete_amss_test.go
```diff
@@ -0,0 +1,339 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			srv := setupServer(t, tt.handler)
+			ts := &temporalsdk_testsuite.WorkflowTestSuite{}
+			env := ts.NewTestActivityEnvironment()
+			env.SetTestTimeout(5 * time.Second)
+
+			// Set up a ticker channel for polling tests and tick twice.
+			ch := make(chan time.Time, 2)
```
djjuhasz, line 305: I think this kind of testing scenario where you are waiting for a timer to fire can now be handled without slowing down the tests or introducing an alternate timing channel by using https://go.dev/blog/synctest.  I haven't used synctest, but I think you could just wrap the test in a `synctest.Run()` function, then add `synctest.Wait()` (maybe twice?) after the `env.ExecuteActivity()` call.

enduro.toml
```diff
@@ -322,3 +296,43 @@ workflowName = "preprocessing"
 # namespace = "default"
 # taskQueue = "poststorage"
 # workflowName = "poststorage"
+
+#########################################################################
```
djjuhasz, line 300: Nice! Good move to explicitly separate the storage service config. 👍

internal/storage/activities/delete_amss_test.go
```diff
@@ -0,0 +1,339 @@
...
+					err := json.NewEncoder(w).Encode(map[string]any{"error_message": "error message"})
+					assert.NilError(t, err)
+				default:
+					t.Fatalf("unexpected request to %s", r.URL.Path)
+				}
+			},
+			wantErr: "approve deletion: error message",
+		},
+		{
+			name: "Deletes AIP polling for status (approved)",
```
djjuhasz: I had a hard time parsing this test name, I think something like "Deletes AIP after polling for status" would be clearer.

internal/storage/activities/delete_amss_test.go
```diff
@@ -0,0 +1,339 @@
...
+			handler: func() http.HandlerFunc {
+				var requested, polled bool
+				return func(w http.ResponseWriter, r *http.Request) {
+					switch r.URL.Path {
+					case "/api/v2/file/" + aipUUID + "/":
+						if !requested {
+							handleAIPInfo(w, r, "DEL_REQ")
```
djjuhasz: Shouldn't the server return a status other than "DEL_REQ" before the deletion is requested?

internal/storage/activities/delete_amss_test.go
```diff
@@ -0,0 +1,339 @@
...
+						requested = true
+					default:
+						t.Fatalf("unexpected request to %s", r.URL.Path)
+					}
+				}
+			}(),
+			want: activities.DeleteFromAMSSLocationActivityResult{Deleted: true},
+		},
+		{
+			name: "Deletes AIP polling for status (rejected)",
```
djjuhasz: I think this name is misleading — the AIP wouldn't be deleted in this scenario, would it?

internal/storage/activities/delete_amss.go
```diff
@@ -17,6 +17,8 @@ import (
 
 type DeleteFromAMSSLocationActivity struct {
 	ssclient *http.Client
```
djjuhasz: Is there some reason to have an `ssclient` field in this struct rather than just using a local variable in the `Execute()` method?  The `ssclient` value is overwritten with a new client every time `Execute()` is called, so I can't see any reason to persist the value outside of the `Execute()` scope.

djjuhasz (changes requested):
Looks good @jraddaoui! I suggested a few improvements, but I think they are all pretty minor.  Feel free to ignore the suggestion to use https://go.dev/blog/synctest if you don't want to get into it right now.

internal/storage/activities/delete_amss.go
jraddaoui: I think it was there to have access to it from the internal functions. But now that you say that, it's an AMSS client specific to the location config, so being part of the activity struct may conflict if the activity is executed in time with another location config. I have moved it out of the struct and we are now passing it to each function.

internal/storage/activities/delete_amss_test.go
jraddaoui: Thanks for noting https://go.dev/blog/synctest, I didn't know about it. I gave it a try but I couldn't make it work, tests never ended. I think the issue is that Temporal's test framework uses mutexes, channels, and goroutines internally, which conflicts with [synctest's bubble isolation model](https://go.dev/blog/synctest#blocking-and-the-bubble) and the bubble doesn't advance. I also tried to execute the activity as a function directly, but it complained about the context for the heartbeat.

jraddaoui:

internal/storage/activities/delete_amss_test.go
djjuhasz: Oh, that's too bad.  Thanks for trying out the synctest package, hopefully it will come in handy in the future.

djjuhasz (approved):
Looks good! :+1:

Issue #1207, opened 2025-05-02
Enhancement: Add a downloadable report that captures AIP deletion information


sallain:
**Is your feature request related to a problem? Please describe.**

Some institutions are required to save a record of any AIP deletion action. This report should be able to be exported from Enduro to save in a separate location, or possibly even printed out. Currently, the record of an AIP deletion is recorded in Enduro on the AIP detail page. While it's certainly possible to just save a copy of the webpage, this isn't a very elegant solution.

**Describe the solution you'd like**

Provide users with a downloadable report in a non-editable format, like a PDF. The report should contain the following information:

General:

* Information about the Enduro instance - URL, anything else?

Request:

* **User details** of the requestor - name, email, whatever information is available that uniquely identifies the user
* **Timestamp** when the requests was made - I think that this would be the start time of the _Review AIP deletion request_ activity
* **Reason** for request

Review:

* **User details** of the reviewer - name, email, whatever information is available that uniquely identifies the user
* **Timestamp** when the review was completed - the end time of the _Review AIP deletion request_ activity
* **Reason** for review
* **Deletion confirmation** from the Storage Service - not sure what we can get here, maybe the timestamp? Name of the location where the AIP was deleted?

**Describe alternatives you've considered**

**Additional context**

fiver-watson:
I have drafted up an initial report proposal for review and discussion, here: 

 * https://docs.google.com/document/d/1Xy5kZFghO99djBo3bc1k2hIoMVg0i3yAFNLDDFwQOXU/edit?tab=t.0

djjuhasz:
We had a discussion in the team Slack channel about where to store the AIP deletion reports and the decision from the discussion was to rename the current "aips" bucket to "enduro-storage" and the current "internal" bucket to "enduro-internal".

The existing "internal" bucket was created from a consolidation of previous buckets as part of issue #1210 after [detailed analysis](https://docs.google.com/document/d/1Ojmr7wrvwtXvmsLYnID6EuIewlP1W4RIllO37_AASzE/edit?tab=t.0#heading=h.i2ywkpx6lge4) by @jraddaoui and subsequent discussion in the Enduro team to consolidate several existing buckets.  The "internal" bucket is used to store failed SIPs, failed PIPs (when the failure occurs during preservation), uploaded SIPs.  After discussion we've decided to change the name of the "internal" bucket to "enduro-ingest" to indicate that the bucket is used by the Enduro Ingest application to store files in a network accessible location.

[...]

PR #1403, opened 2025-11-04
Generate PDF AIP Deletion Report

djjuhasz:
Refs #1207

- Add the pdfcpu package to allow PDF editing
- Use pdfcpu to fill the fields of an AIP Deletion Report PDF form template
- Write the filled AIP Deletion Report to a "reports" bucket
- Add a `ReadDeletionRequest()` method to the storage service
- Add a `ListDeletionRequests()` method to the storage service
- Add a `ReadWorkflow()` method to the storage service
- Replace `ReadAipPendingDeletionRequest()` calls with `ListDeletionRequest()` to find pending deletion requests
- Remove the `ReadAipPendingDeletionRequest()` method from the storage service
- Add configuration to specify the AIP deletion report template file path and reports bucket location and credentials
- Copy the default AIP Delete Report template to the Enduro Docker container


Dockerfile
```diff
@@ -56,9 +56,11 @@ RUN groupadd --gid 1000 enduro \
...
 
 FROM base AS enduro
-COPY --link --from=build-enduro /out/enduro /home/enduro/bin/enduro
-COPY --link --from=build-enduro /src/enduro.toml /home/enduro/.config/enduro.toml
-COPY --link hack/xsd/premis.xsd /home/enduro/premis.xsd
+COPY --link --chown=enduro:enduro --from=build-enduro /out/enduro /home/enduro/bin/enduro
+COPY --link --chown=enduro:enduro --from=build-enduro /src/enduro.toml /home/enduro/.config/enduro.toml
+COPY --link --chown=enduro:enduro assets/Enduro_AIP_deletion_report_v3.tmpl.pdf /home/enduro/Enduro_AIP_deletion_report_v3.tmpl.pdf
+COPY --link --chown=enduro:enduro hack/xsd/premis.xsd /home/enduro/premis.xsd
+USER enduro
```
jraddaoui: Why is this needed now? The user is set in the base before copying.

cmd/enduro/main.go
```diff
@@ -335,6 +337,14 @@ func main() {
...
 	}
 
+	// Set up reports bucket.
+	reportsBucket, err := bucket.NewWithConfig(ctx, cfg.Storage.Reports.Bucket)
+	if err != nil {
+		logger.Error(err, "Error setting up reports bucket.")
+		os.Exit(1)
+	}
+	defer reportsBucket.Close()
+
```
jraddaoui: We discussed the proliferation of buckets on the ingest side and decided to use a single internal bucket for most cases. We already have an internal bucket on the storage side that is only use for AIP review, I think we should use that one.

In some cases MinIO access may be restricted, so we will need to provide a link to the report in the dashboard. In another PR, as we are discussing in the standup.

cmd/enduro/main.go
```diff
@@ -532,6 +546,17 @@ func main() {
 			storage_activities.NewDeleteFromAMSSLocationActivity().Execute,
 			temporalsdk_activity.RegisterOptions{Name: storage.DeleteFromAMSSLocationActivityName},
 		)
+		w.RegisterActivityWithOptions(
+			storage_activities.NewAIPDeletionReportActivity(
+				reportsBucket,
+				clockwork.NewRealClock(),
```
jraddaoui, line 544: Not a big deal since you already did the work, but I think all these clock params complicate the interfaces/structs. In this case, testing the report timestamp with `EquateApproxTime` or similar would have been enough.

enduro.toml
```diff
@@ -164,6 +164,17 @@ bucket = "aips"
 redisAddress = "redis://redis.enduro-sdps:6379"
 redisChannel = "enduro-storage-events"
 
+[storage.reports.bucket]
+endpoint = "http://minio.enduro-sdps:9000"
+pathStyle = true
+accessKey = "minio"
+secretKey = "minio123"
+region = "us-west-1"
+bucket = "reports"
```
jraddaoui: `[storage.internal]` is just above. BTW it's called "aips" because we used it only for AIP review, I'd consider changing its name now.

hack/kube/base/minio-setup-buckets-job.yaml
```diff
@@ -33,5 +33,6 @@ spec:
               mc mb enduro/perma-aips-2 --ignore-existing;
               mc mb enduro/sipsource --ignore-existing;
               mc mb enduro/internal --ignore-existing;
+              mc mb enduro/reports --ignore-existing;
```
jraddaoui: I'd rename "internal" to "ingest-internal" and "aips" to "storage-internal" or something like that. Also, there is a recreate buckets job that also needs to be updated.

internal/storage/reports/aip_deletion.go
```diff
@@ -0,0 +1,137 @@
+package reports
```
jraddaoui, line 1: Can we move this reports package inside internal/storage? I'd like to keep the separation clear, and at some point do the same on ingest (moving specific ingest package inside internal/ingest, like internal/persistence), keeping the in the internal folder only those that are used by both services (like internal/event).

internal/reports/aip_deletion.go
```diff
@@ -0,0 +1,137 @@
...
+
+	"github.com/google/uuid"
+	"github.com/jonboulle/clockwork"
+	"github.com/pdfcpu/pdfcpu/pkg/api"
+)
+
+// AIPDeletionConfig holds the configuration for the AIP Deletion Report.
+type AIPDeletionConfig struct {
+	TemplatePath string
+}
```
jraddaoui: Would it be possible to embed the 119 KB template binary? That will simplify the code and, more importantly, the deployment and configuration. I'd use a boolean flag to enable/disable the report generation.

internal/storage/config.go
```diff
@@ -44,3 +47,8 @@ type LocationConfig struct {
 	Token     string
 	Bucket    string
 }
+
+type ReportsConfig struct {
+	Bucket      *bucket.Config
+	AIPDeletion reports.AIPDeletionConfig
+}
```
jraddaoui, line 60: Using the internal bucket and embedding the template, we'll only need the boolean flag. But I already added a new config setting related to this in https://github.com/artefactual-sdps/enduro/pull/1402.

Considering both new settings, I'd wrap them in an "AIPDeletion" struct/section including "ApproveAMSSRequests" and "GenerateReport" boolean flags.

internal/storage/persistence/telemetry.go
```diff
@@ -249,6 +278,22 @@ func (w *wrapper) UpdateDeletionRequest(
...
+
+	r, err := w.wrapped.ReadDeletionRequest(ctx, id)
+	if err != nil {
+		telemetry.RecordError(span, err)
+		return nil, updateError(err, "ReadDeletionRequest")
+	}
+
+	return r, nil
+}
+
```
jraddaoui: Just below this we still have the ReadAipPendingDeletionRequest method, how is it not complaining about the wrapped missing ReadAipPendingDeletionRequest? Is there something else that needs to be removed?

internal/storage/service_test.go
```diff
@@ -150,7 +150,7 @@ func TestNewService(t *testing.T) {
 	})
 }
 
-func TestServiceSubmit(t *testing.T) {
+func TestService_SubmitAip(t *testing.T) {
```
jraddaoui: Nitpick: `TestServiceSubmitAIP`.

internal/storage/workflows/delete.go
```diff
@@ -12,29 +12,44 @@ import (
...
+	"github.com/artefactual-sdps/enduro/internal/reports"
 	"github.com/artefactual-sdps/enduro/internal/storage"
 	"github.com/artefactual-sdps/enduro/internal/storage/activities"
 	"github.com/artefactual-sdps/enduro/internal/storage/enums"
 	"github.com/artefactual-sdps/enduro/internal/storage/types"
 )
 
 type StorageDeleteWorkflow struct {
-	storagesvc storage.Service
+	internalLocation storage.LocationConfig
```
jraddaoui: Why do we need the internal location config?

internal/storage/workflows/delete_test.go
```diff
@@ -293,7 +331,13 @@ func TestStorageDeleteWorkflow(t *testing.T) {
 			},
 		).Return(nil)
 
-		s.env.ExecuteWorkflow(NewStorageDeleteWorkflow(s.storagesvc).Execute, req)
+		s.env.ExecuteWorkflow(
+			NewStorageDeleteWorkflow(
+				s.internalLocation,
+				reports.AIPDeletionConfig{TemplatePath: "../../../assets/Enduro_AIP_deletion_report_v3.tmpl.pdf"},
+				s.storagesvc).Execute,
```
jraddaoui: Nitpick: `).Execute,` in a new line.

jraddaoui (changes requested):
Really nice @djjuhasz, looking great!

cmd/enduro/main.go
djjuhasz: I originally planned to use the ingest "internal" MinIO bucket, but there's a namespacing issue that storage already has an "internal" location which points to the "aips" bucket. :(

I've made some major changes to the report implementation since choosing to make a new "reports" bucket though, so I'll have another look at reusing the ingest internal bucket.

cmd/enduro/main.go
djjuhasz: Yeah, I wish Go had the ability to override the system time in test environments like you can in Vue/Javascript.  I don't love `EquateApproxTime` because it can be flaky in CI if the "fuzziness" is set too low, and I don't like the imprecision in the test.

enduro.toml
djjuhasz: To get a bit nitpicky, I believe all AIPs go to the "aips" bucket initially, then are moved to the "permanent AIP" bucket, not just AIPs that are waiting on review.

Are you proposing we use the current MinIO "internal" bucket (used for failed SIPs, etc.) as the "internal" location for temporary AIP storage as well or are you thinking we rename "aips" to something else (e.g. "storageInternal") to avoid a namespace conflict with the existing "internal" bucket? 

I'd prefer to keep the storage "internal" bucket separate from the ingest bucket to keep the concerns separate, but that does raise a naming problem — we need a different name than "internal" but nothing immediately comes to mind. :thinking:

hack/kube/base/minio-setup-buckets-job.yaml
djjuhasz: :cry:

internal/storage/persistence/telemetry.go
djjuhasz: Oh yeah, I removed `ReadAipPendingDeletionRequest()` from the `storage.Service` interface, but not the storage `persistence.Storage` interface. Oops.

internal/storage/workflows/delete.go
djjuhasz: Oops, missed it when refactoring.

internal/storage/persistence/telemetry.go
djjuhasz: Fixed.

internal/storage/workflows/delete.go
djjuhasz: Fixed.

internal/storage/reports/aip_deletion.go
djjuhasz: Sure. I thought the "reports" bucket would be shared between ingest (for future reports like SIP ingest) and storage, but since we're eliminating the "reports" bucket anyway I can move the reports package.

internal/reports/aip_deletion.go
djjuhasz: It's possible but it does mean the default template is always included in the Go binary, even in deployments where it's not used.  I'd prefer to keep the binary smaller even though it means extra deployment complexity.

internal/storage/config.go
djjuhasz: You still need `TemplatePath` even if we embed the default Deletion Report template to allow specifying a different template without having to recompile the Enduro binary.

cmd/enduro/main.go
jraddaoui: Not the ingest internal bucket, the storage internal bucket/location/storage. It's set up right above:

https://github.com/artefactual-sdps/enduro/blob/main/cmd/enduro/main.go#L245-L251

enduro.toml
jraddaoui: I think I addressed this question in other comments below and in the one I just made. AIP deletion is part of the storage domain so I'd rename the "aips" bucket and use that one.

internal/reports/aip_deletion.go
jraddaoui: I strongly prefer to embed the template, @djjuhasz, having worked on the Ansible side for the premis.xsd. I really liked how you did it in [preprocessing](https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/internal/premis/xsdv3.go), even in that case, where the file needs to be recreated on disk. In this case we don't even have to recreate the file and I think a boolean to enable/disable the feature is a lot cleaner than setting a path or not.

internal/storage/config.go
jraddaoui: I see, you want the template to be configurable :thinking:, that's thinking ahead but a good idea. That makes me realize this is lacking some documentation and makes me have second thoughts about embedding the template.

I don't want to complicate even more this initial implementation, but I think the ideal would be a mix of what you are saying here and what you already did; embed the default template and allow using a custom one trough configuration. Lets go with what you did in this PR and think about embedding a default template later.

I still think it would be a good idea to wrap the config as I was suggesting:

```toml
[storage.aipDeletion]
reportTemplatePath = "/path/to/template.pdf"
approveAMSSDeletions= true
```

Something like that? I'll create the wrapper in the other PR.

internal/reports/aip_deletion.go
jraddaoui: See my other comment, not sure if we are already planning to use a custom template in the deployments we manage with Ansible, but it's a good idea to allow customization. We can decide later if we want to embed this default template to ease vanilla installs.

hack/kube/base/minio-setup-buckets-job.yaml
jraddaoui: Happy if you have better alternatives :man_shrugging: :smile:

enduro.toml
djjuhasz: Yeah, sorry. I should read all your comments before I start replying. :grimacing:

internal/storage/config.go
djjuhasz: Okay, sure.

hack/kube/base/minio-setup-buckets-job.yaml
djjuhasz: As discussed we'll use "enduro-ingest" (now "internal") and "enduro-storage" (now "aips").

djjuhasz:
@jraddaoui this is ready for another look.

Dockerfile
```diff
@@ -56,9 +56,10 @@ RUN groupadd --gid 1000 enduro \
 USER enduro
 
 FROM base AS enduro
-COPY --link --from=build-enduro /out/enduro /home/enduro/bin/enduro
-COPY --link --from=build-enduro /src/enduro.toml /home/enduro/.config/enduro.toml
-COPY --link hack/xsd/premis.xsd /home/enduro/premis.xsd
+COPY --link --chown=enduro:enduro --from=build-enduro /out/enduro /home/enduro/bin/enduro
+COPY --link --chown=enduro:enduro --from=build-enduro /src/enduro.toml /home/enduro/.config/enduro.toml
+COPY --link --chown=enduro:enduro assets/Enduro_AIP_deletion_report_v3.tmpl.pdf /home/enduro/Enduro_AIP_deletion_report_v3.tmpl.pdf
+COPY --link --chown=enduro:enduro hack/xsd/premis.xsd /home/enduro/premis.xsd
```
jraddaoui, line 62: Since `enduro` is set as the user before, do we really need the `--chown` flag?

hack/kube/base/minio-setup-buckets-job.yaml
```diff
@@ -28,7 +28,7 @@ spec:
               "-c",
               "mc alias set enduro http://minio.enduro-sdps:9000 ${MINIO_USER} ${MINIO_PASSWORD};
               mc mb enduro/sips --ignore-existing;
-              mc mb enduro/aips --ignore-existing;
+              mc mb enduro/enduro-storage --ignore-existing;
```
jraddaoui, line 31: Don't forget about https://github.com/artefactual-sdps/enduro/blob/main/hack/kube/tools/minio-recreate-buckets-job.yaml#L31. Also, are you planning to rename the `internal` bucket to `enduro-ingest` in another PR or just forgot in this one?

internal/storage/persistence/telemetry.go
```diff
@@ -180,6 +180,19 @@ func (w *wrapper) CreateWorkflow(ctx context.Context, workflow *types.Workflow)
 	return nil
 }
 
+func (w *wrapper) ReadWorkflow(ctx context.Context, dbID int) (*types.Workflow, error) {
```
jraddaoui, line 183: Doesn't need to happen here, but it would be nice to have a test for telemetry.go like we have on ingest.

internal/storage/activities/aip_deletion_report.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (a *AIPDeletionReportActivity) Execute(
+	ctx context.Context,
+	params *AIPDeletionReportActivityParams,
+) (*AIPDeletionReportActivityResult, error) {
+	if a.cfg.ReportTemplatePath == "" {
+		return nil, fmt.Errorf("AIP deletion report: template path is not configured")
+	}
+
+	key := fmt.Sprintf("reports/aip_deletion_report_%s.pdf", params.AIPID.String())
```
jraddaoui: Similar to what we have on the ingest side, I think it would be nice to make `reports/` a constant:

https://github.com/artefactual-sdps/enduro/blob/main/internal/ingest/ingest.go#L27-L34

I don't know if `/` could be an issue in other object stores, and we may want to make those prefixes configurable at some point.

internal/storage/service.go
```diff
@@ -199,6 +208,7 @@ func (s *serviceImpl) SubmitAip(
 	if err != nil {
 		return nil, err
 	}
+	bucket = blob.PrefixedBucket(bucket, "aips/")
```
jraddaoui: The same, I think having `aips/` as a constant may be beneficial. It's used in three different places here, safer to update a constant than three strings if we decide to change the prefix at some point.

internal/storage/workflows/delete.go
```diff
@@ -315,3 +334,44 @@ func (w *StorageDeleteWorkflow) deleteAIPFromAMSSLocation(
...
 }
+
+func (w *StorageDeleteWorkflow) generateAIPDeletionReport(
+	ctx temporalsdk_workflow.Context,
+	aipID uuid.UUID,
+	locationSource enums.LocationSource,
+) error {
+	logger := temporalsdk_workflow.GetLogger(ctx)
+
+	logger.Info("Generating AIP deletion report", "aip_id", aipID, "location_source", locationSource)
```
jraddaoui: I think the other log message inside the if below would be enough if the report is not going to be generated. I'd move this one after the if to log it only when it's going to happen.

internal/storage/workflows/delete.go
```diff
@@ -315,3 +334,44 @@ func (w *StorageDeleteWorkflow) deleteAIPFromAMSSLocation(
...
+	if w.cfg.ReportTemplatePath == "" {
+		logger.Info(
+			"Skipping AIP deletion report generation as no template path is configured",
+			"template_path",
+			w.cfg.ReportTemplatePath,
+		)
```
jraddaoui: Why logging the template path if it will always be empty?

internal/storage/workflows/delete_test.go
```diff
@@ -29,9 +34,11 @@ const (
 )
 
 type StorageDeleteWorkflowTestSuite struct {
-	env        *temporalsdk_testsuite.TestWorkflowEnvironment
-	mockCtrl   *gomock.Controller
-	storagesvc *fake.MockService
+	env              *temporalsdk_testsuite.TestWorkflowEnvironment
+	internalLocation storage.LocationConfig
+	mockCtrl         *gomock.Controller
+	reportsBucket    *blob.Bucket
```
jraddaoui: Do we need the internal location config and the reports bucket in the tests? It should be the same bucket, but the config is not used to set up the bucket.

jraddaoui (changes requested):
Thanks @djjuhasz! Just a few questions and minor details, looking great!

Dockerfile
djjuhasz: I was getting a permissions error trying to read the template file before I added the `--chown` flag. The [Dockerfile COPY docs](
https://docs.docker.com/reference/dockerfile/#copy) include this note about file permissions:

> All files and directories copied from the build context are created with a UID and GID of 0 unless the optional --chown flag specifies a given username, groupname, or UID/GID combination to request specific ownership of the copied content.

hack/kube/base/minio-setup-buckets-job.yaml
djjuhasz: I was planning to rename the "internal" bucket in another PR because I wasn't sure how much work it would require.  Thinking about it now though, it probably only needs to be changed here and in the `enduro.toml` file — you've already done the work of prefixing the objects being written to the bucket.

internal/storage/workflows/delete.go
djjuhasz: Yeah, good point.

hack/kube/base/minio-setup-buckets-job.yaml
djjuhasz: Bucket renamed. :)

And I had totally forgotten about "minio-recreate-buckets-job.yaml" thanks for the reminder. :grimacing:

internal/storage/workflows/delete.go
djjuhasz: I removed this log message altogether, I don't think it's very useful when the report is generated successfully and hopefully the error messages give enough info for debugging any problems.

internal/storage/workflows/delete_test.go
djjuhasz: Those aren't needed anymore, they were in previous versions of the AIP report implementation.

Dockerfile
jraddaoui: Right, they are owned by root, good catch @djjuhasz! 

```
$ kubectl -n enduro-sdps exec enduro-64fb5f4b64-c9cql -c enduro -- ls -al /home/enduro
total 96
drwx------ 1 enduro enduro  4096 Nov 14 18:26 .
drwxr-xr-x 1 root   root    4096 Nov  4 17:48 ..
-rw-r--r-- 1 enduro enduro   220 Jul 30 19:28 .bash_logout
-rw-r--r-- 1 enduro enduro  3526 Jul 30 19:28 .bashrc
drwxr-xr-x 3 enduro enduro  4096 Nov 14 18:26 .cache
drwxr-xr-x 2 root   root    4096 Nov 14 18:26 .config
-rw-r--r-- 1 enduro enduro   807 Jul 30 19:28 .profile
drwxr-xr-x 2 root   root    4096 Nov 14 18:26 bin
drwx------ 1 enduro enduro  4096 Nov 14 18:26 logs
-rw-rw-r-- 1 root   root   52845 Jan 21  2025 premis.xsd
```

I wonder why it didn't complain about the other ones.

jraddaoui (approved):
Thanks @djjuhasz!

PR #1404, opened 2025-11-07
Allow poll interval configuration and close response bodies

jraddaoui:
Some feedback from @sevein after #1402.

### DIFF
--- a/internal/storage/activities/delete_amss.go
+++ b/internal/storage/activities/delete_amss.go
@@ -126,6 +126,7 @@ func (a *DeleteFromAMSSLocationActivity) getPipelineUUID(
 	if err != nil {
 		return "", fmt.Errorf("get pipeline UUID: %v", err)
 	}
+	defer resp.Body.Close()
 	if resp.StatusCode >= 400 {
 		return "", fmt.Errorf("get pipeline UUID: response code: %d", resp.StatusCode)
 	}
@@ -170,6 +171,7 @@ func (a *DeleteFromAMSSLocationActivity) requestDeletion(
 	if err != nil {
 		return 0, fmt.Errorf("request deletion: %v", err)
 	}
+	defer resp.Body.Close()
 	if resp.StatusCode >= 400 {
 		return 0, fmt.Errorf("request deletion: response code: %d", resp.StatusCode)
 	}
@@ -210,6 +212,7 @@ func (a *DeleteFromAMSSLocationActivity) approveDeletion(
 	if err != nil {
 		return fmt.Errorf("approve deletion: %v", err)
 	}
+	defer resp.Body.Close()
 	if resp.StatusCode >= 400 {
 		return fmt.Errorf("approve deletion: response code: %d", resp.StatusCode)
 	}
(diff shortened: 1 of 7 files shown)


internal/storage/activities/delete_amss.go
```diff
@@ -72,17 +72,18 @@ func (a *DeleteFromAMSSLocationActivity) Execute(
 	}
 
 	// Set up ticker channel if not provided.
-	if a.tickerCh == nil {
+	tickerCh := a.tickerCh
```
djjuhasz: I think that even with this change, the current functionality will cause concurrent executions of the activity to drain the same channel which will cause two executions to tick at half the ticker interval, see:
https://go.dev/play/p/ab_1291pDTj

If you look at the timings from that example the "activity 1" executions are alternating receives from the ticker channel.

Is there a reason to inject a ticker channel into the activity rather than just passing a `time.Duration` interval, and then starting the ticker inside `Execute()`?

djjuhasz (commented):
I don't think that copying the ticker channel to a new variable is preventing interactions between concurrent executions of the activity.

internal/storage/activities/delete_amss.go
jraddaoui: In that example you are manipulating the ticket channel directly in the activity struct. The ticket can only be provided using a different constructor, intended for tests:

https://github.com/artefactual-sdps/enduro/blob/dev/issue-1401-feedback/internal/storage/activities/delete_amss.go#L36-L46

Production code will use the other constructor:

https://github.com/artefactual-sdps/enduro/blob/dev/issue-1401-feedback/cmd/enduro/main.go#L532

In this case each activity will create its own channel in a local var. Previously I was updating the activity struct, even if the channel was not provided.

jraddaoui: Passing a duration could be another option to what I did in #1402 with the channel. The current approach with two constructors gives us a cleaner activity set up outside tests, but if you really prefer that approach I can give it a try. Let me know ;)

djjuhasz: I think there are a few advantages to passing a poll interval in the activity constructor, rather than passing a `<-chan time.Time`:

1. No need for a separate constructor and code path for tests
2. Easier to understand
3. Makes it easier to set the poll interval by configuration, if we want to do that in the future

I think in general it's good practice to try and minimize the differences between the code run in production and in test.

jraddaoui: Done ;)

internal/storage/activities/delete_amss.go
```diff
@@ -71,18 +59,14 @@ func (a *DeleteFromAMSSLocationActivity) Execute(
 		return &DeleteFromAMSSLocationActivityResult{Deleted: true}, nil
 	}
 
-	// Set up ticker channel if not provided.
-	if a.tickerCh == nil {
-		ticker := time.NewTicker(time.Second * 60)
-		defer ticker.Stop()
-		a.tickerCh = ticker.C
-	}
+	ticker := time.NewTicker(a.pollInterval)
```
djjuhasz, line 65: `time.NewTicker()` will panic if `a.pollInterval` is zero, so I think we should set a minimum value.

djjuhasz (changes requested):
Thanks @jraddaoui, I think this is a better implementation.  :) I just spotted one problem that could cause a panic.

internal/storage/activities/delete_amss.go
jraddaoui: Good catch!

djjuhasz (approved):
LGTM! Thanks @jraddaoui. :+1:

Issue #1405, opened 2025-11-07
Feature: add batch management


sallain:
**Is your feature request related to a problem? Please describe.**

Several of our clients think of their content in batches - that is, they want to process many SIPs at once, and be able to track the progress of those SIPs through the system as a group. Currently in Enduro there is no way to link SIPs together to enable this kind of tracking. It is possible to upload multiple SIPs at once, but they're each processed individually. A user must use manual tracking methods to ensure that a batch is fully ingested.

**Describe the solution you'd like**

Here goes:

* On the Upload SIPs page, add an option to Configure Ingest that has a "Treat this ingest as a batch?" toggle
* If the user activates the toggle, display a text entry field where a user can optionally enter a custom batch ID (if they do not enter a custom ID, the system will generate one)
* When the ingest is launched, all of the SIPs included in the batch will be give a batch ID

<img width="1264" height="1064" alt="Image" src="https://github.com/user-attachments/assets/22ebb34c-163d-450b-b030-e28c89e4cce5" />

* When batch is toggled on and the user clicks Start Ingest, the user is taken to a new Batches page (currently the user is redirected to the SIPs page - if they are starting a non-batch SIP, they should still be taken there)
* The Batches page lists all batches, with their ID, the number of SIPs, who ingested the batch, the Start time, and the Status

<img width="1275" height="920" alt="Image" src="https://github.com/user-attachments/assets/60131c9d-a240-45c2-ac2c-acc7227ae3bb" />

* Each Batch ID links to a Batch detail page
* The Batch details section displays the following fields: Batch ID, Number of SIPs, Ingested by, Started, Completed (if complete), and Batch Status (INGESTED, PENDING, PROCESSING, or ABANDONED) 
* The Related resources section contains a link to the Batch report (TBD)
* The SIPs in Batch section contains a list of all the SIPs in the batch, with the following fields: SIP Name, SIP UUID, Started, Completed, and SIP Status

<img width="1265" height="869" alt="Image" src="https://github.com/user-attachments/assets/d8f6cd8e-52fd-46f2-b3bf-c4b4d23ac4b7" />

* Each SIP Name links to the correct SIP detail page. If a SIP is part of a batch, it has a Batch ID and a Batch status in the SIP details section

<img width="914" height="694" alt="Image" src="https://github.com/user-attachments/assets/f34c7a7b-3fd8-474d-a319-1ddb92fcb8ad" />

[...]

PR #1406, opened 2025-11-11
Add initial API design for ingest batches

jraddaoui:
Based in #1408. Refs #1405.


internal/api/design/ingest.go
```diff
@@ -583,3 +685,50 @@ var SIPSourceObjects = ResultType("application/vnd.enduro.ingest.sipsource.objec
...
+		TypedAttributeUUID("uuid", "Identifier of Batch")
+		Attribute("identifier", String, "Identifier of the Batch")
+		Attribute("sips_count", Int, "Number of SIPs in the Batch")
+		Attribute("status", String, "Status of the Batch", func() {
+			EnumBatchStatus()
+		})
+		Attribute("created_at", String, "Creation datetime", func() {
+			Format(FormatDateTime)
+		})
+		Attribute("started_at", String, "Start datetime", func() {
```
djjuhasz, line 706: Having both "created_at" and "started_at" seems unnecessary to me.

internal/enums/batch_status.go
```diff
@@ -0,0 +1,12 @@
+package enums
+
+/*
+ENUM(
+queued       // Awaiting resource allocation.
+processing   // Undergoing work.
+pending      // Awaiting user decision.
+ingested     // Successfully ingested.
+abandoned    // Ingest abandoned by user.
```
djjuhasz: "canceled    // Ingest canceled by user." is better I think.  Cancellation implies the user made a decision to stop processing the batch, "abandoned" sounds like the user just left and forgot to come back. :)

djjuhasz (approved):
Looks good @jraddaoui. I had a couple of suggestions but no blockers.

internal/enums/batch_status.go
sallain: I'm in favour of cancelled instead of abandoned - at least two of my testers thought abandoned was kind of weird! Looping in @fiver-watson to see what he thinks.

internal/api/design/ingest.go
jraddaoui: We talked about this a couple of days ago, I added a note to the issue with some decisions we made:

https://github.com/artefactual-sdps/enduro/issues/1405#issuecomment-3523047512

In this case we decided to follow the SIP approach, we create it with the queued status and created date, the started date is set when the workflow is started by Temporal, same as setting the processing status.

internal/enums/batch_status.go
jraddaoui: I agree. Canceled is also more consistent with what we have on the storage side for the AIP deletion workflow.

PR #1410, opened 2025-11-12
Add ingest batch schema and persistence methods

jraddaoui:
Based in #1406. Refs #1405.


internal/datatypes/sip.go
```diff
@@ -37,6 +37,9 @@ type SIP struct {
 
 	// Uploader is the user that uploaded the SIP.
 	Uploader *User
+
+	// Batch is the batch this SIP belongs to.
+	Batch *Batch
```
djjuhasz, line 42: I think including the whole `Batch` struct in each SIP is too much repetition of data, especially since Batch in turn embeds `User` (uploader) and we will often be fetching many SIPs at once.  I think it's better to include just the Batch UUID and any fields shown on the SIP list page (e.g. identifier). We can do a secondary Batch query to get more Batch data when it's required.

internal/persistence/ent/client/batch.go
```diff
@@ -0,0 +1,239 @@
...
+//
+// The input Batch "ID" and "CreatedAt" values are ignored; the stored Batch
+// "ID" is generated by the persistence implementation and "CreatedAt" is always
+// set to the current time.
+func (c *client) CreateBatch(ctx context.Context, b *datatypes.Batch) error {
+	// Validate required fields.
+	if b.UUID == uuid.Nil {
+		return newRequiredFieldError("UUID")
+	}
+	if b.Identifier == "" {
```
djjuhasz, line 29: Identifier is optional, isn't it?

internal/persistence/ent/client/batch.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	page, whole := filterBatches(q, f)
+
+	r, err := page.All(ctx)
+	if err != nil {
+		return nil, nil, newDBError(err)
+	}
+
+	for _, i := range r {
```
djjuhasz: Because we already know how big the `res` slice needs to be I think it's more efficient to declare `res` as below, so the memory can be pre-allocated, then to repeatedly append to the slice.

e.g.

```
res := make([]*datatypes.Batch, len(r))
for i, b := range r {
	res[i] = convertBatch(b)
}
```

Ref: https://oilbeater.com/en/2024/03/04/golang-slice-performance/

djjuhasz: Huh, I didn't know about https://github.com/alexkohler/prealloc until reading the above article.  It would probably be good to add it to our `golangci-lint` linters.

internal/persistence/persistence.go
```diff
@@ -52,4 +53,10 @@ type Service interface {
 
 	// ListUsers retrieves a list of users based on the provided filter.
 	ListUsers(context.Context, *UserFilter) ([]*datatypes.User, *Page, error)
+
+	CreateBatch(context.Context, *datatypes.Batch) error
+	UpdateBatch(context.Context, uuid.UUID, BatchUpdater) (*datatypes.Batch, error)
+	DeleteBatch(context.Context, int) error
```
djjuhasz: I think `DeleteBatch` should use the batch UUID for selection to be consistent with the other batch methods that operate on an existing row.

djjuhasz (changes requested):
Thanks @jraddaoui. This looks good overall, just a few notes.

internal/persistence/ent/client/batch.go
jraddaoui: Thanks for pointing that out. I gave it a try, but I was only able to make it complain about:

https://github.com/artefactual-sdps/enduro/blob/main/internal/sipsource/bucket_source.go#L85

And for that I had to change the default settings to enable `for-loops`:

> -forloops (default false) - Report preallocation suggestions on for loops. This is false by default due to there generally being weirder things happening inside for loops (at least from what I've observed in the Standard Library).

Looking at the code, I think that one should not be pre-allocated, but the pages below could be, although it's not complaining about it:

https://github.com/artefactual-sdps/enduro/blob/main/internal/sipsource/bucket_source.go#L129

It doesn't complain about the SIPs/Batches `res` either, I guess not being able to detect the length of `r` through static analysis.

All that and the note from golangci-lint makes a bit hesitant to add this linter.

> IMPORTANT: we don't recommend using this linter before doing performance profiling.
> For most programs usage of prealloc will be a premature optimization.

But I'll fix what you mentioned here and in the list SIPs functions.

djjuhasz: Huh, interesting. Thanks for doing the research into the linter @jraddaoui.  I'm good with not adding the linter based on the underwhelming results and the not about premature optimization.

internal/persistence/persistence.go
jraddaoui: Done, also for SIPs.

internal/persistence/ent/client/batch.go
jraddaoui: The issue says "if they do not enter a custom ID, the system will generate one". TBH I'm not sure where/how we'll generate one but I think it should be required at this point.

djjuhasz: I thought the UUID would be the generated batch ID.

internal/datatypes/sip.go
jraddaoui: I think it's better to follow the same pattern we have with the uploader relation, something we already considered in  other PRs. Then on the API design we are only adding the three needed batch fields to the SIP, but I think internally is better this way. When fetching SIPs we only join to the direct relations (uploader and batch), not to the batch uploader.

But this made me realize I was not considering the batch in some SIP functions, I've fixed that in another commit.

internal/persistence/ent/client/batch.go
jraddaoui: It could be, which is also a required field, so it can be duplicated as the identifier by the caller.

jraddaoui:
Thanks @djjuhasz! This is ready for another look. I'll check what's going on with Tilt CI, as it's failing in other PRs too.

internal/persistence/ent/client/batch.go
djjuhasz: I don't think duplicating data in the database is a good solution.

internal/datatypes/sip.go
djjuhasz: Yeah, I'm regretting my `Uploader` implementation now. I think that eager loading unnecessary data is a bad pattern and one we should avoid.  Maybe it's better to create a "SIPList" struct (or whatever we want to call it) that only contains the data we need for the SIP list.  I think we have some abridged data views in the API that are used for list results.

Anyway, I can let it go for this PR, and maybe we can come back to it later.

internal/persistence/ent/client/batch.go
jraddaoui: The identifier is a user generated field, with an input in the upload form. If none provided, the system should generate one, but we have not decided what would be the generated value if the user doesn't provide one. I think the UUID is a good first choice but this needs to be discussed with the team.

djjuhasz (approved):
👍

PR #1412, opened 2025-11-17
Use JSON payload for add batch request

jraddaoui:
Do not map fields to Params to prevent URL length limitations.

Refs #1405.


djjuhasz (approved):
LGTM 👍

PR #1414, opened 2025-11-17
Persist AIP deletion report key

djjuhasz:
Refs #1207.

- Add a "deletion_report_key" column to the storage "deletion_request" table
- Add a db migration to add the "deletion_report_key" column
- Add a `DeletionReportKey` field to the `types.DeletionRequest` model
- Persist AIP deletion report in AIP deletion report activity


djjuhasz:
@jraddaoui I finally fixed the data race that was causing the tests to fail on this PR.  The changes to fix the data race are pretty significant so it's probably easier to review the two commits individually.

cmd/enduro/main.go
```diff
@@ -544,6 +545,7 @@ func main() {
 				clockwork.NewRealClock(),
 				cfg.Storage.AIPDeletion,
 				storagesvc,
+				pdf.NewPDFCPU(),
```
jraddaoui: It was not the easiest to understand that `pdf.NewPDFCPU()` returns a `pdf.FormFiller`, `pdf.NewPDFCPUFormFiller()` would be easier to get.

internal/storage/activities/aip_deletion_report.go
```diff
@@ -51,38 +60,35 @@ func (a *AIPDeletionReportActivity) Execute(
 	if a.cfg.ReportTemplatePath == "" {
 		return nil, fmt.Errorf("AIP deletion report: template path is not configured")
 	}
+	if _, err := os.Stat(a.cfg.ReportTemplatePath); err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("AIP deletion report: template file does not exist: %s", a.cfg.ReportTemplatePath)
+		}
+		return nil, fmt.Errorf("AIP deletion report: error reading template file: %s", err)
```
jraddaoui: I'd just return the last error in both cases, not a big difference.

internal/storage/activities/aip_deletion_report.go
```diff
@@ -51,38 +60,35 @@ func (a *AIPDeletionReportActivity) Execute(
...
-	loc, err := a.storageSvc.Location(ctx, uuid.Nil)
-	if err != nil {
-		return nil, fmt.Errorf("AIP deletion report: get storage location: %w", err)
-	}
-
-	b, err := loc.OpenBucket(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("AIP deletion report: open storage bucket: %w", err)
+	if err := a.write(ctx, data, key); err != nil {
+		return nil, fmt.Errorf("AIP deletion report: %w", err)
```
jraddaoui: Should we not wrap the error here?

internal/storage/activities/aip_deletion_report.go
```diff
@@ -51,38 +60,35 @@ func (a *AIPDeletionReportActivity) Execute(
...
+		},
+	)
 	if err != nil {
-		return nil, fmt.Errorf("AIP deletion report: open bucket writer: %w", err)
-	}
-	defer w.Close()
-
-	if err := dr.Write(ctx, data, w); err != nil {
-		return nil, fmt.Errorf("AIP deletion report: write report to bucket: %w", err)
+		return nil, fmt.Errorf("AIP deletion report: update deletion request: %w", err)
```
jraddaoui: And here?

internal/storage/activities/aip_deletion_report.go
```diff
@@ -137,3 +144,55 @@ func (a *AIPDeletionReportActivity) loadData(
...
+	b, err := loc.OpenBucket(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("open storage bucket: %w", err)
+	}
+	defer b.Close()
+
+	w, err := b.NewWriter(ctx, key, nil)
+	if err != nil {
+		return nil, fmt.Errorf("open bucket writer: %w", err)
+	}
```
jraddaoui, line 163: And these?

internal/storage/activities/aip_deletion_report.go
```diff
@@ -137,3 +144,55 @@ func (a *AIPDeletionReportActivity) loadData(
...
+	data.ReportTimestamp = a.clock.Now().UTC()
+
+	jsonData, err := json.Marshal(data)
+	if err != nil {
+		return fmt.Errorf("marshal JSON: %w", err)
+	}
+
+	if err := a.formFiller.FillForm(src, bytes.NewReader(jsonData), w); err != nil {
+		return fmt.Errorf("fill form: %w", err)
+	}
```
jraddaoui, line 195: Also these three?

internal/storage/pdfs/pdf.go
```diff
@@ -0,0 +1,10 @@
+package pdf
+
+import "io"
+
+// FormFiller defines the interface for filling PDF forms.
+type FormFiller interface {
+	// FillForm fills the src PDF form fields with JSON formatted data and
+	// writes the resulting filled PDF to dest.
+	FillForm(src io.ReadSeeker, data io.Reader, dest io.Writer) error
+}
```
jraddaoui, line 10: `pdf.FormFiller.FillForm()` feels a little redundant, `pdf.FormFiller.Fill()` or even `pdf.Form.Fill()` sound better to me.

internal/storage/persistence/ent/schema/deletion_request.go
```diff
@@ -79,6 +79,9 @@ func (DeletionRequest) Fields() []ent.Field {
 		field.Int("workflow_id").
 			Positive().
 			Optional(),
+		field.String("deletion_report_key").
+			Annotations(entsql.Annotation{Size: 1024}).
+			Optional(),
```
jraddaoui: Similar, I'd have called the new column/field `report_key` as it's already part of the `deletion_request` table/types.

internal/storage/types/deletion_report_data.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type DeletionReportData struct {
+	DeletionRequestDBID int
```
jraddaoui, line 11: As we said, we should use the UUID as the persistence interface identifier. I know the update expects the DBID, something for another PR.

jraddaoui (approved):
Nice @djjuhasz, thanks! Some things I'd have named differently and the usual question about `%w` vs `%v` in the errors. Feel free to ignore.

internal/storage/activities/aip_deletion_report_test.go
```diff
@@ -79,13 +80,38 @@ func expectLocation(t *testing.T, msvc *fake.MockService) {
...
-func defaultExpects(t *testing.T, msvc *fake.MockService, aipID uuid.UUID) {
+func expectUpdateDeletionRequest(msvc *fake.MockService, drDBID int) {
+	msvc.EXPECT().
+		UpdateDeletionRequest(
+			mockutil.Context(),
+			drDBID,
+			mockutil.Func(
+				"Updates deletion report key",
+				func(updater persistence.DeletionRequestUpdater) error {
+					_, err := updater(&types.DeletionRequest{})
```
jraddaoui: Just realized while working on another tests, we could check the actual values being updated:

http://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/local_activities_test.go#L173-L181

cmd/enduro/main.go
djjuhasz: The small `pdf.FormFiller` interface is inspired by this Go article I read recently: https://rednafi.com/go/interface-segregation/ (it was linked the weekly Go newsletter).  The idea is that `pdf.PDFCPU` could implement other PDF interfaces in addition to the `pdf.FormFiller` interface, rather than having one big `pdf.Service` interface (or whatever we chose to call it).

Based on this design `pdf.NewPDFCPU()` doesn't return a `pdf.FormFiller`, it returns a struct that implements `pdf.FormFiller`, but could implement other interfaces as well.

internal/storage/activities/aip_deletion_report.go
djjuhasz: Sure.

internal/storage/activities/aip_deletion_report.go
djjuhasz: Oh yeah, I got the `%w` mixed up with `%v`. :(

internal/storage/pdfs/pdf.go
djjuhasz: `FormFiller` is an interface though, so I wouldn't expect to see a `pdf.FormFiller.FillForm()` call. :thinking: 

Maybe it would be better to name the package `pdfs` though, so we could do:

```
pdf := pdfs.NewPDFCPU()
pdf.FillForm(src, data, dest)
```

internal/storage/activities/aip_deletion_report.go
djjuhasz: Using the last error message in both cases changes: 
"AIP deletion report: template file does not exist: non_existent.tmpl.pdf", 

to: 
"AIP deletion report: error reading template file: stat non_existent.tmpl.pdf: no such file or directory"

Which is definitely a bit harder to read.  I think I'll stick with the two separate messages, though I will shorten down the second one.

internal/storage/pdfs/pdf.go
djjuhasz: I renamed the package `pdf` -> `pdfs`

internal/storage/activities/aip_deletion_report_test.go
djjuhasz: Ah, cool! :sunglasses:

PR #1416, opened 2025-11-18
Fix Tilt CI

jraddaoui:
Main fix was increasing the [readiness timeout](https://docs.tilt.dev/api.html#api.ci_settings). But I also saw the following error while testing:

```
[event: pod enduro-sdps/enduro-a3m-0] The node was low on resource: ephemeral-storage.
```

So I lowered the eviction threshold in the Tilt CI cluster to 1%, from the defaults nodefs.available<10% and imagefs.available<15%:

https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#hard-eviction-thresholds

Also upgraded Tilt to its latest version.

### DIFF
--- a/.github/workflows/tilt-ci.yml
+++ b/.github/workflows/tilt-ci.yml
@@ -23,6 +23,6 @@ jobs:
         with:
           setup-tools: |
             tilt
-          tilt: v0.33.11
+          tilt: v0.35.2
       - name: Check tilt ci
         run: timeout 600 tilt ci
--- a/.github/workflows/tilt-ci.yml
+++ b/.github/workflows/tilt-ci.yml
@@ -26,4 +26,4 @@ jobs:
             tilt
           tilt: v0.35.2
       - name: Check tilt ci
-        run: timeout 600 tilt ci
+        run: tilt ci
--- a/.github/workflows/tilt-ci.yml
+++ b/.github/workflows/tilt-ci.yml
@@ -18,6 +18,7 @@ jobs:
             --registry-create sdps-ci-registry
             --no-lb
             --k3s-arg "--no-deploy=traefik,servicelb,metrics-server@server:*"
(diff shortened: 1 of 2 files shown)


.github/workflows/tilt-ci.yml
```diff
@@ -18,11 +18,12 @@ jobs:
             --registry-create sdps-ci-registry
             --no-lb
             --k3s-arg "--no-deploy=traefik,servicelb,metrics-server@server:*"
+            --k3s-arg "--kubelet-arg=eviction-hard=imagefs.available<1%,nodefs.available<1%@server:*"
```
djjuhasz, line 21: I'm a bit worried that the low 1% filesystem eviction limit will cause us problems down the line with the whole node running out of disk space and crashing.  

Looking up the [Github hosted runners](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) I think our standard runner have 14GB of disk space which only give us 140 MB (1%) of buffer between the eviction limit and the disk filling up. A 140MB buffer might be enough, but if it fills up too quickly I think we run the risk of the entire node crashing.

Is it possible the a3m pod was running out of ephemeral space because of excessive logging due to other problems?  If not maybe we should look into why it's filling up the available space so quickly?

djjuhasz (approved):
Thank you so much for looking into the Tilt CI problems @jraddaoui, and for this solution.  I added one note about the lowered eviction limits, but I think we can try out the PR as is, and see how it works out.

.github/workflows/tilt-ci.yml
jraddaoui: Thanks @djjuhasz, yeah, I even thought about adding a new Tilt env var to avoid including some components in CI. Hopefully we don't hit that limit for a while and when we do, if it fails or if the node crashes entirely, we can take another look.

PR #1421, opened 2025-11-20
Add tests for ingest service add SIP endpoint

jraddaoui:
Also, remove goaWrapper from ingest service and move tests to ingest_test package. Aims to normalize with the storage service implementation, and the usage of the new service constructor and params.

Refs #1405.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -54,7 +54,7 @@ func HTTPServer(
 	}
 
 	// Ingest service.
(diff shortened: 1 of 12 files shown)


jraddaoui:
@djjuhasz, I think it will make more sense reviewing commit by commit ;)

djjuhasz (approved):
LGTM. Thanks @jraddaoui.

PR #1422, opened 2025-11-20
Add batch workflow skeleton

jraddaoui:
Unrelated but based in #1421, as I need both for a following PR. Refs #1405.

### DIFF
--- a/internal/ingest/workflows.go
+++ b/internal/ingest/workflows.go
@@ -9,55 +9,94 @@ import (
 	temporalsdk_api_enums "go.temporal.io/api/enums/v1"
 	temporalsdk_client "go.temporal.io/sdk/client"
 
+	"github.com/artefactual-sdps/enduro/internal/datatypes"
 	"github.com/artefactual-sdps/enduro/internal/enums"
 )
 
 const (
-	// The name of the SIP processing workflow.
+	// BatchWorkflowName is the name of the Batch processing workflow.
+	BatchWorkflowName = "batch-workflow"
+
+	// ProcessingWorkflowName is the name of the SIP processing workflow.
 	ProcessingWorkflowName = "processing-workflow"
 
-	// The name of the signal for reviewing a SIP/AIP.
+	// ReviewPerformedSignalName is the name of the signal for reviewing a SIP/AIP.
 	ReviewPerformedSignalName = "review-performed-signal"
 )
 
-type ReviewPerformedSignal struct {
-	Accepted   bool
-	LocationID *uuid.UUID
-}
+type (
+	BatchWorkflowRequest struct {
+		// Batch contains the Batch details.
+		Batch datatypes.Batch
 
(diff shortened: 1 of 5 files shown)


internal/workflow/batch.go
```diff
@@ -0,0 +1,46 @@
...
+	return &BatchWorkflow{
+		cfg:       cfg,
+		rng:       rng,
+		ingestsvc: ingestsvc,
+		tc:        tc,
+	}
+}
+
+func (w *BatchWorkflow) Execute(ctx temporalsdk_workflow.Context, req *ingest.BatchWorkflowRequest) (e error) {
+	w.logger = temporalsdk_workflow.GetLogger(ctx)
```
djjuhasz: This doesn't seem thread-safe — if `Execute()` is run from multiple threads they will be overwriting `w.logger` on each other.

internal/ingest/workflows.go
```diff
@@ -18,6 +19,9 @@ const (
 
 	// The name of the signal for reviewing a SIP/AIP.
 	ReviewPerformedSignalName = "review-performed-signal"
+
+	// The name of the Batch processing workflow.
+	BatchWorkflowName = "batch-workflow"
```
djjuhasz: Nitpick: I'd put the `BatchWorkflowName` first in the `const` declaration because it comes first alphabetically.

internal/ingest/workflows.go
```diff
@@ -60,6 +64,21 @@ type ProcessingWorkflowRequest struct {
 	Extension string
 }
 
+type BatchWorkflowRequest struct {
+	// The Batch details.
```
djjuhasz: Declaration comments should be complete sentences, e.g. "// Batch represents a batch of SIPs." See: https://go.dev/wiki/CodeReviewComments#comment-sentences. 

I know the `ProcessingWorkflowRequest` comments don't conform to this rule, we should go back and rewrite them as sentences at some point.

djjuhasz (changes requested):
Looks pretty good @jraddaoui. I think I spotted one data race and flagged a couple of minor style issues.

internal/workflow/batch.go
jraddaoui: Good catch, I'll fix in the processing workflow too!

internal/ingest/workflows.go
jraddaoui: Re-ordered the entire file and put all types into a `type ()` section.

internal/ingest/workflows.go
jraddaoui: Done for all comments in this file.

djjuhasz (approved):
Looks great, thanks @jraddaoui.

Using `state.logger` is a creative solution for passing around the logger. 😁  I guess we can't set `w.logger` in the activity constructor because we don't have the context. 🤔

PR #1423, opened 2025-11-20
Replace v-html usages with safer rendering patterns

sevein:
I think I've found functionally equivalent alternatives to v-html in all the places we were using it, except for the content-embedding logic on the home page. I've moved that part into a dedicated component called SafeHtml to give it better visibility. I haven't been able to test any of this since I don't currently have a running instance, so please review and test it when you can.


dashboard/src/components/EmailLinkedText.vue
```diff
@@ -0,0 +1,29 @@
...
+<template>
+  <span>
+    <template v-for="(part, index) in parts" :key="index">
+      <template v-if="isEmail(part)">
+        <a :href="'mailto:' + part">{{ part }}</a>
+      </template>
+      <template v-else>
+        {{ part }}
+      </template>
+    </template>
```
sevein, line 27: I did not find this intuitive, and if that's also the case for you, please scroll down and look up the test file (`EmailLinkedText.test.ts`) because it's much easier to see the results that way.

dashboard/src/components/InstitutionLogo.vue
```diff
@@ -1,22 +1,24 @@
...
-  <div v-else-if="props.logo" class="d-none d-sm-block mx-3" v-html="imgTag" />
 </template>
+
+<style scoped>
+img {
+  max-height: 30px;
+  height: auto;
+  width: auto;
+}
+</style>
```
sevein, line 24: Looks like this is the safe replacement for the old `height="30"`, so we don't have to write that twice.

jraddaoui:
Thanks @sevein! I gave it a try locally in prod and dev mode, and I added a couple of commits to fix some icons and a Vue warning. I'll merge!

PR #1424, opened 2025-11-20
Implement add Batch API endpoint

jraddaoui:
Based in #1422. Refs #1405.

### DIFF
--- a/internal/ingest/goa_batch.go
+++ b/internal/ingest/goa_batch.go
@@ -3,8 +3,13 @@ package ingest
 import (
 	"context"
 	"errors"
+	"fmt"
+
+	"github.com/google/uuid"
 
 	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
+	"github.com/artefactual-sdps/enduro/internal/datatypes"
+	"github.com/artefactual-sdps/enduro/internal/enums"
 )
 
 var ErrNotImplemented error = goaingest.MakeNotImplemented(errors.New("not implemented"))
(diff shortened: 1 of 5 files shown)


internal/ingest/goa_batch.go
```diff
@@ -13,7 +18,69 @@ func (svc *ingestImpl) AddBatch(
...
+	}
+	b := &datatypes.Batch{
+		UUID:       bUUID,
+		Status:     enums.BatchStatusQueued,
+		Identifier: identifier,
+		SIPSCount:  len(payload.Keys),
+	}
+
+	if claims != nil {
+		b.Uploader = &datatypes.User{
```
djjuhasz: Isn't this generating a new user for every batch created? I think we want to link to an existing user when possible.

internal/ingest/goa_batch.go
```diff
@@ -13,7 +18,69 @@ func (svc *ingestImpl) AddBatch(
...
+			UUID:    uuid.Must(uuid.NewRandomFromReader(svc.rander)),
```
djjuhasz, line 54: It seems strange to me to generate a UUID for the user that may or may not be used, I assumed this was going to create a new user every time. 😕  I'm assuming the `perSvc..CreateBatch()` method isn't generating a UUID when needed because we don't want to inject a random number generator, but I think we should review this decision in respect to the making the API more intuitive.

djjuhasz (approved):
LGTM @jraddaoui. I made one note about how we are handling user UUIDs when creating batches, but it's not a blocker.

internal/ingest/goa_batch.go
jraddaoui: Yeah, we discussed this in https://github.com/artefactual-sdps/enduro/pull/1315#discussion_r2257879992 and other threads from that PR, I'm happy to revisit at some point.

PR #1426, opened 2025-11-21
Update Batch status and dates from workflow

jraddaoui:
Add failed status, use time.Time for all dates and update Batch from workflow. The state doesn't make a lot of sense yet, just trying to split the work into smaller PRs.

Refs #1405.

### DIFF
--- a/internal/workflow/batch.go
+++ b/internal/workflow/batch.go
@@ -1,12 +1,15 @@
 package workflow
 
 import (
+	"errors"
+	"fmt"
 	"io"
 
 	temporalsdk_client "go.temporal.io/sdk/client"
 	temporalsdk_workflow "go.temporal.io/sdk/workflow"
 
 	"github.com/artefactual-sdps/enduro/internal/config"
+	"github.com/artefactual-sdps/enduro/internal/enums"
 	"github.com/artefactual-sdps/enduro/internal/ingest"
 )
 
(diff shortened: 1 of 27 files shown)


internal/workflow/batch.go
```diff
@@ -32,13 +35,53 @@ func NewBatchWorkflow(
...
+
+	defer func() {
+		// Update final batch status and add completion date.
+		state.batch.CompletedAt = temporalsdk_workflow.Now(ctx)
+		state.batch.Status = enums.BatchStatusIngested
+		if e != nil {
+			state.batch.Status = enums.BatchStatusFailed
+		}
+		if err := w.updateBatch(ctx, state); err != nil {
+			e = errors.Join(e, err)
```
djjuhasz, line 55: I don't think it's possible to modify the return value of `e` in this deferred function because the surrounding `Execute()` function returns `e` before the defer runs.

From https://go.dev/blog/defer-panic-and-recover:
"The list of saved [defer] calls is executed after the surrounding function returns."

Code test:
https://go.dev/play/p/7tRrgRB32dS

So if you want an `updateBatch()` error to be visible to a developer I think you'll need to log it.

internal/workflow/batch.go
```diff
@@ -32,13 +35,53 @@ func NewBatchWorkflow(
...
+		}
+	}()
+
+	// Update batch status to Processing and add start date.
```
djjuhasz: Nitpick: the actual value of "processing" is not capitalized, so I would quote the actual value here, e.g.

```
// Update batch status to "processing" and add start date.
```

internal/workflow/local_activities_test.go
```diff
@@ -287,3 +287,88 @@ func TestUpdateSIPLocalActivity(t *testing.T) {
...
+				svc.EXPECT().
+					UpdateBatch(
+						ctx,
+						batchUUID,
+						mockutil.Func(
+							"should update batch",
+							func(updater persistence.BatchUpdater) error {
+								b, err := updater(&datatypes.Batch{})
+								assert.NilError(t, err)
+								assert.DeepEqual(t, b.Status, enums.BatchStatusProcessing)
```
djjuhasz: Can't you just do:
```
assert.DeepEqual(t, b, &datatypes.Batch{
    Status:  enums.BatchStatusProcessing,
    StartedAt: startedAt,
    CompletedAt: completedAt
})
```

If not, then I think `assert.Equal(t, b.Status, enums.BatchStatusProcessing)` is sufficient.

djjuhasz (approved):
I didn't spot any major problems @jraddaoui, but I noted a few minor things.  Thanks!

internal/workflow/batch.go
jraddaoui: It needs to be named it in the function signature. And you don't even have to use that variable in the returns, but it will be set in the deferred functions. We have been doing this in a lot of places.

Code test fixed:
https://go.dev/play/p/bGZTazK2ZfC

djjuhasz: Wow, I totally expected:
```
func run() error {
    var e error
    ...
    return e
}
```

to be 100% equivalent to:
```
func run() (e error) {
    ...
    return e
}
```

But you're right, the second version allows the defer to modify the returned error. :exploding_head:

PR #1429, opened 2025-11-24
Enable revive:package-directory-mismatch linter

sevein:
While revisiting https://github.com/artefactual-sdps/enduro/pull/1048, I noticed that the internal package `github.com/artefactual-sdps/enduro/internal/persistence/ent/client` declares its name as `entclient` instead of `client`. Go allows this kind of mismatch, but it can be misleading. I've enabled the [`package-directory-mismatch`](https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md#package-directory-mismatch) rule in revive to warn us if this happens again.

### DIFF
--- a/.bine.json
+++ b/.bine.json
@@ -33,7 +33,7 @@
     {
       "name": "golangci-lint",
       "url": "https://github.com/golangci/golangci-lint",
-      "version": "2.4.0",
+      "version": "2.6.2",
       "asset_pattern": "{name}-{version}-{goos}-{goarch}.tar.gz"
     },
     {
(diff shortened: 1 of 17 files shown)


.bine.json
```diff
@@ -33,7 +33,7 @@
     {
       "name": "golangci-lint",
       "url": "https://github.com/golangci/golangci-lint",
-      "version": "2.4.0",
+      "version": "2.6.2",
```
sevein, line 36: The upgrade is required since `package-directory-mismatch` isn't available in golangci-lint v2.4.0.

jraddaoui (approved):
Thanks @sevein!

PR #1433, opened 2025-11-24
Wait for Batch signal after SIP validation

jraddaoui:
Add a `BatchUUID` param to the processing workflow request to indicate
the SIP belongs to a Batch. Mark the SIP as validated and wait for Batch
signal in the processing workflow. Set the workflow and SIP statuses to
canceled if the signal indicates to not continue, continue processing
otherwise.

Add `Canceled` status for ingest Workflows and SIPs and `Validated`
status for SIPs.

Refs #1405.


djjuhasz (approved):
Looks great! Thanks @jraddaoui! 👍

PR #1435, opened 2025-11-24
Bump minimum Tilt version to v0.35.0

djjuhasz:
Commit 9f36343 added `ci_settings(readiness_timeout)` to the Tiltfile, which was introduced in Tilt v0.35.0:
https://github.com/tilt-dev/tilt/releases/tag/v0.35.0


jraddaoui (approved):
Thanks @djjuhasz!!

PR #1437, opened 2025-11-25
Start SIP workflows from Batch workflow

jraddaoui:
Create SIPs, start their workflows and wait for completion.

Based in #1433. Refs #1405.

### DIFF
--- a/internal/workflow/batch.go
+++ b/internal/workflow/batch.go
@@ -4,11 +4,15 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"time"
 
+	"github.com/google/uuid"
+	temporalapi_enums "go.temporal.io/api/enums/v1"
 	temporalsdk_client "go.temporal.io/sdk/client"
 	temporalsdk_workflow "go.temporal.io/sdk/workflow"
 
 	"github.com/artefactual-sdps/enduro/internal/config"
+	"github.com/artefactual-sdps/enduro/internal/datatypes"
 	"github.com/artefactual-sdps/enduro/internal/enums"
 	"github.com/artefactual-sdps/enduro/internal/ingest"
 )
@@ -38,8 +42,8 @@ func (w *BatchWorkflow) Execute(ctx temporalsdk_workflow.Context, req *ingest.Ba
 	state := newBatchWorkflowState(ctx, req)
 	state.logger.Info(
 		"Starting batch workflow",
-		"batch_uuid", req.Batch.UUID.String(),
-		"batch_identifier", req.Batch.Identifier,
+		"uuid", req.Batch.UUID.String(),
+		"identifier", req.Batch.Identifier,
 		"source_id", req.SIPSourceID.String(),
 		"keys", req.Keys,
 	)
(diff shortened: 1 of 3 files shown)


internal/workflow/batch.go
```diff
@@ -85,3 +114,92 @@ func (w *BatchWorkflow) updateBatch(ctx temporalsdk_workflow.Context, state *bat
...
+		return fmt.Errorf("create SIP: %v", err)
+	}
+
+	// Start processing workflow for the SIP, keeping track of the workflow future and execution.
+	var we temporalsdk_workflow.Execution
+	processingCtx := temporalsdk_workflow.WithChildOptions(ctx, temporalsdk_workflow.ChildWorkflowOptions{
+		Namespace:         w.cfg.Temporal.Namespace,
+		TaskQueue:         w.cfg.Temporal.TaskQueue,
+		WorkflowID:        fmt.Sprintf("%s-%s", ingest.ProcessingWorkflowName, sipUUID.String()),
+		ParentClosePolicy: temporalapi_enums.PARENT_CLOSE_POLICY_TERMINATE,
```
djjuhasz, line 159: For asynchronous child workflows the docs say you need to use an "abandon" `ParentClosePolicy`: https://docs.temporal.io/develop/go/child-workflows#async-child-workflows

internal/workflow/batch.go
```diff
@@ -85,3 +114,92 @@ func (w *BatchWorkflow) updateBatch(ctx temporalsdk_workflow.Context, state *bat
...
+	for _, sd := range state.sipDetails {
+		selector.AddFuture(sd.workflowFuture, func(f temporalsdk_workflow.Future) {
+			// Ignore error and result, we just want to know when it's done.
+			_ = f.Get(ctx, nil)
+			state.sipsCompleted++
+		})
+	}
+
+	// Wait for all SIP workflows to complete or context cancellation.
+	for !state.allSIPsCompleted() {
```
djjuhasz: I think you can just do `for range len(state.sipDetails) {` rather than checking `state.allSIPsCompleted()` on each iteration, since selected future always increments `sipsCompleted`.

djjuhasz (commented):
Pretty cool @jraddaoui.  It took me a while to figure out how the fan-out / fan-in logic worked and which parts were acting asynchronously, but I think I get it now.

I just had one note about the child workflow close policy, and one suggestion for the fan-in loop.

internal/workflow/batch.go
jraddaoui: That only determines what happens to the child workflow when the parent ends. We use the two already in the processing workflow, for the preprocessing child workflow we want to block and wait until it ends and finish it if the parent for some reason ends ([PARENT_CLOSE_POLICY_TERMINATE](https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/processing.go#L1022)), but for the poststorage we want them to keep running even if the parent ends because we don't care about the result ATM ([PARENT_CLOSE_POLICY_ABANDON](https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/processing.go#L1089)). Something that the documentation doesn't say is that abandon also requires a disconnected context to work properly. Nevertheless, in both cases the child workflow is started by Temporal by the time you get the workflow execution details and it runs asynchronously.

In this case we are waiting for the child workflows to end, we don't care about the result for now, but we want to keep control of them. We may need to improve the logic in the processing workflow to handle termination/cancellation properly, but I think terminate is going to be better to handle batch integrity properly.

internal/workflow/batch.go
jraddaoui: Nice! I've removed `sipsCompleted` and `allSIPsCompleted` from the state.

internal/workflow/batch.go
djjuhasz: Huh, I don't understand why the docs would have a separate "Async child workflows" section if all workflow run asynchronously, but the child workflow docs are hard to understand in general IMHO. In any case I agree that it's better to terminate any SIP workflows that are still running if the batch workflow fails.

djjuhasz (approved):
👍

jraddaoui:
Thanks @djjuhasz!

internal/workflow/batch.go
jraddaoui: I think they say async meaning the parent is not running at the same time as the child. But yeah, those docs could be better.

PR #1438, opened 2025-11-27
Implement CreateTasks with CreateBulk (take two)

sevein:
This pull request is a refined second take on https://github.com/artefactual-sdps/enduro/pull/1048. These are the improvements over the previous implementation:

- No slice returned from persistence: The new `CreateTasks` API updates the original task pointers in-place with database-generated IDs, rather than returning a new slice. This reduces allocations and simplifies the interface.
- Idiomatic iterator usage: Uses the `TaskSequence` type alias (`iter.Seq[*datatypes.Task]`), making the API easier to understand.
- Single transaction for all batches: All tasks are inserted in a single transaction, ensuring atomicity and consistency.
- Workflow ID caching: Caches workflow database IDs to avoid redundant lookups when multiple tasks share the same workflow UUID.
- Event emission via tee pattern: The ingest layer uses a tee pattern to capture task pointers for event publishing after persistence, eliminating the need for the persistence layer to return tasks.
- Renamed batching helper: The batching function is now called `chunk` instead of `batch` (name already taken!).

**UPDATE:** after review we don't see real gains from the iterator approach. Producers get a tiny ergonomic tweak, but at our scale (~100 tasks) there's no meaningful performance benefit, and the extra complexity isn't justified. See https://github.com/artefactual-sdps/enduro/pull/1438#pullrequestreview-3542189466 for more details.


sevein:
I've just tested the a3m path and it seems to be working fine. Now we see the bulk insert when saving the tasks:

image

sevein:
It isn't directly related, but the traces also show that we're publishing events one at a time. That is probably another area for improvement.

image

jraddaoui (approved):
Thanks @sevein, this looks great to me!

It took me a while to understand the four levels `source -> tee -> chunk -> consumer` and how breaking the loop in the consumer propagates backward as `false` through `yield` to stop the entire generation.

I gave it a try locally and the preprocessing workflow tasks and the Archivematica jobs are also being recorder as expected.

I think @djjuhasz was looking at this already but had focus on other release related things, so I'd wait for his feedback too.

sevein:
Thanks for taking the time!

> It took me a while to understand the four levels source -> tee -> chunk -> consumer and how breaking the loop in the consumer propagates backward as false through yield to stop the entire generation.

Yeah, I'm not fond of the complexity required but hopefully is mostly reusable code? Maybe `tee` can be also implemented as a generic helper or we can find some other way to publish the events like passing an onCreated callback to persistence.

I asked Nano Banana Pro to generate a diagram from this pull request, I should have done this earlier!

image

internal/persistence/ent/client/chunk.go
```diff
@@ -0,0 +1,31 @@
+package client
+
+import "iter"
+
+var defaultTaskBatchSize = 100
```
djjuhasz, line 5: Is there some reason to call this "defaultTaskBatchSize" instead of "defaultBatchSize"? It seems like `chunk` is designed to work with any type, not just tasks.

internal/persistence/ent/client/chunk_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			setDefaultTaskBatchSize(t, tc.size)
+			seq := chunk(slices.Values(tc.values))
+
+			got := [][]int{}
+			for batch := range seq {
+				cp := make([]int, len(batch))
+				copy(cp, batch)
```
djjuhasz, line 73: Why the copy?

internal/ingest/task_test.go
```diff
@@ -50,10 +52,18 @@ func TestCreateTask(t *testing.T) {
...
-					CreateTask(mockutil.Context(), &task).
+					CreateTasks(mockutil.Context(), gomock.Any()).
 					DoAndReturn(
-						func(ctx context.Context, task *datatypes.Task) error {
-							task.ID = 1
+						func(
+							ctx context.Context,
+							seq persistence.TaskSequence,
+						) error {
+							var seen []*datatypes.Task
```
djjuhasz: What's the purpose of `seen`?

internal/ingest/task_test.go
```diff
@@ -50,10 +52,18 @@ func TestCreateTask(t *testing.T) {
...
+							seq(func(t *datatypes.Task) bool {
+								seen = append(seen, t)
+								return true
+							})
+							seen[0].ID = 1
```
djjuhasz: And why do this?

internal/persistence/ent/client/task.go
```diff
@@ -69,6 +70,94 @@ func (c *client) CreateTask(ctx context.Context, task *datatypes.Task) error {
...
+			builders = append(builders, builder)
+			originals = append(originals, task)
+		}
+
+		created, err := tx.Task.CreateBulk(builders...).Save(ctx)
+		if err != nil {
+			return rollback(tx, newDBErrorWithDetails(err, "create tasks"))
+		}
+
+		if len(created) != len(originals) {
```
djjuhasz, line 139: This seems unnecessary to me. I can't imagine a scenario where the number of rows created would be different than the number of originals without an error being returned.

djjuhasz:
@sevein I am working my way through this PR, but it's pretty slow going.  I'll come back it tomorrow.

internal/persistence/persistence.go
```diff
@@ -39,6 +44,15 @@ type Service interface {
 	CreateWorkflow(context.Context, *datatypes.Workflow) error
 
 	CreateTask(context.Context, *datatypes.Task) error
+	// CreateTasks persists all tasks yielded by the sequence. For very large
+	// sequences, the transaction may remain open for an extended period,
+	// potentially causing lock contention or timeout issues.
+	//
+	// Implementors must consume the sequence in batches and stop on the first
```
djjuhasz, line 51: It seem overly prescriptive to me to specify the implementation details like this. What if the back-end is a data store that doesn't allow batch inserts or support transaction locks?  I get why you've implemented the Ent/MySQL method as you have, but I don't see why a different implementation shouldn't make different choices if it leads to better performance.

internal/persistence/telemetry.go
```diff
@@ -126,6 +126,19 @@ func (w *wrapper) CreateTask(ctx context.Context, task *datatypes.Task) error {
 	return nil
 }
 
+func (w *wrapper) CreateTasks(ctx context.Context, seq TaskSequence) error {
```
djjuhasz, line 129: It would be nice to have a unit test for this method.

internal/workflow/localact/save_preprocessing_tasks.go
```diff
@@ -39,16 +40,22 @@ func SavePreprocessingTasksActivity(
...
-	for _, t := range params.Tasks {
-		task := preprocTaskToTask(t)
-		task.WorkflowUUID = params.WorkflowUUID
-		// TODO: Create deterministic UUIDs and make activities idempotent.
-		task.UUID = uuid.Must(uuid.NewRandomFromReader(params.RNG))
-
-		if err := params.Ingestsvc.CreateTask(ctx, &task); err != nil {
-			return &res, fmt.Errorf("SavePreprocessingTasksActivity: %v", err)
+	seq := persistence.TaskSequence(func(yield func(*datatypes.Task) bool) {
+		for _, t := range params.Tasks {
```
djjuhasz, line 44: There doesn't seem to be any advantage in this case to using an iterator, as we already have the slice of `[]*datatypes.Task` in memory.  I understand that an iterator is useful in other situations and it doesn't make sense to have a different `CreateTasks()` method that takes a slice, I'm just noting that I don't think it's a benefit here.

internal/a3m/a3m_test.go
```diff
@@ -69,18 +70,15 @@ func TestCreateAIPActivity(t *testing.T) {
...
-				return true
-			})
-			assert.Equal(t, len(got), 1)
-			assert.DeepEqual(t, got[0], &datatypes.Task{
-				UUID:   taskUUID,
-				Status: enums.TaskStatusDone,
-				StartedAt: sql.NullTime{
-					Time:  time.Date(2009, time.November, 10, 23, 0, 0, 0, time.UTC),
-					Valid: true,
+			got := slices.Collect(seq)
```
djjuhasz, line 73: Why is `slices.Collect(seq) better here than then calling `seq()` directly? I'm just curious for my own learning.

internal/a3m/a3m.go
```diff
@@ -182,28 +183,38 @@ func saveTasks(
...
+				Name:   job.Name,
+				Status: jobStatusToTaskStatus[job.Status],
+				StartedAt: sql.NullTime{
+					Time:  job.StartTime.AsTime(),
+					Valid: true,
+				},
+				WorkflowUUID: wUUID,
+			}
+
+			if !yield(&task) {
```
djjuhasz, line 205: Is there a memory or performance advantage to using an iterator here, given that the `&task` is yielded back to the caller, then stored as an slice via `tee()`?

internal/am/job_tracker.go
```diff
@@ -143,18 +147,23 @@ func (jt *JobTracker) saveTasks(ctx context.Context, jobs []amclient.Job) (int,
 			return 0, err
 		}
 		task.WorkflowUUID = jt.workflowUUID
+		tasks = append(tasks, task)
```
djjuhasz, line 170: Why isn't the task created and yielded in an iterator here like it is in the a3m case?

internal/am/job_tracker.go
```diff
@@ -143,18 +147,23 @@ func (jt *JobTracker) saveTasks(ctx context.Context, jobs []amclient.Job) (int,
...
-			return 0, err
-		}
+	if len(tasks) == 0 {
+		return 0, nil
+	}
 
-		// Add this job ID to the list of savedIDs.
-		jt.savedIDs[job.ID] = struct{}{}
-		count++
+	if err := jt.ingestsvc.CreateTasks(ctx, persistence.TaskSequence(slices.Values(tasks))); err != nil {
```
djjuhasz, line 178: Again, I don't understand the benefit to passing a value from an in-memory slice to an iterator, then saving the `&datatype.Task()` to an array after yielding it back to the caller. :confused:

internal/timerange/timerange_test.go
```diff
@@ -88,7 +88,10 @@ func TestIsInstant(t *testing.T) {
 	t.Run("Returns false when start and end time are not equal", func(t *testing.T) {
 		t.Parallel()
 
-		r, err := timerange.New(time.Now(), time.Now())
+		start := time.Date(2024, 9, 17, 0, 0, 0, 0, time.UTC)
+		end := time.Date(2024, 9, 18, 0, 0, 0, 0, time.UTC)
+
+		r, err := timerange.New(start, end)
```
djjuhasz, line 94: Oh, thanks for fixing this!  Good catch. :)

internal/am/job_tracker.go
```diff
@@ -98,22 +104,36 @@ func NewJobTracker(
...
 		savedIDs:     make(map[string]struct{}),
+		tracer:       tracer,
 	}
 }
 
 // SaveTasks queries the Archivematica jobs list endpoint to get a
 // list of completed jobs related to the transfer or ingest identified by
 // unitID, then saves any new jobs as tasks.
 func (jt *JobTracker) SaveTasks(ctx context.Context, unitID string) (int, error) {
+	ctx, span := jt.tracer.Start(ctx, "JobTracker.SaveTasks")
```
djjuhasz, line 115: Nice! It's great to get telemetry on this.

internal/ingest/task.go
```diff
@@ -4,23 +4,47 @@ import (
...
+func (svc *ingestImpl) CreateTasks(
+	ctx context.Context,
+	seq persistence.TaskSequence,
+) error {
+	// Tee the sequence to capture task pointers while forwarding them to the
+	// persistence layer. This allows us to access the tasks (with their
+	// database-generated IDs) after persistence completes, so we can publish
+	// events without consuming the sequence twice or changing the persistence
+	// API.
+	var tasks []*datatypes.Task
```
djjuhasz, line 30: This could probably benefit from pre-allocating the slice with a larger capacity to minimized the number of copies to a new slice when the capacity is exceeded.  Considering the number of tasks passed through here is usually between one and hundred maybe something like `tasks := make([]*datatypes.Task, 0, 100)`?

djjuhasz (commented):
Thanks for submitting this @sevein, it's definitely interesting to learn more about Go iterators, and doing the task inserts in a batch is a very nice optimization.

On the other hand this doesn't seem like a good case for using an iterator to me, because we are getting the source data as a single big dataset, rather than a stream, and then we have to `tee()` the task pointers to a new slice in the ingest service anyway to publish the creation events.  Even if we can improve the event system so we don't have to return the created `datatype.Task` pointers, I don't think the extra complexity of using an iterator would provide a noticeable performance improvement  considering the cardinality of a "large" tasks insert is on the order of 100 rows. 🤔

internal/persistence/ent/client/chunk_test.go
sevein: Because the slice yielded by chunk (`batch` in the loop) is reused to avoid extra allocations, under the assumption that it isn't kept after persisting, but the test does keep it so it must copy.

E.g. this fails because slices.Collect stores the slices without copying:

```go
	for name, tc := range tests {
		t.Run(name, func(t *testing.T) {
			setDefaultTaskBatchSize(t, tc.size)
			seq := chunk(slices.Values(tc.values))
			assert.DeepEqual(t, slices.Collect(seq), tc.expected)
		})
	}
```

internal/persistence/ent/client/chunk.go
sevein: Yeah totally!

internal/ingest/task.go
sevein: I think you're right!

internal/persistence/ent/client/task.go
sevein: Makes sense.

internal/persistence/persistence.go
sevein: Good point, I should only state the required semantics (all-or-nothing, stop on first error...), perhaps something like the following:

```
// CreateTasks persists all tasks yielded by seq atomically: if it returns nil,
// every task was written and updated in place with generated fields (e.g. DB IDs);
// if any error occurs, no tasks are left written.
//
// The implementation should stop on the first error and must not consume seq
// more than once.
```

internal/workflow/localact/save_preprocessing_tasks.go
sevein: We start from `[]preprocessing.Task`, so the iterator lets us map each entry to a `*datatype.Task` and stream it without first building a `[]*datatypes.Task`. That's perhaps the small usability gain from the perspective of the producer if anything.

internal/a3m/a3m_test.go
sevein: Just less boilerplate I guess, I was just a bit annoyed by having to write this so many times:

```go
var got []*datatypes.Task
seq(func(t *datatypes.Task) bool {
	got = append(got, t)
	return true
})
```

internal/am/job_tracker.go
sevein: Yeah fair; this should mirror the pattern in a3m.go where I build a TaskSequence directly over jobs (converting and yielding and recording IDs) without the intermediate tasks slice.

sevein:
> On the other hand this doesn't seem like a good case for using an iterator to me, because we are getting the source data as a single big dataset, rather than a stream, and then we have to tee() the task pointers to a new slice in the ingest service anyway to publish the creation events. Even if we can improve the event system so we don't have to return the created datatype.Task pointers, I don't think the extra complexity of using an iterator would provide a noticeable performance improvement considering the cardinality of a "large" tasks insert is on the order of 100 rows. 🤔

Thanks for the review, David! You're right, because of the tee for events, the iterator adds almost no real benefit. Maybe there's a small producer ergonomics win, but questionable. I'll park this PR and come back with the simpler slice-based version. If we ever need a `CreateTasksIter` then this could be a good starting point. It was a fun exercise. Compared to #1048, it's wild how much better LLMs have gotten with Go generics/iterators (and a reminder they'll happily help build things we might not actually need 😆).

PR #1439, opened 2025-11-27
Add API "list deletion requests" endpoint

djjuhasz:
Refs #1207.

- Add a "list deletion requests" endpoint to the API.
- Rename old storage service `ListDeletionRequests()` method to `ListDeletionRequestsInternal()` to avoid conflicting with the Goa generated interface
- Add a storage service `ListDeletionRequests()` method that conforms to the Goa interface
- Add adapters to convert Goa structs to internal structs, and vice versa


internal/api/auth/attributes.go
```diff
@@ -14,6 +14,7 @@ const (
 	IngestSIPSourcesObjectsListAttr = "ingest:sipsources:objects:list"
 	IngestUsersListAttr             = "ingest:users:list"
 	StorageAIPSCreateAttr           = "storage:aips:create"
+	StorageAIPSDeletionListAttr     = "storage:aips:deletion:list"
```
jraddaoui, line 17: I was considering `deletion` an action prefix more than an entity in the API. If we want to consider it an entity we should use the plural in the attributes.

internal/api/design/storage.go
```diff
@@ -333,6 +333,32 @@ var _ = Service("storage", func() {
 			})
 		})
 	})
+	Method("list_deletion_requests", func() {
```
jraddaoui, line 336: It's an AIPs prefixed URL, we should name the method "list_aip_deletion_requests". Additionally, that will avoid the conflict with the internal function.

internal/api/design/storage.go
```diff
@@ -729,3 +755,41 @@ var AIPTask = ResultType("application/vnd.enduro.storage.aip.task", func() {
...
+	Required(
+		"uuid",
+		"aip_uuid",
+		"workflow_uuid",
+		"reason",
+		"requested_at",
+		"requester",
+		"status",
+	)
+})
```
jraddaoui, line 795: Similar to the action vs entity issue, I was trying to avoid introducing a `DeletionRequest` type/entity in the API, and I think this is too much just to expose the report key.

Also, looking at https://github.com/artefactual-sdps/enduro/pull/1440 and thinking ahead to the actual report download, we’ll only use it on the AIP view page. I think it would be much better to include a `deletion_report_key` field directly in the AIP type. This would only require a single join in the read AIP persistence method, which I think is much better than requiring another API request and database query to list/filter deletion requests. In that PR you’re not even showing the key value, just using it to decide whether the link should be displayed, and we shouldn’t need to send it for the download.

That approach also simplifies the attributes needed and avoids the entity vs action issue I mentioned. In the other PR you’re checking the `storage:aips:deletion:list` attribute to show the link, but I think we should use another attribute for that and the actual download, `storage:aips:deletion:downloadreport` (or similar).

jraddaoui (changes requested):
Thanks @djjuhasz, the code looks good. But I have some concerns that this approach might add unnecessary complexity to the API and persistence layer, and it’s also a bit different from how we’ve been handling deletion requests in general. Please, check my comments below and let me know your thoughts. And sorry we didn’t discuss this earlier.

internal/api/design/storage.go
djjuhasz: How would we join the `aip` table to the `deletion_request` table to get the report key, considering it's a one-to-many relationship?

internal/api/design/storage.go
djjuhasz: I considered using `ListAIPDeletionRequests()` for the public interface and `ListDeletionRequests()` for the interna method, but I think the distinction between the two is too subtle and it's not obvious from the names how they are different.

internal/api/auth/attributes.go
djjuhasz: Right, it should probably be "storage:aips:deletion_requests:list" or "storage:deletion_requests:list".

internal/api/design/storage.go
jraddaoui: Same as listing all deletion requests, filtering by status and getting the first one. I think in Ent it will be something like this:

```go
q := c.c.AIP.Query().WithDeletionRequests(func(drq *db.DeletionRequestQuery) {
    drq.Where(deletionrequest.StatusEQ(enums.DeletionRequestStatusPending))
})
```

Then on conversion, check if there are related deletion requests and get the key from the first one.

internal/api/design/storage.go
jraddaoui: `ListAIPDeletionRequests` makes more sense for the AIP considering the path, the attributes and all. Not a big deal how to name the internal one, that was just to avoid changing it. After reviewing all, I just think we should remove it.

internal/api/auth/attributes.go
jraddaoui: We added `sipsources` all together, that's why I thought about `downloadreport` for consistency below, this would be the same. But I think we should remove it and keep treating deletion like an action prefix more than an entity.

internal/api/design/storage.go
djjuhasz: I think creating bloated ORM objects with multiple joins is a bad design pattern that leads to loading a bunch of data that isn't necessary most of the time.

I also just don't understand why you wouldn't want to have a API endpoint to list deletion requests. I can see several uses for it (like reporting) and in general I think API access to application data is useful.

djjuhasz: I think the better solution here is to move the "report_key" field from the `deletion_request` table to the `aip` table (with a rename to "deletion_report_key"), as there will only ever be one deletion report for a given AIP.  I'm willing to do that as a next iteration after we do the current release.

jraddaoui: Yeah, I agree, after seeing the entire implementation I think that would have been a better choice. Apart from what I said already in the previous comments, modifying access control attributes has deployment implications that we'll need to notify. I'd like to avoid adding new attributes that need to be modified in the next release.

jraddaoui: Now that I think about it, adding a data migration to move existing report keys from one table to the other may not be the easiest. And I think SFA is going live with this release for the Kanton's data migrations, I doubt they will be doing deletions during the migration, but they won't get another release for two months. If we really want to move the column it's also something that would be better for this release, where we are on time to add a migration without worrying about existing data.

jraddaoui: If we decide to maintain the join, we can limit the selected fields in the query to avoid bloating ORM objects:

```go
q := c.c.AIP.Query().WithDeletionRequests(func(drq *db.DeletionRequestQuery) {
    drq.Where(deletionrequest.StatusEQ(enums.DeletionRequestStatusPending))
    drq.Select(deletionrequest.FieldReportKey)
})
```

internal/api/auth/attributes.go
djjuhasz: If "deletion" is intended to be an action rather than entity I think it would make more sense to use the verb form of "delete" rather than the noun "deletion".

jraddaoui: Right, it would be more like a prefix (a property?), the actual action would be request, review, and download report. I'd prefer not to change it in this release and just add the new one for the report download, we can revisit in the future.

internal/api/design/storage.go
jraddaoui: > I also just don't understand why you wouldn't want to have a API endpoint to list deletion requests. I can see several uses for it (like reporting) and in general I think API access to application data is useful.

This endpoint's design is limited to list deletion requests from an individual AIP, I don't see a lot of future value on it the way it is right now. If we need reporting we can add something similar then, for now I'd prefer to not introduce a new entity and keep treating deletion request as part of the AIP.

I like the idea of moving the report key to the AIP table as you suggested, and I'd love to see that happening before release. Let me know if you think it's going to be tight to get that done in the next couple of days, I'm happy to help if needed.

djjuhasz:
Replaced by PR #1443.

PR #1440, opened 2025-11-28
Show download link for deletion report

djjuhasz:
Refs #1207.

On the Dashboard AIP detail page show a link to download the deletion report, if the AIP has been deleted and the user has permission to list deletion requests.


djjuhasz:
Replaced by PR #1443.

PR #1441, opened 2025-11-28
Implement AIP deletion report download from Dashboard

djjuhasz:
Refs #1207.

- Add deletion report download API endpoints
- Implement deletion report download storage service methods
- Add deletion report download calls to Dashboard "deletion report" button


djjuhasz:
@jraddaoui the deletion report download is working, so I'd like to merge this and the other open #1207 PRs so we can get the feature in to the next Enduro release.  I will work on adding tests next.

djjuhasz:
Screencast: 
[Screencast from 2025-11-28 11-20-58.webm](https://github.com/user-attachments/assets/306aeff3-15c4-4814-9d54-d954651b335f)

dashboard/src/components/AipReportsCard.vue
```diff
@@ -1,25 +1,48 @@
...
     return;
   }
-  // TODO: download deletion report.
+
+  try {
+    await client.storage.storageDownloadDeletionReportRequest({
+      uuid: uuid,
+    });
+    window.open(
+      getPath() + `/storage/deletion-reports/${uuid}/download`,
```
jraddaoui, line 25: I mentioned this endpoint in the feedback from https://github.com/artefactual-sdps/enduro/pull/1439. This is adding another resource type to the API ("deletion-reports"), but the request is using the deletion request UUID. As I commented there I think we should try to avoid these two new resources and keep handling everything from the AIP. I think it would make more sense with the current API design to use `/aips/{uuid}/download-deletion-report`, following the existing `/aips/{uuid}/deletion-request`, `/aips/{uuid}/deletion-review` and `/aips/{uuid}/deletion-cancel` endpoints.

jraddaoui: Or even `/aips/{uuid}/deletion-report`.

dashboard/src/components/AipReportsCard.vue
```diff
@@ -1,25 +1,48 @@
...
+    }
+    error.value = errorMsg;
+    setTimeout(() => (error.value = ""), 5000);
+  }
 };
 
 // Fetch the deletion report key if the user has permission.
 if (
-  authStore.checkAttributes(["storage:aips:deletion:list"]) &&
+  authStore.checkAttributes(["storage:deletion_report:download"]) &&
```
jraddaoui, line 45: Similar here, maintaining the current approach of the AIP and deletion prefixes, and using the shortest version of the endpoint: `storage:aips:deletion:report`. Otherwise, `storage:aips:deletion:downloadreport`? We have `ingest:sipsources:objects:list` using `sipsources`. If not, I'd prefer using `-` in the API attributes, like we do in the API URLs, `storage:aips:deletion:download-report`.

internal/storage/download_deletion_report_test.go
```diff
@@ -0,0 +1,372 @@
...
+				return
+			}
+
+			assert.NilError(t, err)
+			assert.DeepEqual(t, res, tt.wantRes)
+		})
+	}
+}
+
+// func TestDownloadAip(t *testing.T) {
```
jraddaoui, line 180: Forgot this :wink:

jraddaoui (changes requested):
This looks great @djjuhasz, thanks! Just renaming the endpoint and the attribute, following the feedback from https://github.com/artefactual-sdps/enduro/pull/1439.

djjuhasz:
Replaced by PR #1443.

PR #1443, opened 2025-12-02
Add dashboard link to download AIP deletion report

djjuhasz:
Refs #1207.

- Move `deletion_request.report_key` to `aip.deletion_report_key` in database
- Update deletion report generation to save report key to AIP
- Rename API method for signaling AIP upload is completed to avoid confusion with `UpdateAIP()` method
- Add deletion report download API endpoints
- Add deletion report download link to Dashboard AIP details page


cmd/enduro-a3m-worker/main.go
```diff
@@ -319,7 +319,7 @@ func main() {
 			storageHttpClient.ListAips(),
 			storageHttpClient.CreateAip(),
 			storageHttpClient.SubmitAip(),
-			storageHttpClient.UpdateAip(),
+			storageHttpClient.SubmitAipComplete(),
```
jraddaoui, line 322: Good call!

dashboard/src/components/AipReportsCard.vue
```diff
@@ -0,0 +1,71 @@
...
+
+const aipStore = useAipStore();
+const authStore = useAuthStore();
+const reportKey = ref("");
+const error = ref("");
+
+const downloadDeletionReport = async () => {
+  if (!aipStore.current?.uuid) {
+    return;
+  }
```
jraddaoui, line 16: The UUID is required, I'd look at `aipStore.current?.deletionReportKey`, or remove the check entirely since we already check that to display the template.

dashboard/src/components/AipReportsCard.vue
```diff
@@ -0,0 +1,71 @@
...
+    }
+    error.value = errorMsg;
+    setTimeout(() => (error.value = ""), 5000);
+  }
+};
+
+// Set the deletion report key if the current AIP has been deleted, and the user
+// has permission to download deletion reports.
+if (
+  authStore.checkAttributes(["storage:aips:deletion-report"]) &&
```
jraddaoui: **IMPORTANT:** We should use `storage:aips:deletion:report`, like the existing `storage:aips:deletion:request` and  `storage:aips:deletion:review`, that way you can use the `storage:aips:deletion:*` attribute to allow the three.

dashboard/src/components/AipReportsCard.vue
```diff
@@ -0,0 +1,71 @@
...
+  aipStore.current &&
+  aipStore.isDeleted
+) {
+  reportKey.value = aipStore.current.deletionReportKey || "";
+}
+</script>
+
+<template>
+  <div v-if="reportKey" class="card mb-3">
```
jraddaoui: Why not checking directly in the template and removing the `reportKey` ref?

```vue
<div
    v-if="
      aipStore.current?.deletionReportKey &&
      authStore.checkAttributes(['storage:aips:deletion:report'])
    "
    class="card mb-3"
  >
```

dashboard/src/pages/storage/aips/[id]/index.vue
```diff
@@ -48,6 +49,7 @@ onMounted(() => {
       </div>
       <div class="col-md-6">
         <AipLocationCard />
+        <AipReportsCard v-if="aipStore.isDeleted" />
```
Comment (jraddaoui) [inline dashboard/src/pages/storage/aips/[id]/index.vue:52]: Doesn't hurt to check if the AIP is deleted here, but I think the checks in the component are enough.

internal/api/design/storage.go
```diff
@@ -156,7 +156,7 @@ var _ = Service("storage", func() {
 		Error("not_available")
 		Error("not_valid")
 		HTTP(func() {
-			POST("/aips/{uuid}/update")
+			POST("/aips/{uuid}/submit-complete")
```
jraddaoui, line 159: What do you think about using PUT over the same path? `PUT("/aips/{uuid}/submit")`

internal/api/design/storage.go
```diff
@@ -391,6 +391,68 @@ var _ = Service("storage", func() {
 			Response("not_found", StatusNotFound)
 		})
 	})
+	Method("aip_deletion_report_request", func() {
+		Description("Request access to download a deletion report")
+		Security(JWTAuth, func() {
+			Scope(auth.StorageAIPSDeletionReportAttr)
+		})
+		Payload(func() {
+			AttributeUUID("uuid", "UUID of the deletion report to download")
```
jraddaoui: **IMPORTANT:** UUID of the AIP ...

internal/api/design/storage.go
```diff
@@ -391,6 +391,68 @@ var _ = Service("storage", func() {
...
+			Response("not_valid", StatusBadRequest)
+			Response("internal_error", StatusInternalServerError)
+		})
+	})
+	Method("aip_deletion_report", func() {
+		Description("Download deletion report by UUID")
+		// Disable JWTAuth security (it validates the previous method cookie).
+		NoSecurity()
+		Payload(func() {
+			AttributeUUID("uuid", "UUID of the deletion report to download")
```
jraddaoui: **IMPORTANT:** Same, it's the AIP UUID.

internal/storage/download_deletion_report.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/artefactual-sdps/enduro/internal/auditlog"
+	"github.com/artefactual-sdps/enduro/internal/storage/enums"
+)
+
+func deletionReportReader(
+	ctx context.Context,
+	s *serviceImpl,
+	aip *goastorage.AIP,
+) (*blob.Reader, error) {
+	if aip.DeletionReportKey == nil || *aip.DeletionReportKey == "" {
```
jraddaoui: We could check `aip.Status != enums.AIPStatusDeleted.String()` here too, updating a little the error message.

internal/storage/download_deletion_report.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, goastorage.MakeNotValid(errors.New("invalid UUID"))
+	}
+
+	aip, err := s.ReadAip(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	if aip.Status != enums.AIPStatusDeleted.String() || aip.DeletionReportKey == nil {
+		return nil, goastorage.MakeNotValid(errors.New("deletion report is not available for download"))
+	}
```
jraddaoui: Then remove this check.

internal/storage/download_deletion_report.go
```diff
@@ -0,0 +1,139 @@
...
+
+	aip, err := s.ReadAip(ctx, id)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if aip.Status != enums.AIPStatusDeleted.String() || aip.DeletionReportKey == nil {
+		return nil, nil, goastorage.MakeNotValid(errors.New("deletion report is not available for download"))
+	}
```
jraddaoui: And this one.

internal/storage/download_deletion_report.go
```diff
@@ -0,0 +1,139 @@
...
+
+	r, err := deletionReportReader(ctx, s, aip)
+	if err != nil {
+		return nil, nil, goastorage.MakeInternalError(errors.New("error reading deletion report"))
+	}
+
+	filename, ok := strings.CutPrefix(ReportPrefix+"/", *aip.DeletionReportKey)
```
jraddaoui: **IMPORTANT:** The slash is already part of the `ReportPrefix` constant.

internal/storage/persistence/ent/client/client.go
```diff
@@ -44,6 +47,11 @@ func (c *Client) CreateAIP(ctx context.Context, goaaip *goastorage.AIP) (*goasto
 		SetObjectKey(goaaip.ObjectKey).
 		SetStatus(status)
 
+	// Set optional fields.
+	if goaaip.DeletionReportKey != nil {
+		q.SetDeletionReportKey(*goaaip.DeletionReportKey)
+	}
+
```
jraddaoui, line 54: Hardly unlikely that this is set on creation.

internal/storage/persistence/ent/client/client.go
```diff
@@ -185,6 +193,83 @@ func (c *Client) UpdateAIPLocationID(ctx context.Context, aipID, locationID uuid
...
+	if up.Status != "" && up.Status != dbAIP.Status {
+		q.SetStatus(up.Status)
+		updated = true
+	}
+
+	if up.LocationUUID != nil && up.LocationUUID.String() != locUUID.String() {
+		loc, err := c.c.Location.Query().
+			Where(location.UUID(*up.LocationUUID)).
+			Only(ctx)
```
jraddaoui, line 237: Not sure how much better is to use `OnlyID`.

internal/storage/persistence/ent/client/client.go
```diff
@@ -185,6 +193,83 @@ func (c *Client) UpdateAIPLocationID(ctx context.Context, aipID, locationID uuid
...
+		if err != nil {
+			if db.IsNotFound(err) {
+				return nil, fmt.Errorf("load location: %w", ErrNotFound)
+			}
+			return nil, fmt.Errorf("load location: %v", err)
+		}
+		q.SetLocation(loc)
```
jraddaoui: And `setLocationID` here.

internal/storage/persistence/ent/client/client.go
```diff
@@ -185,6 +193,83 @@ func (c *Client) UpdateAIPLocationID(ctx context.Context, aipID, locationID uuid
...
+
+	if up.DeletionReportKey != nil && *up.DeletionReportKey != dbAIP.DeletionReportKey {
+		q.SetDeletionReportKey(*up.DeletionReportKey)
+		updated = true
+	}
+
+	// If no changes were made return the existing AIP.
+	if !updated {
+		return convertDBAIP(dbAIP), nil
+	}
```
jraddaoui, line 257: Nice!

internal/storage/persistence/ent/client/client.go
```diff
@@ -185,6 +193,83 @@ func (c *Client) UpdateAIPLocationID(ctx context.Context, aipID, locationID uuid
...
+	s, err := q.Save(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	res := convertDBAIP(s)
+
+	// Add the location UUID to the returned AIP because the AIP returned by the
+	// db update doesn't include the location.
+	res.LocationUUID = &locUUID
```
jraddaoui: **IMPORTANT:** We should check that `locUUID` is not the zero UUID, or declare it as a pointer directly.

internal/storage/persistence/migrations/20251128222016_move_deletion_report_key_column.up.sql
```diff
@@ -0,0 +1,12 @@
+-- Add "deletion_report_key" column to "aip" table
+ALTER TABLE `aip` ADD COLUMN `deletion_report_key` varchar(1024) NULL;
+
+-- Copy existing report keys from "deletion_request" to "aip"
+UPDATE `aip`,`deletion_request`
+SET `aip`.`deletion_report_key` = `deletion_request`.`report_key`
+WHERE `deletion_request`.`aip_id` = `aip`.`id`
+AND `deletion_request`.`report_key` IS NOT NULL
+AND `deletion_request`.`report_key` != '';
```
jraddaoui, line 9: Nice! I was not expecting this to be that simple.

internal/storage/types/aip.go
```diff
@@ -0,0 +1,32 @@
...
+
+	// Status of the AIP.
+	Status enums.AIPStatus
+
+	// LocationUUID is the UUID of the location where the AIP is stored.
+	LocationUUID *uuid.UUID
+
+	// DeletionReportKey is the object store key for the AIP deletion report.
+	DeletionReportKey *string
+}
```
jraddaoui, line 32: I thought we had this type and the DB conversion.

jraddaoui (changes requested):
Thanks for doing this @djjuhasz, it's looking great! I have marked the important feedback below, nothing big, and feel free to ignore all the other comments.

dashboard/src/components/AipReportsCard.vue
djjuhasz: This value check is necessary so Typescript doesn't throw an error when calling:

```
client.storage.storageAipDeletionReportRequest({
      uuid: aipStore.current.uuid,
});
```

dashboard/src/components/AipReportsCard.vue
djjuhasz: Sure.

jraddaoui:
I gave it a try locally and it works great @djjuhasz, I just noticed one more thing. We are not emitting a WebSocket event for this AIP update, and we only have [AIP updated events for specific fields](https://github.com/artefactual-sdps/enduro/blob/main/internal/api/design/storage_monitor.go#L15-L16). This can happen in another PR, not sure if you'll have time before release, not a huge deal.

internal/storage/types/aip.go
djjuhasz: Yeah, I was kind of surprised we didn't have `types.AIP` already, but I think all of the AIP methods predate the `types` package.

dashboard/src/components/AipReportsCard.vue
djjuhasz: Yeah, good idea.

jraddaoui:
I could also give it a try to the WebSocket event tomorrow, I may have it ready for review by the time you start your day. Let me know.

dashboard/src/pages/storage/aips/[id]/index.vue
Comment (djjuhasz) [reply dashboard/src/pages/storage/aips/[id]/index.vue:52]: My hope is if the conditional evaluates false here that the component isn't loaded and it saves a bit of resources, but that may be naïve.

internal/api/design/storage.go
djjuhasz: I don't know much about the PUT method, but I don't think this usage matches the intended use case for PUT: 

> The PUT HTTP method creates a new resource or replaces a representation of the target resource with the request [content](https://developer.mozilla.org/en-US/docs/Glossary/HTTP_Content).
-- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/PUT

Maybe [PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/PATCH) would be more appropriate, or a POST with a parameter to indicate the upload is complete?

djjuhasz:
@jraddaoui sure, I'm fine with you having a crack at emitting a WebSocket event for the report key update. :+1:

internal/api/design/storage.go
jraddaoui: Yeah, feels a bit odd not being over the resource directly `PUT("/aips/{uuid}")`.

dashboard/src/components/AipReportsCard.vue
jraddaoui: I needed this change to make the component reactive when we update the current AIP in the store after receiving the WS event. I'll remove it from the other PR if you include it in this one.

djjuhasz: Changed.

internal/storage/download_deletion_report.go
djjuhasz: I refactored all of the common code for opening the deletion report file into the `deletionReportReader()` method.

internal/storage/persistence/ent/client/client.go
djjuhasz: Yeah, true but I prefer to add it just so it's not a surprise to callers that the deletion_report_key is not writable from `CreateAIP()`

internal/storage/persistence/ent/client/client.go
djjuhasz: We need the `location.UUID` too, but I've added a `Select(location.FieldID, location.FieldUUID).` to limit the fields returned.

internal/storage/persistence/ent/client/client.go
djjuhasz: I declared it as a pointer.

djjuhasz:
@jraddaoui I think I've addressed all your feedback, let me know if I've missed anything.

jraddaoui (approved):
All looks great @djjuhasz, thanks!

PR #1445, opened 2025-12-03
Add WebSocket event for AIP update

jraddaoui:
Based on #1443. Refs #1207.


dashboard/src/components/AipReportsCard.vue
```diff
@@ -38,20 +37,16 @@ const downloadDeletionReport = async () => {
...
-
-// Set the deletion report key if the current AIP has been deleted, and the user
-// has permission to download deletion reports.
-if (
-  authStore.checkAttributes(["storage:aips:deletion:report"]) &&
-  aipStore.current &&
-  aipStore.isDeleted
-) {
-  reportKey.value = aipStore.current.deletionReportKey || "";
-}
```
jraddaoui: I added a comment in the other PR about this, I'll remove it from here if you add it there.

internal/storage/persistence/persistence.go
```diff
@@ -22,7 +22,12 @@ type Storage interface {
 	CreateAIP(ctx context.Context, aip *goastorage.AIP) (*goastorage.AIP, error)
 	ListAIPs(ctx context.Context, payload *goastorage.ListAipsPayload) (*goastorage.AIPs, error)
 	ReadAIP(ctx context.Context, aipID uuid.UUID) (*goastorage.AIP, error)
-	UpdateAIP(ctx context.Context, aipID uuid.UUID, updater AIPUpdater) (*types.AIP, error)
+	// TODO: normalize type usage between *goastorage.AIP and *types.AIP.
+	// For now, we return both types from this method to minimize changes and be able to publish an event
+	// with the *goastorage.AIP representation from the storage service. We should consider this alongside
+	// the ingest service implementation, and decide if we want to use our own types or goa-generated types
+	// in the persistence layer.
+	UpdateAIP(ctx context.Context, aipID uuid.UUID, updater AIPUpdater) (*types.AIP, *goastorage.AIP, error)
```
jraddaoui, line 30: Now I remember why we didn't have our own AIP type and conversion (re: https://github.com/artefactual-sdps/enduro/pull/1443#discussion_r2583142534), the storage service and its persistence layer started with the idea to reuse Goa types internally. I added this big TODO comment here, and I'd love to revisit and normalize this in the future for both the ingest and the storage services. For now, this method is a bit odd returning both types, but that's all I could do in this time without changing a lot the current implementation.

internal/storage/service.go
```diff
@@ -387,7 +387,17 @@ func (s *serviceImpl) UpdateAIP(
 	aipID uuid.UUID,
 	updater persistence.AIPUpdater,
 ) (*types.AIP, error) {
-	return s.storagePersistence.UpdateAIP(ctx, aipID, updater)
+	aip, goaaip, err := s.storagePersistence.UpdateAIP(ctx, aipID, updater)
```
djjuhasz, line 390: Why not just call `aipAsGoa(ctx, aip)` here, rather than in the `persistence.UpdateAIP()` method?

djjuhasz: Oh I see, `aipAsGoa()` needs the ent representation of the AIP. :(

jraddaoui: `aipAsGoa` expects an Ent type and `aip` is `types.AIP` in here. I didn't want to add another conversion method from `types.AIP` to `goastorage.AIP`.

internal/storage/persistence/persistence.go
djjuhasz: I think this is okay for now, but we should definitely discuss if we want to use our own types vs. Goa types soon.

djjuhasz (approved):
I think this works as a quick fix @jraddaoui until we can make a final decision regarding custom types vs. goa types in the persistence layer.  Just a couple of notes:

1. "WebSocket" is misspelled in the commit message
2. I think we also need to consider if we want to add logic to only emit an event if the method actually updates the AIP.  As it is now an update might not change the AIP at all, but we would still emit an event.  It's difficult now to check whether the AIP is updated in the service method, because we load the existing AIP and update it in the persistence method.

jraddaoui:
Thanks @djjuhasz! Sounds good, maybe returning that `updated` boolean you added inside the method. We can do that when we address one of the TODOs from this PR.

PR #1446, opened 2025-12-03
Fix AIP deletion report not written to bucket

jraddaoui:
Keep the bucket open in write() until after the writer closes. Return errors from writer.Close so failures surface. Update the workflow status to error when report generation fails.

Refs #1207.


djjuhasz (approved):
LGTM 👍

PR #1448, opened 2025-12-04
Improve persistence `UpdateAIP()` method

djjuhasz:
Refs #1207.

- Use a database transaction to ensure data consistency after update
- Fix panic when updating the AIP location from no location to a location

### DIFF
--- a/internal/storage/persistence/ent/client/client.go
+++ b/internal/storage/persistence/ent/client/client.go
@@ -202,51 +202,57 @@ func (c *Client) UpdateAIP(
 	updater persistence.AIPUpdater,
 ) (*types.AIP, *goastorage.AIP, error) {
 	var updated bool
-	var locUUID *uuid.UUID
+	var locUUID uuid.UUID
+
+	// Start a database transaction.
+	tx, err := c.c.Tx(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("begin transaction: %v", err)
+	}
 
 	// Load existing AIP.
-	dbAIP, err := c.c.AIP.Query().
+	dbAIP, err := tx.AIP.Query().
 		WithLocation().
 		Where(aip.AipID(aipID)).
 		Only(ctx)
 	if err != nil {
 		if db.IsNotFound(err) {
-			return nil, nil, fmt.Errorf("load AIP: %w", ErrNotFound)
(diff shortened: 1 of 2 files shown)


internal/storage/persistence/ent/client/client.go
```diff
@@ -223,27 +229,27 @@ func (c *Client) UpdateAIP(
...
+	q := tx.AIP.UpdateOneID(dbAIP.ID)
 
 	if up.Status != "" && up.Status != dbAIP.Status {
 		q.SetStatus(up.Status)
 		updated = true
 	}
 
-	if up.LocationUUID != nil && up.LocationUUID.String() != locUUID.String() {
+	if !uuidsEqual(up.LocationUUID, locUUID) {
 		loc, err := c.c.Location.Query().
```
sevein: Should we use the tx handle?

```suggestion
		loc, err := tx.Location.Query().
```

internal/storage/persistence/ent/client/client.go
```diff
@@ -381,3 +390,13 @@ func (c *Client) LocationAIPs(ctx context.Context, locationID uuid.UUID) (goasto
...
+
+func uuidsEqual(a, b *uuid.UUID) bool {
+	if a == nil && b == nil {
+		return true
+	}
+	if a != nil && b != nil && *a == *b {
+		return true
+	}
+	return false
+}
```
sevein: I've been experimenting with a version that drops `uuidsEqual` and relies on direct UUID comparisons (array types are comparable). I put together this [patch](https://gist.github.com/sevein/4dbf4cb00296a29e1d98cd78929e19c8). Not sure it's an improvment but using direct UUID comparisons feels cleaner: we just check `up.LocationUUID != nil` and compare it to the current UUID, which preserves the existing "leave unchanged unless a new UUID is provided" semantics without the helper.

jraddaoui:
Thanks @djjuhasz and @sevein, changes and provided feedback look good to me!

internal/storage/persistence/ent/client/client.go
djjuhasz: Oops, missed that one. :smile:

internal/storage/persistence/ent/client/client.go
djjuhasz: Nice! :+1:

PR #1450, opened 2025-12-05
Add SBOM GitHub Action workflow

sbreker:
Add new GitHub Action workflow that will generate an SBOM file using Trivy and upload it to Dependency Track.

This workflow will run on push to main, and when a new release is created.

This workflow requires the secrets DEPENDENCY_TRACK_URL and DEPENDENCY_TRACK_API_KEY be configured in GitHub to allow upload to Dependency Track.

### DIFF
--- a/.github/workflows/generate-sbom.yml
+++ b/.github/workflows/generate-sbom.yml
@@ -1,0 +1,70 @@
+name: SBOM (Dependency-Track)
+
+on:
+  push:
+    branches:
+      - main
+  release:
+    types: [published]
+
+concurrency:
+  group: sbom-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+env:
+  PROJECT_NAME:
+    ${{ github.repository_owner }}-${{ github.event.repository.name }}
+  PROJECT_VERSION:
+    ${{ (github.event_name == 'release' && github.event.release.tag_name) || github.ref_name }}
+
+jobs:
+  sbom:
+    runs-on: ubuntu-latest
+    env:
+      DEPENDENCY_TRACK_URL: ${{ secrets.DEPENDENCY_TRACK_URL }}
+      DEPENDENCY_TRACK_API_KEY: ${{ secrets.DEPENDENCY_TRACK_API_KEY }}
+    steps:
+      - name: Validate Dependency-Track secrets
+        run: |
+          if [ -z "${DEPENDENCY_TRACK_URL}" ] || [ -z "${DEPENDENCY_TRACK_API_KEY}" ]; then
+            echo "Dependency-Track secrets not configured; set DEPENDENCY_TRACK_URL and DEPENDENCY_TRACK_API_KEY."
+            exit 1
+          fi
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.PROJECT_VERSION }}
+
+      - name: Set up Trivy cache
+        uses: actions/cache@v4
+        with:
+          path: .trivycache
+          key: ${{ runner.os }}-trivy-${{ hashFiles('**/go.sum', '**/package-lock.json', '**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-trivy-
+
+      - name: Generate SBOM (CycloneDX JSON)
+        uses: aquasecurity/trivy-action@0.33.1
+        env:
+          TRIVY_NO_PROGRESS: "true"
+        with:
+          scan-type: fs
+          format: cyclonedx
+          output: sbom.cdx.json
+          cache-dir: .trivycache
+          exit-code: '0'
+
+      - name: Upload to Dependency-Track
+        run: |
+          curl --fail-with-body --silent --show-error \
+            -X POST "${DEPENDENCY_TRACK_URL%/}/api/v1/bom" \
+            -H "X-Api-Key: ${DEPENDENCY_TRACK_API_KEY}" \
+            -F "autoCreate=true" \
+            -F "projectName=${PROJECT_NAME}" \
+            -F "projectVersion=${PROJECT_VERSION}" \
+            -F "parentName=artefactual-sdps-enduro" \
+            -F "bom=@sbom.cdx.json"


.github/workflows/generate-sbom.yml
```diff
@@ -0,0 +1,85 @@
...
+            '**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-trivy-
+
+      - name: Generate SBOM (CycloneDX XML)
+        run:
+          trivy fs --cache-dir .trivycache --format cyclonedx --output
+          sbom.cdx.xml .
+
+      - name: Upload SBOM artifact
```
djjuhasz: Where is the SBOM being uploaded to?

djjuhasz (approved):
Looks good to me @sbreker :+1:

.github/workflows/generate-sbom.yml
```diff
@@ -0,0 +1,85 @@
+name: SBOM (Dependency-Track)
+
+on:
+  push:
+    branches:
+      - main
+  release:
+    types: [published]
```
jraddaoui, line 8: Why do we need both? Are they tracked differently in Dependency-Track?

.github/workflows/generate-sbom.yml
jraddaoui: As an artifact of the workflow, I think that you can download it from the run (never tested it). But why do we need this? Should we look at it or something? Wouldn't it be enough to just upload it to Dependency-Track directly?

.github/workflows/generate-sbom.yml
```diff
@@ -0,0 +1,85 @@
...
+    runs-on: ubuntu-latest
+    env:
+      DEPENDENCY_TRACK_URL: ${{ secrets.DEPENDENCY_TRACK_URL }}
+      DEPENDENCY_TRACK_API_KEY: ${{ secrets.DEPENDENCY_TRACK_API_KEY }}
+    steps:
+      - name: Download SBOM artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: sbom
+          path: .
```
jraddaoui: Why using two jobs and upload/download as an artifact? If we need the artifact in the workflow run, by using a single job we remove the need to download the artifact. If we don't need the artifact, we can remove its upload/download entirely and just `Upload to Dependency-Track` directly in the `generate-sbom` job.

.github/workflows/generate-sbom.yml
```diff
@@ -0,0 +1,85 @@
...
+
+      - name: Upload to Dependency-Track
+        if:
+          ${{ env.DEPENDENCY_TRACK_URL != '' && env.DEPENDENCY_TRACK_API_KEY !=
+          '' }}
```
jraddaoui: I guess you are using an auto-formatter or something that cuts long lines in YAML files? I think they are cut not in the best places, I'd prefer the long lines or something a bit more readable.

jraddaoui (changes requested):
Nice @sbreker! Can we use a single job to at least avoid the artifact download? Do we really need that artifact?

.github/workflows/generate-sbom.yml
```diff
@@ -0,0 +1,85 @@
...
+        with:
+          ref: ${{ env.PROJECT_VERSION }}
+
+      - name: Set up cache
+        uses: actions/cache@v4
+        with:
+          path: .trivycache
+          key:
+            ${{ runner.os }}-trivy-${{ hashFiles('**/go.sum',
+            '**/package-lock.json') }}
```
jraddaoui: I guess we don't care about the docs deps at `docs/requirements.txt`, right?

.github/workflows/generate-sbom.yml
sbreker: The SBOMs are a snapshot of the dependencies and their versions at a point in time. In Dependency Track we can maintain separate projects for each of the versions so that we can see what vulnerabilities affect any of our released versions. E.g. in AtoM we can view a dependency analysis of qa/2.x, stable/2.9x, stable/2.8.x, stable/2.7.x etc. In this same way, this config above will maintain an SBOM for main, and each release. This could potentially accumulate a large amount of versioned projects in DT.... I am relying on your advice on whether you think this is overkill, or if the intent is not like AtoM and clients upgrade right away when there is a new Enduro version (perhaps we will not see clients on older versions like AtoM).

.github/workflows/generate-sbom.yml
sbreker: I missed these docs dependencies - unintentional omission on my part. I will add this to be monitored as well

.github/workflows/generate-sbom.yml
jraddaoui: Nice, thanks @sbreker! That is good, I don't think it's overkill.

.github/workflows/generate-sbom.yml
sbreker: Fixed. Prettier was running on save...

.github/workflows/generate-sbom.yml
sbreker: Initially it was good to see the SBOM for debugging. You are right though - the SBOM can be retrieved if needed via the Dependency Track UI. I am refactoring this to perform this as a single job without retaining the SBOM artifact.

.github/workflows/generate-sbom.yml
sbreker: image

sbreker:
Thanks for the feedback @jraddaoui and @djjuhasz! I have completed some refactoring - let me know if you have any additional feedback.

.github/workflows/generate-sbom.yml
jraddaoui: Nice!

.github/workflows/generate-sbom.yml
```diff
@@ -0,0 +1,70 @@
...
+          output: sbom.cdx.json
+          cache-dir: .trivycache
+          exit-code: '0'
+
+      - name: Validate Dependency-Track secrets
+        run: |
+          if [ -z "${DEPENDENCY_TRACK_URL}" ] || [ -z "${DEPENDENCY_TRACK_API_KEY}" ]; then
+            echo "Dependency-Track secrets not configured; set DEPENDENCY_TRACK_URL and DEPENDENCY_TRACK_API_KEY."
+            exit 1
+          fi
```
jraddaoui: Should this be the first step? I don't see a point doing all the previous steps if we are going to fail here.

jraddaoui (approved):
Thanks @sbreker! I think we should validate the secrets sooner, but other than that it LGTM!

.github/workflows/generate-sbom.yml
sbreker: Good point. Updated in latest commit.

PR #1452, opened 2025-12-06
Implement CreateTasks using slice-based batching

sevein:
Adopt slice-based CreateTasks across ingest workflows, superseding PR #1438.


internal/a3m/a3m.go
```diff
@@ -175,12 +186,7 @@ func saveTasks(
 	ctx, span := tracer.Start(ctx, "saveTasks")
 	defer span.End()
 
-	jobStatusToTaskStatus := map[transferservice.Job_Status]enums.TaskStatus{
-		transferservice.Job_STATUS_UNSPECIFIED: enums.TaskStatusUnspecified,
-		transferservice.Job_STATUS_COMPLETE:    enums.TaskStatusDone,
-		transferservice.Job_STATUS_PROCESSING:  enums.TaskStatusInProgress,
-		transferservice.Job_STATUS_FAILED:      enums.TaskStatusError,
-	}
+	tasks := make([]*datatypes.Task, 0, len(jobs))
```
jraddaoui: Same as in the other PR, nitpicking, use `make([]*datatypes.Task, len(jobs))` and add using the index from the loop below.

internal/a3m/a3m.go
```diff
@@ -165,6 +168,14 @@ func (a *CreateAIPActivity) Execute(
 	return result, nil
 }
 
+// taskStatusesByA3mJobStatuses maps a3m job statuses to Enduro task statuses.
+var taskStatusesByA3mJobStatuses = map[transferservice.Job_Status]enums.TaskStatus{
```
jraddaoui: I think the name we better before, or what I'd understand better, `a3mJobToTaskStatuses`. Not a big deal for sure.

internal/ingest/task.go
```diff
@@ -25,6 +25,26 @@ func (svc *ingestImpl) CreateTask(ctx context.Context, task *datatypes.Task) err
...
 }
 
+func (svc *ingestImpl) CreateTasks(ctx context.Context, tasks []*datatypes.Task) error {
+	if len(tasks) == 0 {
+		return nil
+	}
+
+	err := svc.perSvc.CreateTasks(ctx, tasks)
+	if err != nil {
+		return fmt.Errorf("tasks: create: %v", err)
```
jraddaoui: Feels wrong to change the error prefix to a plural version, maybe `task: create bulk` and change the one above to `task: create one` or leave as it is. Or just remove the prefix from the error messages and use the function name directly. Not a big deal either.

internal/persistence/ent/client/task.go
```diff
@@ -13,59 +14,116 @@ import (
...
-	if task.UUID == uuid.Nil {
-		return newRequiredFieldError("UUID")
-	}
-	if task.Name == "" {
-		return newRequiredFieldError("Name")
-	}
-	if task.WorkflowUUID == uuid.Nil {
-		return newRequiredFieldError("WorkflowUUID")
+	if task == nil {
+		return newRequiredFieldError("task")
```
jraddaoui: You are looking for this in the loop from CreateTasks, I think it would be okay to remove it here.

internal/persistence/telemetry_test.go
```diff
@@ -17,6 +21,101 @@ import (
...
 
+func newTracerWithRecorder(t *testing.T) (*tracetest.SpanRecorder, *otelsdktrace.TracerProvider) {
+	t.Helper()
+
+	recorder := tracetest.NewSpanRecorder()
+	tp := otelsdktrace.NewTracerProvider(otelsdktrace.WithSpanProcessor(recorder))
+	t.Cleanup(func() { _ = tp.Shutdown(context.Background()) })
+
+	return recorder, tp
+}
```
jraddaoui, line 31: Nice!

internal/persistence/ent/client/task_test.go
```diff
@@ -128,6 +133,146 @@ func TestCreateTask(t *testing.T) {
...
+			Status:       enums.TaskStatusInProgress,
+			WorkflowUUID: wUUID,
+		}
+	}
+
+	assert.NilError(t, svc.CreateTasks(ctx, tasks))
+	assert.Equal(t, 1, cdrv.begins)
+	assert.Equal(t, 2, len(cdrv.execs))
+}
+
```
jraddaoui, line 276: I think the test driver approach adds too much complexity for what it does. I think we could live with a normal test, checking that `CreateTasks` is called twice. If you still want to maintain a version of this test, I think it would be simpler with a logging hook, ChatGPT-Codex provided this solution:

```go
// ...existing code...
func TestCreateTasksUsesBatching(t *testing.T) {
    t.Parallel()

    ctx := context.Background()
    dsn := fmt.Sprintf("file:%s?mode=memory&cache=shared&_fk=1", t.Name())

    sqlDB, err := sql.Open("sqlite3", dsn)
    assert.NilError(t, err)
    t.Cleanup(func() { sqlDB.Close() })

    var begins, inserts int
    entc := enttest.NewClient(
        t,
        enttest.WithOptions(
            db.Driver(entsql.OpenDB(dialect.SQLite, sqlDB)),
            db.Log(func(args ...any) { // hook: every Exec/Query goes through here
                q := strings.ToUpper(fmt.Sprint(args...))
                if strings.Contains(q, "BEGIN") {
                    begins++
                }
                if strings.Contains(q, "INSERT") {
                    inserts++
                }
            }),
        ),
    )
    t.Cleanup(func() { entc.Close() })

    svc := entclient.New(logr.Discard(), entc)
    // ...existing code to seed SIP/workflow and build tasks...
    assert.NilError(t, svc.CreateTasks(ctx, tasks))
    assert.Equal(t, 1, begins)
    assert.Equal(t, 2, inserts)
}
// ...existing code...
```

jraddaoui (changes requested):
Thanks @sevein!

internal/ingest/task.go
sevein: Makes sense. I went with `tasks: bulk create: %v` so it sounds like we're performing a creation op in bulk as opposed to a "bulk" object.

internal/persistence/ent/client/task_test.go
sevein: Good call.

internal/filenotify/filenotify_test.go
```diff
@@ -45,22 +45,26 @@ func TestPollerEvent(t *testing.T) {
...
-	case e := <-w.Errors():
-		err = fmt.Errorf("got unexpected error waiting for events %v: %v", eType, e)
-	case <-time.After(pollInterval * 2):
-		err = fmt.Errorf("timeout waiting for event %v", eType)
-	}
-	return err
+	},
+		poll.WithTimeout(pollInterval*20),
+		poll.WithDelay(pollInterval),
+	)
```
jraddaoui, line 69: Nice!

jraddaoui (approved):
Thanks @sevein! I think I've never seen the poller test error, but the changes on it LGTM.

PR #1454, opened 2025-12-08
Do not load observability resources in Tilt CI

jraddaoui:
Tilt CI was still failing, complaining about ephemeral storage:

```
mysql-create… │ [event: pod enduro-sdps/mysql-create-locations-dr9fm] The node was low on resource: ephemeral-storage.
mysql-create… │ ┊ Scheduled - <1s
mysql-create… │ ┊ Initialized - <1s
mysql-create… │ ┊ Ready - (…) Pending
Error: back-off 2m40s restarting failed container=create-locations pod=mysql-create-locations-dr9fm_enduro-sdps(3138c6c4-c8b1-4365-9c5d-8b151c3fc8eb)
```

Removing the observability components reduces the cluster size by not having to download those images and not having to collect all their logs, which also makes the Tilt CI output cleaner.


hack/kube/overlays/dev-a3m/kustomization.yaml
```diff
@@ -15,8 +15,3 @@ images:
   - name: ghcr.io/artefactual-sdps/enduro-dashboard:main
     newName: enduro-dashboard
     newTag: dev
-patches:
-  - target:
-      kind: StatefulSet
-      name: enduro-a3m
-    path: enduro-a3m-patch.yaml
```
jraddaoui, line 22: No need to patch a resource from the same layer, I moved the A3M_DEBUG env var to the resource directly.

djjuhasz (approved):
Thanks @jraddaoui, LGTM. 👍

PR #1455, opened 2025-12-08
Upgrade docs dependencies

jraddaoui:
Running `pip-compile --upgrade`. Fixes https://github.com/artefactual-sdps/enduro/security/dependabot/130.


djjuhasz (approved):
Thanks @jraddaoui 👍

PR #1456, opened 2025-12-09
Wait for SIPs validation in Batch workflow

jraddaoui:
Poll for SIP statuses until they are all validated or in a final status,
signal processing workflows to continue or stop processing. Poll
for SIP statuses until they are all ingested or in another final status,
update final Batch status based on the final SIP statuses.

Also, to allow processing SIPs from batches in parallel we need to increase
the a3m and AM workers capacity. Use the default limits from Temporal
for:

- MaxConcurrentSessionExecutionSize (a3m and AM)
- MaxConcurrentActivityExecutionSize (AM)

Use the capacity configuration to set both values to allow concurrent
sessions executing activities in parallel in the AM worker. Keep the
a3m worker limited to a single activity as a3m fails with concurrent
transfers.

Refs #1405.


jraddaoui:
@djjuhasz, this is a bit bigger PR than the previous ones, but it's split in multiple commits if you prefer to review one by one. @sevein, I'd love to get your input on this one too if you have some time, specially about the signal/session issue. I hope I have added enough comments in code to make the situation clear.

internal/workflow/activities/poll_sip_statuses.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+type PollSIPStatusesActivityResult struct {
+	AllExpectedStatus bool
+}
+
+type PollSIPStatusesActivity struct {
+	ingestsvc    ingest.Service
+	pollInterval time.Duration
+}
```
sevein, line 46: Suggested docstring for the activity struct:

```go
// PollSIPStatusesActivity polls the ingest service until all SIPs in a batch
// reach a final state, failing immediately if the SIP count differs from
// ExpectedSIPCount or any SIP reports an invalid final status.
```

And maybe moving this into a file-level variable would make things clearer too:

```go
// finalStatuses lists all SIP statuses that are considered terminal. Any SIP
// reporting one of these statuses is treated as final by the polling logic,
// except the caller's ExpectedStatus, which is checked separately for success.
var finalStatuses = []enums.SIPStatus{
    enums.SIPStatusFailed,
    enums.SIPStatusError,
    enums.SIPStatusCanceled,
}
```

internal/workflow/activities/poll_sip_statuses.go
```diff
@@ -0,0 +1,116 @@
...
+func (a *PollSIPStatusesActivity) checkSIPStatuses(
+	ctx context.Context,
+	batchUUID uuid.UUID,
+	expectedSIPCount int,
+	expectedStatus enums.SIPStatus,
+) (*checkResult, error) {
+	// Query all SIPs for this batch.
+	result, err := a.ingestsvc.ListSips(ctx, &goaingest.ListSipsPayload{
+		BatchUUID: ref.New(batchUUID.String()),
+		Limit:     ref.New(50_000),
```
sevein: Maybe extract this into a file-level constant to make the intent clearer, e.g.:

```go
// maxListSIPsLimit defines the maximum number of SIP records we request when
// polling batch status. This activity currently loads all SIPs into memory,
// but we may switch to a stats-based or aggregated query approach in the future,
// in which case this limit will either go away or be significantly reduced.
const maxListSIPsLimit = 50_000
```

internal/workflow/processing.go
```diff
@@ -1320,6 +1320,10 @@ func (w *ProcessingWorkflow) updateSIPProcessing(ctx temporalsdk_workflow.Contex
 	).Get(activityOpts, nil)
 }
 
+// waitForBatch waits for a signal indicating that all SIPs in the Batch have been validated. If the signal
+// indicates that not all SIPs in the Batch have been validated successfully, the workflow status is set to
+// canceled and an error is returned. This function blocks the worker session until the signal is received,
+// this requires that the available workers have enough capacity to handle all SIPs in the running Batches.
```
sevein: I'm still trying to understand the problem and its implications. An alternative rewrite of the last sentence:

> This function blocks the worker session until the signal is received, which means the workers must have enough remaining session capacity to accommodate all SIPs in the running batches. Otherwise, the system may stall; for example, with a single worker replica configured with capacity 1000, a batch of 1001 SIPs would exceed that limit and prevent progress.

Would that be accurate?

internal/workflow/processing.go
```diff
@@ -1365,6 +1369,9 @@ func (w *ProcessingWorkflow) waitForBatch(ctx temporalsdk_workflow.Context, stat
 	}()
 
 	// Wait for a Batch signal.
+	// TODO: Consider releasing the session:
+	// - Put the SIP in the internal bucket while waiting for the signal.
+	// - Add configuration to indicate there is only one session worker.
```
sevein: I wonder if this might be a good use case for [workflow.RecreateSession](https://pkg.go.dev/go.temporal.io/sdk/workflow#RecreateSession).

jraddaoui: I looked exactly for something like that to avoid this capacity issue, no idea how I didn't find it :man_facepalming:. I'll give it a try, thanks @sevein!

internal/workflow/processing.go
jraddaoui: Yeah, that's right. But with the session recreation it won't be needed.

internal/workflow/activities/poll_sip_statuses.go
jraddaoui: Good call!

internal/workflow/activities/poll_sip_statuses.go
jraddaoui: I was trying to match [this constant](https://github.com/artefactual-sdps/enduro/blob/main/internal/entfilter/filter.go#L17), I'll reuse it here. Good point about aggregated queries to avoid the limit, that's one of the issues of having some many layers internally, I'd need to add such method to the ingest and the persistence services. Not a huge deal, but something for another PR.

internal/workflow/activities/poll_sip_statuses.go
jraddaoui: Done, and moved to https://github.com/artefactual-sdps/enduro/pull/1458.

internal/workflow/activities/poll_sip_statuses.go
jraddaoui: Done, and moved to https://github.com/artefactual-sdps/enduro/pull/1458.

internal/workflow/processing.go
```diff
@@ -205,20 +205,20 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
 		var sessErr error
 		maxAttempts := 5
 
-		activityOpts := temporalsdk_workflow.WithActivityOptions(ctx, temporalsdk_workflow.ActivityOptions{
+		ctx = temporalsdk_workflow.WithActivityOptions(ctx, temporalsdk_workflow.ActivityOptions{
```
jraddaoui, line 208: Adding them to the main context allows the recreate session to work without setting this activity options again.

internal/workflow/processing.go
```diff
@@ -287,12 +287,10 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
...
 // SessionHandler runs activities that belong to the same session.
 func (w *ProcessingWorkflow) SessionHandler(
-	sessCtx temporalsdk_workflow.Context,
-	attempt int,
-	state *workflowState,
+	ctx, sessCtx temporalsdk_workflow.Context, attempt int, state *workflowState,
 ) error {
 	// Cleanup session files on exit.
-	defer w.sessionCleanup(sessCtx, state)
+	defer func() { w.sessionCleanup(sessCtx, state) }()
```
jraddaoui, line 293: I was surprised by this: https://go.dev/play/p/zngUowXB5Az

internal/workflow/processing.go
jraddaoui: Updated the comment, noting a possible edge case now that we recreate the session.

internal/workflow/processing.go
jraddaoui: It was not the easiest but I could make it work. It's quite ugly as it now requires to pass the main workflow context down to be able to recreate the session with it, and it also requires to return the new session context back for the cleanup activities. See https://github.com/artefactual-sdps/enduro/pull/1456/commits/22fc5b87425f3135dbbbbfcf89a65ab8099e7b09.

One thing that took me a while to realize:

- Sessions are maintained using an activity (`internalSessionCreationActivity`) that heartbeats during the entire session and takes a slot from `MaxConcurrentActivityExecutionSize`.
- Setting the worker as a session worker creates another task queue (`a3m__internal_session_creation`).
- The initial session creation activity runs in that task queue and creates another one assigned to the current worker (`8710c9d0-7349-46ff-a0db-96997ee57689@enduro-a3m-0`).
- Even with `MaxConcurrentActivityExecutionSize = 1`, the initial session is maintained in `a3m__internal_session_creation` and the functional activities can run in `8710c9d0-7349-46ff-a0db-96997ee57689@enduro-a3m-0` sequentially at the same time as the session activity runs.
- When we recreate the session, the `internalSessionCreationActivity` goes into the `8710c9d0-7349-46ff-a0db-96997ee57689@enduro-a3m-0` task queue instead of into `a3m__internal_session_creation` like the initial one.
- With `MaxConcurrentActivityExecutionSize = 1` the recreated session takes that slot and blocks all functional activities to run.
- This requires to set `MaxConcurrentActivityExecutionSize` to the double of `MaxConcurrentSessionExecutionSize`, to allow each recreated session to run the `internalSessionCreationActivity` and the functional activities of each session.

I still think this is better than not recreating and keeping the initial session, as that requires to consider the amount of SIPs per batch and configure the capacity accordingly.

PS. While working on this I found this Temporal sample: https://github.com/temporalio/samples-go/tree/main/worker-specific-task-queues. This could be an alternative that does pretty much the same as sessions but doesn't require to maintain nor recreate one. But I'm not sure if that may affect load balancing and it's quite a different approach to what were doing. Maybe something to test if we found issues with this implementation.

internal/workflow/processing.go
sevein: Yeah, that's a common trap unfortunately. I've been beaten by it before. To their credit, the spec reads:

> Each time a "defer" statement executes, the function value and parameters to the call are evaluated as usual and saved anew but the actual function is not invoked.

jraddaoui:
Thanks for the review and feedback in this PR @sevein. I have divided it in multiple PRs to make it a bit easier for @djjuhasz's review, sorry I didn't organize this a bit better sooner.

PR #1458, opened 2025-12-10
Add poll SIP statuses activity

jraddaoui:
This activity will be used in the batch workflow to poll for all SIP statuses at two points, waiting for all SIPs to reach the "validated" and the "ingested" statuses.

Refs #1405.

### DIFF
--- a/internal/workflow/activities/poll_sip_statuses.go
+++ b/internal/workflow/activities/poll_sip_statuses.go
@@ -1,0 +1,125 @@
+package activities
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/google/uuid"
+	"go.artefactual.dev/tools/ref"
+	temporal_tools "go.artefactual.dev/tools/temporal"
+
+	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
+	"github.com/artefactual-sdps/enduro/internal/entfilter"
+	"github.com/artefactual-sdps/enduro/internal/enums"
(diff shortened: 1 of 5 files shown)


jraddaoui:
@djjuhasz, if you need more context, https://github.com/artefactual-sdps/enduro/pull/1462 is a follow up to this PR where the activity is actually being used.

internal/workflow/activities/poll_sip_statuses.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/artefactual-sdps/enduro/internal/ingest"
+)
+
+// finalStatuses lists all SIP statuses that are considered terminal. Any SIP
+// reporting one of these statuses is treated as final by the polling logic,
+// except the caller's ExpectedStatus, which is checked separately for success.
+var finalStatuses = []enums.SIPStatus{
+	enums.SIPStatusFailed,
+	enums.SIPStatusError,
+	enums.SIPStatusCanceled,
```
djjuhasz, line 26: Should `enums.SIPStatusIngested` be added to this list as a terminal status? If the expected status is `enums.SIPStatusValidated` then an ingested SIP might indicate an error.

djjuhasz (approved):
LGTM @jraddaoui. I just had one question whether "ingested" should be considered a terminal SIP status.

internal/workflow/activities/poll_sip_statuses.go
jraddaoui: Yeah, hardly unlikely with the signal and all, but it won't hurt. If the expected status is "ingested" it will be checked directly before looking into this array.

Issue #1459, opened 2025-12-11
Problem: concurrent ingest workflows using Archivematica fail deleting the transfer


jraddaoui:
**Describe the bug**

Running multiple ingest workflows can fail deleting the transfer from the AM transfer source location after a successful ingest.

**To Reproduce**

- Using AM as preservation system.
- Set `am.capacity` to higher than one.
- Create SIPs that contain the same top-level folder inside.
- Ingest multiple SIPs at the same time.

**Expected behavior**

Ingest succeeds for all SIPs.

**Screenshots**

```
2025-12-11T18:53:27.982Z	V(2)	enduro-am-worker.temporal-client	log/with_logger.go:79	error	{"Namespace": "default", "TaskQueue": "am", "WorkerID": "1@enduro-am-86445d98c5-mn5gr@", "WorkflowID": "processing-workflow-49703e15-7648-4166-9613-ddc191d0afe8", "RunID": "fa4ff1ff-0746-4f3f-b7a1-6cade0595d72", "ActivityType": "DeleteTransferActivity", "Attempt": 10, "Error": "delete transfer: path: \"small\": SFTP: unable to remove \"small\": file does not exist", "error": "Activity error."}
go.temporal.io/sdk/log.(*withLogger).Error
	go.temporal.io/sdk@v1.33.1/log/with_logger.go:79
go.temporal.io/sdk/internal.(*activityTaskHandlerImpl).Execute
	go.temporal.io/sdk@v1.33.1/internal/internal_task_handlers.go:2286
go.temporal.io/sdk/internal.(*activityTaskPoller).ProcessTask
	go.temporal.io/sdk@v1.33.1/internal/internal_task_pollers.go:1100
go.temporal.io/sdk/internal.(*baseWorker).processTaskAsync.func1
	go.temporal.io/sdk@v1.33.1/internal/internal_worker_base.go:453
```

<img width="1871" height="923" alt="Image" src="https://github.com/user-attachments/assets/4dc098bc-7ad0-4617-acb7-85a0eb714697" />

**Additional context**

The error occurs deleting the transfer, but copying it to the same path while another transfer is processing could also lead to issues, even if it's not reported as an error.

The SIP and workflow are marked as error while the AIP has been created and it's accessible. There is no recollection of the error in any task from the workflow. See screenshot above.

PR #1460, opened 2025-12-11
Fix concurrency issues with Archivematica and a3m

jraddaoui:
- Remove capacity configuration from a3m.
- Prevent file collisions in AM transfer source.
- Increase default AM capacity to 10.

Refs #1459 and #1405.

### DIFF
--- a/cmd/enduro-a3m-worker/main.go
+++ b/cmd/enduro-a3m-worker/main.go
@@ -223,13 +223,12 @@ func main() {
 
 	// Activity worker.
 	{
-		logger.V(1).Info("a3m worker config", "capacity", cfg.A3m.Capacity)
-
 		done := make(chan struct{})
 		workerOpts := temporalsdk_worker.Options{
-			DisableWorkflowWorker:              true,
-			EnableSessionWorker:                true,
-			MaxConcurrentSessionExecutionSize:  cfg.A3m.Capacity,
+			DisableWorkflowWorker: true,
+			EnableSessionWorker:   true,
+			// Allow only one concurrent session as a3m is not concurrency safe.
+			MaxConcurrentSessionExecutionSize:  1,
 			MaxConcurrentActivityExecutionSize: 1,
 			Interceptors: []temporalsdk_interceptor.WorkerInterceptor{
 				temporal_tools.NewLoggerInterceptor(logger),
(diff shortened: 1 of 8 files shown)


jraddaoui:
For the concurrency issue with a3m, I tried with a semaphore approach to lock access from multiple sessions to the same a3m instance, code mostly generated by Claude: https://github.com/artefactual-sdps/enduro/commit/2cb5e8fb66a5cde9977925922744fc6e03df3f90. While this works, I think it adds to much complexity to "fake" concurrency against a3m, so I added a comment in the config that suggests deploying multiple workers and a3m instances to enable parallelism, something we can easily do in the dev env and in Kubernetes in general by adding more replicas of the enduro-a3m StatefulSet.

djjuhasz (approved):
The code LGTM @jraddaoui. Going from sending one AM SIP at a time to ten seem like a big jump to me, but I don't know whether AM will try and process all of the SIPs concurrently, or just queue them.

One minor note that there's a typo "transfger" in the commit message for the second commit.

jraddaoui:
Thanks @djjuhasz! I checked [in Slack](https://artefactual.slack.com/archives/C07DN6XEW/p1765560062216219) and it looks like they normally use 20 as max concurrent packages in AM configuration, so I'll match that value.

djjuhasz:
@jraddaoui I don't really see the advantage of sending 20 SIPs to AM all at once if it only has one worker configured, which I think is the standard deployment, and it seems likely to cause more disk full errors if the SIPs aren't tiny.  I guess we'll see how it plays out in staging / testing but I think this default is going to be impractical in most real-world deployments.

jraddaoui:
I'm not sure what you mean, that's why I asked in Slack and most real-world deployments (at least the ones we manage) have 20 setup as ARCHIVEMATICA_MCPSERVER_MCPSERVER_CONCURRENT_PACKAGES, and "recent AM has a pool or workers for mcp-client". Check the conversation linked above and follow-up there if you think this is wrong, we can easily change the default in another PR.

djjuhasz:
I just think that there are probably not many AM operators that are actually sending it 20 SIPs to process concurrently, but I may be wrong. I doubt that ARCHIVEMATICA_MCPSERVER_MCPSERVER_CONCURRENT_PACKAGES is a good indication of the number of packages that Archivematica can practically process at one time.  But again, we'll see how it goes when we try batch processing in testing.

PR #1461, opened 2025-12-11
Release session while waiting for batch signal

jraddaoui:
Complete the current session before waiting for the batch signal to allow other sessions to execute in the worker. Recreate the session after in the same worker to continue having access to the local files.

The recreated session takes an activity slot in the same worker where the functional activities run. Set `MaxConcurrentActivityExecutionSize` to the double of `MaxConcurrentSessionExecutionSize` in both workers to avoid deadlocks in this situation.

Comes from #1456. Based in #1460. Refs #1405.

### DIFF
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -205,12 +205,12 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
 		var sessErr error
 		maxAttempts := 5
 
-		activityOpts := temporalsdk_workflow.WithActivityOptions(ctx, temporalsdk_workflow.ActivityOptions{
+		ctx = temporalsdk_workflow.WithActivityOptions(ctx, temporalsdk_workflow.ActivityOptions{
 			StartToCloseTimeout: time.Minute,
 			TaskQueue:           w.cfg.Preservation.TaskQueue,
 		})
 		for attempt := 1; attempt <= maxAttempts; attempt++ {
-			sessCtx, err := temporalsdk_workflow.CreateSession(activityOpts, &temporalsdk_workflow.SessionOptions{
+			sessCtx, err := temporalsdk_workflow.CreateSession(ctx, &temporalsdk_workflow.SessionOptions{
 				CreationTimeout:  forever,
 				ExecutionTimeout: forever,
 			})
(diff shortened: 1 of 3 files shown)


internal/workflow/processing.go
```diff
@@ -205,20 +205,20 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
 		var sessErr error
 		maxAttempts := 5
 
-		activityOpts := temporalsdk_workflow.WithActivityOptions(ctx, temporalsdk_workflow.ActivityOptions{
+		ctx = temporalsdk_workflow.WithActivityOptions(ctx, temporalsdk_workflow.ActivityOptions{
```
jraddaoui, line 208: https://github.com/artefactual-sdps/enduro/pull/1456#discussion_r2607996005

internal/workflow/processing.go
```diff
@@ -287,12 +287,10 @@ func (w *ProcessingWorkflow) Execute(ctx temporalsdk_workflow.Context, req *inge
...
 // SessionHandler runs activities that belong to the same session.
 func (w *ProcessingWorkflow) SessionHandler(
-	sessCtx temporalsdk_workflow.Context,
-	attempt int,
-	state *workflowState,
+	ctx, sessCtx temporalsdk_workflow.Context, attempt int, state *workflowState,
 ) error {
 	// Cleanup session files on exit.
-	defer w.sessionCleanup(sessCtx, state)
+	defer func() { w.sessionCleanup(sessCtx, state) }()
```
jraddaoui, line 293: https://github.com/artefactual-sdps/enduro/pull/1456#discussion_r2607997336

jraddaoui (commented):
@djjuhasz, check https://github.com/artefactual-sdps/enduro/pull/1456#discussion_r2602287081 for a bit more info about the concurrent activities change. About moving the contexts around, I thought about adding them to the state, but it feels odd to hide the context that way.

djjuhasz (approved):
Wow, good work @jraddaoui.  The batch processing signalling is a lot more complicated than I anticipated when we started discussing handling batches and this is a clever way to keep processing on the same worker when all the SIPs validate without having to keep the session open the whole time.

jraddaoui:
Thanks @djjuhasz!

PR #1462, opened 2025-12-11
Wait for SIPs validation in Batch workflow

jraddaoui:
Poll for SIP statuses until they are all validated or in a final status, signal processing workflows to continue or stop processing. Poll for SIP statuses until they are all ingested or in another final status, update final Batch status based on the final SIP statuses.

Comes from #1456. Refs #1405.

### DIFF
--- a/internal/workflow/batch.go
+++ b/internal/workflow/batch.go
@@ -15,6 +15,7 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/datatypes"
 	"github.com/artefactual-sdps/enduro/internal/enums"
 	"github.com/artefactual-sdps/enduro/internal/ingest"
+	"github.com/artefactual-sdps/enduro/internal/workflow/activities"
 )
 
 type BatchWorkflow struct {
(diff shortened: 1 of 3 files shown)


internal/workflow/batch.go
```diff
@@ -81,17 +82,63 @@ func (w *BatchWorkflow) Execute(ctx temporalsdk_workflow.Context, req *ingest.Ba
...
-	// - Wait for all SIPs validation and send signals to continue/stop processing.
-	// - Check for all SIPs final status and update batch status.
-	// - Handle retention periods here or in child workflows.
+	// TODO: consider errors in child workflows and handle failures, retries, or compensations.
 
-	// Wait for all SIP workflows to complete.
-	if err := w.waitForWorkflowsCompletion(ctx, state); err != nil {
+	defer func() {
+		// Always wait for all SIP workflows to complete from this point.
+		if err := w.waitForWorkflowsCompletion(ctx, state); err != nil {
```
djjuhasz, line 89: I don't understand the need for this deferred `waitForWorkflowsCompletion()` with the current code. The two "PollSIPStatuesActivity" executions will block until the SIPs reach the desired status or fail. After each SIP reaches `ingested` status the SIP workflow should be complete.

Is there some reason to wait for all of the SIP workflows to return after the batch workflow fails or completed and before doing cleanup?

djjuhasz (changes requested):
@jraddaoui looks good overall, I'm just not sure that we need to both poll all the SIPs for status and wait all the SIP workflows to complete.

internal/workflow/batch_test.go
```diff
@@ -229,5 +557,5 @@ func (s *BatchWorkflowTestSuite) TestBatchError() {
 		},
 		SIPSourceID: sourceID,
 		Keys:        []string{batchSIP1Key, batchSIP2Key},
-	}, true)
+	}, nil, nil, true)
 }
```
jraddaoui, line 561: I'll update the tests in a following PR to match the setup/expectations/tests pattern we added in the processing workflow tests.

internal/workflow/batch.go
jraddaoui: Yes, we run them with PARENT_CLOSE_POLICY_TERMINATE so when the batch workflow ends it terminates all child workflows that may still be running. Right now the final SIP status update is one of the last things that happen in the processing workflow, but that may not always be the case. Also, in the future, we may want to get some information from the child workflow results.

djjuhasz (approved):
:+1:

PR #1463, opened 2025-12-12
Ensure tracing is resilient

sevein:
In 8f0059 the telemetry env vars were moved into Kustomize patches under
hack/kube/overlays/observability, but that overlay's kustomization.yaml only
listed the observability stack (Prometheus, Grafana Alloy, Tempo, Grafana)
and never rendered the enduro, enduro-a3m, or enduro-am resources those patches
targeted. Because Kustomize only patches resources it renders, the telemetry
env injections were no-ops, so the deployments launched without tracing env
vars and no spans reached Grafana.

The new approach keeps the telemetry env vars directly in the base/dev
manifests so tracing is always enabled, and the application's tracer setup now
fails fast on invalid config but uses a retrying OTLP exporter for runtime
connectivity. If the observability stack is temporarily unreachable, the app
stays up and keeps retrying until it can export spans.

4557f53314eaa89b33ecb4f42bb1e336b468eb88 changes the config in Tempo to stop logging received spans (related to https://github.com/artefactual-sdps/enduro/pull/1454).


internal/telemetry/traces.go
```diff
@@ -26,26 +26,29 @@ func TracerProvider(
...
+	logger = logger.WithValues("enabled", cfg.Traces.Enabled, "addr", cfg.Traces.Address)
 	if !cfg.Traces.Enabled || cfg.Traces.Address == "" {
-		logger.V(1).
-			Info("Tracing system is disabled.", "enabled", cfg.Traces.Enabled, "addr", cfg.Traces.Address, "sampling-ration", cfg.Traces.SamplingRatio)
+		logger.V(1).Info("Tracing system is disabled.")
 		shutdown := func(context.Context) error { return nil }
 		return noop.NewTracerProvider(), shutdown, nil
 	}
 
-	conn, err := grpc.DialContext( //nolint SA1019: grpc.DialContext is deprecated: use NewClient instead.
```
jraddaoui, line 36: Nice to see this deprecation warnings go.

internal/telemetry/traces.go
```diff
@@ -26,26 +26,29 @@ func TracerProvider(
...
-		ctx,
-		cfg.Traces.Address,
-		grpc.WithTransportCredentials(insecure.NewCredentials()),
-		grpc.WithBlock(), //nolint SA1019: grpc.WithBlock is deprecated: this DialOption is not supported by NewClient.
-	)
-	if err != nil {
-		return nil, nil, fmt.Errorf("TracerProvider: connect to collector: %v", err)
+	if _, _, err := net.SplitHostPort(cfg.Traces.Address); err != nil {
+		return nil, nil, fmt.Errorf("TracerProvider: invalid traces address %q: %w", cfg.Traces.Address, err)
```
jraddaoui, line 39: Even with this check, somebody could enter a valid wrong address. Now with the unlimited retries it will never fail, so we don't get an error right away if it's not pointing to the correct address.

internal/telemetry/traces.go
```diff
@@ -26,26 +26,29 @@ func TracerProvider(
...
+	exporter, err := otlptracegrpc.New(ctx,
+		otlptracegrpc.WithEndpoint(cfg.Traces.Address),
+		otlptracegrpc.WithInsecure(),
+		otlptracegrpc.WithRetry(otlptracegrpc.RetryConfig{
+			Enabled:         true,
+			InitialInterval: time.Second,
+			MaxInterval:     30 * time.Second,
+			MaxElapsedTime:  0,
+		}),
+	)
```
jraddaoui, line 54: Can we log something while retrying to know when the configuration is wrong?

jraddaoui (commented):
Nice @sevein, I noticed this yesterday, but I was focusing on the other PRs, thanks for taking a look at it.

I have my conflicts about this approach, I like the retry idea but it obscures possible configuration issues. Would there be a way to show something in the logs? If that's not possible, I'm not sure if I'd prefer to log an error without failing the process to know about the issue.

sevein:
Makes sense, I'll see if I can inject our logger or something like that.

sevein:
@jraddaoui e3b41b4 looks like it does the trick.

image

---

I used Info instead of Error because the latter is way too verbose with the full traceback. What do you think? As I see it, it's an error, but not a deal breaker.

> Things that users care about when using your software.
> * Mostly errors, using logger.Error(). Use logger.Info() very sparely.
(https://enduro.readthedocs.io/dev-manual/logging/#how-do-i-choose-my-v-levels-from-the-logr-docs)

jraddaoui (approved):
Nice @sevein, thanks! I agree, better with info.

PR #1465, opened 2025-12-12
Disable login form in Grafana

sevein:
This change disables the Grafana username/password login form and redirects users directly to the Keycloak authentication flow.


jraddaoui (approved):
:ok_hand:

PR #1466, opened 2025-12-12
Remove raw SQL queries

sevein:
Putting the agent on a debt detox!

### DIFF
--- a/internal/ingest/convert.go
+++ b/internal/ingest/convert.go
@@ -46,7 +46,7 @@ func sipIngestAuditEvent(s *datatypes.SIP) *auditlog.Event {
 func workflowToGoa(w *datatypes.Workflow) *goaingest.SIPWorkflow {
 	var startedAt string
 	if w.StartedAt.Valid {
-		startedAt = w.StartedAt.Time.In(time.UTC).Format(time.RFC3339)
+		startedAt = w.StartedAt.Time.Format(time.RFC3339)
 	}
 
 	res := &goaingest.SIPWorkflow{
--- a/internal/ingest/convert.go
+++ b/internal/ingest/convert.go
@@ -46,10 +46,10 @@ func sipIngestAuditEvent(s *datatypes.SIP) *auditlog.Event {
 func workflowToGoa(w *datatypes.Workflow) *goaingest.SIPWorkflow {
 	var startedAt string
 	if w.StartedAt.Valid {
-		startedAt = w.StartedAt.Time.Format(time.RFC3339)
+		startedAt = w.StartedAt.Time.In(time.UTC).Format(time.RFC3339)
 	}
 
-	return &goaingest.SIPWorkflow{
+	res := &goaingest.SIPWorkflow{
 		UUID:        w.UUID,
 		TemporalID:  w.TemporalID,
 		Type:        w.Type.String(),
@@ -58,6 +58,15 @@ func workflowToGoa(w *datatypes.Workflow) *goaingest.SIPWorkflow {
 		CompletedAt: db.FormatOptionalTime(w.CompletedAt),
 		SipUUID:     w.SIPUUID,
 	}
+
+	if len(w.Tasks) > 0 {
+		res.Tasks = make([]*goaingest.SIPTask, 0, len(w.Tasks))
+		for _, task := range w.Tasks {
+			res.Tasks = append(res.Tasks, taskToGoa(task))
+		}
+	}
+
+	return res
 }
 
 // taskToGoa returns the API representation of a task.
@@ -69,7 +78,7 @@ func taskToGoa(task *datatypes.Task) *goaingest.SIPTask {
 
 		// TODO: Make Goa StartedAt a pointer to a string to avoid having to
 		// convert a null time to an empty (zero value) string.
-		StartedAt: ref.DerefZero(db.FormatOptionalTime(task.CompletedAt)),
+		StartedAt: ref.DerefZero(db.FormatOptionalTime(task.StartedAt)),
 
 		CompletedAt:  db.FormatOptionalTime(task.CompletedAt),
 		Note:         &task.Note,
--- a/internal/ingest/goa.go
+++ b/internal/ingest/goa.go
@@ -2,9 +2,7 @@ package ingest
 
 import (
 	"context"
-	"database/sql"
 	"errors"
-	"fmt"
 
 	"github.com/google/uuid"
 	"go.artefactual.dev/tools/ref"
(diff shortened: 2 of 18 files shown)


jraddaoui:
Thanks @sevein! We were just talking about this yesterday, I'll take a look later today. I guess you forgot to remove the draft status from the PR :wink:.

internal/ingest/convert.go
```diff
@@ -69,7 +78,7 @@ func taskToGoa(task *datatypes.Task) *goaingest.SIPTask {
 
 		// TODO: Make Goa StartedAt a pointer to a string to avoid having to
 		// convert a null time to an empty (zero value) string.
-		StartedAt: ref.DerefZero(db.FormatOptionalTime(task.CompletedAt)),
+		StartedAt: ref.DerefZero(db.FormatOptionalTime(task.StartedAt)),
```
jraddaoui, line 81: :man_facepalming: I recall seeing this issue in the UI, but we changed how the task are displayed and completely forgot about it, thanks!

internal/ingest/goa.go
```diff
@@ -177,56 +175,22 @@ func (svc *ingestImpl) ListSipWorkflows(
 		return nil, goaingest.MakeNotValid(errors.New("invalid UUID"))
 	}
 
-	s, err := svc.perSvc.ReadSIP(ctx, sipUUID)
-	if err == sql.ErrNoRows {
+	if _, err := svc.perSvc.ReadSIP(ctx, sipUUID); err == persistence.ErrNotFound {
 		return nil, &goaingest.SIPNotFound{UUID: payload.UUID, Message: "SIP not found"}
 	} else if err != nil {
 		return nil, goaingest.MakeNotAvailable(errors.New("cannot perform operation"))
```
jraddaoui, line 181: Not for this PR, but we are starting to return internal errors (500) from these cases.

internal/ingest/ingest.go
```diff
@@ -146,14 +143,12 @@ func (svc *ingestImpl) UpdateSIP(
 }
 
 func (svc *ingestImpl) SetStatus(ctx context.Context, id uuid.UUID, status enums.SIPStatus) error {
```
jraddaoui, line 145: Also not for this PR, but we should rename to `SetSIPStatus`. But we also mentioned the possibility to just have a general `UpdateSIP` method/event.

internal/ingest/ingest.go
```diff
@@ -163,19 +158,18 @@ func (svc *ingestImpl) SetStatus(ctx context.Context, id uuid.UUID, status enums
...
-
-	if err := svc.updateRow(ctx, query, args); err != nil {
-		return err
+	_, err := svc.perSvc.UpdateSIP(ctx, id, func(s *datatypes.SIP) (*datatypes.SIP, error) {
+		s.Status = enums.SIPStatusProcessing
+		if !startedAt.IsZero() {
+			s.StartedAt = sql.NullTime{
+				Time:  startedAt,
+				Valid: true,
+			}
```
jraddaoui, line 167: Something else we discussed the other day, now without raw SQL, we plan to move to `time.Time` all over.

internal/ingest/ingest.go
```diff
@@ -163,19 +158,18 @@ func (svc *ingestImpl) SetStatus(ctx context.Context, id uuid.UUID, status enums
...
+		}
+		return s, nil
+	})
+	if err != nil {
+		return fmt.Errorf("error updating SIP: %v", err)
```
jraddaoui, line 172: Another thing not for this PR, we should stop using "error" in the error messages.

internal/persistence/ent/client/convert.go
```diff
@@ -114,3 +101,42 @@ func convertBatch(batch *db.Batch) *datatypes.Batch {
 
 	return &b
 }
+
+// convertWorkflow converts an entgo `db.Workflow` representation
+// to a `datatypes.Task` representation.
```
jraddaoui: ... `datatypes.Workflow` representation.

internal/persistence/ent/client/convert.go
```diff
@@ -114,3 +101,42 @@ func convertBatch(batch *db.Batch) *datatypes.Batch {
...
+func convertWorkflow(dbw *db.Workflow) *datatypes.Workflow {
+	w := &datatypes.Workflow{
+		ID:          dbw.ID,
+		UUID:        dbw.UUID,
+		TemporalID:  dbw.TemporalID,
+		Type:        dbw.Type,
+		Status:      enums.WorkflowStatus(uint(dbw.Status)), // #nosec G115 -- constrained value.
```
jraddaoui, line 113: Another mental note not for this PR, move this to a proper enum like we did with the SIP statuses.

internal/persistence/ent/client/convert.go
```diff
@@ -114,3 +101,42 @@ func convertBatch(batch *db.Batch) *datatypes.Batch {
...
+		CompletedAt: nullTime(dbw.CompletedAt),
+	}
+
+	if dbw.Edges.Sip != nil {
+		w.SIPUUID = dbw.Edges.Sip.UUID
+	}
+
+	// Only populate Tasks if they were loaded, preserving nil vs empty semantics.
+	if dbw.Edges.Tasks != nil {
+		w.Tasks = make([]*datatypes.Task, 0, len(dbw.Edges.Tasks))
```
jraddaoui: Nitpicking, not a big deal, but knowing the length of tasks there is a small improvement using `make([]*datatypes.Task, len(dbw.Edges.Tasks))` and using the index instead of append.

internal/persistence/ent/client/convert.go
```diff
@@ -114,3 +101,42 @@ func convertBatch(batch *db.Batch) *datatypes.Batch {
...
+
+	return w
+}
+
+func nullTime(t time.Time) sql.NullTime {
+	if t.IsZero() {
+		return sql.NullTime{}
+	}
+	return sql.NullTime{Time: t.UTC(), Valid: true}
+}
```
jraddaoui, line 142: Nice! until we get rid of `sql.NullTime`.

internal/persistence/ent/client/workflow.go
```diff
@@ -66,3 +69,96 @@ func (c *client) CreateWorkflow(ctx context.Context, w *datatypes.Workflow) erro
...
+) (*datatypes.Workflow, error) {
+	tx, err := c.ent.BeginTx(ctx, nil)
+	if err != nil {
+		return nil, newDBError(err)
+	}
+
+	dbw, err := tx.Workflow.Query().
+		Where(workflow.ID(id)).
+		WithSip().
+		WithTasks().
```
jraddaoui: I think we don't want the tasks on update, there may be a big set of them. At least in the WS event and in the dashboard we don't care about them:

https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/monitor-ingest.ts#L70-L73

internal/persistence/ent/client/workflow.go
```diff
@@ -66,3 +69,96 @@ func (c *client) CreateWorkflow(ctx context.Context, w *datatypes.Workflow) erro
...
+	if up.StartedAt.Valid {
+		q.SetStartedAt(up.StartedAt.Time.UTC())
+	} else {
+		q.ClearStartedAt()
+	}
+	if up.CompletedAt.Valid {
+		q.SetCompletedAt(up.CompletedAt.Time.UTC())
+	} else {
+		q.ClearCompletedAt()
+	}
```
jraddaoui, line 116: Having conflicts about clearing, I guess it could be good for restarts, if we use the same workflow record.

internal/persistence/ent/client/workflow.go
```diff
@@ -66,3 +69,96 @@ func (c *client) CreateWorkflow(ctx context.Context, w *datatypes.Workflow) erro
...
+
+	dbw, err = tx.Workflow.Query().
+		Where(workflow.ID(dbw.ID)).
+		WithSip().
+		WithTasks().
+		Only(ctx)
+	if err != nil {
+		return nil, rollback(tx, newDBError(err))
+	}
```
jraddaoui: We don't need to read the workflow again, use the one returned by `Save`. We don't care about the tasks, but we care about the SIP UUID. Check the storage version:

https://github.com/artefactual-sdps/enduro/blob/main/internal/storage/persistence/ent/client/workflow.go#L67-L113

internal/persistence/ent/client/workflow.go
```diff
@@ -66,3 +69,96 @@ func (c *client) CreateWorkflow(ctx context.Context, w *datatypes.Workflow) erro
...
+	if err != nil {
+		return nil, newDBError(err)
+	}
+
+	return convertWorkflow(dbw), nil
+}
+
+func (c *client) ListWorkflowsBySIP(ctx context.Context, sipUUID uuid.UUID) ([]*datatypes.Workflow, error) {
+	dbw, err := c.ent.Workflow.Query().
+		WithSip().
```
jraddaoui: Small improvement for the list, just fetch the SIP UUID `.WithSip(func(q *db.SIPQuery) { q.Select(sip.FieldUUID) })` (untested).

jraddaoui (changes requested):
SO nice to see this go, thanks @sevein!!

internal/ingest/ingest.go
sevein: That makes a lot of sense!

internal/persistence/ent/client/convert.go
sevein: Thanks, this shows I'm a bit rusty, the pre-sized slice trick is easy to forget!

internal/persistence/ent/client/workflow.go
sevein: Nice, good catch!

sevein:
@jraddaoui I have addressed all your comments and observations. Please let me know if I missed anything.

These are the non-actionable comments shared in this pull request for future reference:
* https://github.com/artefactual-sdps/enduro/pull/1466#discussion_r2620315499
  » Return internal server error.
* https://github.com/artefactual-sdps/enduro/pull/1466#discussion_r2620325398
 » Replace SetStatus with SetSIPStatus or UpdateSIP.
* https://github.com/artefactual-sdps/enduro/pull/1466#discussion_r2620329180
  » Avoid using sql.NullTime in domain types.
* https://github.com/artefactual-sdps/enduro/pull/1466#discussion_r2620339213
  » Stop using "error ..." prefix in error messages
* https://github.com/artefactual-sdps/enduro/pull/1466#discussion_r2620349449
  » Adopt enums in workflow db type

jraddaoui (approved):
Thanks for the summary @sevein! LGTM!

PR #1469, opened 2025-12-16
Allow upload as batch from SIP source

jraddaoui:
Refs #1405.

### DIFF
--- a/dashboard/src/components/SIPUploadSource.vue
+++ b/dashboard/src/components/SIPUploadSource.vue
@@ -5,8 +5,10 @@ import { useRouter } from "vue-router/auto";
 
 import { api, client } from "@/client";
 import { humanFileSize } from "@/composables/format";
+import { useAuthStore } from "@/stores/auth";
 import IconBundle from "~icons/clarity/bundle-line";
 
+const authStore = useAuthStore();
 const router = useRouter();
 
 // Hardcoded source ID that must match the one in the backend configuration.
@@ -17,6 +19,12 @@ const selectedSips = ref<string[]>([]);
 const nextCursor = ref<string | undefined>(undefined);
 const listContainer = ref<HTMLElement>();
 const errorMessage = ref<string | null>(null);
+
+// At this point, the user must have at least one of the permissions
+// to create SIP or batches. Determine whether to default to batch ingest.
+// The switch will be disabled if the user doesn't have both permissions.
+const isBatch = ref(!authStore.checkAttributes(["ingest:sips:create"]));
+const batchID = ref("");
 
 const { execute, isLoading } = useAsyncState(
   async (cursor?: string | undefined) => {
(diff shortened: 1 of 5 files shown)


dashboard/src/components/SIPUploadSource.vue
```diff
@@ -18,6 +20,12 @@ const nextCursor = ref<string | undefined>(undefined);
 const listContainer = ref<HTMLElement>();
 const errorMessage = ref<string | null>(null);
 
+// At this point, the user must have at least one of the permissions
+// to create SIP or batches. Determine whether to default to batch ingest.
+// The switch will be disabled if the user doesn't have both permissions.
+const isBatch = ref(!authStore.checkAttributes(["ingest:sips:create"]));
+const batchID = ref("");
+
 const { execute, isLoading } = useAsyncState(
```
sevein, line 29: I suspect you know but I've just realized while testing that ingestListSipSourceObjects is seeing a server error (panic) when len(page) is zero:

```go
// The next token is the key of the last object of the page.
next := []byte(page[len(page)-1].Key)
```

dashboard/src/components/__tests__/SIPUploadSource.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    ],
+    config: {
+      globalProperties: {
+        $filters: { formatDateTime: vi.fn() },
+      },
+    },
+  },
+});
+
+describe("SIPUploadSource.vue", () => {
```
sevein, line 51: One error-path test would be cool. I've tried this:

```ts
  it("shows an error when ingest fails", async () => {
    ingestListSipSourceObjects.mockResolvedValueOnce({
      objects: [{ key: "sip-1", size: 123, modTime: "2024-01-01T00:00:00Z" }],
    });
    ingestAddSip.mockRejectedValueOnce(new Error("boom"));

    const wrapper = mount(SIPUploadSource, mountOptions());
    await flushPromises();

    await wrapper.get("#cb-sip-1").setValue(true);
    await wrapper.get("button.btn-primary").trigger("click");
    await flushPromises();

    expect(push).not.toHaveBeenCalled();
    expect(wrapper.text()).toContain("Failed to start ingest.");
  });
```

dashboard/src/components/SIPUploadSource.vue
```diff
@@ -164,7 +182,41 @@ const clickSip = (key: string) => {
     </div>
```
sevein, line 182:
```html
    <div class="form-text text-end">
      Each SIP uploaded will be ingested separately in its own workflow.
    </div>
```

... is still accurate?

sevein (approved):
LGTM

dashboard/src/components/SIPUploadSource.vue
jraddaoui: Yeah, I noticed that a while ago, I will create an issue for it. Thanks!

dashboard/src/components/SIPUploadSource.vue
jraddaoui: :thinking: Kind of, I'll show/hide it based on `isBatch`.

PR #1470, opened 2025-12-17
Add ShowBatch and ListBatches implementation to ingest service

jraddaoui:
Refs #1405.

### DIFF
--- a/internal/ingest/goa_batch.go
+++ b/internal/ingest/goa_batch.go
@@ -13,8 +13,6 @@ import (
 	"github.com/artefactual-sdps/enduro/internal/persistence"
 )
 
-var ErrNotImplemented error = goaingest.MakeNotImplemented(errors.New("not implemented"))
-
 func (svc *ingestImpl) AddBatch(
 	ctx context.Context,
 	payload *goaingest.AddBatchPayload,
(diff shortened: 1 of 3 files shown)


internal/ingest/goa_batch.go
```diff
@@ -87,12 +86,50 @@ func (svc *ingestImpl) ListBatches(
...
+
+	pf, err := listBatchesPayloadToBatchFilter(payload)
+	if err != nil {
+		return nil, err
+	}
+
+	r, pg, err := svc.perSvc.ListBatches(ctx, pf)
+	if err != nil {
+		svc.logger.Error(err, "ListBatches")
+		return nil, ErrInternalError
```
jraddaoui, line 101: Related to the TODO from the other PR, when we normalize to return internal errors, it would be nice to create a middleware to do this consistently, like you did in other project.

internal/ingest/goa_batch.go
```diff
@@ -88,7 +86,32 @@ func (svc *ingestImpl) ListBatches(
 	ctx context.Context,
 	payload *goaingest.ListBatchesPayload,
 ) (*goaingest.Batches, error) {
-	return nil, ErrNotImplemented
+	if payload == nil {
```
sevein, line 89: Is this necessary? Maybe goa guarantees that the pointer can never be nil... not something that I remember off the top of my head.

sevein (approved):
👍

internal/ingest/goa_batch.go
jraddaoui: Yeah, similar to the Temporal activities params, I really doubt they will ever be nil. Anyway, just in case, it doesn't hurt.

PR #1471, opened 2025-12-17
Add batch list and view pages to dashboard

jraddaoui:
Refs #1405.

### DIFF
--- a/dashboard/src/pages/ingest/batches/[id]/index.vue
+++ b/dashboard/src/pages/ingest/batches/[id]/index.vue
@@ -52,6 +52,16 @@ const statuses = [
     description:
       "The SIP workflow encountered a system error and ingest was aborted.",
   },
+  {
+    status: api.EnduroIngestSipStatusEnum.Canceled,
+    description:
+      "The SIP workflow has been canceled as part of a failed batch.",
+  },
+  {
+    status: api.EnduroIngestSipStatusEnum.Validated,
+    description:
+      "The SIP has passed validation and it is waiting for other SIPs in the batch to validate.",
+  },
 ];
 
 onMounted(() => {
--- a/dashboard/src/pages/ingest/batches/[id]/index.vue
+++ b/dashboard/src/pages/ingest/batches/[id]/index.vue
@@ -137,7 +137,7 @@ onMounted(() => {
                   data-bs-title="Toggle legend"
                   @click="toggleLegend"
                 >
-                  <IconInfo style="font-size: 1.2em" aria-hidden="true" />
+                  <IconInfo class="fs-6" aria-hidden="true" />
                   <span class="visually-hidden">Toggle SIP status legend</span>
                 </button>
               </span>
--- a/dashboard/src/pages/ingest/batches/[id]/index.vue
+++ b/dashboard/src/pages/ingest/batches/[id]/index.vue
@@ -1,0 +1,157 @@
+<script setup lang="ts">
+import Tooltip from "bootstrap/js/dist/tooltip";
+import { onMounted, ref } from "vue";
+
+import { api } from "@/client";
+import StatusBadge from "@/components/StatusBadge.vue";
+import StatusLegend from "@/components/StatusLegend.vue";
+import UUID from "@/components/UUID.vue";
+import uploader from "@/composables/uploader";
+import { useAuthStore } from "@/stores/auth";
+import { useBatchStore } from "@/stores/batch";
+import IconInfo from "~icons/akar-icons/info-fill";
+
+const authStore = useAuthStore();
+const batchStore = useBatchStore();
+
+const el = ref<HTMLElement | null>(null);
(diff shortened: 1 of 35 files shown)


dashboard/src/pages/ingest/sips/[id]/index.vue
```diff
@@ -51,6 +51,19 @@ onMounted(() => {
...
+              v-if="authStore.checkAttributes(['ingest:batches:read'])"
+              :to="{
+                name: '/ingest/batches/[id]/',
+                params: { id: sipStore.current.batchUuid },
+              }"
+            >
+              {{ sipStore.current.batchIdentifier }}
+            </RouterLink>
+            <span v-else>{{ sipStore.current.batchIdentifier }}</span>
+          </dd>
```
Comment (sevein) [inline dashboard/src/pages/ingest/sips/[id]/index.vue]: Isn't this a good use case for `<template v-if="...">...</template>`?

dashboard/src/pages/ingest/sips/[id]/index.vue
```diff
@@ -51,6 +51,19 @@ onMounted(() => {
           <dd>
             {{ uploader(sipStore.current) }}
           </dd>
+          <dt v-if="sipStore.current.batchUuid">Batch</dt>
```
Comment (sevein) [inline dashboard/src/pages/ingest/sips/[id]/index.vue]: Wouldn't be nice to have a store getter such as the following?

    hasActiveBatch: (state) => Boolean(state.current?.batchUuid),

dashboard/src/pages/ingest/batches/[id]/index.vue
```diff
@@ -0,0 +1,153 @@
...
+import { useBatchStore } from "@/stores/batch";
+import IconInfo from "~icons/akar-icons/info-fill";
+
+const authStore = useAuthStore();
+const batchStore = useBatchStore();
+
+const el = ref<HTMLElement | null>(null);
+let tooltip: Tooltip | null = null;
+
+let showLegend = ref(false);
```
Comment (sevein) [inline dashboard/src/pages/ingest/batches/[id]/index.vue]: We're only mutating the inner value, this should be a const, right?

dashboard/src/pages/ingest/batches/[id]/index.vue
```diff
@@ -0,0 +1,153 @@
...
+const toggleLegend = () => {
+  showLegend.value = !showLegend.value;
+  if (tooltip) tooltip.hide();
```
Comment (sevein) [inline dashboard/src/pages/ingest/batches/[id]/index.vue]: tooltip?.hide();

dashboard/src/pages/ingest/batches/[id]/index.vue
```diff
@@ -0,0 +1,153 @@
...
+                Status
+                <button
+                  ref="el"
+                  class="btn btn-sm btn-link text-decoration-none ms-auto p-0"
+                  type="button"
+                  data-bs-toggle="tooltip"
+                  data-bs-title="Toggle legend"
+                  @click="toggleLegend"
+                >
+                  <IconInfo style="font-size: 1.2em" aria-hidden="true" />
```
Comment (sevein) [inline dashboard/src/pages/ingest/batches/[id]/index.vue]: Can we use here the `fs-5` class?

dashboard/src/pages/ingest/batches/index.vue
```diff
@@ -0,0 +1,491 @@
+<script setup lang="ts">
```
sevein, line 1: There's some leaking of store and domain details into this component; it'd be cool if it were more purely responsible for wiring the UI to the route and the stores, only expressing intentions like "apply these filters" or "go to page 2", instead of knowing how filters are represented internally or when query params like page need to be added or removed.

sevein (approved):
LGTM. I've tested and works great.

dashboard/src/pages/ingest/sips/[id]/index.vue
Comment (jraddaoui) [reply dashboard/src/pages/ingest/sips/[id]/index.vue]: Now with the template it is only checked once, so I don't see a lot of value for a getter yet.

dashboard/src/pages/ingest/batches/[id]/index.vue
Comment (jraddaoui) [reply dashboard/src/pages/ingest/batches/[id]/index.vue]: Fixed in other pages too.

dashboard/src/pages/ingest/batches/[id]/index.vue
Comment (jraddaoui) [reply dashboard/src/pages/ingest/batches/[id]/index.vue]: Fixed in other pages too.

dashboard/src/pages/ingest/batches/[id]/index.vue
Comment (jraddaoui) [reply dashboard/src/pages/ingest/batches/[id]/index.vue]: It makes it a bit bigger, which changes the row height, `fs-6` makes it a little smaller but I think that works.

Comment (jraddaoui) [reply dashboard/src/pages/ingest/batches/[id]/index.vue]: I have also removed all style attributes from the templates.

dashboard/src/pages/ingest/batches/index.vue
jraddaoui: That's a good point, right now it's the same for SIPs and AIPs, we should revisit this in the future for the three pages and stores.

PR #1480, opened 2026-01-13
Add batch post-storage child workflow

djjuhasz:
Add a hook to call one or more post-storage workflows after batch ingest succeeds.

### DIFF
--- a/internal/batch/config.go
+++ b/internal/batch/config.go
@@ -1,0 +1,24 @@
+package batch
+
+type Config struct {
+	// Poststorage contains the configuration for an optional post-storage
+	// workflow that runs after all SIPs in a batch have been ingested and
+	// stored. If nil, no post-storage workflow will be executed.
+	Poststorage *PostStorageConfig
+}
+
+type PostStorageConfig struct {
+	// Namespace is the Temporal namespace used by the batch post-storage
+	// workflow. A unique Namespace will isolating the post-storage workflow
+	// from other Temporal workflows running on the same server. If no Namespace
+	// is specified, the default namespace will be used.
+	Namespace string
+
+	// TaskQueue is the Temporal task queue that will be used to schedule
+	// batch post-storage workflow execution. A Temporal worker must be
+	// listening on the specified task queue to execute the workflow.
+	TaskQueue string
+
+	// WorkflowName is the name of the batch post-storage Temporal workflow.
+	WorkflowName string
+}
--- a/internal/workflow/batch.go
+++ b/internal/workflow/batch.go
@@ -11,6 +11,7 @@ import (
 	temporalsdk_client "go.temporal.io/sdk/client"
 	temporalsdk_workflow "go.temporal.io/sdk/workflow"
 
+	"github.com/artefactual-sdps/enduro/internal/batch"
 	"github.com/artefactual-sdps/enduro/internal/config"
 	"github.com/artefactual-sdps/enduro/internal/datatypes"
 	"github.com/artefactual-sdps/enduro/internal/enums"
@@ -135,6 +136,13 @@ func (w *BatchWorkflow) Execute(ctx temporalsdk_workflow.Context, req *ingest.Ba
 	// Fail workflow if not all SIPs reached "ingested" status.
 	if !pollIngestedResult.AllExpectedStatus {
 		return fmt.Errorf("not all SIPs reached %q status", enums.SIPStatusIngested)
+	}
+
+	// Run post-storage child workflow, if one is configured.
+	if w.cfg.Batch.Poststorage != nil {
+		if err := w.postStorageWorkflow(ctx, *w.cfg.Batch.Poststorage, state); err != nil {
+			return err
+		}
 	}
 
 	// TODO: handle retention period.
@@ -268,6 +276,47 @@ func (w *BatchWorkflow) waitForWorkflowsCompletion(ctx temporalsdk_workflow.Cont
 			return fmt.Errorf("waiting for workflows: %v", err)
 		}
 	}
+
+	return nil
+}
+
+func (w *BatchWorkflow) postStorageWorkflow(
+	ctx temporalsdk_workflow.Context,
+	cfg batch.PostStorageConfig,
+	state *batchWorkflowState,
+) error {
+	state.logger.Info(
+		"Starting post-storage workflow",
+		"Batch ID", state.batch.UUID.String(),
+		"Workflow name", cfg.WorkflowName,
+	)
+
+	childCtx := temporalsdk_workflow.WithChildOptions(
+		ctx,
+		temporalsdk_workflow.ChildWorkflowOptions{
+			Namespace:         cfg.Namespace,
+			TaskQueue:         cfg.TaskQueue,
+			WorkflowID:        fmt.Sprintf("%s-%s", cfg.WorkflowName, state.batch.UUID.String()),
+			ParentClosePolicy: temporalapi_enums.PARENT_CLOSE_POLICY_TERMINATE,
+		},
+	)
+
+	var res batch.PostStorageResult
+	err := temporalsdk_workflow.ExecuteChildWorkflow(
+		childCtx,
+		cfg.WorkflowName,
+		&batch.PostStorageParams{SIPs: state.SIPs()},
+	).Get(childCtx, &res)
+	if err != nil {
+		return fmt.Errorf("batch %q post-storage workflow: %v", state.batch.UUID.String(), err)
+	}
+
+	state.logger.Info(
+		"Post-storage workflow completed",
+		"Batch ID", state.batch.UUID.String(),
+		"Status", res.Status,
+		"Message", res.Message,
+	)
 
 	return nil
 }
--- a/internal/workflow/batch_test.go
+++ b/internal/workflow/batch_test.go
@@ -14,6 +14,7 @@ import (
 	temporalsdk_workflow "go.temporal.io/sdk/workflow"
 	"go.uber.org/mock/gomock"
 
+	"github.com/artefactual-sdps/enduro/internal/batch"
 	"github.com/artefactual-sdps/enduro/internal/config"
 	"github.com/artefactual-sdps/enduro/internal/datatypes"
 	"github.com/artefactual-sdps/enduro/internal/enums"
@@ -66,6 +67,12 @@ func (s *BatchWorkflowTestSuite) SetupWorkflowTest(cfg config.Configuration) {
 		activities.NewPollSIPStatusesActivity(ingestsvc, time.Microsecond).Execute,
 		temporalsdk_activity.RegisterOptions{Name: activities.PollSIPStatusesActivityName},
 	)
+	if cfg.Batch.Poststorage != nil {
+		s.env.RegisterWorkflowWithOptions(
+			postStorageChildWorkflow,
+			temporalsdk_workflow.RegisterOptions{Name: cfg.Batch.Poststorage.WorkflowName},
+		)
+	}
 
 	s.workflow = NewBatchWorkflow(cfg, rng, ingestsvc, nil)
 }
@@ -81,6 +88,15 @@ func (s *BatchWorkflowTestSuite) processingChildWorkflow(
 	s.childSignals = append(s.childSignals, signal)
 
 	return nil
+}
+
+// postStorageChildWorkflow is a no-op workflow that will be replaced with a
+// mock in test.
+func postStorageChildWorkflow(
+	ctx temporalsdk_workflow.Context,
+	params *batch.PostStorageParams,
+) (*batch.PostStorageResult, error) {
+	return nil, nil
 }
 
 func (s *BatchWorkflowTestSuite) AfterTest(suiteName, testName string) {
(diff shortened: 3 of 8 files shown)


djjuhasz:
@jraddaoui it would be great to get any early feedback you have on this implementation.

internal/workflow/batch.go
```diff
@@ -137,6 +138,15 @@ func (w *BatchWorkflow) Execute(ctx temporalsdk_workflow.Context, req *ingest.Ba
 		return fmt.Errorf("not all SIPs reached %q status", enums.SIPStatusIngested)
 	}
 
+	// Run post-storage child workflow, if one is configured.
```
jraddaoui, line 141: "If any is configured"?

internal/workflow/batch.go
```diff
@@ -137,6 +138,15 @@ func (w *BatchWorkflow) Execute(ctx temporalsdk_workflow.Context, req *ingest.Ba
...
+	if len(w.cfg.Batch.PostStorageWorkflows) > 0 {
+		for _, cc := range w.cfg.Batch.PostStorageWorkflows {
+			if err := w.postStorageWorkflow(ctx, cc, state); err != nil {
+				return err
+			}
+		}
+	}
```
jraddaoui: This is going to start and wait for each child workflow sequentially, stopping at the first failure. I think we should do similar to what we do with the SIP workflows, starting all of them and then waiting for the results (if we want to).

As we talked yesterday, right now there is no way to report this types of error in the API/UI (no workflows and tasks related to the batch) and, if you look at the processing workflow implementation for poststorage, we start the workflows without caring about the results. I think that may be a better first implementation approach in here too.

internal/workflow/batch.go
```diff
@@ -271,3 +281,41 @@ func (w *BatchWorkflow) waitForWorkflowsCompletion(ctx temporalsdk_workflow.Cont
...
+
+func (w *BatchWorkflow) postStorageWorkflow(
+	ctx temporalsdk_workflow.Context,
+	cfg batch.PostStorageConfig,
+	state *batchWorkflowState,
+) error {
+	state.logger.Info(
+		"Starting post-storage workflow",
+		"workflow_name", cfg.WorkflowName,
+	)
```
jraddaoui, line 292: Would be nice to include the batch UUID to differentiate from other batches.

internal/workflow/batch.go
```diff
@@ -271,3 +281,41 @@ func (w *BatchWorkflow) waitForWorkflowsCompletion(ctx temporalsdk_workflow.Cont
...
+			SIPs:  state.SIPs(),
+		},
+	)
+
+	var res batch.PostStorageResult
+	err := cwf.Get(childCtx, res)
+	if err != nil {
+		return fmt.Errorf("post-storage workflow %q: %v", cfg.WorkflowName, err)
+	}
+
```
jraddaoui, line 313: Do we want to check the results here and fail if the status is error? As mentioned above I think it would be better to start with an abandon implementation. If not, we should try to start each workflow in parallel and wait for all results. Not sure what we would want to do with the batch if any of the poststorage workflows fail right now.

To be clear, in the end I'd prefer to wait for all poststorage child workflows and report the results in both the processing and the batch workflows. But that could be done as a follow-up addressing both cases

internal/workflow/batch_test.go
```diff
@@ -48,6 +49,10 @@ type BatchWorkflowTestSuite struct {
 
 	// childSignals holds the signals received by the child processing workflows.
 	childSignals []ingest.BatchSignal
+
+	// postStorageChildWorkflow is the mock implementation of the post-storage
+	// workflow used in tests.
+	postStorageChildWorkflow func(temporalsdk_workflow.Context, *batch.PostStorageParams) error
```
jraddaoui: This will return a result too, right?

internal/workflow/batch_test.go
```diff
@@ -117,14 +132,15 @@ func TestBatchWorkflow(t *testing.T) {
...
-// - Batch status update (ingested).
+// - Update batch status (processing).
+// - Create a SIP for each key.
+// - Start a child processing workflow for each SIP.
+// - Poll SIP statuses until all are validated.
+// - Signal SIP workflows to continue processing.
+// - Poll SIP statuses until all are ingested.
+// - Wait for all child workflows to complete.
+// - Update batch status (ingested).
+// - Run post-storage child workflows.
```
jraddaoui, line 144: We should add the config and mock the post-storage calls in the test, I guess this is just WIP ;)

jraddaoui (commented):
This is looking great so far @djjuhasz, thanks!

internal/workflow/batch.go
djjuhasz: I think we want to show some sort of error in the UI if the child workflow fails so the user is not left totally in the dark and can take corrective action if necessary.

I think running the post-storage workflows in parallel could cause problems depending on what the workflows do - if they modify the packages in any way then there could be race conditions and it may be desirable to stop execution if workflow 1 fails rather than execute workflow 2.  I think really this is a good reason to just implement a single post-storage workflow for now, as we don't have a use case currently for running multiple post-storage workflows, and because it avoids the question of whether to run multiple workflow sequentially or concurrently.

internal/workflow/batch_test.go
djjuhasz: Oh, yes should do.

internal/workflow/batch_test.go
djjuhasz: Yes, working on it. :)

internal/workflow/batch.go
jraddaoui: Yeah, good call. We may want to do the same for poststorage in the processing workflow. A custom child workflow should be able to do it all, even orchestrate/parallelize other workflows/activities by itself.

internal/workflow/batch.go
djjuhasz: It's only one now. ;)

internal/workflow/batch_test.go
djjuhasz: Done.

internal/workflow/batch_test.go
djjuhasz: I've added a mock post-storage workflow function and have confirmed that it is being called.  I am not testing that the parameters passed the workflow matched expectations, but I don't think that is really that useful to test.

djjuhasz:
@jraddaoui I think this is ready to merge, unless you notice any problems.

internal/workflow/batch.go
djjuhasz: I think this is okay now that it's just one post-storage workflow?

jraddaoui: I thought you were going to check `res.Status` and fail the Batch based on that. Something that can happen in another PR when/if we figure out how to display workflow details/errors for batches.

internal/batch/config.go
```diff
@@ -0,0 +1,11 @@
+package batch
+
+type Config struct {
+	PostStorageWorkflow *PostStorageConfig
```
jraddaoui: Looking at the TOML file, this should be `PostStorage` instead of `PostStorageWorkflow`.

internal/workflow/batch.go
```diff
@@ -271,3 +279,49 @@ func (w *BatchWorkflow) waitForWorkflowsCompletion(ctx temporalsdk_workflow.Cont
...
+		ctx,
+		temporalsdk_workflow.ChildWorkflowOptions{
+			Namespace:         cfg.Namespace,
+			TaskQueue:         cfg.TaskQueue,
+			WorkflowID:        fmt.Sprintf("%s-%s", cfg.WorkflowName, state.batch.UUID.String()),
+			ParentClosePolicy: temporalapi_enums.PARENT_CLOSE_POLICY_TERMINATE,
+		},
+	)
+
+	cwf := temporalsdk_workflow.ExecuteChildWorkflow(
```
jraddaoui: Nitpicking, we are just calling Get directly, no need to have a variable for the workflow future.

internal/workflow/batch_test.go
jraddaoui: I think it's useful to check the parameters when possible. Changes to the workflow state during development may affect the call parameters, it would be nice to catch that ASAP.

I think you found that a bit complex here because you followed the SIP child workflows pattern, but that shouldn't be needed here, check my following comment.

internal/workflow/batch_test.go
```diff
@@ -66,6 +71,22 @@ func (s *BatchWorkflowTestSuite) SetupWorkflowTest(cfg config.Configuration) {
...
+			) (*batch.PostStorageResult, error) {
+				s.postStorageCallCount++
+				return &batch.PostStorageResult{
+					Status:  batch.PostStorageSuccess,
+					Message: "Batch post-storage workflow executed successfully",
+				}, nil
+			},
+			temporalsdk_workflow.RegisterOptions{Name: cfg.Batch.PostStorageWorkflow.WorkflowName},
+		)
+	}
```
jraddaoui: I'd not mock the workflow this way as it makes it harder to check the call parameters and mock the returned result. The reason why I mocked the processing child workflows this way here is to be able to access and check the signals sent from the batch workflow, but that required to also check the call parameters differently.

That's not necessary in this workflow mock, I'd follow the usual way we have to mock activities and workflows in this case, check how we do it for the preprocessing child workflow in the processing workflow tests.

internal/workflow/batch_test.go
```diff
@@ -103,6 +124,7 @@ func (s *BatchWorkflowTestSuite) ExecuteAndValidateWorkflow(
 
 	s.Equal(s.childRequests, childRequests)
 	s.Equal(s.childSignals, childSignals)
+	s.Equal(1, s.postStorageCallCount)
```
jraddaoui: How is this one in all cases, if it's only configured in one test?

internal/workflow/batch.go
```diff
@@ -137,6 +138,13 @@ func (w *BatchWorkflow) Execute(ctx temporalsdk_workflow.Context, req *ingest.Ba
 		return fmt.Errorf("not all SIPs reached %q status", enums.SIPStatusIngested)
 	}
 
+	// Run post-storage child workflow, if one is configured.
+	if w.cfg.Batch.PostStorageWorkflow != nil {
+		if err := w.postStorageWorkflow(ctx, *w.cfg.Batch.PostStorageWorkflow, state); err != nil {
+			return err
+		}
+	}
```
jraddaoui: Oh, I see. This is no longer a slice, so we should check it differently.

jraddaoui (changes requested):
Thanks @djjuhasz! Now we need to do "the same" in the processing workflow, I'll create an issue for that.

internal/batch/config.go
djjuhasz: Good catch, "PostStorage" matches the existing preprocessing config too, which is better I think.

djjuhasz: I think it actually needs to be "Poststorage" (no caps for "storage") because the TOML var isn't camel-cased.

internal/workflow/batch.go
djjuhasz: @jraddaoui are you okay with this as-is, or are you requesting a change?

jraddaoui: If we don't change this, it will try to execute a poststorage child workflow even if none is configured, as you can see in the tests.

djjuhasz: Why do you think the poststorage child workflow is being run when no workflow is configured? If `w.cfg.Batch.PostStorageWorkflow` is not set then it should be `nil` and the child workflow should not be run, unless there's some sort of cfg state sharing going on between tests. 🤔

djjuhasz: If I explicitly remove the poststorage config from the "batch" test then the child workflow code is skipped as expected.

jraddaoui: https://github.com/artefactual-sdps/enduro/pull/1480#discussion_r2695158437

How is `s.Equal(1, s.postStorageCallCount)` passing in all test cases then?

jraddaoui: Maybe that's why we usually don't use pointers in the config for structs?

jraddaoui: Right, it's not set back to 0 in https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/batch_test.go#L86-L90. I guess that doesn't matter if you are changing how it's mocked :wink:.

djjuhasz: I don't think the pointer in the `batch.Config` struct is an issue in the tests.  

The reason `s.Equal(1, s.postStorageCallCount)` is passing is all the tests is because `s.postStorageCallCount` is shared state for the whole test suite.  The `TestBatch()` increments the counter to 1, then all the other tests confirm that `s.postStorageCallCount == 1`.  I guess that if the tests ran out of order then the evaluation might happen before `s.postStorageCallCount` is incremented, but I don't know how the `github.com/stretchr/testify/suite` code handles test run order or concurrency.

In any case, I've removed the `s.postStorageCallCount` variable from the tests altogether now and used the `OnWorkflow()` mocking functionality instead.

internal/workflow/batch_test.go
```diff
@@ -66,6 +67,12 @@ func (s *BatchWorkflowTestSuite) SetupWorkflowTest(cfg config.Configuration) {
 		activities.NewPollSIPStatusesActivity(ingestsvc, time.Microsecond).Execute,
 		temporalsdk_activity.RegisterOptions{Name: activities.PollSIPStatusesActivityName},
 	)
+	if cfg.Batch.Poststorage != nil {
+		s.env.RegisterWorkflowWithOptions(
+			postStorageChildWorkflow,
+			temporalsdk_workflow.RegisterOptions{Name: cfg.Batch.Poststorage.WorkflowName},
+		)
+	}
```
jraddaoui, line 75: Not a big deal, but it should be okay to register it in all cases.

jraddaoui (approved):
Thanks @djjuhasz!

internal/workflow/batch_test.go
djjuhasz: If I remove the conditional registration the tests panic if `cfg.Batch.Poststorage` is nil in the `RegisterOptions{}` struct.  I could set the post-storage workflow name to a constant, but I think we're going to have to register the post-storage workflow conditionally like this in the workers, so this is a more accurate reflection of the real code.

Which brings up the problem that I haven't done any registration of the post-storage workflow in the worker code. :thinking:

djjuhasz: Oh, it looks like the post-storage workflow registration will happen in the post-storage worker, so it's not part of the base Enduro code at all.  That's very clever of us. :smile:

PR #1481, opened 2026-01-14
Strip Go binaries by default, with opt‑out for debug builds

jraddaoui:
### Summary
- Added optional stripping (`-s`) to builds to reduce binary size while preserving normal runtime behavior.
- Introduced `STRIP=1` (default) in `hack/build_dist.sh` and Docker builds; set `STRIP=0` to keep full symbols when needed.

### Why
- The project ships three Go binaries, and unstripped builds were ~150MB each due to full symbol/DWARF data.
- Stripping is a low risk optimization for release artifacts; debuggable builds remain available via a toggle.

### Results
- `enduro`: 164MB → 138MB (‑26MB)
- `enduro-a3m-worker`: 148MB → 125MB (‑23MB)
- `enduro-am-worker`: 148MB → 125MB (‑23MB)

### Notes
- Stack traces still include function/file/line info because `pclntab` remains; only DWARF/symbol tables are removed.
- Use `STRIP=0` when you need Delve/core‑dump analysis or richer external profiling.

### DIFF
--- a/hack/build_dist.sh
+++ b/hack/build_dist.sh
@@ -8,6 +8,7 @@
 set -eu
 
 MODULE_PATH="${MODULE_PATH:-github.com/artefactual-sdps/enduro}"
+STRIP="${STRIP:-1}"
 
 IFS=".$IFS" read -r major minor patch < internal/version/VERSION.txt
 version_path=${MODULE_PATH}/internal/version
@@ -36,5 +37,8 @@ EOF
 fi
 
 ldflags="-X ${VERSION_PATH}.Long=${LONG} -X ${VERSION_PATH}.Short=${SHORT} -X ${VERSION_PATH}.GitCommit=${GIT_HASH}"
+if [ "$STRIP" = "1" ]; then
+	ldflags="-s $ldflags"
+fi
 
 exec go build -ldflags "$ldflags" "$@"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -18,6 +18,7 @@ builds:
   flags:
   - -trimpath
   ldflags:
+  - -s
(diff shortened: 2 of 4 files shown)


scollazo (commented):
lgtm!

scollazo (approved):

hack/build_dist.sh
```diff
@@ -36,5 +37,8 @@ EOF
 fi
 
 ldflags="-X ${VERSION_PATH}.Long=${LONG} -X ${VERSION_PATH}.Short=${SHORT} -X ${VERSION_PATH}.GitCommit=${GIT_HASH}"
+if [ "$STRIP" = "1" ]; then
+	ldflags="-s $ldflags"
```
djjuhasz, line 41: https://www.codingexplorations.com/blog/reducing-binary-size-in-go-strip-unnecessary-data-with-ldflags-w-s suggests that the "-w" linking flag is required to strip the DWARF symbol table (whatever that is 😜). Did you try including "-w" as well?

djjuhasz (approved):
LGTM @jraddaoui.  I did a quick search to learn more about the "-s" linking option and several articles mentioned using "-w" as well, so I added a question as to whether you had tried that out.

hack/build_dist.sh
djjuhasz: Here's another post by a Go crypto dev: https://words.filippo.io/shrink-your-go-binaries-with-this-one-weird-trick/

djjuhasz: Oops, nevermind it looks like the "-s" linker flag implicitly sets the "-w" flag as well:
https://pkg.go.dev/cmd/link

Maybe this has changed since those article were written.

PR #1484, opened 2026-01-15
Fix DEPENDENCY_TRACK_URL replacement variable

djjuhasz:
Attempt to fix "The requested URL returned error: 503" error when sending SBOM to dependencytrack in Github CI.


djjuhasz:
The problem is at the server end, not with this workflow.

PR #1485, opened 2026-01-16
Switch to DependencyTrack Github upload action

djjuhasz:
Use the recommended `DependencyTrack/gh-upload-sbom` Github Action instead of a CURL call to upload the Enduro SBOM to DependencyTrack.

### DIFF
--- a/.github/workflows/generate-sbom.yml
+++ b/.github/workflows/generate-sbom.yml
@@ -15,22 +15,17 @@ permissions:
   contents: read
 
 env:
-  PROJECT_NAME:
-    ${{ github.repository_owner }}-${{ github.event.repository.name }}
-  PROJECT_VERSION:
-    ${{ (github.event_name == 'release' && github.event.release.tag_name) || github.ref_name }}
+  PROJECT_NAME: ${{ github.repository_owner }}-${{ github.event.repository.name }}
+  PROJECT_VERSION: ${{ (github.event_name == 'release' && github.event.release.tag_name) || github.ref_name }}
 
 jobs:
   sbom:
     runs-on: ubuntu-latest
-    env:
-      DEPENDENCY_TRACK_URL: ${{ secrets.DEPENDENCY_TRACK_URL }}
-      DEPENDENCY_TRACK_API_KEY: ${{ secrets.DEPENDENCY_TRACK_API_KEY }}
     steps:
       - name: Validate Dependency-Track secrets
         run: |
-          if [ -z "${DEPENDENCY_TRACK_URL}" ] || [ -z "${DEPENDENCY_TRACK_API_KEY}" ]; then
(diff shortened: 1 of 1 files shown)


djjuhasz:
I've set this PR to "Draft" as the "generate-SBOM" action is currently disabled due to an issue with our Dependency Track server.

.github/workflows/generate-sbom.yml
```diff
@@ -15,22 +15,17 @@ permissions:
...
     runs-on: ubuntu-latest
-    env:
-      DEPENDENCY_TRACK_URL: ${{ secrets.DEPENDENCY_TRACK_URL }}
-      DEPENDENCY_TRACK_API_KEY: ${{ secrets.DEPENDENCY_TRACK_API_KEY }}
     steps:
       - name: Validate Dependency-Track secrets
         run: |
-          if [ -z "${DEPENDENCY_TRACK_URL}" ] || [ -z "${DEPENDENCY_TRACK_API_KEY}" ]; then
-            echo "Dependency-Track secrets not configured; set DEPENDENCY_TRACK_URL and DEPENDENCY_TRACK_API_KEY."
+          if [ -z "${DEPENDENCY_TRACK_HOSTNAME}" ] || [ -z "${DEPENDENCY_TRACK_API_KEY}" ]; then
```
sbreker, line 27: This holds a URL - so I think this name change may confuse things.

.github/workflows/generate-sbom.yml
```diff
@@ -56,15 +51,15 @@ jobs:
...
-        run: |
-          curl --fail-with-body --silent --show-error \
-            -X POST "${DEPENDENCY_TRACK_URL%/}/api/v1/bom" \
-            -H "X-Api-Key: ${DEPENDENCY_TRACK_API_KEY}" \
-            -F "autoCreate=true" \
-            -F "projectName=${PROJECT_NAME}" \
-            -F "projectVersion=${PROJECT_VERSION}" \
-            -F "parentName=artefactual-sdps-enduro" \
-            -F "bom=@sbom.cdx.json"
+        uses: DependencyTrack/gh-upload-sbom@v3
```
sbreker, line 57: Good idea to change this to use the action 👍

.github/workflows/generate-sbom.yml
djjuhasz: @sbreker the script expects the hostname to be provided without protocol (http:// or https://), see examples at https://github.com/DependencyTrack/gh-upload-sbom.

I added a new DEPENDENCY_TRACK_HOSTNAME Github variable to avoid parsing the host name from the DEPENDENCY_TRACK_URL.

sbreker: Ah I see - it has changed because of using the action. Ok - I think you can delete the DEPENDENCY_TRACK_URL var then as this project is the only one using it in the artefactual-sdps org.

sbreker (approved):
This looks great - thanks @djjuhasz

PR #1486, opened 2026-01-16
Remove worker conditionals that are always false

djjuhasz:
Remove conditional error checks from worker code that will always evaluate as false.


jraddaoui (approved):
👍

Issue #1394, opened 2025-10-20
Feature: Allow for options other than ssh to copy files to Archivematica transfer source


scollazo:
**Is your feature request related to a problem? Please describe.**

Enduro copies the processed files to the Archivematica transfer source using ssh

This makes configuration/deployment slightly complex, as client and server keys needs to be properly configured.

**Describe the solution you'd like**

The ability to copy files locally between Enduro processing folder, and Archivematica transfer source, when both folders are in the same filesystem.

Also,  Enduro could have the ability to use the same folder for processing than Archivematica uses as transfer source to avoid having to copy files from one place to the other.  A sample implementation of this feature is available in the old enduro, at https://github.com/artefactual-labs/enduro/blob/34bad222abe03dc1bdae9d2bd4c8717a11f391aa/internal/workflow/activities/bundle.go#L63 

**Describe alternatives you've considered**

**Additional context**

This might also speed up a bit the enduro processing, as it's reduce the number of times files needs to be copied around.

PR #1487, opened 2026-01-19
Use ambox in development

jraddaoui:
https://github.com/sevein/ambox/blob/dev/ambox/hack/box/README.md


Makefile
```diff
@@ -216,7 +216,7 @@ test-race:
 
 tilt-am-knownhosts:  # @HELP Update Archivematica Storage Service known_hosts.
 tilt-am-knownhosts: HOST ?= host.k3d.internal
-tilt-am-knownhosts: PORT ?= 12322
+tilt-am-knownhosts: PORT ?= 64022
```
jraddaoui: Do we still need `tilt-am-knownhosts`? Is it going to work like it is right now without exposing this port outside the cluster?

docs/src/dev-manual/archivematica.md
```diff
@@ -1,115 +1,43 @@
...
+## Quick start
 
-In order to configure the connection to the SFTP server and the AM API, you'll
-need to create three `.secret` files inside the `dev-am` overlay folder. These
-files are used by a [Kustomize secret generator] to add the `enduro-am-secret`
-inside the cluster, and they are not tracked in the repository.
+1. Make sure Enduro uses Archivematica:
 
-!!! note
+   `ENDURO_PRES_SYSTEM=am`
```
jraddaoui: I'd note this is in the Tilt env file.

docs/src/dev-manual/archivematica.md
```diff
@@ -1,115 +1,43 @@
...
 
-    The secret generator is used via Tilt so users CANNOT manually apply
-    the Kubernetes commands for these operations. Understanding the way these
-    files are used is NOT required to work with Archivematica.
+2. Ensure the preservation task queue is set to AM:
 
-### Quick checklist for configuration files
+   `[preservation] taskQueue = "am"`
```
jraddaoui: And this in the Enduro config file.

docs/src/dev-manual/devel.md
```diff
@@ -201,9 +201,20 @@ Enables live updates on code changes for the enduro services.
 ### ENDURO_PRES_SYSTEM
 
 Determines the preservation system between Archivematica (`am`) and a3m
-(`a3m`), defaults to `a3m`. Check the [Working with Archivematica] docs if you
+(`a3m`), defaults to `am`. Check the [Working with Archivematica] docs if you
 are planning to use Archivematica as preservation system.
```
jraddaoui: Nitpicking, now it's the default, something like "Check the [Working with Archivematica] docs for more details" may be better.

docs/src/dev-manual/devel.md
```diff
@@ -201,9 +201,20 @@ Enables live updates on code changes for the enduro services.
...
 
+### Using a3m in development
+
+To use a3m instead of Archivematica in the dev environment (a3m is a lightweight
+Archivematica derivative, but it has seen little adoption and is largely
+unmaintained):
+
+1. Set `ENDURO_PRES_SYSTEM=a3m` in `.tilt.env`.
+2. Set `[preservation] taskqueue = "a3m"` in `enduro.toml`.
```
jraddaoui, line 214: Something like this is what I mean for the quick start section.

docs/src/dev-manual/devel.md
```diff
@@ -201,9 +201,20 @@ Enables live updates on code changes for the enduro services.
...
+3. If you want to build a local a3m image, set `LOCAL_A3M=true` and clone the
+   `a3m` repository as a sibling of this repo.
```
jraddaoui: We should note that it is set in `.tilt.env` too.

docs/src/dev-manual/devel.md
```diff
@@ -190,7 +190,7 @@ located in the root of the project. Example:
 
 ```text
 TRIGGER_MODE_AUTO=true
-ENDURO_PRES_SYSTEM=a3m
+ENDURO_PRES_SYSTEM=am
```
jraddaoui: This is an example, I'd keep it as `a3m` to give more sense to the LOCAL_A3M var below.

enduro.toml
```diff
@@ -139,9 +139,9 @@ workflowType = "create aip"
 retentionPeriod = "0"
 
 # Change the taskqueue setting to your prefered preservation system, by default
-# it is a3m. Use 'am' to configure Archivematica instead.
+# it is am. Use 'a3m' to configure a3m instead.
```
jraddaoui: Making sure am is understood as Archivematica, I'd say `by default it is 'am' (Archivematcia).` or something like that.

hack/kube/components/ambox/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  ports:
+    - name: am-dashboard
+      port: 64080
+      targetPort: 64080
+    - name: am-ss
+      port: 64081
+      targetPort: 64081
+    - name: am-sftp
+      port: 64022
+      targetPort: 64022
```
jraddaoui: The manifest organization could be improved a lot, but for now I think it would be better to have all the `ambox` related manifests in a single YAML like we do for other resources and place it directly in the `dev-am` overlay. Also, while doing that, the overlay seems like a better place too for the `enduro-am-secret.yaml` that is currently living under components/dev.

hack/kube/overlays/dev-am/enduro-am-secret.yaml
```diff
@@ -4,9 +4,24 @@ metadata:
   name: enduro-am-secret
 type: Opaque
 stringData:
-  api_user: "artefactual"
-  api_key: ""
-  sftp_host: "localhost"
-  sftp_user: "artefactual"
-  known_hosts: ""
-  id_ed25519: ""
+  user: "test"
```
jraddaoui, line 7: Oh, this was wrong, thanks!

hack/kube/overlays/dev-am/enduro-am.yaml
```diff
@@ -27,6 +27,33 @@ spec:
...
+              # Generate known_hosts with readable perms for the worker user.
+              ssh-keyscan -H -p 64022 ambox.enduro-sdps > /ssh/known_hosts
+              chown 1000:1000 /ssh/known_hosts
+              chmod 0644 /ssh/known_hosts
+          volumeMounts:
+            - name: ssh-secret
+              mountPath: /ssh-secret
+              readOnly: true
+            - name: ssh-volume
+              mountPath: /ssh
```
jraddaoui: How is this going to work? Does it require to restart this resource if `ambox` is restarted? Is there a way to generate fixed values for the know hosts in `ambox`? Maybe in https://github.com/sevein/ambox/blob/dev/ambox/hack/box/etc/sftpgo/sftpgo.json#L82-L84?

jraddaoui (commented):
Thanks @sevein!

This is a huge improvement compared to what we have to set up to work with Archivematica and it will be a great help when we tackle https://github.com/artefactual-sdps/enduro/issues/1394.

May main concern is how we deal with the known hosts situation, would it work between restarts? Also, working against AM dev branch will be okay most of the time but, will it be possible to build `ambox` images for specific AM/AMSS versions?

sevein:
Thanks for the feedback, that makes sense. I'll update the PR soon.

hack/kube/overlays/dev-am/enduro-am.yaml
sevein: > Is there a way to generate fixed values for the know hosts in ambox? Maybe in https://github.com/sevein/ambox/blob/dev/ambox/hack/box/etc/sftpgo/sftpgo.json#L82-L84?

Good idea! I've done this in `ambox:1.0.8`.

jraddaoui (commented):
Awesome, thanks @sevein!! I can't approve because I created the PR, but it looks great to me, merge when you can.

sevein (approved):

Issue #1220, opened 2025-05-13
Problem: it's not possible to delete an AIP without API authentication


jraddaoui:
**Is your feature request related to a problem? Please describe.**

In #1076 we added the possibility to delete an AIP to Enduro. The initial implementation required a two step process to request and review the AIP deletion, where the user that requested the deletion can't be the one approving/rejecting it. We are also adding a way for the requester to cancel their own request, where there will be a similar check making sure it's the same user in this case (re: #1174).

The problem comes when we disable authentication in Enduro's API, as there is no way to identify the user for this process. The initial implementation just made authentication a requirement for the two endpoints involved:

- [/aips/{uuid}/deletion-request](https://github.com/artefactual-sdps/enduro/blob/main/internal/storage/deletion_request.go#L15-L19)
- [/aips/{uuid}/deletion-review](https://github.com/artefactual-sdps/enduro/blob/main/internal/storage/deletion_request.go#L63-L67)

But, what would you do without authentication? Does it makes sense to have a review process in that case? Should the deletion happen automatically? What about cancellation if there is no review and the deletion happens automatically?

**Describe the solution you'd like**

---

**Describe alternatives you've considered**

---

**Additional context**

---

jraddaoui:
I'm working on https://github.com/artefactual-sdps/enduro/issues/1483 and, when the batch workflow is canceled I want to be able to use the existing functionality to delete the already stored AIPs from the ingested SIPs. Our services coordinate via the public API (ingest workflow -> storage API) and for that integration we run a second API without authentication (see: https://github.com/artefactual-sdps/enduro/issues/995). That meant deletion requests could not be sent at all, so I decided to tackle this issue in https://github.com/artefactual-sdps/enduro/pull/1491.

## Functional changes

### New config

- `storage.aipDeletion.allowUnauthenticated` (default: `true`):
  - `true`: deletion request/review/cancel proceed without auth.
  - `false`: all unauthenticated deletion operations are rejected.

### API behavior

- AIP deletion endpoints now allow unauthenticated requests when `allowUnauthenticated = true`.
- When auth is disabled and `allowUnauthenticated = false`, deletion endpoints return a 400 response with the "unauthenticated AIP deletion is disabled" message.
- Unauthenticated requests are attributed to an "unknown" user identity for reporting fields.

### Dashboard behavior

- When auth is disabled, Approve/Reject buttons are visible even if Cancel is available.
[...]

PR #1491, opened 2026-01-21
Allow unauthenticated AIP deletion

jraddaoui:
Allow AIP deletion request/review/cancel to proceed without API authentication when allowed in the configuration, use an "unknown" actor identity for the persistence layer and reports in those cases.

Refs #1220.


jraddaoui:
See https://github.com/artefactual-sdps/enduro/issues/1220#issuecomment-3779437850 for more details.

Issue #1499, opened 2026-01-29
Problem: Upload from SIP source shows an error if there are no SIPs in the bucket


jraddaoui:
**Describe the bug**

The upload from SIP source page shows an error alert if there are no SIPs in the bucket. The `ingest/sip-sources/e6ddb29a-66d1-480e-82eb-fcfef1c825c5/objects` API endpoint seems to be returning a 500 (Internal Server Error) response.

**To Reproduce**

- Configure a SIP source bucket without any object
- Visit the upload from source page
- See the error alert (screenshot below)

**Expected behavior**

The API doesn't return a internal error and the error alert is not shown.

**Screenshots**

<img width="1043" height="658" alt="Image" src="https://github.com/user-attachments/assets/d95e53a1-8ba7-40a9-a2cf-8b0740481063" />

PR #1500, opened 2026-01-29
Fix nil page handling for empty SIP sources

jraddaoui:
Refs #1499.


djjuhasz (approved):
LGTM! Thanks @jraddaoui. 👍

Issue #1478, opened 2026-01-12
Enhancement: Add the option to run a post-preservation child workflow for batches


djjuhasz:
**Is your feature request related to a problem? Please describe.**

It may be helpful or necessary for some institutions to take some action at the end of a batch preservation workflow, such as creating a consolidated report of the batch outcomes.  Creating a batch level report now would require forking and customizing the Enduro code, which makes it difficult to manage future updates to the base project.

**Describe the solution you'd like**

Add a call point to the batch workflow that can call one or more "post-preservation" child workflows based on application configuration, as we've done with with "post-storage" child workflow in the SIP ingest workflow (processing).

**Describe alternatives you've considered**

We considered several other alternatives to running a child workflow before developing the initial "pre-processing" child workflow, but the child workflow solution has worked well and there is no compelling reason to consider alternate approaches for this case.

PR #1501, opened 2026-01-29
Standardize child workflows

djjuhasz:
See the individual commit messages for a detailed summary of the changes.


internal/childwf/config.go
```diff
@@ -0,0 +1,59 @@
...
+	WorkflowName string
+}
+
+type Configs []Config
+
+func (c Configs) GetByName(name string) *Config {
+	for _, cw := range c {
+		if cw.WorkflowName == name {
+			return &cw
+		}
```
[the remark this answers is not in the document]
djjuhasz: Fixed.

docs/src/admin-manual/configuration.md
```diff
@@ -1153,9 +1153,14 @@ underlying general Enduro code. For more information, see:
...
 [preprocessing]
-enabled = false
 extract = false
 sharedPath = "/home/enduro/preprocessing"
+
+[[childworkflows]]
+enabled = false
+namespace = "default"
+taskQueue = "preprocessing"
+workflowName = "preprocessing"
```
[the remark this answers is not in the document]
djjuhasz: Updated.

internal/config/config.go
```diff
@@ -63,11 +62,10 @@ type Configuration struct {
 	InternalAPI     api.Config
 	API             api.Config
 	BagIt           bagcreate.Config
-	Batch           batch.Config
+	Childworkflows  childwf.Configs
```
jraddaoui: PascalCase, `ChildWorkflows`.

internal/config/config.go
```diff
@@ -87,7 +85,6 @@ func (c *Configuration) Validate() error {
 		c.A3m.Validate(),
 		c.API.Auth.Validate(),
 		c.BagIt.Validate(),
-		c.Preprocessing.Validate(),
```
jraddaoui, line 90: Missing call to `c.ChildWorkflows.Validate()`.

internal/poststorage/poststorage.go
```diff
@@ -1,11 +1,5 @@
...
 
-type Config struct {
-	Namespace    string
-	TaskQueue    string
-	WorkflowName string
-}
-
 type WorkflowParams struct {
 	AIPUUID string
 }
```
jraddaoui: Similar to the batch postprocessing params, it would be nice to remove the `preprocessing` and `poststorage` packages and move everything to the `childwf` package.

internal/preprocessing/preprocessing.go
```diff
@@ -43,19 +24,3 @@ type WorkflowResult struct {
...
-	if c.SharedPath == "" {
-		return errors.New("sharedPath is required in the [preprocessing] configuration")
-	}
-	if c.Temporal.Namespace == "" || c.Temporal.TaskQueue == "" || c.Temporal.WorkflowName == "" {
-		return errors.New(
-			"namespace, taskQueue and workflowName are required in the [preprocessing.temporal] configuration",
-		)
-	}
-	return nil
-}
```
jraddaoui, line 61: If you configure a preprocessing child workflow the shared path is required, I'd love to not loose that validation. Now that they are separated configs, maybe move that check to the full config validation.

internal/workflow/batch.go
```diff
@@ -315,9 +313,13 @@ func (w *BatchWorkflow) waitForWorkflowsCompletion(ctx temporalsdk_workflow.Cont
 
 func (w *BatchWorkflow) postStorageWorkflow(
 	ctx temporalsdk_workflow.Context,
-	cfg batch.PostStorageConfig,
 	state *batchWorkflowState,
 ) error {
+	cfg := w.cfg.Childworkflows.GetByName(childwf.BatchPostStorageName)
+	if cfg == nil || !cfg.Enabled {
+		return nil
+	}
```
jraddaoui: This will get the first batch poststorage config and if it's not enabled it won't look for other possible enabled configs. To be clear, I don't think it should fallback to another config, I think we should not allow multiple configs for the same role, and check that on validation.

internal/workflow/download.go
```diff
@@ -36,7 +36,7 @@ func (w *ProcessingWorkflow) downloadSIP(sessCtx temporalsdk_workflow.Context, s
 	}
 
 	var destinationPath string
-	if w.cfg.Preprocessing.Enabled {
+	if w.cfg.Childworkflows.IsEnabled("preprocessing") {
 		destinationPath = w.cfg.Preprocessing.SharedPath
 	}
```
jraddaoui: Now this could "fail", giving the split in configs and the missing validation.

internal/workflow/processing.go
```diff
@@ -367,7 +367,7 @@ func (w *ProcessingWorkflow) SessionHandler(
 	}
 
 	// Unarchive the transfer if it's not a directory and it's not part of the preprocessing child workflow.
-	if !state.sip.isDir && (!w.cfg.Preprocessing.Enabled || !w.cfg.Preprocessing.Extract) {
+	if !state.sip.isDir && (!w.cfg.Childworkflows.IsEnabled("preprocessing") || !w.cfg.Preprocessing.Extract) {
```
jraddaoui: Also, "preprocessing" should be a constant or part of an enum.

internal/workflow/processing.go
```diff
@@ -1078,29 +1079,27 @@ func (w *ProcessingWorkflow) preprocessing(ctx temporalsdk_workflow.Context, sta
...
+			WorkflowID:        fmt.Sprintf("%s-%s", cfg.WorkflowName, aipUUID),
+			ParentClosePolicy: temporalapi_enums.PARENT_CLOSE_POLICY_ABANDON,
+		},
+	)
+	err := temporalsdk_workflow.ExecuteChildWorkflow(
+		ctx,
+		cfg.WorkflowName,
+		poststorage.WorkflowParams{AIPUUID: aipUUID},
+	).GetChildWorkflowExecution().Get(ctx, nil)
+
```
jraddaoui: As we said, we only want one workflow per role, so I think this is good. However, there is no limitation to configure multiple and I think the documentation doesn't address that possibility. We should prevent the configuration of multiple workflows per role.

internal/workflow/processing_setup_test.go
```diff
@@ -162,13 +170,12 @@ func (s *ProcessingWorkflowTestSuite) SetupWorkflowTest(cfg config.Configuration
...
+		temporalsdk_workflow.RegisterOptions{Name: "poststorage"},
 	)
 	s.env.RegisterWorkflowWithOptions(
-		poststorageChildWorkflow,
-		temporalsdk_workflow.RegisterOptions{Name: "poststorage_2"},
+		batchPostStorageChildWorkflow,
+		temporalsdk_workflow.RegisterOptions{Name: childwf.BatchPostStorageName},
 	)
-
 	s.workflow = NewProcessingWorkflow(cfg, rng, ingestsvc, wsvc)
```
jraddaoui: Why registering a batch poststorage workflow in the processing workflow tests? Removing the second workflow registration should be enough.

Tiltfile
```diff
@@ -67,7 +67,14 @@ if PRES_SYS == 'am':
...
 
-# Preprocessing
+# Child workflows
+CHILD_WORKFLOW_PATHS = os.environ.get("CHILD_WORKFLOW_PATHS", "")
+if CHILD_WORKFLOW_PATHS != "":
+  for path in CHILD_WORKFLOW_PATHS.split(":"):
+    # Load child workflow Tiltfile for Enduro
+    load_dynamic(path + "/Tiltfile.enduro")
+
+# The preprocessing child workflow requires extra setup for a shared directory
```
jraddaoui: This is missing documentation. Also, if both are set (`PREPROCESSING_PATH` and `CHILD_WORKFLOW_PATHS`) we may be calling `load_dynamic(path + "/Tiltfile.enduro")` twice. I'd use a different variable for the preprocessing volume, maybe a boolean fits better now, `MOUNT_PREPROCESSING_VOLUME` or something like that.

internal/preprocessing/preprocessing.go
jraddaoui: Having conflicts about this, what do you think about including both fields in the general child workflow config? The `sharedPath` may become useful for poststorage workflows too, and we could document that `extract` is only for the `preprocessing` role.

jraddaoui (changes requested):
Thanks @djjuhasz, this is looking good! We should probably discuss this type of change a bit earlier next time. There are a few things I would have done differently, but at this point it may be heavy to rework everything.

---

## Child workflows config

With this changes we’re effectively forcing workflow names (`preprocessing`, `poststorage`, `batch-poststorage`) because lookup uses `WorkflowName`. I think a cleaner approach would be to separate role from workflow name and treat missing config as disabled. Allowing `workflowName` to differ from `role` gives more descriptive names and avoids forcing renames in existing extensions.

**Option A: keep the list, add role:**

```toml
[[childworkflows]]
role = "preprocessing"
namespace = "default"
taskQueue = "preprocessing"
workflowName = "my-preprocess-wf"
```

**Option B: role keyed tables:**

```toml
[childworkflows.preprocessing]
namespace = "default"
taskQueue = "preprocessing"
workflowName = "my-preprocess-wf"
```

In both options, absence could mean disable, but I'm okay if you want to keep the enabled flag. Option B also prevents configuring multiple workflows per role without needing extra validation, but it means changing the config type to a map (e.g., `map[string]Config` or `map[RoleEnum]Config`) or a wrapper struct keyed by role. `GetByName` could become `GetByRole` and we could even default `workflowName` to `role` if unset, but I think it's okay to require all fields.

Additionally, we said we want only one workflow per role, but the current list allows multiple entries per role/name, and `GetByName/GetByRole` would just return the first match. We should either validate uniqueness or use a role keyed structure to prevent duplicates. It would also help to validate roles against an enum or constants so typos don’t silently disable workflows.

Finally, I think it would be nice for the About API to return both `role` and `workflowName`, so the UI can show something like `workflowName (role)`.

---

## Batch poststorage params

I think we can avoid the new `BPS*` types and conversion helpers. Temporal uses the default JSON converter, and unknown fields are ignored when decoding into a smaller struct. So we can keep sending `datatypes.Batch` and `datatypes.SIP` and let child workflows define slim types with just the fields they need. If we add JSON tags with `omitempty` to `datatypes.SIP` and `datatypes.Batch`, we can also prevent empty fields from being sent and reduce payload noise, but it shouldn't hurt to send a few extra fields like the `status`.

djjuhasz:
@jraddaoui thanks for the code review and your feedback. I'd like to respond to a few of your suggestions below.

Firstly, sorry to surprise you with such a big change.  The change grew organically out of trying to avoid adding a "batch poststorage" section to the "about" component of the Dashboard, and it got complex pretty quickly.  As I said in the code comments and in our stand-up, I did actually have more changes originally and I dialed them back to *reduce* the complexity. :sweat_smile: 

### Child workflows

I'm okay with adding a role field as you've suggested, but I do want to mention that I originally had a "name" field with a similar purpose which I eliminated because it was just repeating the WorkflowName for our three existing child workflows ("preprocessing", "poststorage", and "batch-poststorage"). At this point I can't imagine a use-case for multiple workflows with the same role. For example, why would we need to add multiple workflows with a role of preprocessing?  As we discussed when I added the "batch-poststorage" child workflow, adding multiple child workflows with the same role raises questions of whether they should be run serially or in parallel and how to a critical error in one child workflow should impact other workflows with the same role.  But maybe you don't mean to have multiple workflows with the same "role"?

I'm not sure what you mean by "With this changes we’re effectively forcing workflow names".  The workflow name field was already required for the preprocessing and poststorage configs, so I'm not sure how this change would force anything we didn't require already. I'm also not clear what "more descriptive" names we would want to give to the child workflows.  If we want something more human readable for the "about" component (e.g. "batch poststorage") maybe we should call the field "description" or "name" rather than "role"?  I'd love it if you could provide an example of what you'd like to see in the "about" component for one of our existing child workflows, so I can see what you are envisioning.

I like the idea of role keyed tables for the child workflow config (Option B).  I didn't realize you could assign a dot separated section header like "[childworkflows.preprocessing]" to a map key. :open_mouth:  As you've pointed out the automatic enforcement of a unique role key is a nice feature, and it may also speed up lookups slightly. 

I'm totally on board with removing the "enabled" field and having the existence of a child workflow configuration be sufficient and necessary for enabling the child workflow.  I only added the "enabled" field because it was present in the preprocessing config and I wanted consistency with all of the child workflow configs.

### Batch Poststorage params

My main reason for introducing new SIP and Batch types for the BPS params is to avoid sending the User and Batch data with each SIP.  Including the User and Batch more than doubles the payload of each SIP and is extremely repetitive.  I don't really expect many institutions will ingest a huge number of SIPs in a batch (more than a few dozen seems unlikely) but I do think we still have to be conscious of Temporal's payload limits  when sending data between workflows and activities.  

Adding a "json:omitempty" annotation to the fields might reduce the payload a bit, but I think it mostly would just prevent sending the field name and an empty value? This won't really help with removing the Batch and User data either, as they should be populated for all SIPs in a batch I think.

I'm open to alternatives to adding BPS specific structs here. Maybe I could just add `datatypes.SimpleSIP` and `datatypes.SimpleBatch` structs that omit the related User and Batch data?  I'm open to different ideas or struct names.

internal/poststorage/poststorage.go
djjuhasz: Done.

jraddaoui:
Thanks @djjuhasz!

- **Roles:** yes, I do not mean multiple workflows per role, let's keep the one workflow per role solution we discussed. Option B enforces that nicely as you noted.
- **Forcing names:** I meant that lookup by `WorkflowName` requires the configured workflow name to be "preprocessing", etc. We already have a deployed child workflow with name "ais", so tying lookup to the name would break it unless we rename. The about dialog could show `ais (poststorage)`; I agree a human readable description field could be a good addition in the future.
- **BPS params:** fair point about `User` and `Batch` fields, I didn't think about them, so I'm okay keeping the new types. An alternative could be setting `User/Batch` to nil before sending so we don't proliferate types, but either approach is fine with me.

internal/childwf/batch_poststorage.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"github.com/google/uuid"
+
+	"github.com/artefactual-sdps/enduro/internal/datatypes"
+)
+
+const (
+	BatchPostStorageKey  = "batch_poststorage"
+	BatchPostStorageName = "batch-poststorage"
```
jraddaoui: Why keeping both constants? The name will be configurable, right? For the other constant, I'd prefer role, id, name or key, in that order of preference.

Maybe this is something we should bring to the team, thinking than in most cases it could look like `preprocessing (preprocessing)` in the dashboard; so I don't know, even if I'd prefer to not change the existing child workflows names, we may decide that is better to not make their names configurable. There is only a wild edge case that can think of that will require to make them configurable to avoid collisions, which could be solved deploying multiple child workflows workers.

djjuhasz: As to why I added the "BatchPostStorageName", it's used here: https://github.com/artefactual-sdps/enduro/blob/dev/standardize-child-workflows/internal/childwf/config.go#L40.  In that config section we don't have the "BatchPostStorageKey" available, and we need something to compare to `config.WorkflowName` to check if the `sharedPath` is required.  We could use a string constant "preprocessing", but maybe it's actually an argument for adding the "key" to struct instead of making it a map key.

djjuhasz: I've switched to "id" for the workflow config unique identifier and moved it into the `childwf.Config` struct again.  This makes the separate "BatchPostStorageName" constant (and similar) unnecessary, so I've removed it and the equivalent name constants for other child workflows.

I've also removed the configuration ID from the "About" dashboard component, so it just shows the Workflow Name now.

image

(The workflow names in the screenshot are for illustration purposes only)

internal/preprocessing/preprocessing.go
djjuhasz: I've added both the "extract" and "sharedPaths" fields to the `childwf.Config` struct as you suggested.  I tried to think of a way to use generics or an interface to allow variations in the config for different types of workflow but I couldn't think of a way to integrate that with the Viper config parsing.

internal/workflow/batch.go
djjuhasz: I've added validation to ensure child workflow configs have a unique ID.

internal/workflow/download.go
djjuhasz: This has been fixed by combining the preprocessing config with the general `childwf.Config` and validating `SharedPath` is set for a preprocessing workflow.

internal/workflow/processing.go
djjuhasz: I've added validation to ensure each child workflow config has a unique ID, and I've updated the docs.

internal/childwf/config.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	if len(missing) > 0 {
+		return fmt.Errorf("missing required value(s): %s", strings.Join(missing, ", "))
+	}
+
+	return nil
+}
+
+type Configs []Config
```
jraddaoui, line 71: If you liked the config layout including the ID in the section:

```toml
[childWorkflows.preprocessing]
namespace = "default"
taskQueue = "preprocessing"
workflowName = "preprocessing"

[childWorkflows.poststorage]
namespace = "default"
taskQueue = "poststorage"
workflowName = "ais"

[childWorkflows.batchPoststorage]
namespace = "default"
taskQueue = "batch-poststorage"
workflowName = "batch-poststorage"
```

I think you said there was some issues using the map, but the following should work, right?

```go
type Configs struct {
    Preprocessing    *Config
    Poststorage      *Config
    BatchPoststorage *Config
}
```

This will enforce the IDs and the uniqueness without validation; we wouldn't need the `ByID` method nor the constants, and the validation of each individual config could accept a boolean to check for the `SharedPath` value:

```go
func (c Config) Validate(requiresSharedPath bool) error {
}

func (c Configs) Validate() error {
    return errors.Join(
        c.Preprocessing.Validate(true),
        c.Poststorage.Validate(false),
        c.BatchPoststorage.Validate(false),
    )
}
```

Sorry I didn't think about the implications of each option sooner; I think this is better to enforce the configuration but I' m okay if you want to keep using the ID fields; you already changed this a few times.

internal/childwf/config.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Extract the SIP in the child workflow (Preprocessing only).
+	Extract bool
+}
+
+func (c Config) Validate() error {
+	missing := make([]string, 0)
+	if c.ID == "" {
+		missing = append(missing, "ID")
+	}
```
jraddaoui: If we keep this approach, we should validate the `c.ID` matches one of the ID constants.

internal/childwf/config.go
jraddaoui:
```go
func (c Configs) Validate() error {
    return errors.Join(
        c.Preprocessing.Validate(true),
        c.Poststorage.Validate(false),
        c.BatchPoststorage.Validate(false),
    )
}
```

Just writing here as I provide the feedback, that method should check for nil on each individual config :wink:

djjuhasz: I think your proposed `childwf.Config` struct that contains explicit child workflow configs has the same problem as using a map - if you only have the `childwf.Config` struct without extra context (the map key the Configs field name) there's no way to tell which "type" of config you are dealing with unless you use the workflow name.  Using explicit config fields also has the disadvantage that you can't loop over the child configs without using some sort of reflection, which is one of the reasons I wanted to move to a general "child workflow" config paradigm in the first place.  For instance in the "About" endpoint I'd have to go back to explicitly calling each of the child workflow configs instead of just being able to loop through `childwf.Configs`.  I think this will mean a bunch of extra work if we add a new child workflow because we have to explicitly define the config in multiple places throughout the code.

internal/childwf/config.go
djjuhasz: I think if we add value checking it makes sense to define the child workflow types as an enum, so we get a "central" registry of child workflow types and the generated code to enforce the values match the allowed values.

internal/childwf/config.go
jraddaoui: I think the clarity in the TOML configuration, the config validation and how to access each config in workflows is better with the struct fields. I see your point about not being able to loop and I like the API schema and the dashboard changes; but I'd prefer not looping in the about endpoint, and doing something similar to the validation example above, accessing each field to build the response array. This would be more explicit and remove the need of an enum/constants; which is something we'd also need to maintain if we add new child workflows.

internal/childwf/config.go
jraddaoui: Yeah, in my initial feedback I was envisioning this as an enum; if you keep this approach I think that makes sense.

internal/childwf/config.go
djjuhasz: In that case, I think it makes more sense to just throw out this whole PR and add the Batch poststorage as it's own config type, like the "poststorage" config was.  I think it's a mistake though because it requires a lot of work and duplication to add every new child workflow, which is what I was trying to avoid.

jraddaoui: I don't see it like that, we'll get the new unified config, the better config schema, the better API schema, the childwf package, among other things. As I said, I' m okay if you want to keep using the ID fields; you already changed this a few times. We should have discussed this a lot sooner.

Tiltfile
djjuhasz: I've added a MOUNT_PREPROCESSING_VOLUME boolean flag and updated the Tilt env docs.

internal/childwf/config.go
djjuhasz: I've stuck with having the workflow id/role/type in the Config struct.  I named the enum "ChildWorkflowType" because "ID" didn't seem right for a enumerated value.  I prefer "type" to "role" but I can switch to "role" if you feel strongly about it.

djjuhasz:
@jraddaoui this is ready for another review.

docs/src/admin-manual/configuration.md
```diff
@@ -1144,53 +1144,35 @@ The following two sections are for configuring a [child workflow] as part of the
...
-  preprocessing child workflow. This path is required when `enabled` is set to
-  true.
-
-#### Temporal workflow configuration for preprocessing workflow
+!!! note
 
-These settings ensure that Enduro can connect with [Temporal] and trigger the
-configured [child workflow].
+    The `childWorkflows` configuration block can be repeated more than once to
+    add multiple child workflows. This is why the `childWorkflows` header is in double brackets rather than single brackets like other configuration blocks.
```
jraddaoui: Quite a long line here.

docs/src/admin-manual/configuration.md
```diff
@@ -1199,12 +1181,24 @@ workflowName = "preprocessing"
...
+  child workflow is run. In some cases, you may wish to design custom child
+  workflow activities that perform operations on the SIP before it is extracted:
+  for example, calculating a checksum of the zipped package to check for prior
+  duplicate ingests before proceeding. When this value is set to `true` Enduro
+  will skip the extraction task at the beginning of the ingest workflow,
+  allowing you to define when and how extraction occurs in the child workflow.
+* `sharedPath`: The absolute path to the directory that Enduro should use to
+  share the SIP between the primary ingest workflow and the configured
+  preprocessing child workflow. This path is required when `enabled` is set to
+  true.
```
jraddaoui: There is no `enabled` anymore.

docs/src/admin-manual/configuration.md
```diff
@@ -1214,29 +1208,31 @@ archival management system), AIP encryption or replication, and more.
...
+[[childWorkflows]] section to enable the post-storage workflow.
 
 !!! note
 
-    The post-storage configuration block can be repeated more than once to add
-    multiple post-storage child workflows. This is why the `poststorage` header
-    is in double brackets rather than single brackets like other configuration
-    blocks.
+    The `childWorkflows` configuration block can be repeated more than once to
+    add multiple child workflows. This is why the `childWorkflows` header is in double brackets rather than single brackets like other configuration blocks.
```
jraddaoui: Same long line.

docs/src/admin-manual/configuration.md
```diff
@@ -1248,6 +1244,44 @@ section as you configure the values.
...
+
+A post-batch child workflow runs after multiple SIPs are ingested and stored
+using the batch ingest functionality of Enduro. The post-batch workflow can be
+used to report the results of a batch operation, for instance to create a batch
+ingest report that collates SIP and AIP data for all of the SIPs in a batch.
+
+!!! note
+
+    The `childWorkflows` configuration block can be repeated more than once to
+    add multiple child workflows. This is why the `childWorkflows` header is in double brackets rather than single brackets like other configuration blocks.
```
jraddaoui: Same.

docs/src/admin-manual/configuration.md
```diff
@@ -1248,6 +1244,44 @@ section as you configure the values.
...
+
+**Default values**:
```
jraddaoui: There are not default value anymore, right? Maybe "Example configuration".

docs/src/admin-manual/configuration.md
```diff
@@ -1248,6 +1244,44 @@ section as you configure the values.
...
+  at once, so they don't clash.
+* `taskQueue`: In Temporal, a [Task Queue] is a lightweight, dynamically
+  allocated queue that one or more workers can poll for tasks. Temporal supports
+  many kinds of parallelization for a distributed application architecture, and
+  and task queues can be namespaced to support such a set-up. The default value
+  of `postbatch` here differentiates it from the general `global` task queue
+  used by Enduro and other child workflows and ensures that Enduro looks for
+  post-batch child workflow tasks in a dedicated queue.
+* `workflowName`: The name of the configured [child workflow] in Temporal.
+
```
jraddaoui: One of the benefits of these changes is the unified configuration for all child workflows. I think that could translate into this documentation by having a single "Child workflows" section where the available types and the specific preprocessing options are explained, but everything else should be the same and it will avoid a lot of repetition.

docs/src/dev-manual/devel.md
```diff
@@ -210,13 +212,19 @@ but it has seen little adoption and is largely unmaintained. Check the
...
 
-Relative path to a preprocessing child workflow repository. It loads a Tiltfile
-called `Tiltfile.enduro` from that repository and mounts a presistent volume
-claim (PVC) in the preservation system pod. That PVC must be defined in the
-preprocessing and be called `preprocessing-pvc`. Check the [Preprocessing child
-workflow] docs to configure the child workflow execution.
+A colon (:) separated list of relative paths to child workflow repositories. At
+startup Tilt will attempt to load a `Tiltfile.enduro` file from each path which
+will add any workflow specific resources to the Tilt environment (e.g. a child
+worker). See the [Administrator configuration] and [Preprocessing child workflow]
```
jraddaoui: This now only points to the `[Preprocessing child workflow]`, another point to unify them into a single section.

docs/src/dev-manual/devel.md
```diff
@@ -192,6 +192,8 @@ located in the root of the project. Example:
 TRIGGER_MODE_AUTO=true
 ENDURO_PRES_SYSTEM=a3m
 LOCAL_A3M=true
+CHILD_WORKFLOW_PATHS='../preprocessing-acme:../poststorage-acme'
```
jraddaoui: Nitpick: Maybe include `../acme-enduro-workflows`, following the recent repository name.

docs/src/dev-manual/devel.md
```diff
@@ -210,13 +212,19 @@ but it has seen little adoption and is largely unmaintained. Check the
...
+startup Tilt will attempt to load a `Tiltfile.enduro` file from each path which
+will add any workflow specific resources to the Tilt environment (e.g. a child
+worker). See the [Administrator configuration] and [Preprocessing child workflow]
+docs for instructions on configuring the child workflows.
+
+### MOUNT_PREPROCESSING_VOLUME
+
+If MOUNT_PREPROCESSING_VOLUME is truthy (t, true, y, yes, 1) Tilt mounts a
+persistent volume claim (PVC) in the preservation system pod. The PVC must be
+defined in the preprocessing pod and be called `preprocessing-pvc`.
```
jraddaoui: Kubernetes manifests or just manifests, instead of pod.

docs/src/dev-manual/preprocessing.md
```diff
@@ -12,23 +12,21 @@ Check the [Tilt environment configuration].
...
-# extract determines if the SIP extraction happens on the child workflow.
-extract = false
-# sharedPath is the full path to the directory used to share the SIP between workflows,
-# required when enabled is set to true.
-sharedPath = "/home/enduro/preprocessing"
-
-# Temporal configuration to trigger the preprocessing child workflow, all fields are
-# required when enabled is set to true.
-[preprocessing.temporal]
+# Temporal configuration to enable the preprocessing child workflow, all fields
```
jraddaoui: Not only "Temporal configuration...", maybe "Optional configuration..."

docs/src/dev-manual/preprocessing.md
```diff
@@ -12,23 +12,21 @@ Check the [Tilt environment configuration].
...
 taskQueue = "preprocessing"
 workflowName = "preprocessing"
+
+# extract determines where SIP extraction, from a zip or other archive format, happens. When set to "true" extraction happens in the preprocessing child
+# workflow. When set to "false" (the default) extraction happens in the parent
+# ingest workflow.
+extract = true
+
+# sharedPath is the full path to a directory to share files between workflows.
+sharedPath = "/home/enduro/preprocessing"
```
jraddaoui: I think it would be okay to fully remove this documentation file in favor of the general "Child workflows" section.

internal/childwf/postbatch.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func SIPstoBPSSIPs(sips []datatypes.SIP) []*BPSSIP {
+	r := make([]*BPSSIP, len(sips))
+	for i, sip := range sips {
+		s := &BPSSIP{
+			UUID: sip.UUID,
+			Name: sip.Name,
+		}
+		if sip.AIPID.Valid {
```
jraddaoui: I would be nice if we could just change the datatypes.SIP AIPID to `*uuid.UUID`. Doesn't need to happen in this PR.

internal/childwf/batch_poststorage.go
```diff
@@ -0,0 +1,49 @@
...
+
+	return r
+}
+
+func BatchtoBPSBatch(batch datatypes.Batch) *BPSBatch {
+	return &BPSBatch{
+		UUID:      batch.UUID,
+		SIPSCount: batch.SIPSCount,
+	}
+}
```
jraddaoui: This transformation could be placed in the batch workflow state, which contains the SIPs and the Batch, and it could be a single method returning the full BPSParams. Not a big deal though.

internal/childwf/config.go
```diff
@@ -0,0 +1,109 @@
...
+	SharedPath string
+
+	// Extract the SIP in the child workflow (Preprocessing only).
+	Extract bool
+}
+
+func (c Config) Validate() error {
+	errs := c.missingFields()
+
+	if c.Type != "" && !c.Type.IsValid() {
```
jraddaoui, line 36: Doesn't it fail directly if you use an invalid type in the TOML?

internal/childwf/config.go
jraddaoui: I like type!

internal/config/config_test.go
```diff
@@ -27,6 +29,13 @@ defaultPermanentLocationId = "f2cc963f-c14d-4eaa-b950-bd207189a1f1"
 
 [api.auth.oidc.abac]
 rolesMapping = '{"admin": ["*"], "operator": ["ingest:sips:list", "ingest:sips:workflows:list", "ingest:sips:read", "ingest:sips:upload"], "readonly": ["ingest:sips:list", "ingest:sips:workflows:list", "ingest:sips:read"]}'
+
+[[childWorkflows]]
+type = "preprocessing"
+namespace = "default"
+taskQueue = "preprocessing"
+workflowName = "preprocessing"
+sharedPath = "/home/enduro/shared"
```
jraddaoui: The test is not checking this addition.

internal/config/config_test.go
```diff
@@ -173,4 +182,22 @@ func TestConfigValidate(t *testing.T) {
...
+				},
+			},
+		}
+		err := c.Validate()
+		assert.ErrorContains(
+			t,
+			err,
+			`child workflow[0]: missing required value(s): namespace, taskQueue, workflowName, sharedPath`,
+		)
+	})
```
jraddaoui: As long as the other test works (the test from the comment above), I think this is covered by the `childwf` package tests.

internal/workflow/batch.go
```diff
@@ -334,20 +336,23 @@ func (w *BatchWorkflow) postStorageWorkflow(
...
-	var res batch.PostStorageResult
+	var res childwf.Result
 	err := temporalsdk_workflow.ExecuteChildWorkflow(
 		childCtx,
 		cfg.WorkflowName,
-		&batch.PostStorageParams{SIPs: state.SIPs()},
+		&childwf.BPSParams{
+			Batch: childwf.BatchtoBPSBatch(state.batch),
+			SIPs:  childwf.SIPstoBPSSIPs(state.SIPs()),
+		},
```
jraddaoui: Based on previous comments, this could just be `state.BPSParams()`.

internal/workflow/batch.go
```diff
@@ -334,20 +336,23 @@ func (w *BatchWorkflow) postStorageWorkflow(
...
 	).Get(childCtx, &res)
 	if err != nil {
-		return fmt.Errorf("batch %q post-storage workflow: %v", state.batch.UUID.String(), err)
+		return fmt.Errorf("batch %q poststorage workflow: %v", state.batch.UUID.String(), err)
```
jraddaoui: Should this now be "...postbatch workflow"?

internal/workflow/batch.go
```diff
@@ -334,20 +336,23 @@ func (w *BatchWorkflow) postStorageWorkflow(
...
 	}
 
 	state.logger.Info(
-		"Post-storage workflow completed",
+		"Poststorage workflow completed",
```
jraddaoui: Similar.

Tiltfile
```diff
@@ -67,11 +67,19 @@ if PRES_SYS == 'am':
...
-# Preprocessing
-PREPROCESSING_PATH = os.environ.get("PREPROCESSING_PATH", "")
-if PREPROCESSING_PATH != "":
-  # Load preprocessing Tiltfile for Enduro
-  load_dynamic(PREPROCESSING_PATH + "/Tiltfile.enduro")
+# The CHILD_WORKFLOW_PATHS environment variable is a colon-separated list of 
+# paths to child workflow directories. If set, we load each child workflow's
+# Tiltfile.enduro to load resources required by the workflow (e.g. a Temporal 
+# worker).
+CHILD_WORKFLOW_PATHS = os.environ.get("CHILD_WORKFLOW_PATHS", "")
```
jraddaoui, line 74: I'd trim here.

Tiltfile
```diff
@@ -67,11 +67,19 @@ if PRES_SYS == 'am':
...
+if CHILD_WORKFLOW_PATHS != "":
+  for path in CHILD_WORKFLOW_PATHS.split(":"):
```
jraddaoui, line 76: Maybe in here too, not a big deal of course.

jraddaoui (commented):
Thanks @djjuhasz, looking great! There is a bunch of comments below, but nothing major. As we discussed, please wait to finally merge this PR in case we have to do a quick release.

docs/src/dev-manual/preprocessing.md
djjuhasz: I've removed this file altogether.

docs/src/dev-manual/preprocessing.md
djjuhasz: Done.

internal/workflow/batch.go
djjuhasz: yes. :)

Tiltfile
djjuhasz: I trimmed the paths after splitting them, which I should trim any leading or trailing whitespace.

internal/childwf/config.go
djjuhasz: I don't think so.  If I remove this check then an invalid type ("foo") doesn't return an error when reading the config.  Is there somewhere in the config parsing code that you think would automatically check enum values?

internal/config/config_test.go
djjuhasz: Removed.

djjuhasz:
@jraddaoui I think I've addressed all of your feedback. Let me know if I've missed anything.

docs/src/admin-manual/configuration.md
```diff
@@ -61,8 +61,8 @@ packages in the `internal` directory have configurable settings found here.
 
 !!! tip
 
-    Note that some [child workflow](../dev-manual/preprocessing.md) activities
-    may have their own configuration files.
+    Note that some [child workflow] activities may have their +own configuration
```
jraddaoui:
```suggestion
    Note that some [child workflow] activities may have their own configuration
```

docs/src/admin-manual/configuration.md
```diff
@@ -1138,116 +1138,135 @@ samplingRatio = 1.0
...
   needs to be changed if you expect to be running multiple different workflows
-  running at once, so they don't clash.
+  at once, so they don't clash.
 * `taskQueue`: In Temporal, a [Task Queue] is a lightweight, dynamically
   allocated queue that one or more workers can poll for tasks. Temporal supports
   many kinds of parallelization for a distributed application architecture, and
-  and task queues can be namespaced to support such a set-up. The default value
-  of `preprocessing` here differentiates it from the general `global` task queue
+  and task queues can be namespaced to support such a set-up. The value of
+  of `poststorage` here differentiates it from the general `global` task queue
```
jraddaoui:
```suggestion
  and task queues can be namespaced to support such a set-up. The value of
  `poststorage` here differentiates it from the general `global` task queue
```

docs/src/admin-manual/configuration.md
```diff
@@ -1138,116 +1138,135 @@ samplingRatio = 1.0
...
-taskQueue = "preprocessing"
-workflowName = "preprocessing"
+taskQueue = "poststorage"
+workflowName = "poststrage"
 ```
 
+* `type`: The type of child workflow. Type is used to load the correct child
+  workflow configuration in Enduro so only one child workflow of each type may
+  be configured for any instance of Enduro. If multiple child workflows have
+  the same type, an error will be thrown when the configuration is validated.
```
jraddaoui, line 1171: It would be nice to mention the three allowed type values here, even if they are explained below in more detail.

docs/src/dev-manual/devel.md
```diff
@@ -210,13 +212,19 @@ but it has seen little adoption and is largely unmaintained. Check the
...
+startup Tilt will attempt to load a `Tiltfile.enduro` file from each path which
+will add any workflow specific resources to the Tilt environment (e.g. a child
+worker). See the [Administrator configuration] docs for instructions on
+configuring the child workflows.
+
+### MOUNT_PREPROCESSING_VOLUME
+
+If MOUNT_PREPROCESSING_VOLUME is truthy (t, true, y, yes, 1) Tilt mounts a
+persistent volume claim (PVC) in the preservation system pod. The PVC must
+be defined in the preprocessing manifests and be called `preprocessing-pvc`.
```
jraddaoui: I don't know how best to word this, but it's not exactly the preservation system pod, it's the enduro-am-worker or enduro-a3m-worker pod.

internal/childwf/config.go
jraddaoui: I always expect that parsing an invalid value into an enum field will cause an error, but no, that's why have the enum validation in all places. We may be able to create some sort of hook that does that validation for all enums, but nothing to worry in this PR.

jraddaoui (approved):
Thanks @djjuhasz! Just a few really minor details.

docs/src/admin-manual/configuration.md
djjuhasz: Hmm, seems like a potential maintenance hassle to remember to update the list if we add more types. :thinking:

docs/src/dev-manual/devel.md
djjuhasz: I've just listed to the two pod names.

docs/src/admin-manual/configuration.md
djjuhasz: I'm going to leave this as-is for now.

jraddaoui: Sounds good. Having an entire section for each seems like a bigger hassle to me, but it's okay.

jraddaoui:
Refs https://github.com/artefactual-sdps/enduro/issues/1478.

PR #1503, opened 2026-02-05
Get AIP IDs in batch workflow

djjuhasz:
When polling SIP status in the batch workflows also get AIP IDs after ingest is complete so the IDs can be sent to the batch post-storage child workflow for reporting.

### DIFF
--- a/internal/workflow/activities/poll_sip_statuses.go
+++ b/internal/workflow/activities/poll_sip_statuses.go
@@ -36,6 +36,7 @@ type PollSIPStatusesActivityParams struct {
 
 type PollSIPStatusesActivityResult struct {
 	AllExpectedStatus bool
+	SIPIDstoAIPIDs    map[uuid.UUID]uuid.UUID
 }
 
 // PollSIPStatusesActivity polls the ingest service until all SIPs in a batch
@@ -73,7 +74,10 @@ func (a *PollSIPStatusesActivity) Execute(
 				return nil, fmt.Errorf("check SIP statuses: %v", err)
 			}
 			if result.done {
-				return &PollSIPStatusesActivityResult{AllExpectedStatus: result.allExpected}, nil
+				return &PollSIPStatusesActivityResult{
+					AllExpectedStatus: result.allExpected,
+					SIPIDstoAIPIDs:    result.aipIDs,
+				}, nil
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


internal/workflow/activities/poll_sip_statuses.go
```diff
@@ -119,7 +126,23 @@ func (a *PollSIPStatusesActivity) checkSIPStatuses(
 		} else if !slices.Contains(finalStatuses, status) {
 			return &checkResult{done: false}, nil
 		}
+
+		if status == enums.SIPStatusIngested && sip.AipUUID != nil {
+			id, err := uuid.Parse(*sip.AipUUID)
+			if err != nil {
+				return nil, fmt.Errorf("parse AIP UUID: %v", err)
```
jraddaoui, line 133: Hardly unlikely, and I'm not sure if we want to stop the poll, maybe just continue?

internal/workflow/batch.go
```diff
@@ -353,3 +357,16 @@ func (w *BatchWorkflow) postStorageWorkflow(
...
+func (w *BatchWorkflow) updateAIPIDs(state *batchWorkflowState, aipIDs map[uuid.UUID]uuid.UUID) {
+	for _, sd := range state.sipDetails {
+		aipID, ok := aipIDs[sd.sip.UUID]
+		if !ok {
+			continue
+		}
+
+		sd.sip.AIPID = uuid.NullUUID{UUID: aipID, Valid: true}
+	}
+}
```
jraddaoui: I think this method would make more sense in the state, with direct access to the SIP details.

internal/workflow/activities/poll_sip_statuses.go
```diff
@@ -119,7 +126,23 @@ func (a *PollSIPStatusesActivity) checkSIPStatuses(
...
+		done:        true,
+		allExpected: expectedStatusCount == expectedSIPCount,
+	}
+	if len(aipIDs) > 0 {
+		res.aipIDs = aipIDs
 	}
 
-	return &checkResult{done: true, allExpected: expectedStatusCount == expectedSIPCount}, nil
+	return res, nil
 }
```
jraddaoui, line 148: It would be nice to add some tests for the activity changes.

jraddaoui (commented):
Looking great @djjuhasz, thanks!

internal/workflow/activities/poll_sip_statuses.go
djjuhasz: Huh, I would expect it to be very unlikely that a non-nil AIP UUID from the database would fail to parse. :thinking:

jraddaoui: Yeah, that's why I thought continue may be better, not a big deal.

jraddaoui (approved):
👌

Issue #1494, opened 2026-01-23
Problem: There is no service to service authentication option between ingest and storage


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Ingest and storage communicate through the public API. If Enduro requires authentication everywhere (see #1493), internal service calls will fail unless we add a service to service authentication mechanism.

**Describe the solution you'd like**

Introduce a service to service authentication method that allows internal workflows (e.g., batch cancellation cleanup) to call storage endpoints securely without relying on OIDC tokens.

**Describe alternatives you've considered**

Keep API authentication optional for internal communication. This weakens security guarantees and conflicts with the decision to require authentication globally.

**Additional context**

We initially solved this issue running two APIs in the Enduro binary, a public one with authentication/access control and a private one without those options, see #995.

jraddaoui:
We completed two main changes related to API authentication:

### API support for multiple OIDC verifiers

- API auth now supports multiple `[[api.auth.oidc]]` entries instead of a single provider/client pair.
- Each entry creates one verifier. Verification succeeds on the first matching verifier.
- This allows the API to accept tokens from multiple provider/client combinations.
- ABAC and `skipEmailVerifiedCheck` are configured per verifier entry.
- Documentation: https://enduro.readthedocs.io/admin-manual/configuration/#oidc-authentication-providers-configuration
- Pull request: https://github.com/artefactual-sdps/enduro/pull/1511

### OIDC client credentials for ingest to storage requests

- Ingest storage client now supports OIDC client credentials via `[ingest.storage.oidc]`.
- Ingest fetches and refreshes bearer tokens and attaches them to storage API requests.
- Documentation: https://enduro.readthedocs.io/admin-manual/configuration/#ingest-storage-oidc-settings
- Pull request: https://github.com/artefactual-sdps/enduro/pull/1512

### Why this matters

This enables a multiple client setup:
- Public OIDC client for dashboard user access.
- Private OIDC client for ingest to storage s2s requests.
- Same API can verify both token types by configuring both verifier entries.

It also opens the door to:
- Multiple dashboards using the same API with different provider/client pairs.
- Potential future dashboard support for multiple providers in a single deployment (e.g., internal and external access).

PR #1506, opened 2026-02-12
Move storage client to ingest package

jraddaoui:
Move the storage API client contract and constructor into the ingest domain, where the dependency is consumed. This centralizes client creation that was duplicated across the AM and a3m worker mains and removes the storage domain client interface.

Refs #1494.


djjuhasz (approved):
Wow, this seems like a big change conceptually, but it's surprisingly straight-forward in the implementation. I like it! :)

jraddaoui:
Thanks @djjuhasz! This is the good thing about having been thinking about domain separation. If we do a full split of the project, we'd have to figure out how to share the generated client, but it **should** be easy to move the code from one side to the other.

jraddaoui:
FYI, Tilt CI needed some fixes that are also included in this PR. It looks like Docker was upgraded in these runners, requiring a more recent k3d version.

PR #1510, opened 2026-02-16
Move ingest storage config

jraddaoui:
Move storage API integration settings out of `[storage]` and into `[ingest.storage]` so ingest concerns are configured in the ingest domain. Rename `enduroAddress` to `address`. Keep `[storage]` for storage service settings and add `taskQueue` with default.

Refs #1494, follow-up to #1506.

### DIFF
--- a/docs/src/admin-manual/configuration.md
+++ b/docs/src/admin-manual/configuration.md
@@ -527,7 +527,7 @@ workflowType = "create aip"
   values are "create aip" and "create and review aip". The latter review
   workflow also only works if [a3m] is the configured [preservation engine].
 
-### Storage endpoint
+### Ingest storage settings
 
 This element configures the Enduro storage service API endpoint. Even when using
 [Archivematica] as the [preservation engine] (which includes the AM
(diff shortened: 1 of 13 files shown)


docs/src/admin-manual/configuration.md
```diff
@@ -538,12 +538,12 @@ AMSS. Otherwise, this configures the primary local storage service when using
 **Default values**:
 
 ```toml
-[storage]
-enduroAddress = "enduro.enduro-sdps:9002"
+[ingest.storage]
```
jraddaoui, line 541: I had conflicts naming this section, because we don't use the ingest prefix in other sections like we do with the storage specific settings, I initially called it `[storageClient]` to avoid conflicts with storage settings. I thought that, if at some point we normalize the ingest vs storage structure, it would be nice to use that prefix in all ingest related sections; so I called this one `[ingest.storage]`.

Check the changes in `enduro.toml`, where I added some comments explaining each section.

internal/ingest/config.go
```diff
@@ -0,0 +1,12 @@
...
+import "github.com/google/uuid"
+
+type Config struct {
+	Storage StorageConfig
+}
+
+type StorageConfig struct {
+	Address                    string
+	DefaultPermanentLocationID uuid.UUID
+}
```
jraddaoui, line 12: I'll add validation in a following PR.

enduro.toml
```diff
@@ -307,6 +307,20 @@ samplingRatio = 1.0
 # taskQueue = "postbatch"
 # workflowName = "postbatch"
 
+# [ingest.storage] configures ingest as a client of the storage API.
+# Use this section for cross-domain integration values:
+# - where ingest reaches the storage API (`address`)
```
djjuhasz: This is a bit hard to understand. Maybe "where ingest should send storage API calls (`address`)" or maybe just "the address of the storage API".

djjuhasz (approved):
LGTM @jraddaoui. I just had one change to suggest for the commentary.

PR #1511, opened 2026-02-17
Support multiple API OIDC token verifiers

jraddaoui:
Switch API auth from a single OIDC config to OIDCConfigs so the API
can verify access tokens from multiple provider/client pairs.

Build one verifier per OIDC config and evaluate them in order. Return
the first successful claims result, return ErrUnauthorized when all
verifiers fail authorization checks, and join non-authorization errors
when no verifier succeeds.

Update auth config validation to require at least one OIDC entry and
validate ABAC settings per entry. Refresh tests, sample config, and docs
to show `[[api.auth.oidc]]` usage and per-verifier ABAC configuration.

Also:

- Add service to service client to Keycloak

  Add a new OIDC secret client to the Keycloak configuration enabling
  service accounts for service to service authentication.

- Allow client credentials flow in `make auth`

  Allow setting CLIENT_SECRET and use the client credentials flow when
  it's set, fallback to the existing auth code flow if it's unset.

Refs #1494.

### DIFF
--- a/docs/src/admin-manual/configuration.md
+++ b/docs/src/admin-manual/configuration.md
@@ -211,25 +211,31 @@ obtaining an access token from the provider.
 
 ```toml
 [api.auth]
-enabled = true
+enabled = false
 ```
 
-#### OIDC authentication provider configuration
+#### OIDC authentication providers configuration
 
-This setting block is used to configure the OpenID Connect ([OIDC]) provider
-when API authentication is [enabled](#enable-api-authentication). The default
-value assumes that [Keycloak] will be used for handling OIDC single sign-on
-requests.
+These setting blocks are used to configure the OpenID Connect ([OIDC])
+providers used for access token verification when API authentication is
+[enabled](#enable-api-authentication). Multiple OIDC providers can be
+configured with a TOML array-of-tables syntax with the same structure. When
+API authentication is enabled, at least one OIDC provider must be configured to
+verify tokens.
 
 For more details on OIDC configuration, consult the [OIDC specification].
 
-**Default values**:
+**Example configuration**:
 
 ```toml
-[api.auth.oidc]
-providerURL = "http://keycloak:7470/realms/artefactual"
+[[api.auth.oidc]]
+providerURL = "https://idp-public.example.com/realms/enduro"
 clientID = "enduro"
-skipEmailVerifiedCheck = false
+
+[[api.auth.oidc]]
+providerURL = "https://idp-public.example.com/realms/enduro"
+clientID = "enduro-s2s"
+skipEmailVerifiedCheck = true
 ```
 
 * `providerURL`: Defines the OIDC provider URL. This parameter is required when
@@ -246,16 +252,23 @@ skipEmailVerifiedCheck = false
   tokens or UserInfo responses for the verified email claim. When set to
   **true**, this check is skipped.
 
-#### Enable Attribute Based Access Control for the API OIDC authentication
+##### Enable Attribute Based Access Control for the API OIDC authentication
 
 Enduro uses Attribute Based Access Control ([ABAC]) to manage permissions and
-access. When ABAC is enabled for the API, it will check a configurable
+access. When ABAC is enabled for an OIDC verifier, it will check a configurable
 multivalue claim against the defined required attributes based on each
 endpoint's configuration.
 
-**Default values**:
+For each verifier, add a matching `[api.auth.oidc.abac]` section immediately
+after the verifier configuration if ABAC is needed.
+
+**Example configuration**:
 
 ```toml
+[[api.auth.oidc]]
+providerURL = "https://idp-public.example.com/realms/enduro"
+clientID = "enduro"
+
 [api.auth.oidc.abac]
 enabled = true
 claimPath = "enduro"
@@ -263,6 +276,11 @@ claimPathSeparator = ""
 claimValuePrefix = ""
 useRoles = false
 rolesMapping =
+
+[[api.auth.oidc]]
+providerURL = "https://idp-public.example.com/realms/enduro"
+clientID = "enduro-s2s"
+skipEmailVerifiedCheck = true
 ```
 
 * `enabled`: Set to `true` to enable ABAC, or `false` to disable ABAC.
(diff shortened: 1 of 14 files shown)


docs/src/admin-manual/configuration.md
```diff
@@ -211,25 +211,31 @@ obtaining an access token from the provider.
...
 ```
 
-#### OIDC authentication provider configuration
+#### OIDC authentication providers configuration
 
-This setting block is used to configure the OpenID Connect ([OIDC]) provider
-when API authentication is [enabled](#enable-api-authentication). The default
-value assumes that [Keycloak] will be used for handling OIDC single sign-on
-requests.
+This setting blocks are used to configure the OpenID Connect ([OIDC]) providers
```
djjuhasz: "*These* setting blocks..." (emphasis is for proofreading only and should not be used in docs)

docs/src/admin-manual/configuration.md
```diff
@@ -211,25 +211,31 @@ obtaining an access token from the provider.
...
+used for access token verification when API authentication is
+[enabled](#enable-api-authentication). Multiple OIDC providers can be
+configured by using TOML array of tables syntax with the same structure. When
```
djjuhasz: configured *with a* TOML "array of tables" syntax... 

(emphasis is for proofreading only and should not be used in docs)

djjuhasz: A possible alternative to '"array of tables" syntax' is 'array-of-tables syntax' (without quotes)

docs/src/admin-manual/configuration.md
```diff
@@ -246,23 +252,35 @@ skipEmailVerifiedCheck = false
...
 
 Enduro uses Attribute Based Access Control ([ABAC]) to manage permissions and
-access. When ABAC is enabled for the API, it will check a configurable
+access. When ABAC is enabled for an OIDC verifier, it will check a configurable
 multivalue claim against the defined required attributes based on each
 endpoint's configuration.
 
-**Default values**:
+For each verifier, add a matching `[api.auth.oidc.abac]` section immediately
+after it if ABAC is needed.
```
djjuhasz: after *the verifier configuration* if ABAC is needed.

(emphasis is for proofreading only and should not be used in docs)

docs/src/admin-manual/iac.md
```diff
@@ -30,11 +30,26 @@ corsOrigin = "http://localhost"
...
 # moment. When enabled the API verifies the access token submitted with
-# each request. The API client is responsible for obtaining an access
-# token from the provider.
+# each request. Clients are responsible for obtaining an access token
+# from a configured OIDC provider below.
 enabled = true
 
-[api.auth.oidc]
+# Multiple OIDC providers are supported. Add one `[[api.auth.oidc]]` table
+# per provider/client pair. For each, add a matching `[api.auth.oidc.abac]`
```
djjuhasz: For each *provider/client* pair add an [api.auth.oidc.abac] section if ABAC is needed.

(emphasis is for proofreading only and should not be used in docs)

internal/api/auth/config.go
```diff
@@ -42,14 +44,21 @@ func (c Config) Validate() error {
...
 	}
-	if c.OIDC.ABAC.Enabled && c.OIDC.ABAC.ClaimPath == "" {
-		return errors.New("missing OIDC ABAC claim path with ABAC enabled")
-	}
-	if c.OIDC.ABAC.UseRoles && len(c.OIDC.ABAC.RolesMapping) == 0 {
-		return errors.New("missing OIDC ABAC roles mapping with use roles enabled")
+
+	for i := range c.OIDC {
+		if c.OIDC[i].ProviderURL == "" || c.OIDC[i].ClientID == "" {
+			return errors.New("missing OIDC configuration with API auth. enabled")
```
djjuhasz: This error message is a bit confusing imho. I'd suggest "OIDC configuration required when API auth is enabled", or something similar.

internal/api/auth/config.go
```diff
@@ -42,14 +44,21 @@ func (c Config) Validate() error {
...
+		}
+		if c.OIDC[i].ABAC.Enabled && c.OIDC[i].ABAC.ClaimPath == "" {
+			return errors.New("missing OIDC ABAC claim path with ABAC enabled")
```
djjuhasz: Ditto

internal/api/auth/config.go
```diff
@@ -42,14 +44,21 @@ func (c Config) Validate() error {
...
+		}
+		if c.OIDC[i].ABAC.UseRoles && len(c.OIDC[i].ABAC.RolesMapping) == 0 {
+			return errors.New("missing OIDC ABAC roles mapping with use roles enabled")
```
djjuhasz: Same.

internal/api/auth/token_verifier.go
```diff
@@ -26,27 +26,61 @@ func (t *NoopTokenVerifier) Verify(ctx context.Context, token string) (*Claims,
 }
 
 type OIDCTokenVerifier struct {
```
djjuhasz: Shouldn't this be plural because it represents multiple verifiers (e.g. OIDCTokenVerifiers)? Is there a reason to make `verifiers` a field instead of just defining `type OIDCTokenVerifiers []oidcVerifier`?

internal/api/auth/token_verifier.go
```diff
@@ -26,27 +26,61 @@ func (t *NoopTokenVerifier) Verify(ctx context.Context, token string) (*Claims,
...
+
+	verifiers := make([]oidcVerifier, len(cfgs))
+	for i, cfg := range cfgs {
+		// Initialize an OIDC provider.
+		provider, err := oidc.NewProvider(ctx, cfg.ProviderURL)
+		if err != nil {
+			return nil, err
+		}
+
+		// Create an ID token parser, but only trust ID tokens issued to this client id.
```
djjuhasz: token parser or verifier?

djjuhasz (approved):
Looks good @jraddaoui, I've mostly suggested some minor changes to the documentation and comments. I also had a few questions about the OIDCTokenVerifier struct.

internal/api/auth/token_verifier.go
jraddaoui: Initially it was `OIDCTokenVerifierSet` and I kept the single verifier version. In the end I thought it would be nice to maintain the interface even if it has multiple verifiers.

I've changed it now to what you are suggesting.

PR #1512, opened 2026-02-19
Add OIDC auth for ingest storage client

jraddaoui:
Introduce service to service authentication for ingest calls to the storage API using OIDC client credentials, and wire workers to build the storage client with that auth configuration at startup.

Extend ingest.storage configuration with OIDC settings, retry controls, and validation, including required ingest.storage address and default permanent location ID checks.

Refs #1494.


internal/a3m/config_test.go
```diff
@@ -12,6 +12,12 @@ import (
 func TestConfigValidate(t *testing.T) {
 	t.Parallel()
 
+	const baseConfig = `
+[ingest.storage]
+address = "storage-api:9000"
+defaultPermanentLocationId = "f2cc963f-c14d-4eaa-b950-bd207189a1f1"
+`
+
```
jraddaoui, line 20: It looks like these are the first required fields in the entire configuration, which doesn't feel right. This also raises the issue that we have with the config validation testing, we shouldn't be mixing packages in the unit tests like this. You'll see this issue in many tests from this PR.

jraddaoui (commented):
I'm still working on the provider and transport tests, and the documentation, but this is ready for an initial review.

internal/a3m/config_test.go
djjuhasz: I'm not sure what you are saying here — the `ingest.storage` fields are required now so the a3m config test fails to validate if they aren't present?

jraddaoui: Yes.

djjuhasz: Yeah, the config system in Enduro is hard to separate because TOML parsing happens globally but validation happens locally. We can get around the problem of having to include the `ingest.storage` config by declaring an a3m `config` struct directly and running `Validate()` on it but then we aren't testing that the TOML config settings are getting parsed into the config struct correctly, which has bitten me in the past. 😕

internal/ingest/storage_client_auth_provider.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+var _ AccessTokenProvider = (*OIDCAccessTokenProvider)(nil)
+
+func NewOIDCAccessTokenProvider(ctx context.Context, cfg StorageOIDCConfig) (*OIDCAccessTokenProvider, error) {
+	tokenURL := cfg.TokenURL
+	if tokenURL == "" {
+		provider, err := oidc.NewProvider(ctx, cfg.ProviderURL)
+		if err != nil {
+			return nil, fmt.Errorf("discover OIDC provider: %w", err)
```
djjuhasz: Is there a reason you are wrapping the provider error instead of using `%v` like we normally do?

internal/ingest/storage_client_auth_provider.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (p *OIDCAccessTokenProvider) requestToken(ctx context.Context) (*oauth2.Token, error) {
+	var err error
+	for attempt := 1; attempt <= p.retryMaxAttempts; attempt++ {
+		token, err := p.cc.TokenSource(ctx).Token()
+		if err == nil {
+			return token, nil
+		}
+		if !isRetryableErr(err) || attempt == p.retryMaxAttempts {
```
djjuhasz, line 102: Is the `|| attempt == p.retryMaxAttempts` necessary? You could alternatively make the for loop condition `attempt := 1; attempt < p.retryMaxAttempts; attempt++ {` I think?

internal/ingest/storage_client_auth_provider.go
```diff
@@ -0,0 +1,146 @@
...
+	if p.token.Expiry.IsZero() {
+		return false
+	}
+
+	return !p.token.Expiry.After(time.Now().Add(p.tokenExpiryLeeway))
+}
+
+func (p *OIDCAccessTokenProvider) requestToken(ctx context.Context) (*oauth2.Token, error) {
+	var err error
+	for attempt := 1; attempt <= p.retryMaxAttempts; attempt++ {
```
djjuhasz, line 97: I would expect if `p.retryMaxAttempts == 3` that this would make a maximum of 4 attempts: 1 initial attempt + a maximum of 3 retry attempts.  This will only make 3 attempts though.

internal/ingest/storage_client_auth_provider.go
```diff
@@ -0,0 +1,146 @@
...
+		token, err := p.cc.TokenSource(ctx).Token()
+		if err == nil {
+			return token, nil
+		}
+		if !isRetryableErr(err) || attempt == p.retryMaxAttempts {
+			break
+		}
+
+		if wait := p.backoff(attempt); wait > 0 {
+			timer := time.NewTimer(wait)
```
djjuhasz: I think `time.Sleep(wait)` has the same effect, but is easier to understand and less verbose.

djjuhasz: You could also get rid of the conditional altogether, as `time.Sleep(p.backoff(attempt))` will return immediately if backoff evaluates to 0 or a negative number.

djjuhasz (commented):
Looks nice @jraddaoui, I just had a few notes.

internal/a3m/config_test.go
jraddaoui: But we could test that separately, using Viper but not through the config package. Something for another PR ;)

internal/ingest/storage_client_auth_provider.go
jraddaoui: That condition is there to avoid sleeping after the last attempt.

internal/ingest/storage_client_auth_provider.go
jraddaoui: I read it differently, since these settings are under the retry prefix, I interpret `retryMaxAttempts` as the maximum total attempts, including the initial request. If we wanted retries after the first attempt, I'd name it `retryMaxRetries`. This matches how Temporal models it as well, `RetryPolicy.MaximumAttempts` refers to total attempts, not retry attempts only.

internal/ingest/storage_client_auth_provider.go
jraddaoui: The timer/select form is cancelable, so we wouldn't keep waiting when the context is canceled or times out. But the sleep time shouldn't be long here, so I'll go for that in favor of clarity.

internal/ingest/storage_client_auth_provider.go
jraddaoui: My bad, fixed!

internal/ingest/storage_client_auth_provider.go
jraddaoui: This brings a question I have been having all the time. What do you think about allowing this retry configuration? Could we live without it and just use the default values? This retries will happen as part of another request context so we probably don't want to allow many retires or big intervals.

internal/ingest/storage_client_auth_provider.go
djjuhasz: I see. 👍

internal/ingest/storage_client_auth_provider.go
djjuhasz: Yeah, that's a tough call.  Based on the recent need to add more configuration for activity time outs for large SIPs I'm inclined to keep it as-is rather than removing the config.  The defaults should be fine for the vast majority of deployments, but we can't know for sure that some networks might need higher (or lower) values.

internal/ingest/storage_client_auth_provider_test.go
```diff
@@ -0,0 +1,160 @@
...
+			responses: []tokenResponse{
+				{statusCode: http.StatusOK, accessToken: "token-1"},
+			},
+			calls:      2,
+			wantTokens: []string{"token-1", "token-1"},
+		},
+		{
+			name: "Refreshes token that is within leeway window",
+			responses: []tokenResponse{
+				{statusCode: http.StatusOK, accessToken: "token-1", expiresIn: 1},
```
djjuhasz, line 73: So, does the expiry leeway mean that the token *must* be refreshed before the end of the leeway window? 😕  What happens if the second `AccessToken()` call happens after the end of the leeway window?  Why are you setting `expiresIn: 1` here, but `expiresIn: 3600` on the second response? It seems like with a 1s token expiry the second `AccessToken()` call will always happen before the token expires (unless it's a very very slow environment) so the leeway setting is irrelevant to this test.

internal/ingest/storage_client_auth_provider.go
```diff
@@ -0,0 +1,136 @@
+package ingest
```
djjuhasz, line 1: This is adding four "auth" files (including tests) to the `ingest` package, which seem like a lot.  Could this code be moved to a general `auth` package, or is the auth provider and transport code specific to the ingest storage client somehow?

djjuhasz (commented):
Hi @jraddaoui. I don't really understand how the `OIDCAccessTokenProvider.tokenExpiryLeeway` is supposed to interact with the access token expiry. Could you explain the intent?

internal/ingest/storage_client_auth_provider.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (p *OIDCAccessTokenProvider) tokenNeedsRefresh() bool {
+	if p.token == nil || p.token.AccessToken == "" {
+		return true
+	}
+	if p.token.Expiry.IsZero() {
+		return false
+	}
+
+	return !p.token.Expiry.After(time.Now().Add(p.tokenExpiryLeeway))
```
djjuhasz: I'm having a hard time wrapping my head around this conditional, but I think it means that if the token expiry time is before the expiry leeway expires (now + tokenExpiryLeeway) then the token is refreshed, but if not then the cached token is returned?  Is this the intended behaviour?

jraddaoui: Exactly, to say it differently, if the token is going to expire in 30 seconds we ask for a new one.

internal/ingest/storage_client_auth_provider.go
jraddaoui: Right now it's only used by the storage client. The api/auth package contains code to auth requests sent to the API, this auth is to make requests to the API, so I'd not put it there.

How about putting it all, including the client, into a new ingest/storage package? Similar to the main storage package, where we contain what's specific to the storage domain, I think we could do the same at some point for ingest, moving the persistence package and others that are only for the ingest domain there.

internal/ingest/storage_client_auth_provider_test.go
jraddaoui: I'm not sure I understand your questions, without the leeway setting the token that is about to expire won't be refreshed, that's what we are testing here.

internal/ingest/storage_client_auth_provider.go
djjuhasz: Sure, creating an `ingest/storage` package sounds like a good solution.

internal/ingest/storage_client_auth_provider.go
djjuhasz: I was reading it wrong — I thought the cached token would be returned if the current time was after the token expiry time. 

I find the inversion of condition along with the time comparison functions to be very confusing.   For me it would make more sense if the conditional was `time.Now().After(token.Expiry.Add(-1 * p.tokenExpiryLeeway))` but that might just be my brain.  Also, today I learned that the go time package doesn't have a `time.Subtract(d duration)` function. :open_mouth:

internal/ingest/storage_client_auth_provider_test.go
djjuhasz: I misunderstood the condition for refreshing the token, this makes sense to me now.

internal/ingest/storage_client_auth_provider.go
jraddaoui: Thinking it through, we are going to need the transport/provider for the APIS integration. I'll move them to https://github.com/artefactual-labs/gotools soon, so I'll leave them in the ingest package for now. When we move them, only `storage_client.go` will remain here.

internal/ingest/storage_client_auth_provider.go
jraddaoui: Good point, although I think this makes more sense: `p.token.Expiry.Before(time.Now().Add(p.tokenExpiryLeeway))`.

jraddaoui:
Ready for another look @djjuhasz, I'll squash before merge ;)

internal/ingest/storage_client_auth_provider.go
djjuhasz: Works for me. :)

djjuhasz (approved):
LGTM 👍

Issue #1513, opened 2026-02-20
Problem: Batch cancellation cannot delete stored AIPs in a single request


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Batch cancellation (#1483) needs to delete stored AIPs through the storage API. We now use service to service authentication with client credentials (#1494), but the current deletion flow requires a request and a separate review by a distinct identity. A system account cannot satisfy that two identity review model in a single workflow step, so cancellation cannot complete deletion in one API call.

Current constraints:
- Deletion is currently request/review based (`/storage/aips/{uuid}/deletion-request` then `/storage/aips/{uuid}/deletion-review`).
- The requester cannot review their own deletion request.
- Cancellation of a pending deletion request is tied to the requester identity.

**Describe the solution you'd like**

Preferred choice: add a dedicated auto-approved deletion endpoint protected by a specific attribute.

Proposed behavior:
- Add a new storage API endpoint for direct AIP deletion without request/review.
- Require a dedicated attribute/permission for access to this endpoint.
- Use this endpoint for Batch cancellation so deletion can be executed by a single system account.
- Keep existing reviewed deletion endpoints and behavior unchanged.
- Do not generate an AIP deletion report when deletion is performed through this new endpoint.

Optional hardening:
- Add a flag in the API OIDC providers/verifiers configuration to mark a provider as a system account provider.
- When enabled, require both conditions for the endpoint: dedicated permission/attribute and token verified by a system account provider.

This keeps permission/attribute as the primary model while allowing provider trust controls where stronger operational guarantees are required.

**Describe alternatives you've considered**

1. Provider trust only (system account OIDC provider flag) for endpoint access.
This can improve certainty for service workflows but, on its own, shifts authorization away from the API’s ABAC pattern.

2. Two-provider/two-user review process for service deletion.
This preserves strict review semantics but adds significant configuration and operational complexity for a system workflow.

**Additional context**

- Authorization nuance to account for: wildcard `*` attributes satisfy all required scopes, and ABAC disabled verifiers can make scoped checks effectively permissive.
- Existing deletion report generation is tied to the reviewed deletion workflow and approved deletion requests; this direct endpoint will intentionally skip report generation.

scollazo:
Would this approach make RBAC mandatory?

jraddaoui:
[...]

PR #1514, opened 2026-02-23
Add AutoApprove and SkipReport flags to AIP deletion workflow

jraddaoui:
I decided to keep them separated, at some point we may want auto-approval with report generation.

Refs #1513.


djjuhasz (approved):
LGTM 👍

PR #1515, opened 2026-02-23
Add auto-approve AIP deletion endpoint

jraddaoui:
Based on #1514, refs #1513.


internal/api/design/storage.go
```diff
@@ -345,10 +345,37 @@ var _ = Service("storage", func() {
...
+		Error("internal_error")
 		HTTP(func() {
 			POST("/aips/{uuid}/deletion-request")
 			Response(StatusOK)
 			Response("not_found", StatusNotFound)
+			Response("not_valid", StatusBadRequest)
+			Response("internal_error", StatusInternalServerError)
+		})
+	})
+	Method("request_aip_deletion_auto", func() {
```
djjuhasz: I think we can just call this method "delete_aip".  The only reason for the "request" prefix in "request_aip_deletion" is because it required an approval step.

docs/src/admin-manual/iac.md
```diff
@@ -212,6 +212,7 @@ endpoints.
 | GET    | /storage/aips                         | `storage:aips:list`              |
 | POST   | /storage/aips                         | `storage:aips:create`            |
 | GET    | /storage/aips/{uuid}                  | `storage:aips:read`              |
+| POST   | /storage/aips/{uuid}/deletion-auto    | `storage:aips:deletion:auto`     |
```
djjuhasz, line 252: I think we should just make this a `DELETE` request method with the endpoint `/storage/aips/{uuid}` to follow the standard HTTP conventions.  The "deletion-request" prefix was only necessary because of the two step delete process.

internal/api/auth/attributes.go
```diff
@@ -15,6 +15,7 @@ const (
 	IngestSIPSourcesObjectsListAttr = "ingest:sipsources:objects:list"
 	IngestUsersListAttr             = "ingest:users:list"
 	StorageAIPSCreateAttr           = "storage:aips:create"
+	StorageAIPSDeletionAutoAttr     = "storage:aips:deletion:auto"
```
djjuhasz, line 18: And we can make this "storage:aips:delete"

djjuhasz (changes requested):
This looks good @jraddaoui, but I think the API should present a one-step deletion process as a standard DELETE HTTP request.  I think it's okay to keep the internal implementation with the "AutoApprove" switch to keep from branching the code too much.

docs/src/admin-manual/iac.md
jraddaoui: I’m not sure, @djjuhasz. I think of `DELETE` as fully removing a resource from the system, so future requests would return `404`/`410`, but [the standard](https://www.rfc-editor.org/rfc/rfc9110#name-delete) is not completely explicit about it. In our case, this endpoint does not actually remove the AIP from the system; it triggers a workflow, and we still keep the AIP with an updated status. For me, that feels closer to an action (`POST`) than a true resource deletion.

internal/api/design/storage.go
jraddaoui: Good point, we could remove that prefix, but I'd keep that name for a full resource deletion. Maybe just `aip_deletion_auto` to math the endpoint and the attribute.

internal/api/auth/attributes.go
jraddaoui: Check my previous comment, if you feel strongly about using `DELETE` this makes sense. Otherwise I'd keep it as it is.

docs/src/admin-manual/iac.md
jraddaoui: This made me realize, we should be returning `202` from these endpoints.

djjuhasz: As we discussed in person, I think sticking with `POST /storage/aips/{uuid}/deletion-auto` is fine for now and not worth a major overhaul of how we handle the distinction between deleting an AIP resource vs. and AIP file.  In the future we may want to revisit our API endpoint names to better distinguish between delete actions that target the file vs. the resource.

djjuhasz (approved):
Approved. 👍

Issue #1483, opened 2026-01-15
Feature: Add continue/cancel decision after batch ingest completion


jraddaoui:
**Is your feature request related to a problem? Please describe.**

The batch workflow currently fails as soon as not all SIPs reach the `ingested` status, which prevents users from completing a batch when only a small subset of SIPs fail. In `internal/workflow/batch.go` the workflow polls SIPs for `enums.SIPStatusIngested` and immediately errors if any are not ingested, so there is no user decision point to continue or cancel a partially successful batch.

**Describe the solution you'd like**

Introduce a user decision step at the end of the batch workflow after waiting for SIPs to reach `ingested` status. If some SIPs failed, set the batch status to "pending" and pause until a decision is submitted via a new API endpoint (similar to the AIP deletion review flow). The dashboard should call this endpoint with the user’s decision. If the user cancels, the workflow should clean up any remaining processing files and AIPs from the storage service; if the user continues, the workflow should complete normally.

**Describe alternatives you've considered**

Keep the existing behavior (fail the batch if any SIP is not ingested) and require operators to manually clean up and restart processing for successful SIPs. This does not support the "mostly successful batch" use case.

**Additional context**

A similar decision flow already exists for AIP deletions: the delete workflow waits on `storage.DeletionDecisionSignalName` in `internal/storage/workflows/delete.go`, and is driven by the `review_aip_deletion`/`cancel_aip_deletion` endpoints in `internal/api/design/storage.go`. The new batch decision flow could mirror that pattern.

This issue is a follow up to the initial batch workflow implementation from #1405.

jraddaoui:
Initial alert, confirmation dialogs and statuses:

<img width="1034" height="880" alt="Image" src="https://github.com/user-attachments/assets/82e46db1-112c-4e2f-a035-2da3316e6b51" />

After #1494 is addressed and automatic AIP deletion implemented (ignore retry option for now):

<img width="1034" height="880" alt="Image" src="https://github.com/user-attachments/assets/1abf334b-c309-4024-9a37-3febf4ebeccc" />

sallain:
There are still some next steps in the hopper to make this feature complete.

PR #1517, opened 2026-02-26
Clear ingested SIPs when canceling batches

jraddaoui:
Run a cleanup activity after canceled batch workflows complete so ingested SIPs are marked as canceled and their stored AIPs are deleted. The activity paginates SIPs, skips non-relevant SIPs, requests AIP deletions, and polls storage until deletions finish while aggregating errors.

Refs #1483.

### DIFF
--- a/internal/workflow/activities/clear_ingested_sips.go
+++ b/internal/workflow/activities/clear_ingested_sips.go
@@ -1,0 +1,207 @@
+package activities
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/google/uuid"
+	"go.artefactual.dev/tools/ref"
+	"go.artefactual.dev/tools/temporal"
+	goa "goa.design/goa/v3/pkg"
+
+	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
+	goastorage "github.com/artefactual-sdps/enduro/internal/api/gen/storage"
+	"github.com/artefactual-sdps/enduro/internal/enums"
+	"github.com/artefactual-sdps/enduro/internal/ingest"
+	storage_enums "github.com/artefactual-sdps/enduro/internal/storage/enums"
+)
+
+const ClearIngestedSIPsActivityName = "clear-ingested-sips-activity"
+
+type ClearIngestedSIPsActivity struct {
+	ingestsvc     ingest.Service
+	storageClient ingest.StorageClient
+	pollInterval  time.Duration
+}
+
+type ClearIngestedSIPsActivityParams struct {
+	BatchUUID uuid.UUID
+}
+
+type ClearIngestedSIPsActivityResult struct{}
+
+func NewClearIngestedSIPsActivity(
+	ingestsvc ingest.Service,
+	storageClient ingest.StorageClient,
+	pollInterval time.Duration,
+) *ClearIngestedSIPsActivity {
+	return &ClearIngestedSIPsActivity{
+		ingestsvc:     ingestsvc,
+		storageClient: storageClient,
+		pollInterval:  pollInterval,
+	}
+}
+
+func (a *ClearIngestedSIPsActivity) Execute(
+	ctx context.Context,
+	params *ClearIngestedSIPsActivityParams,
+) (*ClearIngestedSIPsActivityResult, error) {
+	h := temporal.StartAutoHeartbeat(ctx)
+	defer h.Stop()
+
+	logger := temporal.GetLogger(ctx)
+	logger.V(1).Info("Executing ClearIngestedSIPsActivity", "params", params)
+
+	// Paginate through all SIPs in the batch.
+	var sipCount int
+	var errs error
+	var aipIDs []string
+	for {
+		result, err := a.ingestsvc.ListSips(ctx, &goaingest.ListSipsPayload{
+			BatchUUID: ref.New(params.BatchUUID.String()),
+			Limit:     ref.New(1000),
+			Offset:    ref.New(sipCount),
+		})
+		if err != nil {
+			return nil, fmt.Errorf("list SIPs: %v", err)
+		}
+
+		sipCount += len(result.Items)
+		if sipCount == 0 {
+			return nil, fmt.Errorf("list SIPs: no SIPs found")
+		}
+
+		for _, sip := range result.Items {
+			status, err := enums.ParseSIPStatus(sip.Status)
+			if err != nil {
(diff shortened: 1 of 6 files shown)


internal/workflow/activities/clear_ingested_sips.go
```diff
@@ -0,0 +1,189 @@
...
+					)
+					continue
+				}
+			}
+
+			// Ignore SIPs that are not ingested or canceled, or if they don't have an AIP UUID.
+			if status != enums.SIPStatusIngested && status != enums.SIPStatusCanceled {
+				continue
+			}
+			if sip.AipUUID == nil || *sip.AipUUID == "" {
```
jraddaoui, line 100: When I started working on this activity, we didn't have the changes that you made to the poll SIP status activity to obtain the AIP IDs, so this was the best way I found to get them while doing the SIP updates. I may have done it differently now, maybe adding a bulk SIP status update method to call it with a local activity, and then leaving this one just for the AIP deletions. But I was too invested in this solution already and there may be some benefits of doing it "one by one". I tried to collect all non-blocking errors even if we don't show them anywhere yet, they will appear in the logs and workflow error.

internal/workflow/activities/clear_ingested_sips.go
```diff
@@ -0,0 +1,189 @@
...
+				if err := a.ingestsvc.SetStatus(ctx, sip.UUID, enums.SIPStatusCanceled); err != nil {
+					errs = errors.Join(
+						errs, fmt.Errorf("set SIP %q status to canceled: %v", sip.UUID.String(), err),
+					)
+					continue
+				}
+			}
+
+			// Ignore SIPs that are not ingested or canceled, or if they don't have an AIP UUID.
+			if status != enums.SIPStatusIngested && status != enums.SIPStatusCanceled {
```
djjuhasz: This is a bit confusing because you just changed the status from "ingested" to "canceled" above.  I think it's clearer to update the status var to "canceled" after the update succeeds, and make this condition `status != enums.SIPStatusCanceled`.

djjuhasz: Alternately you could move this conditional, as-is, above the status update block.

internal/workflow/activities/clear_ingested_sips.go
```diff
@@ -0,0 +1,189 @@
...
+			if err := a.deleteAIP(ctx, params.BatchUUID, *sip.AipUUID); err != nil {
+				errs = errors.Join(errs, err)
+				continue
+			}
+
+			// Keep track of the AIP IDs for polling.
+			aipIDs = append(aipIDs, *sip.AipUUID)
+		}
+
+		sipCount += len(result.Items)
```
djjuhasz: I'd move this expression up to the top of the loop right after `results` is returned.  I had to scroll up there to figure out where the `result` was coming from.

internal/workflow/activities/clear_ingested_sips.go
```diff
@@ -0,0 +1,189 @@
...
+		if sipCount == 0 || result.Page == nil || sipCount >= result.Page.Total {
```
djjuhasz: I'd move `if sipCount == 0` up to, right after the `sipCount += len(result.Items)` expression, and return from the `Execute()` method immediately rather than break the loop.

internal/workflow/activities/clear_ingested_sips.go
```diff
@@ -0,0 +1,189 @@
...
+		UUID:       aipID,
+		Reason:     fmt.Sprintf("Batch %s canceled", batchUUID),
+		SkipReport: ref.New(true),
+	})
+	if err != nil {
+		aip, showErr := a.storageClient.ShowAip(ctx, &goastorage.ShowAipPayload{UUID: aipID})
+		if showErr != nil {
+			return fmt.Errorf("request AIP %q deletion: %v", aipID, errors.Join(err, showErr))
+		}
+		if aip.Status == storage_enums.AIPStatusDeleted.String() {
```
djjuhasz: Why would the the AIP already be deleted at this point?

If it's important to check for  `storage_enums.AIPStatusDeleted` specifically I'd return a specific "AIP already deleted" error from `a.storageClient.AipDeletionAuto` (where we have the AIP status available already), then catch that error and return nil here.  Doing another database read just to get this status seems unnecessary to me.

djjuhasz: You might also just want to catch all `AIP is not stored` goaerrors here, and return nil for any status that is not "stored".

internal/workflow/activities/clear_ingested_sips.go
```diff
@@ -0,0 +1,189 @@
...
+func (a *ClearIngestedSIPsActivity) checkAIPStatuses(ctx context.Context, aipIDs []string) (bool, error) {
+	var errs error
+	for _, aipID := range aipIDs {
+		aip, err := a.storageClient.ShowAip(ctx, &goastorage.ShowAipPayload{UUID: aipID})
+		if err != nil {
+			errs = errors.Join(errs, fmt.Errorf("show AIP %q: %v", aipID, err))
+			continue
+		}
+
+		if aip.Status == storage_enums.AIPStatusStored.String() {
```
djjuhasz: What is the `aip.Status` when the deletion request is happening? "processing"?

Maybe it should error if `aip.Status` is "unspecified" or "pending" too?  If the AIP is waiting on a deletion request somehow this poll could be waiting a very long time.

djjuhasz (changes requested):
Looks good @jraddaoui, just a few notes.

internal/workflow/activities/clear_ingested_sips.go
jraddaoui: We are only polling for AIPs when their deletion request succeeded, and during the deletion process the AIP can have three statuses: "queued", "processing" and "pending". So I have added "unspecified" to this check.

Because we know the request succeeded, we know the AIP is going to have the "pending" status for a short time, but we still need to consider it. Ideally, the AIP deletion workflow shouldn't mark the AIP as "pending" if the auto-approve flag is set, I'll do that in a different PR and then consider the "pending" status in here too.

jraddaoui:
Thanks @djjuhasz, great feedback!

internal/workflow/activities/clear_ingested_sips.go
jraddaoui: See https://github.com/artefactual-sdps/enduro/pull/1525, I'll check for "pending" here too.

djjuhasz (approved):
LGTM! Thanks  @jraddaoui. 👍

PR #1519, opened 2026-02-26
Fix dependency vulnerabilities

jraddaoui:
https://docs.npmjs.com/cli/v11/configuring-npm/package-json#overrides


djjuhasz (approved):
Thanks @jraddaoui 👍

PR #1520, opened 2026-02-26
Add `SIPID` to the preprocessing workflow params

djjuhasz:
This is required for the CVA preprocessing workflow for https://github.com/artefactual-sdps/cva-enduro-workflows/issues/1


jraddaoui (approved):
Nice!

Issue #1449, opened 2025-12-04
Problem: task details show the same message for all tasks


jraddaoui:
**Describe the bug**

When the workflow tasks are automatically updated in the UI they all show the same note.

**To Reproduce**

Start an ingest workflow and go to the SIP page.

**Expected behavior**

Each task should show its own message.

**Screenshots**

Ingest workflow - automatic updates:

<img width="1337" height="859" alt="Image" src="https://github.com/user-attachments/assets/e0f9499c-3a44-45c9-8087-62825d5eea1f" />

Ingest workflow - after page refresh:

<img width="1337" height="859" alt="Image" src="https://github.com/user-attachments/assets/b539685c-d844-44dc-980a-81bfa5d5ca71" />

AIP deletion workflow - automatic updates:

<img width="1337" height="424" alt="Image" src="https://github.com/user-attachments/assets/0ae47ba0-4a82-4f89-a14a-b3d7c95c6bfc" />

AIP deletion workflow - after page refresh:

<img width="1337" height="424" alt="Image" src="https://github.com/user-attachments/assets/306e06c3-ceec-44f2-822b-cdd1b38fdb10" />

**Additional context**

Websocket events seem to be okay:

<img width="988" height="464" alt="Image" src="https://github.com/user-attachments/assets/cb302646-a080-465b-9ebb-db0524fd8b7b" />

Show more link shows the expected content:

<img width="806" height="197" alt="Image" src="https://github.com/user-attachments/assets/a94af923-7ba2-4554-ac82-187e19c8e189" />

sallain:
Wuh-oh

PR #1521, opened 2026-02-27
Make EmailLinkedText react to text prop changes

jraddaoui:
Compute the split parts so updates to the text prop are reflected after mount. This fixes stale task note rendering during live updates.

Refs #1449.


djjuhasz (approved):
👍

Issue #1493, opened 2026-01-23
Problem: Enduro allows to disable authentication in the API and the dashboard


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Enduro currently supports running without API authentication, which can lead to unsafe behavior and ambiguous authorization. After the discussion in https://github.com/artefactual-sdps/enduro/issues/1220, we decided to stop supporting unauthenticated mode and require authentication across the application.

**Describe the solution you'd like**

Remove the option to disable authentication entirely from the API and dashboard. Authentication should be required for all operations, and any related UI or logic that assumes authentication is disabled should be removed or updated accordingly.

**Describe alternatives you've considered**

Keep the unauthenticated mode but restrict/disable sensitive operations like AIP deletion. This still leaves ambiguity in behavior and increases maintenance burden, so we decided to remove the mode instead.

**Additional context**

This decision depends on a separate effort to implement service to service authentication for internal workflows (ingest <-> storage), see #1494.

scollazo:
Making https://github.com/dexidp/dex part of the enduro stack could allow for OIDC authentication out-of-the box.

PR #1523, opened 2026-02-27
Allow unauth auto-approve AIP deletion

jraddaoui:
Until  #1493 is addressed we need to consider systems where authentication is disabled. This also allows to target the unauth internal API from the storage client in environments where an OIDC provider with the client credentials flow is not available (e.g.: Dex).

Refs #1494, #1483 and #1513. Based on #1517.

### DIFF
--- a/internal/storage/deletion_request.go
+++ b/internal/storage/deletion_request.go
@@ -35,8 +35,22 @@ func (s *serviceImpl) AipDeletionAuto(ctx context.Context, payload *goastorage.A
 		payload = &goastorage.AipDeletionAutoPayload{}
 	}
 
+	// Authentication can be disabled for auto-approve.
+	claims := auth.UserClaimsFromContext(ctx)
+	if claims != nil {
+		if err := checkClaims(claims); err != nil {
+			return err
+		}
+	} else {
+		claims = &auth.Claims{
+			Email: "auto-approve@enduro.sys",
+			Sub:   "unauthenticated",
+			Iss:   "unauthenticated",
+		}
+	}
+
 	skipReport := payload.SkipReport != nil && *payload.SkipReport
-	return s.requestAIPDeletion(ctx, payload.UUID, payload.Reason, true, skipReport)
+	return s.requestAIPDeletion(ctx, payload.UUID, payload.Reason, claims, true, skipReport)
 }
 
 func (s *serviceImpl) RequestAipDeletion(ctx context.Context, payload *goastorage.RequestAipDeletionPayload) error {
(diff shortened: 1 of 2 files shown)


djjuhasz:
@jraddaoui this is necessary because Dex doesn't support service-to-service client credentials?

jraddaoui:
Yes, kind of. Check this Slack thread: https://artefactual.slack.com/archives/C05BGJAJ4HZ/p1769642116808239.

internal/storage/deletion_request.go
```diff
@@ -35,31 +35,46 @@ func (s *serviceImpl) AipDeletionAuto(ctx context.Context, payload *goastorage.A
...
 
+	// Authentication can be disabled for auto-approve.
+	claims := auth.UserClaimsFromContext(ctx)
+	if claims != nil {
+		if err := checkClaims(claims); err != nil {
+			return err
+		}
+	} else {
+		claims = &auth.Claims{
+			Email: "unauthenticated@enduro.invalid",
```
djjuhasz: I think seeing this email in the dashboard could scare some operators who might think an unauthenticated user has hacked their Enduro site.  I think something a little gentler would be less scary, e.g. "auto-approve@enduro.sys"

djjuhasz: @jraddaoui and I discussed this internally — I've had second thoughts about the "auto-approve@enduro.sys" email-like string, as it acts like an email in the dashboard, with a mailto: link, but is not a real email address that we control.  @jraddaoui agreed that we could use user identifiers here that are not emails or email-like strings, but that an email is currently required for the the access token claims, and the code would need  to be changed in several places to undo this requirement:

- email check: https://github.com/artefactual-sdps/enduro/blob/main/internal/storage/deletion_request.go#L20-L22
- hardcoded client credentials email: https://github.com/artefactual-sdps/enduro/pull/1517/changes/5b7bcc5c2d49588639e971244ec5c27f29fade34

For now I think it's fine to proceed with "auto-approve@enduro.sys" or something similar and do the changes to the access token claims later.

djjuhasz (approved):
👍

jraddaoui:
I merged this in the wrong branch and then it got lost when rebasing and merging that branch into main. See #1527 for a new version.

internal/storage/deletion_request.go
jraddaoui: Thanks @djjuhasz! See #1527 for a replacement of this PR. In the end I didn't include the hardcoded client credentials email (5b7bcc5) in #1517.

Issue #1498, opened 2026-01-27
Feature: Make certain long-running activity timeout values configurable


fiver-watson:
**Is your feature request related to a problem? Please describe.**

Yes. With large SIPs that contain many files, certain activities can time out, causing unexpected failures during processing. 

From a client report: 

> Those [SIPs] with particularly many files (>8000) fail already at step 15 during the create premis step, in particular it runs into the timeout `ScheduleToClose` during the activity `add_premis_objects.go`. 
> ...
> The SIP which still have a lot of files (>5500) but not quite as many, fail after step 30 or 35 during the “poll transfer activity” or “poll ingest activity”. Our hypothesis is that this activity times out and attempts to retry. The retry then writes the same task uuid again to the enduro database in the task.uuid column, causing the duplicate error and leading the process to fail.

**Describe the solution you'd like**

Right now changing the timeout values for these activities require code changes - making them configurable would allow system administrators to adjust their deployments without needing to alter application code. 

Activities reported for consideration: 

* Create PREMIS file
* Validate File formats
 
**Additional context**

Somewhat related issues: 

* https://github.com/artefactual-sdps/preprocessing-sfa/issues/170
* https://github.com/artefactual-sdps/temporal-activities/issues/54
* https://github.com/artefactual-sdps/enduro-team-tasks/issues/79

jraddaoui:
Spreadsheet with activity configurations: https://docs.google.com/spreadsheets/d/1S4x_qQHPzSwN66YmXORoM6BhMU-a_xzuY0dsRoBFQEI

While we consider making timeouts configurable, there is one timeout we should increase ASAP. The remove paths activity has a  `StartToCloseTimeout` of 1 second (T1, L62), which won't be enough for large SIPs or/and mounted filesystems like NFS. Ideally, we would heartbeat from this activity and add retries, but for now I'll increase the current timeout to 15 minutes.

PR #1526, opened 2026-03-02
Increase remove paths activity StartToCloseTimeout

jraddaoui:
See https://github.com/artefactual-sdps/enduro/issues/1498#issuecomment-3986854718.


djjuhasz (approved):
👍

PR #1527, opened 2026-03-03
Allow unauth auto-approve AIP deletion

jraddaoui:
Until #1493 is addressed we need to consider systems where authentication is disabled. This also allows to target the unauth internal API from the storage client in environments where an OIDC provider with the client credentials flow is not available (e.g.: Dex).

Also, allow AIP deletion requests without email claim. Consider `preferred_username` claim. Get requester/reviewer display name from the email claim if available, falling back to the preferred username and then to the name; at least one of them is required. This removes the requirement to configure an email claim for the client credentials flow, which may not be included by default in some providers.

Refs #1494, #1483 and #1513. Replaces #1523.

### DIFF
--- a/internal/api/auth/claims.go
+++ b/internal/api/auth/claims.go
@@ -8,11 +8,12 @@ import (
 )
 
 type Claims struct {
-	Email         string `json:"email,omitempty"`
-	EmailVerified bool   `json:"email_verified,omitempty"`
-	Name          string `json:"name,omitempty"`
-	Iss           string `json:"iss,omitempty"`
-	Sub           string `json:"sub,omitempty"`
+	Email             string `json:"email,omitempty"`
+	EmailVerified     bool   `json:"email_verified,omitempty"`
+	PreferredUsername string `json:"preferred_username,omitempty"`
+	Name              string `json:"name,omitempty"`
+	Iss               string `json:"iss,omitempty"`
+	Sub               string `json:"sub,omitempty"`
 	// The attributes are parsed from a configured claim and added here,
 	// they are needed in the JSON representation for the MarshalBinary and
 	// UnmarshalBinary methods below. We use the `enduro_internal_attributes`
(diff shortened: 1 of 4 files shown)


internal/storage/deletion_request.go
jraddaoui, line 31: We may want to move this to the access token verification and require it in all request; after we address #1493.

internal/storage/deletion_request.go
```diff
@@ -35,31 +35,46 @@ func (s *serviceImpl) AipDeletionAuto(ctx context.Context, payload *goastorage.A
...
 
+	// Authentication can be disabled for auto-approve.
+	claims := auth.UserClaimsFromContext(ctx)
+	if claims != nil {
+		if err := checkClaims(claims); err != nil {
+			return err
+		}
+	} else {
+		claims = &auth.Claims{
+			Name: "enduro-auto-approve",
```
jraddaoui: Something better?

internal/api/auth/claims.go
```diff
@@ -59,6 +60,22 @@ func (c *Claims) CheckAttributes(required []string) bool {
 	return true
 }
 
+// DisplayName returns the email claim if available, falling back to the
+// preferred username and then to the name. It returns an empty string if none
+// are set.
+func (c *Claims) DisplayName() string {
```
jraddaoui, line 66: We could use this in the audit logs too, something for another PR.

jraddaoui (commented):
Sorry for the mess @djjuhasz, you could ignore the first commit as that is the one you already reviewed in #1523. The second one allows the email claim to be missing, which simplifies the client credentials configuration in the provider; and also changes the unauth user claims to not use an email.

jraddaoui:
This is how it looks with service to service auth enabled and the `enduro-s2s` client credentials Keycloak client. The `service-account-enduro-s2s` value comes in the `preferred_username`.

Screenshot from 2026-03-03 13-21-27

It will look the same with authentication disabled, with `enduro-auto-approve` instead of `service-account-enduro-s2s`.

internal/storage/deletion_request.go
djjuhasz: I'd just format it for human readability, e.g. "Enduro auto-approve"

internal/storage/deletion_request.go
djjuhasz: I have mixed feelings about making a "display name" required. On the one hand it's obviously helpful for UI and logging to have a unique, human friendly name but on the other hand there's so much range in what fields IDPs provide in their claims, and I don't think any of the email, preferred name, or name claims are required by the spec. :(

djjuhasz (approved):
LGTM @jraddaoui. I had one suggestion for the Enduro auto-approve name, but I don't think it really matters that much, so I'll leave it up to you.

internal/storage/deletion_request.go
jraddaoui: Yeah, I have my conflicts too, but I don't see a point on making an audit log without them, unless we fallback to the iss/sub, which is probably what we should be doing to really be able to identify the user.

PR #1539, opened 2026-03-09
Use go.artefactual.dev/tools/clientauth

jraddaoui:
Refs #1494.


internal/config/config.go
```diff
@@ -108,11 +108,6 @@ func Read(config *Configuration, configFile string) (found bool, configFileUsed
 	v.SetDefault("storage.taskqueue", temporal.GlobalTaskQueue)
 	v.SetDefault("temporal.taskqueue", temporal.GlobalTaskQueue)
 	v.SetDefault("upload.maxSize", 4294967296)
-	v.SetDefault("ingest.storage.oidc.tokenExpiryLeeway", ingest.DefaultStorageOIDCTokenExpiryLeeway)
-	v.SetDefault("ingest.storage.oidc.retryMaxAttempts", ingest.DefaultStorageOIDCRetryMaxAttempts)
-	v.SetDefault("ingest.storage.oidc.retryInitialInterval", ingest.DefaultStorageOIDCRetryInitialInterval)
-	v.SetDefault("ingest.storage.oidc.retryMaxInterval", ingest.DefaultStorageOIDCRetryMaxInterval)
-	v.SetDefault("ingest.storage.oidc.retryBackoffCoefficient", ingest.DefaultStorageOIDCRetryBackoffCoefficient)
```
jraddaoui, line 115: Defaults are set at validation in: https://github.com/artefactual-labs/gotools/blob/main/clientauth/provider.go#L53-L70.

internal/ingest/config.go
```diff
@@ -27,18 +20,8 @@ type StorageConfig struct {
...
-	ClientSecret            string
-	Scopes                  []string
-	Audience                string
-	TokenExpiryLeeway       time.Duration
-	RetryMaxAttempts        int
-	RetryInitialInterval    time.Duration
-	RetryMaxInterval        time.Duration
-	RetryBackoffCoefficient float64
+	Enabled                                  bool
+	clientauth.OIDCAccessTokenProviderConfig `mapstructure:",squash"`
```
jraddaoui, line 24: `mapstructure:",squash"` is needed to flatten the `clientauth.OIDCAccessTokenProviderConfig` fields into the parent config level:

https://pkg.go.dev/github.com/go-viper/mapstructure/v2#hdr-Embedded_Structs_and_Squashing

This allow us to reuse the clientauth config and add the enabled flag without another section in the TOML.

djjuhasz: Cool! :)

djjuhasz (approved):
LGTM! 👍

PR #1541, opened 2026-03-09
Pass BatchID to preprocessing child workflow

djjuhasz:
The CVA preprocessing child workflow should upload a metadata file only if the SIP is part of a batch, as the metadata is used to create a CSV for a batch of SIPs. Pass the BatchID (if there is one) to preprocessing so it knows whether to upload the metadata file or not.


jraddaoui (approved):
Good call!

PR #1543, opened 2026-03-10
Remove sql.NullTime from domain types

sevein:
`sql.NullTime` was leaking a persistence concern into the ingest domain model. `datatypes.SIP`, `datatypes.Workflow`, and `datatypes.Task` were using a SQL-specific nullable wrapper to represent optional timestamps, even though those types are not database rows and should not need to know how SQL encodes `NULL`.

This change moves null handling back to the edges:

- Domain types now use plain `time.Time`
- The zero value of `time.Time` represents "timestamp not set"
- Persistence adapters translate zero time <-> SQL `NULL`
- Goa/API conversion translates zero time into omitted optional fields

The important distinction is that "unset" is still preserved, but it is now represented in the domain with Go's own zero value instead of a database wrapper type.

#### Why zero `time.Time` instead of `*time.Time`?

The domain only needs a simple "set or not set" representation and `time.Time{}` already provides that without adding pointer semantics.

#### Why keep nullable handling in the adapters?

Because NULL is a database/API representation detail, not a domain concept. The domain can use zero `time.Time` to mean "unset", and the adapters are the right place to translate that to and from SQL NULL or omitted wire fields.

### DIFF
--- a/internal/datatypes/sip_test.go
+++ b/internal/datatypes/sip_test.go
@@ -1,0 +1,99 @@
+package datatypes
+
+import (
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"gotest.tools/v3/assert"
+
+	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
+	"github.com/artefactual-sdps/enduro/internal/enums"
+)
+
+func TestSIPGoa(t *testing.T) {
+	t.Parallel()
+
+	sipUUID := uuid.New()
+	aipUUID := uuid.New()
+	uploaderUUID := uuid.New()
+	batchUUID := uuid.New()
+	createdAt := time.Date(2025, 6, 23, 14, 57, 12, 0, time.UTC)
+	startedAt := time.Date(2025, 6, 23, 15, 0, 0, 0, time.UTC)
+	completedAt := time.Date(2025, 6, 23, 15, 5, 0, 0, time.UTC)
+
+	type test struct {
+		name string
+		sip  *SIP
+		want *goaingest.SIP
+	}
+
+	for _, tt := range []test{
+		{
+			name: "Converts nil SIP to nil Goa SIP",
+		},
+		{
+			name: "Converts SIP with zero optional times",
+			sip: &SIP{
+				UUID:      sipUUID,
+				Name:      "transfer-1",
+				Status:    enums.SIPStatusQueued,
+				CreatedAt: createdAt,
+			},
+			want: &goaingest.SIP{
+				UUID:      sipUUID,
+				Name:      new("transfer-1"),
+				Status:    enums.SIPStatusQueued.String(),
+				CreatedAt: "2025-06-23T14:57:12Z",
+			},
+		},
+		{
+			name: "Converts SIP with all optional fields",
+			sip: &SIP{
+				UUID:        sipUUID,
+				Name:        "transfer-1",
+				AIPID:       uuid.NullUUID{Valid: true, UUID: aipUUID},
+				Status:      enums.SIPStatusIngested,
+				CreatedAt:   createdAt,
+				StartedAt:   startedAt,
+				CompletedAt: completedAt,
+				FailedAs:    enums.SIPFailedAsPIP,
+				FailedKey:   "failed/pip.7z",
+				Uploader: &User{
+					UUID:  uploaderUUID,
+					Email: "nobody@example.com",
+					Name:  "Test User",
+				},
+				Batch: &Batch{
+					UUID:       batchUUID,
+					Identifier: "batch-1",
+					Status:     enums.BatchStatusPending,
+				},
+			},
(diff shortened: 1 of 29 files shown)


internal/ingest/convert.go
```diff
@@ -44,18 +44,13 @@ func sipIngestAuditEvent(s *datatypes.SIP) *auditlog.Event {
...
-	}
-
 	res := &goaingest.SIPWorkflow{
 		UUID:        w.UUID,
 		TemporalID:  w.TemporalID,
 		Type:        w.Type.String(),
 		Status:      w.Status.String(),
-		StartedAt:   startedAt,
-		CompletedAt: db.FormatOptionalTime(w.CompletedAt),
+		StartedAt:   ref.DerefZero(db.FormatOptionalZeroTime(w.StartedAt)),
```
jraddaoui: Could this just be `StartedAt:   startedAt,` directly?

internal/datatypes/sip_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"gotest.tools/v3/assert"
+
+	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
+	"github.com/artefactual-sdps/enduro/internal/enums"
+)
+
+func TestSIP_Goa(t *testing.T) {
```
jraddaoui: Nitpick: we usually use `TestSIPGoa`.

internal/persistence/ent/client/sip_test.go
```diff
@@ -223,17 +222,14 @@ func TestUpdateSIP(t *testing.T) {
...
-			t, _ := time.Parse(time.RFC3339, "1980-01-01T09:30:00Z")
-			return t
-		}(),
-		Valid: true,
-	}
+	started := time.Now()
+	started2 := func() time.Time {
+		t, _ := time.Parse(time.RFC3339, "1980-01-01T09:30:00Z")
+		return t
+	}()
```
jraddaoui: Why this and not just `time.Date()`?

internal/persistence/ent/client/sip.go
```diff
@@ -149,11 +149,11 @@ func (c *client) UpdateSIP(
...
+	if !up.StartedAt.IsZero() {
+		q.SetStartedAt(up.StartedAt.UTC())
 	}
-	if up.CompletedAt.Valid {
-		q.SetCompletedAt(up.CompletedAt.Time)
+	if !up.CompletedAt.IsZero() {
+		q.SetCompletedAt(up.CompletedAt.UTC())
 	}
 	if up.FailedAs.IsValid() {
 		q.SetFailedAs(up.FailedAs)
```
jraddaoui, line 159: We were not calling `UTC()` here before, if it's really needed we may need it in a lot more places.

internal/persistence/ent/client/sip_test.go
```diff
@@ -600,23 +596,17 @@ func TestListSIPs(t *testing.T) {
...
-		Valid: true,
-	}
+	started := func() time.Time {
+		t, _ := time.Parse(time.RFC3339, "2024-09-25T09:31:11Z")
+		return t
+	}()
+	started2 := func() time.Time {
+		t, _ := time.Parse(time.RFC3339, "2024-09-25T10:03:42Z")
+		return t
+	}()
```
jraddaoui: Same, I know it was like this before, but it feels unnecessary.

internal/persistence/ent/client/task.go
```diff
@@ -153,11 +155,11 @@ func (c *client) UpdateTask(
...
-	if up.StartedAt.Valid {
-		q.SetStartedAt(up.StartedAt.Time)
+	if !up.StartedAt.IsZero() {
+		q.SetStartedAt(up.StartedAt.UTC())
 	}
-	if up.CompletedAt.Valid {
-		q.SetCompletedAt(up.CompletedAt.Time)
+	if !up.CompletedAt.IsZero() {
+		q.SetCompletedAt(up.CompletedAt.UTC())
 	}
```
jraddaoui, line 161: Same about UTC().

internal/persistence/ent/client/workflow.go
```diff
@@ -104,13 +106,13 @@ func (c *client) UpdateWorkflow(
...
 	} else {
 		q.ClearStartedAt()
 	}
-	if up.CompletedAt.Valid {
-		q.SetCompletedAt(up.CompletedAt.Time.UTC())
+	if !up.CompletedAt.IsZero() {
+		q.SetCompletedAt(up.CompletedAt.UTC())
 	} else {
 		q.ClearCompletedAt()
 	}
```
jraddaoui, line 116: Same.

jraddaoui (approved):
Thanks for taking care of this @sevein! Mostly nitpicks below, feel free to ignore.

PR #1546, opened 2026-03-11
Bump version to 0.25.0 and update changelog

jraddaoui:
Also, use changelog links for GitHub release notes. The current release notes only included the latest commit, then we had to update them manually to include links to the functional and full changelogs, but by then the release email is already sent with the previous single commit notes.

The `hack/generate_release_notes.sh` script (and specially its regular expressions) was developed with Codex gpt-5.4, although I did many cleanups and tweaks considering it will only be used by this repository and the release action.


djjuhasz (approved):
I don't understand much of `generate_release_notes.sh` but the results look good! :+1:

PR #1549, opened 2026-03-11
Fix GitHub release notes generation

jraddaoui:
Fetch full git history in the release workflow so the workflow can resolve the current and previous tags correctly. Write generated release notes to the runner's temporary directory instead of the repository so GoReleaser does not fail on a dirty checkout.


jraddaoui:
This made it work, even if it looks like it failed in the commit. That failure was caused by pushing the tag twice; when it tried to save the release assets they were already there from the previous run.

PR #1550, opened 2026-03-12
Remove nightly gosec Github action

djjuhasz:
We are running gosec as part of the golangci-lint suite on pull request and when pushing to main so a nightly check is no longer necessary.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -21,13 +21,3 @@ jobs:
         run: go install golang.org/x/vuln/cmd/govulncheck@latest
       - name: Execute govulncheck
         run: govulncheck ./...
-  gosec:
-    name: GoSec security scan
-    runs-on: ubuntu-latest
-    steps:
-      - name: Check out code
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
-      - name: Scan code
-        uses: securego/gosec@b7b2c7b668f3f2bef8a8ae04d72f0eb60492322c # master
-        with:
-          args: ./...


jraddaoui (commented):
This version would be more up-to-date than the one from golangci-lint, right? We tend to upgrade golangci-lint only when we bump Go versions and it's needed.

I think it doesn't hurt to keep this. If you are worried about so many PRs to update it, maybe we could do that only once a week instead of daily ([here](https://github.com/artefactual-sdps/enduro/blob/main/.github/dependabot.yml#L6)); or follow a similar pattern to the one from govulncheck above, using `go install` instead of the action.

djjuhasz:
@jraddaoui The reason I was looking at the gosec nightly action is because I'm now getting an email when the nightly workflow fails, and it has been consistently failing every night.  In fact it looks like the last time the nightly workflow ran successfully on this repository was **over a year ago: Feb 25, 2025**!  I spent several hours today trying to figure out why the gosec action is failing right now, and I believe it's because the VM it's running in has installed Go 1.26.0 and we are now requiring 1.26.1, so the error report is a false positive due to the Go version in the test runner.

Put all of this together and I have many reasons to want to remove the nightly gosec scan:

1. **Visibility:** The report goes to one, and only one, email box and it's easy to ignore. It also means the burden of maintenance rests on that one person.
2. **Responsibility:** The report is not seen by the person who made the change that raised the warning, so it obfuscates the responsibility for writing secure code by denying immediate feedback.
3. **Reliability:** This particular error report is not due to any actual security issue in our code, it's due to the test deployment environment.  There lots of reasons that this test can report false positives, and it is clearly hard to maintain based on the fact that it hasn't be working for a year.
4. **Testability:** The tests run in a Github specific deployment environment making it impossible (or at least very hard) to test the actions effectively without running the GH workflow. Testing the GH workflow requires pushing changes to our main branch, and then running the action, and repeating until the problem is fixed.  This process is slow, frustrating and creates a lot of noise in the project commit history and for those of us that receive an email every time a workflow fails.

I think if it's important to test our code regularly with the latest version of gosec then we need to find a way to do it in a way that is easier to maintain and test and gives more immediate feedback to more people.

jraddaoui (approved):

PR #1552, opened 2026-03-13
Updated broken links and fixed line lengths

sallain:


README.md
```diff
@@ -8,25 +8,27 @@
...
 
 ## User documentation resources
 
-- [User manual](https://enduro.readthedocs.io/user-manual/)
+- [User manual](https://enduro.readthedocs.io/user-manual/overview/)
 
 ## Development documentation resources
 
-- [Developer manual](https://enduro.readthedocs.io/dev-manual/)
+- [Developer manual](https://enduro.readthedocs.io/dev-manual/docs/)
```
djjuhasz: This links directly to the documentation for generating the Mkdocs docs in the development environment, which I don't think is what most developers will be interested in.   I think the previous link (https://enduro.readthedocs.io/dev-manual/) is fine, as it shows a table of contents for the docs.  Some other alternatives I'd consider:

- https://enduro.readthedocs.io/dev-manual/api/ - this the first "page" in the dev docs. The reader can use the left side bar to navigate to other pages.
- https://enduro.readthedocs.io/dev-manual/devel/ - this is the page that I go to most, and I think is most helpful for a developer that is looking to get a local development version of Enduro up and running.

djjuhasz (changes requested):
Thanks @sallain. I would prefer to link to a different page in the developer manual then the "documentation page".

README.md
sallain: Oh thanks! I didn't realize docs was for making docs :joy: I've reverted the previous link but also added the dev env link, as I think that seems like a thing many folks may look for in the README.

djjuhasz (approved):
LGTM @sallain. 👍  Please make sure to squash the two commits before merging to main.

PR #1553, opened 2026-03-13
Fix User Manual link in README.md

djjuhasz:
User documentation linked to a page that no longer exists. I have updated that to point to the "Overview and first steps" link instead.


djjuhasz:
Closing this because #1552 makes this fix, plus more.

PR #1554, opened 2026-03-13
Count SIP preservation files

djjuhasz:
Two clients have requested reports from Enduro that include a count of the preservation files in the ingested SIP.  Counting the SIPs in a BagIt bag is easy to standardize because the BagIt standard requires all preservation files to be put in the "data" directory, or one of its sub-directories.

The only time we don't have a bagged SIP in Enduro is when a3m is used as the preservation engine, which is not a configuration of Enduro that is supported at this time.

### DIFF
--- a/cmd/enduro-a3m-worker/main.go
+++ b/cmd/enduro-a3m-worker/main.go
@@ -269,6 +269,10 @@ func main() {
 			temporalsdk_activity.RegisterOptions{Name: bagvalidate.Name},
 		)
 		w.RegisterActivityWithOptions(
+			activities.NewCountSIPFilesActivity().Execute,
+			temporalsdk_activity.RegisterOptions{Name: activities.CountSIPFilesActivityName},
+		)
+		w.RegisterActivityWithOptions(
 			activities.NewBundleActivity().Execute,
 			temporalsdk_activity.RegisterOptions{Name: activities.BundleActivityName},
 		)
--- a/cmd/enduro-am-worker/main.go
+++ b/cmd/enduro-am-worker/main.go
@@ -276,6 +276,10 @@ func main() {
 			temporalsdk_activity.RegisterOptions{Name: bagvalidate.Name},
 		)
 		w.RegisterActivityWithOptions(
+			activities.NewCountSIPFilesActivity().Execute,
+			temporalsdk_activity.RegisterOptions{Name: activities.CountSIPFilesActivityName},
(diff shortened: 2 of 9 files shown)


jraddaoui (changes requested):
Thanks @djjuhasz!

Dropping a3m support has been a discussion for a long time, but we never got to a conclusion. Even if a3m is not maintained and we drop support for it at some point, it would be great to not make this dependent on the preservation system.

I know the bundle activity is a bit daunting, but in the end: if the SIP is a bag, it just unbags it in an specific way; if it's not a bag, it just copies the full dir; and if it's a single file, it bundles it into a directory.

So I think we could find a general solution that works for both preservation systems for now. I'd make the activity more general (e.g.: "count-files-activity") and set the path param based on the SIP type. If it's a Bag pass the data subdir, otherwise pass the full SIP path. And I'd call it before the branching for preservation.

Another big question is if we should be counting all the files inside data, not sure if you discussed that with the team, or if you considered excluding the metadata files. Other than that, the code changes look great!

djjuhasz:
@jraddaoui the problem with the counting the preservation objects in a SIP on the a3m path is that we don't know how the SIP is structured.  If the SIP is Bagit Bag, we check that in the current ingest workflow and this activity will count the files in the data directory.  If a single file is submitted as a SIP then the file count is trivial. But if a directory is submitted we have no idea how it is structured — is it a AM standard transfer, just a directory of preservation files, or some other structure (e.g. E-ARK)?  If we don't add more code to detect the type of SIP then the best we can do is count the total number of files in the SIP, which may include metadata files, manifests, etc.

Maybe the best course is to do the preservation object count in the preprocessing child workflow and return the count to Enduro for persistence.  We are more likely to know what kind of SIPs are being submitted in preprocessing because we customize it based on institutional processes.  This will give us more ability to filter out metadata files, manifests and other files that should not be considered preservation files.  This change would mean that a file count would only be done if the institution is running a preprocessing child workflow, but this has been the case for all of our clients to date.

djjuhasz:
@jraddaoui I've updated the "count files" activity to do the count based on whether the SIP is a BagIt Bag or not:

- If the SIP is a Bag, count all the files in the "data" directory
- If the SIP is *not* a Bag, count all the files in the SIP

jraddaoui (approved):
Neat, thanks @djjuhasz!

PR #1555, opened 2026-03-19
Migrate to go.artefactual.dev/ssclient

sevein:
We already use [amclient](https://pkg.go.dev/go.artefactual.dev/amclient), so it would be great to also adopt [ssclient](https://pkg.go.dev/go.artefactual.dev/ssclient).

Compared to amclient, ssclient is built on a different stack: it uses TypeSpec to generate an OpenAPI specification, Kiota for the generated low-level API client (handling transport and encoding), and ssclient-go as a higher-level wrapper that provides a better developer experience. Kiota introduces some runtime dependencies that this project would inherit. I'm not entirely sure how you'll feel about that, although it shouldn't be a big issue in practice.


internal/storage/activities/delete_amss.go
```diff
@@ -2,21 +2,48 @@ package activities
...
+// creating a new one. In that case the activity can continue with the polling
+// path, but it cannot auto-approve the request because AMSS does not return an
+// event ID for the existing request.
 type DeleteFromAMSSLocationActivity struct {
-	approve      bool
+	// The HTTP client to use for AMSS API calls.
+	httpClient *http.Client
+	// Whether to automatically approve deletion requests. If false, the
+	// activity requests deletion when possible, or reuses an existing pending
+	// request, and then polls the AIP status until processing completes.
```
jraddaoui: A bit repetitive, giving the comment above.

internal/storage/activities/delete_amss.go
```diff
@@ -43,224 +78,217 @@ func (a *DeleteFromAMSSLocationActivity) Execute(
...
-		params.AIPUUID.String(),
-	)
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	ssclient *ssclient_client.Client,
+	aipUUID uuid.UUID,
+) (uuid.UUID, error) {
+	pkg, err := ssclient.Packages().Get(ctx, aipUUID)
 	if err != nil {
-		return "", err
+		return uuid.Nil, fmt.Errorf("get pipeline UUID: %w", err)
```
jraddaoui: We are wrapping this error and others below now, is it intended?

internal/storage/ssblob/ssblob.go
```diff
@@ -56,44 +85,48 @@ func OpenBucket(opts *Options) (*blob.Bucket, error) {
...
 		}
 	}
-	switch err {
-	case errNotImplemented:
+	switch {
+	case errors.Is(err, errNotImplemented):
 		return gcerrors.Unimplemented
-	default:
-		return gcerrors.Unknown
 	}
```
jraddaoui: This could be a simple if.

jraddaoui (approved):
Nice, thanks @sevein! I like the new logic in the AMSS AIP deletion activity. I did a quick test locally and API download and deletion seems to be working properly. Feel free to ignore the comments below.

Issue #1384, opened 2025-10-06
Feature: add ability to search for AIP by UUID


sallain:
**Is your feature request related to a problem? Please describe.**

A client has described a scenario where they identify an AIP by finding it in their archival information system. In this system, the AIP's UUID is used as its primary identifier, rather than the AIP name. Therefore they would like to be able to search for the AIP in Enduro by entering the UUID.

**Describe the solution you'd like**

The AIP UUID is already displayed in the AIP browse table, so adding the ability to search by this value seems reasonable. I think we could simply add it to the current text search box, since a user is unlikely to want to search by both the name and UUID.

<img width="420" height="181" alt="Image" src="https://github.com/user-attachments/assets/06b840b9-670b-4d27-b89d-f04aabfeea65" />

**Describe alternatives you've considered**

None

**Additional context**

None.

fiver-watson:
AIP search works well! 

Works the same as name search - you can search based on a partial (i.e. searching for `9b72` would return both `9b72db64-dae9-444a-b2f1-d05925d0ec1e` and `733bd715-d708-42d0-9b72-4800339575b6` as results). 

Name searching also still works!

PR #1566, opened 2026-03-26
Allow filtering AIPs by name or UUID

jraddaoui:
Replace "name" by "query" in the list AIPs request payload and use this
field to filter AIPs by name or UUID.

Refs #1384.


sevein (approved):
LGTM

PR #1569, opened 2026-03-30
Update dashboard dependencies

sevein:
This pull request updates the majority of dependencies used in this project, including most available major-version upgrades. The main exceptions are TypeScript v6, ESLint v10, and @types/node v25.

TypeScript remains on v5 because the current typescript-eslint release does not yet support TypeScript 6. ESLint remains on v9 because eslint-plugin-import is not yet compatible with ESLint 10. @types/node remains on v24 by choice, since this project targets Node 24 and we prefer the Active LTS line over Node 25 Current.

Vite 8 is included in this update and brings a significant improvement in build performance. During installation, npm may emit peer-dependency warnings from transitive packages under vite-plugin-vue-devtools; these are expected with the current dependency graph and can be ignored for now.


djjuhasz (approved):
LGTM. Thanks @sevein. 👍

PR #1570, opened 2026-03-31
Pass SIP name to preprocessing workflow

djjuhasz:
Pass the original SIP file name (AKA "key") to the preprocessing child workflow. I need the original SIP name for a client preprocessing workflow but the name is modified during download to avoid name collisions.

### DIFF
--- a/internal/childwf/preprocessing.go
+++ b/internal/childwf/preprocessing.go
@@ -12,6 +12,9 @@ type PreprocessingParams struct {
 	// BatchID is the identifier of the batch being processed. If the SIP is not
 	// part of a batch, this will equal uuid.Nil.
 	BatchID uuid.UUID
+
+	// SIPName is the original filename of the SIP being processed.
+	SIPName string
 }
 
 type PreprocessingResult struct {
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -1039,6 +1039,7 @@ func (w *ProcessingWorkflow) preprocessing(ctx temporalsdk_workflow.Context, sta
 			RelativePath: relPath,
 			SIPID:        state.sip.uuid,
 			BatchID:      state.req.BatchUUID,
+			SIPName:      state.sip.name,
 		},
 	).Get(preCtx, &ppResult)
 	if err != nil {
--- a/internal/workflow/processing_test.go
+++ b/internal/workflow/processing_test.go
@@ -324,6 +324,7 @@ func (s *ProcessingWorkflowTestSuite) TestChildWorkflows() {
 		&childwf.PreprocessingParams{
 			RelativePath: strings.TrimPrefix(prepDownloadPath+"/"+key, prepSharedPath),
 			SIPID:        sipUUID,
+			SIPName:      sipName,
 		},
 	).Return(
 		&childwf.PreprocessingResult{
@@ -426,6 +427,7 @@ func (s *ProcessingWorkflowTestSuite) TestFailedSIP() {
 		&childwf.PreprocessingParams{
 			RelativePath: strings.TrimPrefix(prepDownloadPath+"/"+key, prepSharedPath),
 			SIPID:        sipUUID,
+			SIPName:      sipName,
 		},
 	).Return(
 		&childwf.PreprocessingResult{


jraddaoui (commented):
LGTM and it doesn't hurt even if it's not needed, but I thought we were avoiding collisions with the parent directory. Are we sure this is needed? Looking at the code:

Download from bucket uses the key:

- https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/download.go#L126-L129
- https://github.com/artefactual-sdps/temporal-activities/blob/main/bucketdownload/activity.go#L25-L26

Download from watcher does that too:

- https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/download.go#L86-L90
- https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/activities/download.go#L54

And also the extraction just places the content alongside the original file without any name change:

- https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/processing.go#L376
- https://github.com/artefactual-sdps/temporal-activities/blob/main/archiveextract/activity.go#L28-L31

What am I missing?

djjuhasz:
@jraddaoui the web upload path makes the SIP key unique by modifying the original name to "SIP_{original_name}-{UUID}.zip" here:
https://github.com/artefactual-sdps/enduro/blob/main/internal/ingest/upload.go#L70

Processing workflow parameters for an uploaded SIP:
```json
{
  "SIPUUID": "a9e5d222-93cf-49c0-9825-dcd66873c68e",
  "SIPName": "SIP_20220915_dept.zip",
  "Type": "create aip",
  "WatcherName": "",
  "SIPSourceID": "00000000-0000-0000-0000-000000000000",
  "RetentionPeriod": 0,
  "CompletedDir": "",
  "Key": "SIP_SIP_20220915_dept-a9e5d222-93cf-49c0-9825-dcd66873c68e.zip",
  "IsDir": false,
  "Extension": ".zip",
  "BatchUUID": "00000000-0000-0000-0000-000000000000"
}
```

So the processing workflow is started with the key: "SIP_SIP_20220915_dept-a9e5d222-93cf-49c0-9825-dcd66873c68e.zip" (original name: "SIP_20220915_dept.zip").

djjuhasz:
Processing workflow execution from a watched bucket sets the key equal to the SIP name given (this is using the Tilt "Minio Upload" button):

```json
  {
    "SIPUUID": "19261363-c140-4599-bf56-90e741d19f73",
    "SIPName": "DavidJ_Test_SIP",
    "Type": "create aip",
    "WatcherName": "dev-minio",
    "SIPSourceID": "00000000-0000-0000-0000-000000000000",
    "RetentionPeriod": 0,
    "CompletedDir": "",
    "Key": "DavidJ_Test_SIP",
    "IsDir": false,
    "Extension": "",
    "BatchUUID": "00000000-0000-0000-0000-000000000000"
  }
```

The original filename was also "SIP_20220915_dept.zip" in this case, but I guess the given "Object name" is used for the "SIPName" and "Key" anyway?

djjuhasz:
Starting ingest from the SIPSource uses the object key as the "SIPName" and "Key" without modification when starting the processing workflow:

```json
{
  "SIPUUID": "52a49e27-87a5-48a7-b877-9777258f7474",
  "SIPName": "SIP_20220915_dept.zip",
  "Type": "create aip",
  "WatcherName": "",
  "SIPSourceID": "e6ddb29a-66d1-480e-82eb-fcfef1c825c5",
  "RetentionPeriod": 0,
  "CompletedDir": "",
  "Key": "SIP_20220915_dept.zip",
  "IsDir": false,
  "Extension": "",
  "BatchUUID": "00000000-0000-0000-0000-000000000000"
}
```

As you've pointed out, the SIP is passed to preprocessing with an unique dir and the unchanged object key in this case.

Preprocessing workflow params:

```json
{
  "RelativePath": "52a49e27-87a5-48a7-b877-9777258f7474/SIP_20220915_dept.zip",
  "SIPID": "52a49e27-87a5-48a7-b877-9777258f7474",
  "BatchID": "00000000-0000-0000-0000-000000000000"
}
```

jraddaoui:
I see, thanks @djjuhasz!

jraddaoui (approved):

PR #1578, opened 2026-04-08
Add Tilt resource dependencies

jraddaoui:
Avoids some resource restarts that would increase the exponential
backoff, making cluster initialization faster.

---

I added an initial commit that is just cosmetic to be able to compare Tilt CI times here, I'll squash before merging.

- [Without resource deps](https://github.com/artefactual-sdps/enduro/actions/runs/24149365566/job/70471789334): 6m 33s
- [With resource deps](https://github.com/artefactual-sdps/enduro/actions/runs/24149780528/job/70473279299): 4m 54s

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -111,7 +111,8 @@ k8s_resource(
   "enduro",
   labels=["Enduro"],
   port_forwards=["9000:9000", "9002:9002"],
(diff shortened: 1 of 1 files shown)


djjuhasz (approved):
Nice! Thanks @jraddaoui. 👍

PR #1579, opened 2026-04-08
Set Tilt dashboard target based on DASHBOARD_DEV

jraddaoui:
Set the Tilt dashboard image target to `builder` only when `DASHBOARD_DEV` is truthy, and to the default target otherwise. This makes the choice explicit through `.tilt.env` instead of requiring manual edits to the Tiltfile.


djjuhasz (approved):
Thanks @jraddaoui. 👍

PR #1585, opened 2026-04-08
Upgrade go-jose from v3.0.4 to v4.1.4

djjuhasz:
github.com/go-jose/go-jose/v4 is the current stable branch, v3 is in
maintenance mode and only receives critical security updates. Rather
than upgrade to v3.0.5 to address CVE-2026-34986 jump directly to
v4.1.4.

- Switch from github.com/go-jose/go-jose/v3 to
  github.com/go-jose/go-jose/v4
- Upgrade chainguard.dev/go-oidctest v0.3.1 to v0.4.0 to break it's
  dependency on go-jose v3


jraddaoui (approved):
Thanks @djjuhasz!

Issue #1382, opened 2025-10-06
Problem: Azure support is not available for SIP source buckets


jraddaoui:
**Is your feature request related to a problem? Please describe.**

In https://github.com/artefactual-sdps/enduro/issues/1231 we added support for Azure buckets, but only to the internal storage. In https://github.com/artefactual-sdps/enduro/issues/1291, we added SIP source buckets, but they don't support Azure as the internal storage.

**Describe the solution you'd like**

Ideally, all configurable buckets should support Azure. We can ignore watched buckets for now as they will require a different notification process apart from the configuration.

**Describe alternatives you've considered**

None, leave as is.

**Additional context**

We have a bucket package in https://github.com/artefactual-labs/gotools that allows configuration and creating the bucket, it would be nice to add Azure support through configuration there. The bucket package could work with environment variables and an Azure specific URL but we want a method to do it through configuration as it may be more secure that using the environment in some cases (encrypted config file vs environment vars when considering a rogue admin user).

jraddaoui:
After https://github.com/artefactual-labs/gotools/pull/35 and https://github.com/artefactual-sdps/enduro/pull/1589, Azure support is now available for the ingest and storage internal buckets and for the SIP source. Remaining buckets that may need Azure support in the future: watched buckets in the ingest domain (requires a notification mechanism) and AIP store locations in the storage side (requires a new type of location config).

jraddaoui:
The documentation has been updated too:

- https://enduro.readthedocs.io/admin-manual/configuration/#internal-location-used-for-storing-staging-aips-and-reports
- https://enduro.readthedocs.io/admin-manual/configuration/#internal-storage-configuration
- https://enduro.readthedocs.io/admin-manual/configuration/#sip-source-location-bucket

PR #1589, opened 2026-04-10
Allow Azure for internal storage and SIP source

jraddaoui:
Remove the Azure integration done for the internal storage bucket. Use the latest gotools/bucket package, which includes the same integration with the addition of client secret auth. Use gotools/bucket for the SIP source and the storage internal bucket too.

Refs #1382. Related to https://github.com/artefactual-labs/gotools/pull/35.


docs/src/admin-manual/configuration.md
```diff
@@ -1040,28 +1104,21 @@ bucket = "internal"
...
 
-If you intend to use Azure for your internal upload space, be sure to remove or
-comment out the S3-like object store parameters in the section
-[above](#internal-storage-configuration---s3-like-bucket-or-filesystem), and
-then add the Azure blob URL connection information there instead. For example,
-if your target Azure storage blob is named "sips", your configuration of the
-section above might look like this when configured properly:
+If you intend to use Azure for your internal upload space, set an Azure Blob
+Storage URL in the section
+[above](#internal-storage-configuration---s3-like-bucket-or-filesystem). For
```
djjuhasz: I think linking to the S3 bucket configuration section here is confusing, I read it several times before I understood the intention. I'd say something like:

---

If you intend to use Azure for your internal upload space, set an Azure bucket `url` using the `azblob://` schema in the `[internalStorage]` section, and remove all other configurations in that section.  Note that the Azure bucket name is included in the URL (the bucket name in the example below is "enduro-ingest") so there is no need for a separate `bucket` config value.

Then add an `[internalStorage.azure]` section to configure Azure authentication via shared key or client secret.

Example Azure shared key configuration:

```toml
[internalStorage]
url = "azblob://enduro-ingest"

[internalStorage.azure]
storageAccount = "my-account"
storageKey = "dGVzdA==" # Must be base64 encoded
```

Example Azure client secret configuration:

```toml
[internalStorage]
url = "azblob://enduro-ingest"

[internalStorage.azure]
tenantID = "my-tenant-id"
clientID = "my-client-id"
clientSecret = "my-secret"
```

docs/src/admin-manual/configuration.md
```diff
@@ -1154,6 +1236,49 @@ bucket = "sipsource"
...
+
+```toml
+[sipsource.bucket]
+url = "azblob://sipsource"
+```
+
+The following subsection provides Enduro with the connection information needed
+to access packages in the target Azure blob store.
+
+**Default values**:
```
djjuhasz: I think it's more helpful to provide example configurations here, rather than the empty default configuration, and to help illustrate that the shared key and client secret configs are exclusive.

djjuhasz (approved):
Looks great @jraddaoui! I'm very happy to see the bucket configurations standardized so I never mix up the "key" and "secret" vs. "accessKey" and "secretKey" configs again. 😆 

I just made a couple of suggestions for the docs, but nothing critical.

docs/src/admin-manual/configuration.md
```diff
@@ -684,15 +695,59 @@ bucket = "aips"
...
+url = "azblob://enduro-storage"
+```
+
+The following subsection provides Enduro with the connection information needed
+to access the target Azure blob store.
+
+**Default values**:
+
+```toml
+[storage.internal.azure]
```
djjuhasz, line 736: I think it's more helpful to provide example configurations here, rather than the empty default configuration, and to help illustrate that the shared key and client secret configs are exclusive.

docs/src/admin-manual/configuration.md
jraddaoui: Thanks @djjuhasz! I updated these sections pretty much as it is here, with one exception. I have included the `storageAccount` in both auth examples, as it's not used only for auth and can be set for both methods.

PR #1593, opened 2026-04-14
Render Temporal k8s manifest from Helm chart

jraddaoui:
Update the Temporal manifest from the final 1.0.0 Temporal Helm chart and use the latest server, admin-tools, and UI images.

Replaces https://github.com/artefactual-sdps/enduro/pull/1565.

---

I prefer this implementation over the one in https://github.com/artefactual-sdps/enduro/pull/1565. The previous approach keeps Temporal behind `helm_resource`, pulls the chart at `tilt up` time, relies on a Python script to patch the schema job, and needs a separate `local_resource` just to keep the UI port forward working. That makes the dev path more fragile: it depends on Helm and Python on every machine, hides the actual Temporal workloads behind one aggregated Tilt resource, and pushes important behavior into render patching rather than manifests.

This version checks in the rendered Temporal manifest and loads it through `kustomize` instead. That gives us deterministic manifests in the repo, removes the Helm/Python runtime requirement from `tilt up`, exposes Temporal services as normal Tilt `k8s_resource`, and lets the UI port forward attach directly to `temporal-web`. The schema `MYSQL_*` compatibility patch is explicit in the manifest, documented, and easy to review when the chart is rerendered.

In short, this approach trades dynamic Helm integration for a simpler and more predictable dev environment, with better Tilt behavior and less hidden logic.


hack/kube/base/mysql.yaml
```diff
@@ -37,6 +37,38 @@ spec:
...
+            exec:
+              command:
+                - sh
+                - -c
+                - >-
+                  mysqladmin ping -h 127.0.0.1 -uroot
+                  -p"$MYSQL_ROOT_PASSWORD"
+            periodSeconds: 10
+            timeoutSeconds: 3
+            failureThreshold: 6
```
jraddaoui, line 71: Explanation for the different values from gpt-5.4:

  startupProbe

  - Asks: has MySQL finished booting yet?
  - It should be the most tolerant one.
  - Here it allows up to about 2 minutes: periodSeconds: 5, failureThreshold: 24.
  - While startup is still failing, Kubernetes does not run liveness/readiness checks.

  readinessProbe

  - Asks: should traffic/dependent workloads use this pod right now?
  - It should react relatively quickly once MySQL becomes usable.
  - Here it checks every 5 seconds with a shorter failure window.

  livenessProbe

  - Asks: is the process stuck badly enough that Kubernetes should restart it?
  - It should be less eager than readiness, because a brief hiccup should make the pod
    unready, not dead.
  - Here it checks less often, every 10 seconds.

  So the intent is:

  - startup: patient during initialization
  - readiness: quick signal for dependents
  - liveness: conservative restart trigger

  Using the same values for all three is usually a sign the probes are not modeling the
  service lifecycle properly.

jraddaoui:
Sorry for the noise, finally ready for review.

hack/helm/temporal/README.md
```diff
@@ -0,0 +1,33 @@
+# Temporal Helm chart rendering
+
+Temporal is checked into the repo as a rendered Kubernetes manifest at
```
djjuhasz: I'm not sure what you are trying to say here, "Temporal is checked into the repo" doesn't make sense to me.  Maybe something like this is clearer:

> The rendered Temporal Kubernetes manifest is included at [hack/kube/base/temporal.yaml](../../kube/base/temporal.yaml) and loaded by kustomize during `tilt up`.

hack/kube/base/temporal.yaml
djjuhasz, line 1: Wow, +1123 lines. 😮 🐳

Tiltfile
```diff
@@ -131,16 +136,53 @@ if PRES_SYS == 'am':
...
     "enduro-a3m",
     labels=["Enduro"],
     trigger_mode=trigger_mode,
-    resource_deps=["temporal"],
+    resource_deps=["temporal-schema-1-0-0-1"],
   )
 
+# Temporal resources
+k8s_resource(
+  "temporal-schema-1-0-0-1",
```
djjuhasz, line 151: Do we need to encode the schema version (I assume 1.0.0.1 is the version) in the resource name? This seem like it will updates harder because we'll have to update the name wherever it's used.

djjuhasz (approved):
LGTM @jraddaoui. I did have one question about the docs, and one about the name of the "temporal-schema-1-0-0-1" resource.

hack/helm/temporal/README.md
jraddaoui: A lot better, thanks!

Tiltfile
jraddaoui: Yeah, sadly that's how the job is created by the chart, and I want to avoid as much manual edits to the manifest as possible.

Issue #1595, opened 2026-04-16
Feature: Count the files in a SIP


djjuhasz:
The number of files in a SIP is a useful and easily understandable metric that can help understand the composition and size of a SIP. The file count can also be a simple check after preservation that the number of files in the AIP is equal to (or greater than, if derivatives are made) the number of files in the SIP.

**Describe the solution you'd like**

Count the number of files in a SIP. The file count can be passed to child workflows for reporting, and should also be added to the SIP details page of the Dashboard.

**Describe alternatives you've considered**

We could only count the SIP files in a preprocessing child workflow and return the number to the parent ingest workflow.  

Pros: 
- allows client customization of how the files are counted 
- avoids the computation cost of counting files when the value is not relevant

Cons:
- if preprocessing is not run or doesn't implement a file counter, then we can't show a file count in UI

**Additional context**

- Two clients have requested reporting on the number of files in a SIP
- Deciding which files to count in a SIP is difficult, and the criteria may be different for different institutions. For instance one institution may want to count metadata files, whereas another may only want to count preservation objects.  It may be beneficial to allow a preprocessing child workflow to return a file count that would override a general count in the ingest workflow.

sallain:
Based on user feedback, we have decided to change the default behaviour of this activity to exclude all metadata directories and their contents. Any directory called `metadata` should be ignored when counting files.

PR #1596, opened 2026-04-16
Add `file_count` column to the SIP table

djjuhasz:
Refs #1595.

- Add a `sip.file_count` column to the db schema
- Add a database migration to add the `sip.file_count` column
- Add a `FileCount` field to `datatypes.SIP`
- Add the "file count" field to SIP create, update, and read methods

### DIFF
--- a/internal/datatypes/sip.go
+++ b/internal/datatypes/sip.go
@@ -38,6 +38,9 @@ type SIP struct {
 
 	// Batch is the batch this SIP belongs to.
 	Batch *Batch
+
+	// FileCount is the number of files in the SIP.
+	FileCount int32
 }
 
 // Goa returns the API representation of the SIP.
--- a/internal/persistence/ent/schema/sip.go
+++ b/internal/persistence/ent/schema/sip.go
@@ -61,6 +61,9 @@ func (SIP) Fields() []ent.Field {
 		field.Int("batch_id").
 			Optional().
 			Positive(),
+		field.Int32("file_count").
+			Optional().
+			NonNegative(),
 	}
 }
 
--- a/internal/db/migrations/20260417140248_add_sip_file_count_column.up.sql
+++ b/internal/db/migrations/20260417140248_add_sip_file_count_column.up.sql
@@ -1,0 +1,2 @@
+-- Modify "sip" table
+ALTER TABLE `sip` ADD COLUMN `file_count` int NULL;
--- a/internal/db/migrations/atlas.sum
+++ b/internal/db/migrations/atlas.sum
@@ -1,4 +1,4 @@
-h1:+w0cFlo2J6po1A+Xh2d1s3AOIbBR68sVVU6MdKrVrX0=
+h1:SCNNo6ElQOyVdaaaPahyx3vTXaqfIPmBtHTqGdL0bGE=
 1570659451_init.up.sql h1:zyiKKl39RqMxuEhop5jeeiPTxPiSSq00Tn6u06gyNmk=
 1710442322_nullable_aip_id.up.sql h1:vL4eG5YELXr3k4ymhHuRD/R7KpNt3/DNRhH26t83x3A=
 20250207193001_rename_package_table.up.sql h1:d2RjfIturPoFYMMtFocrMvvjEXEqDXDdxQRttcknX/0=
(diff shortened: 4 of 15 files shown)


internal/db/migrations/20260416214859_add_sip_file_count_column.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Modify "sip" table
+ALTER TABLE `sip` ADD COLUMN `file_count` bigint NULL DEFAULT 0;
```
jraddaoui: BIGINT? 😮 https://dev.mysql.com/doc/refman/8.4/en/integer-types.html

internal/persistence/ent/schema/sip.go
```diff
@@ -61,6 +61,9 @@ func (SIP) Fields() []ent.Field {
 		field.Int("batch_id").
 			Optional().
 			Positive(),
+		field.Int("file_count").
+			Optional().
+			Default(0),
```
jraddaoui: I hate that `Int()` defaults to BIGINT. I made the same mistake in the SIPs count from the batch table. I don't know which variant creates an INT in the SQL, maybe `Int32()`.

Also, having a default, I'd consider removing `Optional()`. And I'd add `Possitive()` to make it unsigned and increase the max value of a normal INT column.

internal/datatypes/sip.go
```diff
@@ -38,6 +38,9 @@ type SIP struct {
 
 	// Batch is the batch this SIP belongs to.
 	Batch *Batch
+
+	// FileCount is the number of files in the SIP.
+	FileCount int
```
jraddaoui: A BIGINT may not fit into an int, right? I made the same mistake in the Batch SIPs count field.

jraddaoui (approved):
Thanks @djjuhasz! We use BIGINT in other columns and I don't know the performance impact, but we may want to reduce this column size. Up to you, if you decide to change it, it may be a good idea to update the SIPs count in the batch table too.

internal/persistence/ent/schema/sip.go
djjuhasz: Yes, I agree that mapping `Int()` to `BIGINT` is unexpected and bad.  It looks like `Int32()` does map to a MySQL `INT` type (See: https://github.com/ent/ent/blob/master/dialect/sql/schema/mysql.go#L147).

I chose `Optional()` because it is described as being "optional on create" (https://github.com/ent/ent/blob/master/schema/field/numeric.go#L213) — I don't want to require providing a `FileCount` when creating a new SIP because we don't know the count at creation.

I don't think `Positive()` does what you are intending, I think it's an ent validator that forces the value to be 1 or greater: https://github.com/ent/ent/blob/master/schema/field/numeric.go#L160.  If we want a  validator that checks the value is zero or greater we can use [`NonNegative()`](https://github.com/ent/ent/blob/master/schema/field/numeric.go#L170).  If we want to make the field type "INT UNSIGNED" then we can define the ent field as `UInt32()`.  I'd prefer to use `NonNegative()` because the Go crowd seems to discourage using unsigned integer types except in very specific cases and the MySQL `UNSIGNED` attribute is non-standard SQL.

internal/db/migrations/20260416214859_add_sip_file_count_column.up.sql
djjuhasz: SO BIGINT! BIGINT IS BIG! :whale: :stuck_out_tongue_winking_eye:

internal/datatypes/sip.go
djjuhasz: True. :(  I've updated this to int32 now, to match the ent `Int32()` type.

internal/persistence/ent/schema/sip.go
djjuhasz: I guess we probably shouldn't blame ent for choosing `Int()` -> `BIGINT` as Go's `int` may be `int32` or `int64` depending on the platform. Mapping `Int()` to `BIGINT` guarantees that an `int64` value won't overflow when it's saved to MySQL. :thinking:

djjuhasz: Okay, after a few iterations I've settled on:

```go
field.Int32("file_count").
	Optional().
	NonNegative(),
```

Which translates to a MySQL definition of:

```SQL
ALTER TABLE `sip` ADD COLUMN `file_count` int NULL;
```

I think it makes sense to allow `file_count` to be NULL in the database when the file count is undetermined.  On the other hand, I don't think it's worth defining the `datatypes.SIP` representation as `FileCount *int32` to support `nil` values because I don't think it's worth the extra hassle of dealing with a pointer in the Go code when a zero value is good enough to represent an unknown number of files.

PR #1598, opened 2026-04-17
Persist SIP file count

djjuhasz:
Refs #1595.

- Change all file count variables to int32 to avoid conversion issues
- Add `FileCount` to `updateSIPLocalActivity()` parameters
- Make all `updateSIPLocalActivity()` parameters optional, except for the SIP UUID
- Call `updateSIPLocalActivity()` to persist the file count

I've kept the call to persist the file count separate from the `countSIPfiles()` function to make it easier to use a file count passed from preprocessing in the future.


jraddaoui (approved):
LGTM!

PR #1599, opened 2026-04-17
Add SIP file_count to ingest API

djjuhasz:
Refs #1595.

- Add "file_count" to Goa ingest SIP schema
- Generate updated Goa and dashboard code to add "file_count"
- Pass "file_count" from persistence layer to Goa API layer


jraddaoui (approved):
👌

PR #1601, opened 2026-04-20
Allow Enduro pods to load config from a K8s secret

jraddaoui:
Mount the `enduro-config` secret into Enduro pods so cluster manifests can supply runtime configuration, and have Tilt create that secret to support custom local configuration.

### DIFF
--- a/hack/kube/base/enduro.yaml
+++ b/hack/kube/base/enduro.yaml
@@ -27,6 +27,11 @@ spec:
       containers:
         - name: enduro
           image: ghcr.io/artefactual-sdps/enduro:main
+          command:
+            - /home/enduro/bin/enduro
+          args:
+            - --config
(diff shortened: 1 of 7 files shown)


hack/kube/base/enduro.yaml
```diff
@@ -73,7 +78,15 @@ spec:
           ports:
             - containerPort: 9000
             - containerPort: 9002
+          volumeMounts:
+            - name: enduro-config
+              mountPath: /home/enduro/enduro-config
+              readOnly: true
```
jraddaoui, line 84: Secret volumes are always mounted as read only. This required to use a different config path, to avoid issues with PDFCPU and other tools that may write to the default config dir (`$HOME/.config`).

djjuhasz (approved):
Cool. Thanks @jraddaoui. 👍

PR #1602, opened 2026-04-21
Pass SIP file counts to postbatch workflow

djjuhasz:
Refs #1595.

- Get SIP file_count when polling SIP statuses in the batch workflow
- Update the SIP data in the `batch.state` with the DB data return from the `PollSIPStatusesActivity`
- Pass the SIP file counts to the postbatch workflow

### DIFF
--- a/internal/workflow/activities/poll_sip_statuses.go
+++ b/internal/workflow/activities/poll_sip_statuses.go
@@ -11,6 +11,7 @@ import (
 	temporal_tools "go.artefactual.dev/tools/temporal"
 
 	goaingest "github.com/artefactual-sdps/enduro/internal/api/gen/ingest"
+	"github.com/artefactual-sdps/enduro/internal/datatypes"
 	"github.com/artefactual-sdps/enduro/internal/entfilter"
 	"github.com/artefactual-sdps/enduro/internal/enums"
 	"github.com/artefactual-sdps/enduro/internal/ingest"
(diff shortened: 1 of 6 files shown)


internal/workflow/activities/poll_sip_statuses.go
```diff
@@ -35,8 +36,14 @@ type PollSIPStatusesActivityParams struct {
...
 type PollSIPStatusesActivityResult struct {
+	// AllExpectedStatus is true if all SIPs in the batch have the
+	// ExpectedStatus.
 	AllExpectedStatus bool
-	SIPIDstoAIPIDs    map[uuid.UUID]uuid.UUID
+
+	// SIPs is a sparse list of the SIPs in this batch, containing only the
+	// AIPID and FileCount fields, so they can be passed to the postbatch
+	// workflow without needing to query the ingest service again.
+	SIPs []datatypes.SIP
```
jraddaoui: Should we still return a map (`map[uuid.UUID]datatypes.SIP`) to speed up the state update?

internal/workflow/batch_state.go
```diff
@@ -96,15 +95,16 @@ func (s *batchWorkflowState) postbatchSIPs() []*childwf.PostbatchSIP {
...
-		aipID, ok := aipIDs[sd.sip.UUID]
-		if !ok {
-			continue
+		for _, sip := range sips {
+			if sd.sip.UUID == sip.UUID {
+				sd.sip.FileCount = sip.FileCount
+				sd.sip.AIPID = sip.AIPID
+				break
+			}
 		}
```
jraddaoui: If you decide to not return a map from the activity, I'd still build that map here, looping over the `sips` once instead of multiple times for each SIP detail.

jraddaoui (changes requested):
Thanks @djjuhasz! We could keep using a map or build one to speed up the state update.

internal/workflow/activities/poll_sip_statuses.go
djjuhasz: Yeah, good idea. :+1:

jraddaoui (approved):
Thanks @djjuhasz!

Issue #1536, opened 2026-03-09
Problem: child workflows cannot surface human intervention decision points through Enduro


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Enduro supports extending some base workflows with custom child workflows, but there is currently no general way for a child workflow to communicate to the parent workflow that it has reached a decision point requiring human intervention. In practice, this means a custom child workflow may need to pause and wait for a user decision, but Enduro has no standard way to surface that state, present the available options, and send the selected decision back so the child workflow can continue.

For the first implementation, this is needed for the `preprocessing` child workflow. However, the same problem can also apply to other child workflows in the future, such as `poststorage` or `postbatch`.

**Describe the solution you'd like**

Add a generic mechanism for child-to-parent and parent-to-child communication using Temporal signals.

When a custom child workflow reaches a point where it requires human intervention, it should signal Enduro's parent workflow with a structured payload describing the decision point. The parent workflow should store that state and expose it through the existing dashboard/API so the user can see that the workflow is waiting for intervention and which options are available. After the user submits a decision, the parent workflow should signal the child workflow back with the selected option so it can resume.

This could be implemented incrementally. In the first iteration, the child workflow only needs to send a simple message describing the decision point plus a list of available decision options. That would allow Enduro to render a generic pending alert and present the available options.

To let Enduro expose this state and receive the user's decision through the dashboard/API, there are two main options:

1. Store the human intervention state in Enduro's database and have the dashboard/API read from there, while the API writes the selected decision back into the parent workflow flow that will signal the child.
2. Use Temporal queries to expose the current waiting state from the parent workflow, and Temporal updates or signals to submit the selected decision back into the parent workflow.

Either approach could work for the first iteration. The important part is to establish the generic signal contract between child and parent, and between parent and child, so the communication path is standardized. Since child workflows are executed while the parent waits for their result, we do not expect multiple concurrent human intervention requests from the same child execution in this first iteration.

**Describe alternatives you've considered**

An alternative approach would be to rely on the customization support being added to the dashboard in this [PoC PR](https://github.com/artefactual-sdps/enduro/pull/1398).

[...]

PR #1608, opened 2026-04-24
Support preprocessing workflow decision requests

jraddaoui:
Handle decision requests from the preprocessing child workflow while the processing workflow waits for it to complete. Add decision signal names and payload types, record a pending decision task, update SIP and workflow statuses, and forward the selected response back to preprocessing.

Refs #1536.

### DIFF
--- a/internal/childwf/signals.go
+++ b/internal/childwf/signals.go
@@ -1,0 +1,24 @@
+package childwf
+
+const (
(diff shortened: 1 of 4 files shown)


djjuhasz (approved):
LGTM! Thanks @jraddaoui.

PR #1612, opened 2026-04-27
Move dashboard specs next to source files

sevein:
Move dashboard specs next to their source files to match the convention already used by the Go side of the project, where tests live beside the code they exercise. This makes ownership clearer and keeps behavior and verification visible in the same directory.

I've also added dashboard coverage reporting to Codecov by running Vitest with coverage in the dashboard workflow, uploading the generated dashboard/coverage/clover.xml report with a dedicated dashboard flag, and updating Codecov configuration so backend and dashboard coverage are tracked separately while still showing both in PR comments.


jraddaoui (approved):
This is great, thanks @sevein!

djjuhasz:
@sevein why, `.spec.ts` instead of `.test.ts`? This naming convention seems less intuitive to me. Googling `.spec.ts` it sounds like an Angular naming convention?

sevein:
> @sevein why, .spec.ts instead of .test.ts? This naming convention seems less intuitive to me. Googling .spec.ts it sounds like an Angular naming convention?

Yeah, I've seen both. I don't have a strong opinion. I think I picked .spec.ts because I recently looked at some [nuxt/ui](https://github.com/nuxt/ui/tree/v4/test/components) stuff, and that project seems to use `*.spec.ts`. It should be easy to change!

PR #1618, opened 2026-04-29
Handle SIP child workflow decisions

jraddaoui:
Add Temporal query/update handling for child workflow decision requests, expose SIP decision GET/POST endpoints, and show pending decisions in the dashboard. The dashboard now renders the decision message and options, submits the selected response through the parent workflow update, refreshes decision state from workflow status updates, and surfaces submission errors.

Refs #1536. See individual commits for more details.


jraddaoui:
Some notes I was going to leave as review comments, but I think they are better here in case you review this commit by commit:

- I decided to expose the decision request and receive the response through workflow state, using Temporal queries and updates. We can treat this as a first increment and move the decision state to the database later. But I would like to try this approach first because it keeps custom decision requests isolated from the persistence model. I expect it to scale well while decisions are queried only for the current SIP view, but we may want to move the state into the database later if we need direct events, or query multiple decisions for browsing/reporting.
- We already have a "review" access control attribute, but I think this is a different action. I added "decision" for both the GET and POST endpoints, and named the path "/decision". Happy to hear other naming suggestions.
- The API design still returns 409 in some cases where I think it should return 500. This PR adds a helper reused by existing endpoints, and I did not want to change their API behavior here. I think we should revisit this more broadly in a separate PR.
- Because we are not using the persistence layer to store these decisions, there are no websocket events for them. The dashboard watches workflow status changes and fetches or clears the decision when the status changes.

internal/workflow/processing.go
```diff
@@ -60,6 +60,50 @@ func NewProcessingWorkflow(
...
+	if err := temporalsdk_workflow.SetQueryHandler(
+		ctx,
+		ingest.ChildDecisionQueryName,
+		func() (ingest.ChildDecisionQueryResult, error) {
+			if state.childDecisionRequest == nil {
+				return ingest.ChildDecisionQueryResult{}, nil
+			}
+
+			req := *state.childDecisionRequest
+			req.Options = slices.Clone(req.Options)
```
djjuhasz: Why are you cloning `req.Options` then reassigning it?

internal/workflow/processing.go
```diff
@@ -1264,19 +1311,18 @@ func (w *ProcessingWorkflow) updateStatuses(
...
-	if !slices.Contains(req.Options, decision.Option) {
-		return childwf.DecisionResponse{}, fmt.Errorf("invalid child workflow decision option %q", decision.Option)
+	if err := temporalsdk_workflow.Await(ctx, func() bool {
+		return state.childDecisionResponse != nil
+	}); err != nil {
+		return childwf.DecisionResponse{}, err
 	}
-	return decision, nil
+
+	return *state.childDecisionResponse, nil
```
djjuhasz: I don't see any advantage to returning the `childDecisionResponse` here, it's already been updated in `state` and you can get it directly in the caller.

internal/ingest/workflows.go
```diff
@@ -56,6 +63,14 @@ type (
 		RetentionPeriod time.Duration
 	}
 
+	ChildDecisionQueryResult struct {
+		// Pending indicates whether a child workflow decision request is active.
+		Pending bool
```
djjuhasz: Is Pending being used? It doesn't seem to be checked anywhere. It seems like checking if `Request != child.DecisionRequest{}` should be enough to check whether there is an active decision request.

internal/workflow/processing.go
```diff
@@ -60,6 +60,50 @@ func NewProcessingWorkflow(
...
+				return errors.New("no child workflow decision request is pending")
+			}
+			if state.childDecisionResponse != nil {
+				return errors.New("child workflow decision response already submitted")
+			}
+			if !slices.Contains(state.childDecisionRequest.Options, decision.Option) {
+				return fmt.Errorf("invalid child workflow decision option %q", decision.Option)
+			}
+
+			state.childDecisionResponse = &decision
```
djjuhasz, line 95: Doesn't this mean that state needs a mutex? looks like setting the state will race if state is being mutated simultaneously by another update or in the workflow.

internal/workflow/processing.go
```diff
@@ -1187,6 +1234,15 @@ func (w *ProcessingWorkflow) handleChildDecisionRequest(
 		return err
 	}
 
+	request := req
+	request.Options = slices.Clone(req.Options)
+	state.childDecisionRequest = &request
```
djjuhasz: This also seems racy.

internal/workflow/processing.go
djjuhasz, line 1215: I think this also needs to check that `state.childDecisionRequest` and `state.childDecisionResponse` are nil, otherwise a subsequent child decision request will overwrite the preceding one, and could lead to some weird race conditions.

djjuhasz, line 1247: I can't see anywhere that `e` is set except at the end of this deferred function, so I don't think this can ever evaluate true.

docs/src/admin-manual/iac.md
```diff
@@ -241,6 +241,8 @@ endpoints.
 | POST   | /ingest/sips/upload                   | `ingest:sips:upload`             |
 | GET    | /ingest/sips/{uuid}                   | `ingest:sips:read`               |
 | POST   | /ingest/sips/{uuid}/confirm           | `ingest:sips:review`             |
+| GET    | /ingest/sips/{uuid}/decision          | `ingest:sips:decision`           |
```
djjuhasz, line 244: I'm inclined to just require `ingest:sip:review` or even `ingest:sips:create` for the two decision endpoints. I worry that if the permissions are too fine grained that the user that ingests the SIP will not be able to decide whether to make a decision about whether to continue processing, which could be a major administrative hassle.

djjuhasz (changes requested):
I've run out of time today to review the fourth and final commit of the Dashboard changes, but I wanted to submit the comments I have so far.  I think the main problem is the shared workflow state can be updated from multiple signals, and is therefore prone to race conditions.  I don't think we can assume that the child workflow or API clients will be well behaved and will send signals only once and in the correct order.

internal/workflow/processing.go
jraddaoui: This was mostly defensive. `DecisionRequest.Options` is a slice, so copying the struct alone would leave the query result sharing the same array as workflow state. Nothing currently mutates the options, so I'll remove it.

internal/ingest/workflows.go
jraddaoui: `Pending` is used by `findChildDecision` in the API endpoints to make the query result explicit, instead of inferring state from an empty request. A direct `Request != childwf.DecisionRequest{}` check would not work because it contains a slice. And using a pointer would work, but I think the explicit flag is clearer.

internal/workflow/processing.go
jraddaoui: I don't think we need a mutex here. Because of Temporal's deterministic execution model, workflow state is not mutated by multiple goroutines at the same time. Updates and workflow code are handled through the workflow task execution, so this is not the same kind of shared memory access that requires a mutex.

internal/workflow/processing.go
jraddaoui: This should not be racy for the same reason as the previous comment. That said, a second child decision request could still overwrite the active one, so I added a guard before recording the request.

internal/workflow/processing.go
jraddaoui: Agreed. I added a check that rejects a new child decision request if either `state.childDecisionRequest` or `state.childDecisionResponse` are already set.

jraddaoui: This is using a named return value, so the deferred function can see errors returned later in the function.

docs/src/admin-manual/iac.md
jraddaoui: That is the main reason we use ABAC, to support fine grained permissions. I agree this can add some administrative hassle, but we already have the `*` wildcard and role mappings in Enduro to help with that. And similar mappings can also be managed in the OIDC provider.

I think reviewing an AIP and responding to a custom child workflow decision are different actions. In the case we are working on now, the decision requires checking an external system, so I think the extra granularity is worth it.

internal/ingest/goa.go
```diff
@@ -218,25 +219,114 @@ func (svc *ingestImpl) ConfirmSip(ctx context.Context, payload *goaingest.Confir
...
+		svc.logger.Error(err, "find child decision: query workflow", "workflow_id", temporalID)
+		return ChildDecisionQueryResult{}, ErrInternalError
+	}
+
+	var decision ChildDecisionQueryResult
+	if err := encoded.Get(&decision); err != nil {
+		svc.logger.Error(err, "find child decision: decode query result", "workflow_id", temporalID)
+		return ChildDecisionQueryResult{}, ErrInternalError
+	}
+	if !decision.Pending {
```
djjuhasz: This is the only place we are checking `decision.Pending` as far as I can see.  The Pending value is not exposed through the API, so there's no need to consider ergonomics for API clients. This test could just as easily be `decision.Request.Message != ""` and it would avoid the need for the `Pending` field as well as avoiding any potential confusion about what happens if `Pending = false` but `Request` is not empty.

djjuhasz: Actually without the `Pending` field we could just remove `ChildDecisionQueryResult` altogether and pass `childwf.DecisionRequest` directly.  In that case the test could be simplified to `decision.Message != ""`.

internal/ingest/workflows.go
djjuhasz: I added a new comment in the `findChildDecision` endpoint about this.

internal/workflow/processing.go
djjuhasz: Wow! That is pretty fancy. It works well in when I test it in dev, so I will trust that Temporal has figured out the details of updating the workflow state without races. :crossed_fingers:

docs/src/admin-manual/iac.md
djjuhasz: Yeah, I get that we want to support fine grained permissions, but I do worry that having too many options will lead to option overload and bad configurations that will make the system practically unusable.  I can't imagine a workplace where they want an archivist that is doing an ingest to stop and get a supervisor to approve or cancel an ingest, but maybe I am being naive.

internal/ingest/goa.go
jraddaoui: Sure.

jraddaoui: Done!

dashboard/src/components/WorkflowCollapse.vue
```diff
@@ -133,16 +165,22 @@ const showTasks = computed(() => {
       :aria-labelledby="'wf' + index + '-heading'"
       data-bs-parent="#workflows"
     >
+      <SipDecisionAlert
```
djjuhasz, line 168: It would be nice to combine all of the "decision" alerts (decision, SIP review, AIP deletion) into a single alert/notification system at some point, to avoid having three different conditional blocks.  That is a job for another day though.

djjuhasz (approved):
Looks great! Thanks @jraddaoui!

Issue #1621, opened 2026-05-05
Problem: Share custom metadata between child workflows


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Enduro supports custom child workflows that extend processing and batch behavior, including preprocessing, poststorage, and postbatch workflows. These workflows may need to exchange custom metadata that Enduro does not understand or use directly.

The first known use case is to return a few fields from a preprocessing child workflow and pass them to the poststorage child workflow for the same SIP. Future cases may need to pass metadata from poststorage back to the processing workflow, from processing workflows to the batch workflow, and from batch state into the postbatch child workflow.

The current child workflow contracts do not provide a generic way to carry this data.

**Describe the solution you'd like**

Add an opaque custom metadata field to child workflow params and results so Enduro can carry custom data between child workflows without interpreting it.

Use a small JSON metadata type, for example:

```go
type CustomMetadata map[string]json.RawMessage
```

Add this metadata to the relevant workflow contracts:

- `childwf.PreprocessingResult`
- `childwf.PostStorageParams`
- the common child workflow result type used by poststorage and postbatch, when poststorage is changed to return a result
- a new processing workflow result returned to the batch workflow
- `childwf.PostbatchSIP` for SIP metadata
- optionally `childwf.PostbatchParams` for batch metadata

The expected flow is:

1. The preprocessing child workflow returns `CustomMetadata`.
2. The processing workflow stores the metadata in workflow state.
3. The processing workflow passes the metadata to the poststorage child workflow.
4. The poststorage child workflow may return updated or additional metadata.
5. The processing workflow returns metadata for each SIP to the batch workflow.
6. The batch workflow passes metadata for each SIP to the postbatch child workflow.

Enduro should treat this metadata as opaque data. It should copy it between workflows but should not validate, interpret, display, or make business decisions based on custom fields.

**Describe alternatives you've considered**

Use blob storage for large shared artifacts. This is a good companion approach, but it should not replace direct Temporal metadata for small fields. Blob storage is better suited for generated reports, validation output, inventories, manifests, and other data that may exceed Temporal payload limits or should be managed with separate retention and access policies.

Use a custom database shared by the custom workflows. This can be appropriate when custom metadata is mutable, queryable, or belongs entirely to the customer extension domain. In that case, Enduro should pass stable references or identifiers through the workflow metadata rather than owning the schema.

The recommended approach is to use Temporal workflow params and results for small metadata, and use external references inside that metadata for large or independently managed data.

[...]

PR #1622, opened 2026-05-05
Pass custom preprocessing metadata to poststorage

jraddaoui:
Carry opaque custom metadata returned by the preprocessing child workflow in processing workflow state and pass it to the poststorage child workflow params.

Refs #1621.


djjuhasz (approved):
LGTM 👍

PR #1623, opened 2026-05-05
Wait for poststorage workflow result

jraddaoui:
Make poststorage child workflow execution wait for its typed result so the parent can record returned preservation tasks, merge custom metadata, and fail or error according to the child outcome. Reuse the preservation task save path shared with preprocessing.

Based in #1622. Refs #1621.


internal/workflow/processing.go
```diff
@@ -1318,35 +1304,97 @@ func (w *ProcessingWorkflow) waitForChildDecisionResponse(
...
+) error {
+	if len(tasks) == 0 {
+		return nil
+	}
+
+	opts := withLocalActivityOpts(ctx)
+	var res localact.SavePreprocessingTasksActivityResult
+	return temporalsdk_workflow.ExecuteLocalActivity(
+		opts,
+		localact.SavePreprocessingTasksActivity,
```
djjuhasz: We should change this activity name from `SavePreprocessingTasksActivity` to `SavePreservationTasksActivity` and update the params and result names too.

djjuhasz (approved):
Looks good @jraddaoui.  We should update the name of the `SavePreprocessingTasksActivity` to make it general for all child workflow preservation tasks, but that doesn't need to happen in this PR.

internal/workflow/processing.go
jraddaoui: Good point, thanks @djjuhasz. I decided to use `SaveChildwfTasksActivity`, and going down the rabbit whole I changed the enum too.

djjuhasz (approved):
Thanks @jraddaoui, looks good! 👍

PR #1624, opened 2026-05-06
Pass custom metadata to postbatch

jraddaoui:
Add a processing workflow result carrying custom metadata so batch workflows can collect SIP metadata from completed processing children. Forward that metadata into postbatch SIP params.

Based on #1623. Refs #1621.


djjuhasz (approved):
LGTM @jraddaoui. 👍

Issue #1625, opened 2026-05-06
Problem: Child workflow contracts are not shareable


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Child workflow projects need to share the same wire contract definitions as Enduro. Those contracts currently live in an internal package, so external child workflow implementations cannot import them directly and must duplicate types for workflow parameters, results, outcomes, metadata, and task records.

**Describe the solution you'd like**

Expose the child workflow contract types from a root `childwf` package that can be imported by other child workflow projects. The package should contain the shared workflow parameter and result types for preprocessing, poststorage, and postbatch workflows, along with shared outcome, task, workflow type, and metadata definitions.

Move unrelated child workflow configuration into an internal package so the public `childwf` package remains focused on the contract surface.

**Describe alternatives you've considered**

Child workflow projects could continue duplicating the same structs and enum values locally, but this increases drift risk and makes contract changes harder to coordinate.

**Additional context**

This should cover the contract surface used by child workflow implementations without exposing Enduro runtime configuration or internal service details.

PR #1626, opened 2026-05-06
Expose childwf package

jraddaoui:
Move child workflow contracts into a pkg/childwf package. Remove the internal/childwf package, which had become only configuration and would now conflict with the shared contract name. Move child workflow config into internal/config, and move API server configuration there as well so about can use config types directly without creating a config -> api -> about -> config import cycle.

Based on #1624. Refs #1625.


jraddaoui:
I'm getting internal errors trying to do my own review. A couple of extra notes @djjuhasz:

I decided to put the package in the root instead of in a pkg subfolder to make the imports clearer. But giving our current project structure having multiple non Go packages in the root, I'd be happy to move it to `pkg/childwf` if you think that's better.

Now that child workflow projects will be importing this new package, I think it's a good moment to change `PreservationTasks` to just `Tasks` in the preprocessing and poststorage child workflows. I was planning to do that when I added them to the poststorage result, but I wanted consistency with the preprocessing result and I didn't what to break the contract. Let me know your thoughts!

djjuhasz (approved):
Cool! Thanks @jraddaoui. 👍

djjuhasz:
@jraddaoui I would prefer to have a `pkg/` directory to distinguish exported Go packages from all the other stuff in the project root directory, but I'm fine with the current organization if you'd prefer to avoid longer import paths.

jraddaoui:
Thanks @djjuhasz! I've done both things, move it to `pkg/childwf` and rename `PreservationTasks` to `Tasks` in the results.

I'm also testing if we can use [apidiff](https://pkg.go.dev/github.com/joelanford/go-apidiff) here, based in [gotools](https://github.com/artefactual-labs/gotools/blob/main/.github/workflows/test.yml#L35-L48). It looks like it [ignores internal packages](https://github.com/joelanford/go-apidiff/blob/v0.8.3/pkg/diff/run.go#L224-L227) and I hope it doesn't complain much about other Go packages living in `hack/` or somewhere else. Let's give it a try.

jraddaoui:
It seems to be working okay, see https://github.com/artefactual-sdps/enduro/pull/1627.

jraddaoui:
@djjuhasz, while importing this package in a child workflow projects, I realized we have those `evenlog` packages and `EventOutcome` enums you created in a few projects. I think it's a good fit here too, so I've added abfff95.

djjuhasz (approved):
LGTM @jraddaoui. 👍

djjuhasz:
@jraddaoui I just suddenly had the thought that maybe we should make the childwf package it's own project / repo so if we have to change the API we can version it separately from the Enduro project versioning. :thinking:

jraddaoui:
Thanks @djjuhasz! I thought about that and it may be the best choice after all. But I want to see how it goes aligning the versions of the package and the main consumer of those interfaces. Let's give it a try, we can always move it later.

PR #1636, opened 2026-05-13
Exclude metadata files from SIP file count

djjuhasz:
Refs #1595.

- Exclude any files in a metadata directory when counting files in a SIP
- Add a top-level "metadata" directory containing a "checksum.sha256" file to the "testdata/standard_transfer/small" test SIP
- Update the bundle activity test to account for the new test SIP directory and file


internal/workflow/activities/count_sip_files.go
```diff
@@ -38,7 +38,8 @@ func NewCountSIPFilesActivity() *CountSIPFilesActivity {
 // Execute counts the number of preservation files in the SIP at params.Path.
 //
 // If the SIP is a BagIt Bag, Execute counts the files in the "data" directory.
-// If the SIP type is unknown, Execute counts all the files in the SIP.
+// If the SIP type is unknown, Execute counts all the files in the SIP except
+// for files in a "metadata" directory.
```
jraddaoui: In any "metadata" directory? Not sure how to indicate it will skip directories named "metadata" at all levels.

internal/workflow/activities/count_sip_files.go
```diff
@@ -62,6 +63,10 @@ func (a *CountSIPFilesActivity) Execute(
 		if err != nil {
 			return err
 		}
+		if d.IsDir() && d.Name() == "metadata" {
+			// Skip the metadata directory and its contents.
```
jraddaoui: Similar here, "the metadata directory"  seems to imply we are looking just for one.

jraddaoui (approved):

internal/workflow/activities/count_sip_files.go
djjuhasz: I've tried to make it clearer in this comment and the following one that any and all "metadata" directories will be skipped.

Issue #1637, opened 2026-05-14
Problem: Dashboard resource views show stale resource while loading a new one


jraddaoui:
**Describe the bug**

After successfully loading a resource in a dashboard view page, such as a SIP or AIP detail page, navigating away and then returning to the same view type for a different resource briefly shows the previously loaded resource while the new resource request is still loading.

This can make the page appear to be showing the wrong SIP, AIP, or other resource, and risks users reading or acting on stale resource metadata before the new request completes.

**To Reproduce**

1. Open a dashboard resource detail page, such as a SIP or AIP view page.
2. Wait for the resource to load successfully.
3. Navigate away from the page.
4. Navigate to the same view page type for a different resource.
5. Observe that the previously loaded resource remains visible briefly until the new resource finishes loading.

**Expected behavior**

When navigating to a resource view for a different resource, the dashboard should clear or replace the old resource state immediately and show an appropriate loading state until the requested resource has loaded.

The previous resource should not remain visible for the new resource route.

**Screenshots**

https://github.com/user-attachments/assets/76e304ae-ee5b-4feb-95ff-55b0015cbabe

PR #1638, opened 2026-05-18
Reset resource stores on unmount

jraddaoui:
Clear selected and listed resource state when dashboard resource views are unmounted. This prevents stale SIP, batch, AIP, location, and related user state from leaking into the next resource view.

Refs #1637.


djjuhasz (approved):
This workaround looks good @jraddaoui, but it seems like we should probably not be using a state store in the first place for state that we don't want persisting between different app "pages" (e.g. AIP detail to AIP list).  

It would be a big change to convert all of our stores that keep state for transitory resources (e.g. aip, sip, batch) to regular Vue reactive elements, so I think using `$reset` is fine for now, but we may want to consider it down the road.

PR #1644, opened 2026-05-20
Show loading state on AIP download button

### LINKED ISSUE #1629: Problem: no indication that an AIP download has started
**Describe the bug**

It can take several seconds for a large AIP to start downloading from the Archivematica storage service, but there's no indication after clicking the "Download" button that the download has started.  The lack of feedback after clicking the button makes it difficult to tell if the button is broken or whether the download will happen.

**To Reproduce**

1. Go to the AIP details page of a large AIP in the Enduro Dashboard
2. In the "Location" panel, click on "Download"
3. Wait patiently for download to start

**Expected behavior**

Some visual indication that the download button has been clicked and the download has started.

**Screenshots**

https://github.com/user-attachments/assets/dcd3ce6c-ed17-485b-8b42-395b079cf23c

**Additional context**

None.

djjuhasz:
Fixes #1629.

Disable the Download button and display a spinner with "Downloading..." text while the download API request is in-flight, preventing duplicate clicks.

### DIFF
--- a/dashboard/src/components/AipLocationCard.spec.ts
+++ b/dashboard/src/components/AipLocationCard.spec.ts
@@ -41,7 +41,7 @@ describe("AipLocationCard.vue", () => {
         </div></span></p>
         <div data-v-09f60ec4="">
           <transition-stub data-v-09f60ec4="" mode="out-in" appear="false" persisted="false" css="true">
-            <div data-v-09f60ec4="" class="d-flex flex-wrap gap-2"><button data-v-09f60ec4="" type="button" class="btn btn-primary btn-sm"> Download </button>
+            <div data-v-09f60ec4="" class="d-flex flex-wrap gap-2"><button data-v-09f60ec4="" type="button" class="btn btn-primary btn-sm">Download</button>
               <!--v-if--><button data-v-09f60ec4="" type="button" class="btn btn-primary btn-sm"> Delete </button>
             </div>
           </transition-stub>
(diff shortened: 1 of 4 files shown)


jraddaoui (approved):
LGTM! If you fully reviewed the changes, I think we shouldn't include the agent as coauthor. VSCode was adding them automatically by default, but they made the new setting opt-in instead of opt-out in recent versions.

djjuhasz:
@jraddaoui I added another commit with unit tests for the AIPstore download() method, so could you take a look at that commit too please? I'll squash the commits before merging.

I'm torn about removing the Claude co-author tag from the commit. I did review all of the code, but Claude *wrote* it all. I don't really feel comfortable saying I "authored" the code, if anything I was more of an editor or reviewer than the author.

jraddaoui (approved):
Thanks @djjuhasz!

It would be nice to do the same in other download buttons like the failed SIP/PIP and the AIP deletion report.

jraddaoui:
About the co-author part, it’s quite a big topic, but my take on it is that Co-authored-by has an existing meaning in Git: another author contributed and can be identified as responsible for part of the change. An AI tool can generate code, but it cannot take responsibility for the result, explain the intent later, or maintain it.

If you reviewed the full change, understood it, and are comfortable committing it, then I think you are the author of the commit, even if Claude helped produce the first version of the code. In that sense, it is similar to accepting generated code from an IDE, a template, or a code generator, just at a larger scale.

I do think it is reasonable to disclose AI assistance, but I’d prefer not to use Co-authored-by for that unless the project has decided that is the convention. Something like Assisted-by: Claude would feel more accurate to me. But if we do this, I think we should agree on it as a project/team convention, not decide commit by commit.

jraddaoui:
About the VSCode changes: https://github.com/microsoft/vscode/issues/314311

djjuhasz:
@jraddaoui I agree that AI attribution is a big topic, and one we need to discuss as a company in a different forum than here.  That said, for this PR I'd like to leave the "co-authored-by: Claude" in this commit, for the following reasons:

1. For me being an "author" primarily means writing the code.  
2. I think there is a significant difference between using and IDE or code completion and an AI agent writing a substantial portion of the code, as in this case.
3. I agree that the committer of the code is responsible for ensuring the code is well-written, maintainable, and safe, but I think that's a separate responsibility from "authoring" the code.
4. I do not believe I wrote this code. I reviewed it, and take responsibility for its quality, but it was written by Claude.
5. I think disclosing AI authorship is required to be open about how our software is created, and for us to maintain the trust of our community.
6. If nothing else, I think it's important to disclose to you, as a reviewer, that the code is written by AI so you can take that into account when reviewing the code. I think this information is important for a reviewer to hold a PR creator accountable if they haven't exercised sufficient attention and care in reviewing AI generated code.

I know we don't have a company policy on this yet, but in the absence of a generally policy I would like to default to being open and explicitly disclose AI code authorship.

PR #1650, opened 2026-05-25
Add docs for batch workflow

sallain:
In this PR I've added a section for upload using the batch functionality, updated screenshots, and added a section to the glossary.

Tbh it might be a bit rough - it was one of those where I just got tired of looking at it so decided to just push and see how it goes.


docs/src/user-manual/glossary.md
```diff
@@ -48,6 +48,14 @@ description of the AIP and its components is available in the Archivematica
 documentation - see: [AIP
 structure](https://www.archivematica.org/docs/latest/user-manual/archival-storage/aip-structure/#aip-structure).
 
+## Batch
+
+A group of SIPs that are related in some way. By identifying a group
+of SIPs as a batch during upload, a user can then use the Batch pages in Enduro
+to track all of the SIPs in the batch and take actions based on the status of
+the batch and/or individual SIPs. Child workflows can be triggered based on the
+status of a batch.
```
fiver-watson, line 57: Might be good to mention something about how users can optionally assign an ID and that will be added as an accession number to the METS?

docs/src/user-manual/ingest/submitting-content.md
```diff
@@ -4,7 +4,10 @@ Depending on how Enduro is configured during installation, there are several
 different ways that operators can initiate an ingest workflow. Currently, in all
 cases Enduro needs one or more
 [SIPs](../glossary.md#submission-information-package-sip) to initiate an ingest
-workflow. Below are instructions on the 3 different ways that operators can
+workflow. SIPs can be uploaded individually or you can upload several SIPs
+together and have Enduro treat them as a batch.
```
fiver-watson, line 8: Not critical, but bc this is the first time in the docs that a batch is being mentioned I think it makes sense to link to the glossary term here

docs/src/user-manual/ingest/submitting-content.md
```diff
@@ -112,16 +116,16 @@ package exceeds the configured limit.
...
 Another method of initiating ingest is by selecting packages previously uploaded
 to a [source location]. A **SIP source location** is an object storage location
 or filesystem directory for the deposit of SIPs that can then be ingested by an
 Enduro operator. The [SIP source configuration] must be set up by a system
-administrator. Unlike a [watched location] where an action is automatically
-triggered upon deposit, a source location is intended for asynchronous manual
-follow-up. In this case, Enduro operators can select zipped packages found in a
-configured source location and initiate ingest via the user interface.
+administrator. Unlike a [watched location], where an action is automatically
+triggered upon deposit, a source location is intended to hold SIPs until a user
```
fiver-watson, line 126: ha, updating this from "for asynchronous manual follow-up" to "to hold SIPs until a user manually starts the ingest" was a gooooood call

docs/src/user-manual/ingest/submitting-content.md
```diff
@@ -166,13 +170,81 @@ watched location [example shown below](#example---upload-via-minio).
...
+
+    A count of selected SIPs will be shown above the right side of the picker.
+    You can click the "Select all" link to select all SIPs in the source
+    location, and/or use the "Clear all selections" link to uncheck any
+    previously selected SIPs.
+
+    ![2 SIPs selected from a source location](../screenshots/sip-source-start-upload.png)
+
+4. To group the selected SIPs as a batch, click on the toggle for "Treat this
+   ingest as a batch?" You can optionally give your batch a custom identifier,
```
fiver-watson, line 224: Might be good to have a note or tip admonition here about the identifier showing up in the metadata, so people are aware and can consider how they might use that information for ongoing management and batch identification etc?

fiver-watson (changes requested):
Hey, thanks for starting this! Actually what you have so far looks pretty good to me; I've only added a few minor nitpicks here and there for you to consider. 

The one thing I don't see mentioned at all is some info on what happens if 1 or more SIPs in a batch fails validation. I know we don't have the final version of the various checks and options we _want_ to eventually have in place, but we should mention what's possible now.

PR #1651, opened 2026-05-25
Support filesystem bucket configuration

jraddaoui:
Switch the sample config to filesystem internal storage and SIP source
buckets, with links to the relevant documentation. Update the gotools
dependency so fileblob support is registered centrally, and adjust tests
to cover filesystem bucket behavior.

Add shared documentation for bucket configuration, including filesystem
URLs, S3-compatible options, Azure URL auth sections, and fileblob URL
parameters.

**Add shared internal storage volume**

Add an enduro-internal-storage PVC to the development Kubernetes
manifests and mount it into the API and worker pods. Initialize the
ingest, storage, and SIP source directories on the volume before the
containers start so file:// buckets can open against a fresh PVC.

Create the same runtime directories in the container base image for
non-Kubernetes runs. Move the shared PREMIS schema copy into the base
stage and make the worker image copies owned by the enduro user for
consistent runtime permissions.

Refs #1660.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,25 +63,27 @@ RUN apt-get update && apt-get install -y --no-install-recommends libxml2-utils
 RUN groupadd --gid 1000 enduro \
 	&& useradd --uid 1000 --gid 1000 -m enduro
 USER enduro
+COPY --link --chown=1000:1000 hack/xsd/premis.xsd /home/enduro/premis.xsd
+RUN ["mkdir", "-m", "700", "-p", \
+    "/home/enduro/logs", \
+    "/home/enduro/internal-storage/ingest", \
+    "/home/enduro/internal-storage/storage", \
+    "/home/enduro/internal-storage/sip-source"]
 
 FROM base AS enduro
 COPY --link --chown=1000:1000 --from=build-enduro /out/enduro /home/enduro/bin/enduro
 COPY --link --chown=1000:1000 --from=build-enduro /src/enduro.toml /home/enduro/.config/enduro.toml
 COPY --link --chown=1000:1000 assets/Enduro_AIP_deletion_report_v3.tmpl.pdf /home/enduro/Enduro_AIP_deletion_report_v3.tmpl.pdf
-COPY --link --chown=1000:1000 hack/xsd/premis.xsd /home/enduro/premis.xsd
-RUN ["mkdir", "-m", "700", "-p", "/home/enduro/logs"]
 CMD ["/home/enduro/bin/enduro", "--config", "/home/enduro/.config/enduro.toml"]
 
 FROM base AS enduro-a3m-worker
-COPY --link --from=build-enduro-a3m-worker /out/enduro-a3m-worker /home/enduro/bin/enduro-a3m-worker
-COPY --link --from=build-enduro-a3m-worker /src/enduro.toml /home/enduro/.config/enduro.toml
-COPY --link hack/xsd/premis.xsd /home/enduro/premis.xsd
+COPY --link --chown=1000:1000 --from=build-enduro-a3m-worker /out/enduro-a3m-worker /home/enduro/bin/enduro-a3m-worker
+COPY --link --chown=1000:1000 --from=build-enduro-a3m-worker /src/enduro.toml /home/enduro/.config/enduro.toml
 CMD ["/home/enduro/bin/enduro-a3m-worker", "--config", "/home/enduro/.config/enduro.toml"]
 
 FROM base AS enduro-am-worker
-COPY --link --from=build-enduro-am-worker /out/enduro-am-worker /home/enduro/bin/enduro-am-worker
-COPY --link --from=build-enduro-am-worker /src/enduro.toml /home/enduro/.config/enduro.toml
-COPY --link hack/xsd/premis.xsd /home/enduro/premis.xsd
+COPY --link --chown=1000:1000 --from=build-enduro-am-worker /out/enduro-am-worker /home/enduro/bin/enduro-am-worker
+COPY --link --chown=1000:1000 --from=build-enduro-am-worker /src/enduro.toml /home/enduro/.config/enduro.toml
 CMD ["/home/enduro/bin/enduro-am-worker", "--config", "/home/enduro/.config/enduro.toml"]
 
 FROM ${TARGET}
--- a/docs/src/admin-manual/configuration.md
+++ b/docs/src/admin-manual/configuration.md
@@ -523,8 +523,7 @@ workflowType = "create aip"
 * `redisAddress`: Binds Redis to a specific address and port.
 * `redisList`: The name of the queue that Redis should use for the watched
   location SIP deposit events.
-* `endpoint`: API endpoint for the target MinIO instance. Used by Enduro to read
-  contents of a watched bucket, or for any other MinIO interactions.
+* `endpoint`: API endpoint for the target S3-compatible object store.
 * `pathStyle`: Currently Amazon Web Services support two different URL
   construction methods when interacting with an object store bucket via API. The
   "path-style" method constructs the bucket's access URL using the configured
@@ -533,7 +532,7 @@ workflowType = "create aip"
   supported so if using an S3-like object store, ensure this is set to `true`.
 * `key`: Username for accessing the configured S3-like bucket.
 * `secret`: Password for accessing the configured S3-like bucket.
-* `region`:  = AWS S3 buckets are created in a specific region. When interacting
+* `region`: AWS S3 buckets are created in a specific region. When interacting
   with S3, you can specify the region during the bucket creation process. For
   a full list of available regions and the syntax to specify them, consult the
   [AWS S3 documentation][S3-regions].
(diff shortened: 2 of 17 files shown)


jraddaoui:
@djjuhasz, see https://gist.github.com/sevein/0cb4b7656a970885abdb98a1c06609bf?permalink_comment_id=6168327#gistcomment-6168327.

Dockerfile
```diff
@@ -63,25 +63,27 @@ RUN apt-get update && apt-get install -y --no-install-recommends libxml2-utils
 RUN groupadd --gid 1000 enduro \
 	&& useradd --uid 1000 --gid 1000 -m enduro
 USER enduro
+COPY --link --chown=1000:1000 hack/xsd/premis.xsd /home/enduro/premis.xsd
+RUN ["mkdir", "-m", "700", "-p", \
+    "/home/enduro/logs", \
+    "/home/enduro/internal-storage/ingest", \
```
djjuhasz, line 69: I don't know how it would work in practice to share these directories between multiple containers (e.g. enduro and enduro-am-worker). I think if a bind mount was mounted to `/home/enduro` at run time the source directory structure would obscure these directories. It looks like an empty Docker volume mounted to `/home/enduro` would inherit this directory structure, so I guess that is use-case your are envisioning when creating the directories in the image?

docs/src/admin-manual/configuration.md
```diff
@@ -545,6 +544,139 @@ workflowType = "create aip"
...
+write, or delete permissions required by that bucket's role. When Enduro
+processes run on different hosts or containers, the filesystem path must refer
+to shared storage that is mounted at the same path for every process that needs
+to access the bucket.
+
+For `file://` URLs, `metadata=skip` prevents `fileblob` from writing `.attrs`
+metadata sidecar files. The `no_tmp_dir=true` query parameter keeps temporary
+write files next to the final object, which avoids cross-device rename failures
+when the system temporary directory is on another filesystem. Use
+`no_tmp_dir=true` for filesystem buckets that are written through `fileblob`.
```
djjuhasz: I think it would be helpful to note that bucket specific guidance on when to use the `no_tmp_dir=true` option is provided below, so readers aren't wondering which buckets are "written through `fileblob`".

docs/src/admin-manual/configuration.md
```diff
@@ -545,6 +544,139 @@ workflowType = "create aip"
...
+
+Each bucket can be configured with either a URL or provider-specific fields.
+When `url` is set, it takes precedence over the S3-compatible fields listed
+below. The URL form supports local filesystem locations, S3-compatible buckets,
+and [Azure Blob Storage][Azure] containers. S3-compatible buckets can
+alternatively be configured with the endpoint, credentials, region, and bucket
+fields below.
+
+Use the examples below with the section you are configuring. Replace
+`[example.bucket]` with `[storage.internal]`, `[internalStorage]`, or
```
djjuhasz, line 564: The similarity between the `[storage.internal]` and `[internalStorage]` config section names is really an unfortunate choice on our part. 😬  I wonder if we can add `[ingest.internal]` (or an even better name) while still supporting `[internalStorage]` for backwards compatibility?

I'm not suggesting changing the naming in this PR, but it's something I'd like to address at some point.

docs/src/admin-manual/configuration.md
```diff
@@ -545,6 +544,139 @@ workflowType = "create aip"
...
+```
+
+For filesystem-backed storage, the directory referenced by `url` must exist
+before Enduro starts, and the process user running Enduro must have the read,
+write, or delete permissions required by that bucket's role. When Enduro
+processes run on different hosts or containers, the filesystem path must refer
+to shared storage that is mounted at the same path for every process that needs
+to access the bucket.
+
+For `file://` URLs, `metadata=skip` prevents `fileblob` from writing `.attrs`
```
djjuhasz: I think we need to provide more context here about why `file://` URLs have these extra config parameters. I think it would be helpful to have a preamble about how Enduro implements filesystem-backed storage with the [fileblob](https://pkg.go.dev/github.com/google/go-cloud/blob/fileblob) package and describe how it has the default behaviours of writing `.attrs` metadata files, and writing files to the tmp directory first to make file copies atomic.  I don't know if we want to include a table like https://gist.github.com/sevein/0cb4b7656a970885abdb98a1c06609bf#52-fileblob-limitations, but I personally found that very helpful for understanding why the `metadata=skip` and `no_tmp_dir=true` options may be required.

docs/src/admin-manual/configuration.md
```diff
@@ -648,113 +780,35 @@ migrate = true
 
 #### Internal location used for storing staging AIPs and reports
 
-These settings are used to configure a local object store bucket or filesystem
-directory where the storage service places reports, and staging AIPs when the
+These settings are used to configure an internal filesystem directory or object
+store bucket where the storage service places reports, and staging AIPs when the
```
djjuhasz: Suggestion: "...and stages AIPs when [a3m] is the configured [preservation engine]."

docs/src/admin-manual/configuration.md
```diff
@@ -648,113 +780,35 @@ migrate = true
...
-* Shared key authentication using `storageKey`.
-* Client secret authentication using `tenantID`, `clientID`, and
-`clientSecret`.
-
-If both authentication modes are configured, the client secret settings take
-precedence.
+For Archivematica deployments, this internal location is used for AIP deletion
+reports while final AIPs are stored in Archivematica Storage Service. The a3m
+AIP staging flow still requires an internal location backend that supports
+signed URLs.
```
djjuhasz, line 819: So a filesystem-backed "file://" URL can not be used for this bucket when using a3m?

docs/src/admin-manual/configuration.md
```diff
@@ -1045,125 +1099,37 @@ maxSize = 4294967296
...
-`clientSecret`.
-
-If both authentication modes are configured, the client secret settings take
-precedence.
+For filesystem-backed storage, Enduro writes API-uploaded SIPs and failed
+packages to this bucket, and later reads them for workflow processing or
+operator downloads. The process user must have read and write access. When the
+API and worker processes run on different hosts or containers, the filesystem
+path must refer to shared storage that is mounted at the same path for every
+process that needs to read or write the bucket. Because Enduro writes to this
```
djjuhasz: Suggestion: "...read from or write to the bucket."

docs/src/admin-manual/configuration.md
```diff
@@ -1045,125 +1099,37 @@ maxSize = 4294967296
...
+bucket through `fileblob`, include `no_tmp_dir=true` in `file://` URLs when the
```
djjuhasz: Edit: "in the `file://` URL..."

docs/src/admin-manual/configuration.md
```diff
@@ -648,113 +780,35 @@ migrate = true
...
+url = "file:///home/enduro/internal-storage/storage?metadata=skip&no_tmp_dir=true"
 ```
 
-If `url` is configured, it takes precedence over the individual S3-like
-configuration settings listed below.
+For filesystem-backed storage, Enduro writes to this bucket, so the process
+user must have write access. If multiple Enduro processes need to access this
+bucket, the filesystem path must refer to storage shared by those processes.
+Because Enduro writes to this bucket through `fileblob`, include
+`no_tmp_dir=true` in `file://` URLs when the system temporary directory may be
```
djjuhasz: edit: "include `no_tmp_dir=true` in the `file://` URL when..."

docs/src/admin-manual/configuration.md
```diff
@@ -1188,111 +1154,28 @@ name = "MinIO SIP Source"
...
-* `tenantID`: Azure tenant ID for client secret authentication.
-* `clientID`: Azure client ID for client secret authentication.
-* `clientSecret`: Azure client secret for client secret authentication.
-
-When using Azure Blob Storage, Enduro supports two authentication modes in this
-subsection:
+For a filesystem-backed SIP source, the API lists objects from this bucket, and
+workers download selected objects for ingest. Depending on SIP retention
+settings, workers may also delete source objects after ingest, so the process
+user must be able to perform each configured operation. When the API and worker
```
djjuhasz: I think it's better to make the required permissions explicit: "the process user must be able to *list* the directory contents, and *read* and *delete* files.

internal/ingest/download_sip_test.go
```diff
@@ -56,7 +60,7 @@ func setupBucket(t *testing.T) *blob.Bucket {
 func TestDownloadSipRequest(t *testing.T) {
 	t.Parallel()
 
-	bucket := setupBucket(t)
```
djjuhasz, line 59: We don't use `bucket` until `t.Run()` so I think it's better to do the setup in that function.

docs/src/dev-manual/devel.md
```diff
@@ -268,6 +268,18 @@ cURL by running the following make target:
...
+### Add a SIP to the filesystem SIP source
+
+When the local development configuration uses the filesystem-backed SIP source,
+copy a local SIP into the shared internal-storage PVC through the `enduro` pod:
+
+```bash
+POD=$(kubectl -n enduro-sdps get pod -l app=enduro -o jsonpath='{.items[0].metadata.name}')
+
+kubectl -n enduro-sdps -c enduro cp ./sip.zip \
+  "$POD":/home/enduro/internal-storage/sip-source/sip.zip
```
djjuhasz, line 280: Huh, I'm going to miss the MinIO web UI for uploading SIPs to the SIP source. :(

djjuhasz (changes requested):
Looks good @jraddaoui. I requested some changes to the docs, mostly about providing more context for the fileblob options.

Dockerfile
jraddaoui: My thinking was that you could mount separate volumes for ingest and storage. But your point is a good one too. When using a single shared volume, the intended mount point is `/home/enduro/internal-storage`. Mounting over `/home/enduro` would obscure the binaries, config, and other image contents.

docs/src/admin-manual/configuration.md
jraddaoui: Yeah, totally agree. I think it would be nice to wrap all ingest specific config like that. https://github.com/artefactual-sdps/enduro/pull/1653 does some reorg in the config file moving into that direction, but without adding the ingest wrapper yet.

docs/src/admin-manual/configuration.md
jraddaoui: A bit hesitant to add the table, but I added the preamble.

docs/src/admin-manual/configuration.md
jraddaoui: Until https://github.com/artefactual-sdps/enduro/pull/1653.

jraddaoui:

djjuhasz (approved):
LGTM @jraddaoui. 👍 

The preamble describing the "fileblob://" implementation is really good, thanks for adding that. 🙏

PR #1653, opened 2026-05-26
Stage a3m AIPs through a shared bucket

jraddaoui:
Stage a3m AIPs through a bucket shared by the ingest and storage
domains instead of sending them through the storage signed URL
upload flow. The a3m worker now opens the ingest storage AIP
staging bucket, prefixes it with the fixed aips/ prefix and passes it
to the upload activity, which writes the AIP then creates the storage
AIP record with a pending status.

Add the [a3m.aipStaging] configuration section and document that
it must point to the same bucket root as [storage.internal]. Update
the sample config and admin manual to describe the shared bucket
requirement and the fixed aips/ prefix.

Refs #1660.

### DIFF
--- a/docs/src/admin-manual/configuration.md
+++ b/docs/src/admin-manual/configuration.md
@@ -550,6 +550,7 @@ Several Enduro settings use the same bucket configuration shape for filesystem
 or object store locations:
 
 * `[storage.internal]`
+* `[a3m.aipStaging]`
 * `[internalStorage]`
 * `[sipsource.bucket]`
 
@@ -561,8 +562,8 @@ alternatively be configured with the endpoint, credentials, region, and bucket
 fields below.
 
 Use the examples below with the section you are configuring. Replace
-`[example.bucket]` with `[storage.internal]`, `[internalStorage]`, or
-`[sipsource.bucket]` as appropriate.
+`[example.bucket]` with `[storage.internal]`, `[internalStorage]`,
+`[a3m.aipStaging]`, or `[sipsource.bucket]` as appropriate.
 
 **Example filesystem URL configuration**:
 
@@ -644,6 +645,7 @@ value. Then add the matching Azure authentication section:
 
 * `[storage.internal]` uses `[storage.internal.azure]`.
 * `[internalStorage]` uses `[internalStorage.azure]`.
+* `[a3m.aipStaging]` uses `[a3m.aipStaging.azure]`.
 * `[sipsource.bucket]` uses `[sipsource.bucket.azure]`.
 
 **Example Azure shared key configuration**:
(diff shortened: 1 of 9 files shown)


jraddaoui:
@djjuhasz, see https://gist.github.com/sevein/0cb4b7656a970885abdb98a1c06609bf?permalink_comment_id=6169695#gistcomment-6169695.

docs/src/admin-manual/configuration.md
```diff
@@ -746,6 +748,26 @@ tokenExpiryLeeway = "30s"
...
+worker writes completed AIPs before the storage domain reads them. Configure
+this bucket with the shared
+[bucket configuration options](#bucket-configuration-options), using
+`[ingest.storage.aipStaging]` as the bucket section. For Azure Blob Storage, use
+`[ingest.storage.aipStaging.azure]` as the matching authentication section.
+
+For a3m deployments, this bucket must point to the same bucket root as
+`[storage.internal]`. Staged AIPs are written under the fixed `aips/` prefix
+inside the bucket.
+
```
djjuhasz: Suggestion: "¶For Archivematica deployments this bucket configuration is ignored".

internal/ingest/config.go
```diff
@@ -17,6 +18,7 @@ type StorageConfig struct {
 	Address                    string
 	DefaultPermanentLocationID uuid.UUID
 	OIDC                       StorageOIDCConfig
+	AIPStaging                 bucket.Config
```
djjuhasz: Maybe it's better to put this in the `a3m` config if it's only relevant for a3m deployments?

internal/workflow/activities/upload.go
```diff
@@ -3,77 +3,71 @@ package activities
...
 
-		fi, err := f.Stat()
-		if err != nil {
-			return &UploadActivityResult{}, err
-		}
+	return &UploadActivityResult{}, err
+}
 
-		uploadReq.ContentLength = fi.Size()
+func (a *UploadActivity) uploadToAIPStagingBucket(ctx context.Context, params *UploadActivityParams) error {
```
djjuhasz, line 54: Maybe `copyToAIPStagingBucket` is more accurate, since the AIP may or may not be uploaded to a remote bucket?

internal/workflow/processing_setup_test.go
```diff
@@ -249,7 +249,7 @@ func (s *ProcessingWorkflowTestSuite) setupA3mWorkflowTest(
 		temporalsdk_activity.RegisterOptions{Name: a3m.CreateAIPActivityName},
 	)
 	s.env.RegisterActivityWithOptions(
-		activities.NewUploadActivity(nil).Execute,
+		activities.NewUploadActivity(nil, memblob.OpenBucket(nil)).Execute,
```
djjuhasz: Will `nil` work here instead of `memblob.OpenBucket(nil)`? I don't think the activity actually gets called in test, so passing an open bucket may not be necessary.

djjuhasz (approved):
Looks good to me @jraddaoui.  I had a few questions, but no blockers.

internal/workflow/activities/upload.go
jraddaoui: I'll do this in a different PR. I think it would be better to split this activity, leaving only the AIP creation part here, and using the shared [bucketupload](https://github.com/artefactual-sdps/temporal-activities/tree/main/bucketupload) activity for the copy part.

internal/workflow/processing_setup_test.go
jraddaoui: We were prefixing the bucket in the activity constructor, we are now passing a prefixed bucket directly.

PR #1656, opened 2026-05-28
Update Goa to v3.28.0

sevein:
Goa v3.27.x adds first-class support for standard Bearer token authentication:

* Use `BearerSecurity` and `BearerToken` in the API design so generated code and OpenAPI specs describe `Authorization: Bearer` consistently.
* Set bearer security at the API level and keep `NoSecurity` on endpoints that authenticate with short-lived ticket cookies.
* Keep `x-required-scopes` extensions so generated docs expose Enduro authorization attributes.

Note: we have to wait until goa v3.27.1 is released.


jraddaoui:
Thanks @sevein!

I was just testing v3.27.0 in another project and noticed `BearerSecurity` / `BearerToken`. Why do we need to wait for v3.27.1 here? Is it because of the `openapi:extension:x-required-scopes` meta?

Also, would that meta be shown in https://enduro.readthedocs.io/dev-manual/api? Do you think we could drop the IAC table we currently update manually in https://enduro.readthedocs.io/admin-manual/iac/#required-attributes?

I noticed that you are using Redoc in other API spec documentation pages, and I like it. Maybe this information would be displayed better there?

sevein:
> I was just testing v3.27.0 in another project and noticed BearerSecurity / BearerToken. Why do we need to wait for v3.27.1 here? Is it because of the openapi:extension:x-required-scopes meta?

I think that we could go with v3.27.0 alone, but I noticed that without the fix in v3.27.1 the cookie-ticket endpoints inherit the API-level bearer security in the spec instead of emitting `security: []`. I guess that could be a problem if some generated clients or docs recommend sending bearer auth unnecessarily.

> Also, would that meta be shown in https://enduro.readthedocs.io/dev-manual/api? Do you think we could drop the IAC table we currently update manually in https://enduro.readthedocs.io/admin-manual/iac/#required-attributes?

Yeah, swagger-ui seems to show them. Try with the following document in https://editor.swagger.io/.

```yaml
openapi: 3.0.3
info:
  title: x-required-scopes example
  version: 1.0.0
servers:
  - url: https://example.com
security:
  - bearer_header_Authorization: []
paths:
  /ingest/sips:
    get:
      summary: List SIPs
      operationId: listSips
      security:
        - bearer_header_Authorization: []
      x-required-scopes:
        - ingest:sips:list
      responses:
        "200":
          description: OK
  /ingest/monitor:
    get:
      summary: Open ingest monitor WebSocket
      operationId: ingestMonitor
      security: []
      responses:
        "101":
          description: Switching Protocols
components:
  securitySchemes:
    bearer_header_Authorization:
      type: http
      scheme: bearer
      description: Bearer token authentication
```

Not sure about Redocly. Dan seems to like swagger-ui better btw.

jraddaoui (approved):
LGTM, thanks @sevein!

PR #1658, opened 2026-06-03
Use local filesystem storage for a3m AIPs

jraddaoui:
- Add `local filesystem` as a storage location source and rename the MinIO source to S3, including migrations and regenerated API/dashboard artifacts.
- Configure local a3m development to use local filesystem AIP storage under the shared internal-storage volume.
- Make AIP deletion activity naming backend neutral and allow deletion from local filesystem locations.
- Narrow the Tilt flush helper so it recreates only the Enduro MySQL databases, preserving Temporal state and internal-storage contents.
- Update development docs with internal-storage copy/list/delete examples and notes about the narrower flush behavior.

Refs #1660.


internal/storage/enums/location_source.go
```diff
@@ -3,9 +3,10 @@ package enums
 /*
 ENUM(
 unspecified
-minio
+s3
 sftp
 amss
+local filesystem
```
djjuhasz: I think "filesystem" is specific enough, I can't imagine adding other filesystem options in the future, and even if we did we could add a qualifier then (e.g. "remote filesystem").

docs/src/dev-manual/devel.md
```diff
@@ -268,16 +268,76 @@ cURL by running the following make target:
...
+
+```bash
+NAMESPACE=enduro-sdps
+CONTAINER=enduro
+INTERNAL_STORAGE=/home/enduro/internal-storage
+POD=$(kubectl -n "$NAMESPACE" get pod -l app=enduro \
+  -o jsonpath='{.items[0].metadata.name}')
+```
+
+Copy a single local file into one of the internal folders:
```
djjuhasz: I think it would be helpful to distinguish the source and destination locations in the command.  Maybe this description could be something like: 

> Copy a single local file ("./sip.zip") into the "sip-source" folder in the Enduro pod:

docs/src/dev-manual/devel.md
```diff
@@ -268,16 +268,76 @@ cURL by running the following make target:
...
+`enduro` pod at `/home/enduro/internal-storage`. The volume contains internal
+folders used by different local workflows:
+
+- `ingest`: internal storage for the ingest domain.
+- `storage`: internal storage for the storage domain.
+- `sip-source`: filesystem-backed SIP source.
+- `perma-aips`: local permanent AIP storage.
+
+Use the `enduro` pod as an access point for copying, listing, and removing files
+in any of those folders:
```
djjuhasz: I'd end this sentence in a full stop, then add some description of the following code chunk like:

> Declare the following variables related to the Enduro Kubernetes pod in the same BASH script as the subsequent `kubectl` commands:

docs/src/dev-manual/devel.md
```diff
@@ -268,16 +268,76 @@ cURL by running the following make target:
...
+
+kubectl -n "$NAMESPACE" exec "$POD" -c "$CONTAINER" -- \
+  ls -la "$INTERNAL_STORAGE/sip-source"
+
+kubectl -n "$NAMESPACE" exec "$POD" -c "$CONTAINER" -- \
+  ls -la "$INTERNAL_STORAGE/perma-aips"
+```
+
+Copy a stored AIP from the internal-storage volume to your host. The source path
+uses the storage object key. If you download the same AIP from the Enduro UI,
```
djjuhasz, line 332: Hmm, at some point we should probably add the file extension to the storage object key so someone interacting with just the filesystem knows the AIP file format.

djjuhasz (approved):
LGTM @jraddaoui, I just had a few minor notes.

PR #1659, opened 2026-06-05
Check for duplicate SIP submissions

### LINKED ISSUE #1617: Feature: check if an incoming SIP has already been ingested

When processing a large volume of packages, it can be easy to upload the same package twice. This could also happen if package deposit is automated or in a variety of other situations. In any case, processing the same package twice is an unnecessary use of compute resources and storage space and should be avoided.

I would like Enduro to check to see if the package has already been ingested. This could be done by computing a checksum for the package and recording this in a database, to be checked against in the future. All incoming packages will be compressed, so computing a checksum should be fast. Any checksum that is a repeat of one already recorded in the database will trigger a failure.

This should be optional :wink: 

1. This is a pretty high-level check that will not catch very similar packages. It would be possible to be much more specific - checking individual file checksums, for example, or checking certain metadata elements.  However, it's difficult with a "similiar" matching strategy to determine exactly when two SIPs are similar enough to count as duplicates.
2. Issue #851 similarly checks for duplicate SIPs by comparing the SIP names instead of their checksums

**Additional context**

- This feature was initially implemented just for SFA in their pre-processing child workflow: https://github.com/artefactual-sdps/preprocessing-sfa/issues/106.  See that issue for more information on why we are moving the feature to Enduro.
- A problem with this initial implementation is it prevents a re-ingest attempt for a SIP when the original ingest attempt fails (https://github.com/artefactual-sdps/enduro-team-tasks/issues/79).
- Another issue with the original implementation is a SIP can not be re-ingested after the created AIP is deleted (#1165).
- Both of the above problems are difficult to solve in the SFA implementation because the preprocessing workflow uses its own database to track the SIP checksums, and Enduro can not update this database if ingest fails after preprocessing or if the AIP is deleted.
- We've decided that preventing accidental ingest of the same SIP twice is a good feature to have in Enduro itself, as long as the check can be turned off with configuration when an institution needs to be able to ingest the same SIP multiple times.

djjuhasz:
Fixes #1617.

Check that the checksum of a newly submitted SIP does doesn't match a previously ingested SIP to prevent duplicate submissions.  A previously submitted SIP that has failed or has been cancelled may be resubmitted successfully, but if a duplicate SIP has a status of pending, queued, processing, or validated it will trigger the duplicate SIP error.

SIPs that are submitted as a directory will not be checked for duplicate contents.

The duplicate checker can be disabled by setting the `ingest.allowDuplicates` configuration to `true`.

### DIFF
--- a/cmd/enduro-a3m-worker/main.go
+++ b/cmd/enduro-a3m-worker/main.go
@@ -273,6 +273,14 @@ func main() {
 			temporalsdk_activity.RegisterOptions{Name: activities.GetSIPExtensionActivityName},
 		)
 		w.RegisterActivityWithOptions(
+			activities.NewCalcFileChecksumActivity().Execute,
+			temporalsdk_activity.RegisterOptions{Name: activities.CalcFileChecksumActivityName},
+		)
+		w.RegisterActivityWithOptions(
+			activities.NewCheckDuplicateSIPActivity(ingestsvc).Execute,
+			temporalsdk_activity.RegisterOptions{Name: activities.CheckDuplicateSIPActivityName},
+		)
+		w.RegisterActivityWithOptions(
 			archiveextract.New(cfg.ExtractActivity).Execute,
 			temporalsdk_activity.RegisterOptions{Name: archiveextract.Name},
 		)
--- a/cmd/enduro/main.go
+++ b/cmd/enduro/main.go
@@ -508,7 +508,12 @@ func main() {
 
 		// Ingest processing workflow and activities.
 		w.RegisterWorkflowWithOptions(
-			workflow.NewProcessingWorkflow(cfg, rand.Reader, ingestsvc, wsvc).Execute,
+			workflow.NewProcessingWorkflow(
+				cfg,
+				rand.Reader,
+				ingestsvc,
+				wsvc,
+			).Execute,
 			temporalsdk_workflow.RegisterOptions{Name: ingest.ProcessingWorkflowName},
 		)
 		w.RegisterActivityWithOptions(
--- a/enduro.toml
+++ b/enduro.toml
@@ -84,8 +84,8 @@ claimPathSeparator = ""
 # to a value of ["*"] when `claimValuePrefix = "enduro:"`. The default "" will
 # not filter any value.
 claimValuePrefix = ""
-# Consider the values obtained from the claim as roles and use the `rolesMapping`
-# config below to map them to Enduro attributes.
+# Consider the values obtained from the claim as roles and use the 
+# `rolesMapping` config below to map them to Enduro attributes.
 useRoles = false
 # A JSON formatted string specifying a mapping from expected roles to Enduro
 # attributes. JSON format:
@@ -332,6 +332,19 @@ url = "file:///home/enduro/internal-storage/sip-source?metadata=skip"
 # namespace = "default"
 # taskQueue = "postbatch"
 # workflowName = "postbatch"
+
+# ingest configures the ingest service.
+[ingest]
+
+# allowDuplicates toggles whether a SIP can be ingested more than once.
+# The default value (false) will stop ingest with a content error when a SIP
+# archive file (e.g. zip) is submitted that has the same checksum as a
+# previously ingested SIP. SIPs submitted as a directory are not checked for
+# duplicate contents.
+#
+# See https://enduro.readthedocs.io/admin-manual/configuration/#allowduplicates 
+# for more details.
+allowDuplicates = false
 
 # [ingest.storage] configures ingest as a client of the storage API.
 # Use this section for cross-domain integration values:
--- a/internal/datatypes/checksum.go
+++ b/internal/datatypes/checksum.go
@@ -1,0 +1,10 @@
+package datatypes
+
+type ChecksumAlgo string
+
+const ChecksumAlgoSHA256 ChecksumAlgo = "SHA-256"
+
+type Checksum struct {
+	Algorithm ChecksumAlgo
+	Hash      string
+}
--- a/internal/ingest/config.go
+++ b/internal/ingest/config.go
@@ -10,6 +10,24 @@ import (
 )
 
 type Config struct {
+	// AllowDuplicates toggles whether a SIP can be ingested more than once.
+	// The default value (false) will stop ingest with a content error when a
+	// SIP archive file (e.g. zip) is submitted that has the same checksum as a
+	// previously ingested SIP. SIPs submitted as a directory are not checked
+	// for duplicate contents.
+	//
+	// A SIP is only considered a duplicate if the checksum matches an existing
+	// SIP with a status of: "ingested", "pending", "processing", "queued", or
+	// "validated". If the SIP status is "error", "failed" or "canceled" the
+	// SIP will be ignored when checking for duplicates.
+	//
+	// A checksum is calculated and stored for every SIP archive ingested by
+	// Enduro, regardless of this setting. When `allowDuplicates` is false, a
+	// new ingest's checksum will be checked against all previously ingested SIP
+	// checksums, even if `allowDuplicates` was true when the old SIPs were
+	// ingested.
+	AllowDuplicates bool
+
 	Storage StorageConfig
 }
 
--- a/internal/persistence/ent/schema/sip.go
+++ b/internal/persistence/ent/schema/sip.go
@@ -64,6 +64,12 @@ func (SIP) Fields() []ent.Field {
 		field.Int32("file_count").
 			Optional().
 			NonNegative(),
+		field.String("checksum_algorithm").
+			Optional(),
+		// checksum_hash is the hash of a SIP archive (e.g. zip file) before it
+		// is extracted for processing.
+		field.String("checksum_hash").
+			Optional(),
 	}
 }
 
@@ -101,5 +107,7 @@ func (SIP) Indexes() []ent.Index {
 			StorageKey("sip_uploader_id_idx"),
 		index.Fields("batch_id").
 			StorageKey("sip_batch_id_idx"),
+		index.Fields("checksum_algorithm", "checksum_hash").
+			StorageKey("sip_checksum_idx"),
 	}
 }
--- a/internal/workflow/activities/calc_file_checksum.go
+++ b/internal/workflow/activities/calc_file_checksum.go
@@ -1,0 +1,66 @@
+package activities
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/artefactual-sdps/enduro/internal/datatypes"
+)
+
+const CalcFileChecksumActivityName = "calc-file-checksum-activity"
+
+type (
+	CalcFileChecksumActivityParams struct {
+		Path string
+	}
+	CalcFileChecksumActivityResult struct {
+		Checksum datatypes.Checksum
+	}
+	CalcFileChecksumActivity struct{}
+)
+
+func NewCalcFileChecksumActivity() *CalcFileChecksumActivity {
+	return &CalcFileChecksumActivity{}
+}
+
+// Execute calculates the SHA-256 checksum of the file at params.Path.
+//
+// If params.Path is a directory, Execute returns an error.
+func (a *CalcFileChecksumActivity) Execute(
+	ctx context.Context,
+	params *CalcFileChecksumActivityParams,
+) (*CalcFileChecksumActivityResult, error) {
+	fi, err := os.Stat(params.Path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("calculate file checksum: file not found: %s", params.Path)
+		}
+		return nil, fmt.Errorf("calculate file checksum: stat file: %v", err)
+	}
+
+	if fi.IsDir() {
+		return nil, fmt.Errorf("calculate file checksum: not a file: %s", params.Path)
+	}
+
+	f, err := os.Open(params.Path)
+	if err != nil {
+		return nil, fmt.Errorf("calculate file checksum: open file: %v", err)
+	}
+	defer f.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return nil, fmt.Errorf("calculate file checksum: compute hash: %v", err)
+	}
+
+	return &CalcFileChecksumActivityResult{
+		Checksum: datatypes.Checksum{
+			Algorithm: datatypes.ChecksumAlgoSHA256,
+			Hash:      hex.EncodeToString(h.Sum(nil)),
+		},
+	}, nil
+}
--- a/internal/workflow/processing.go
+++ b/internal/workflow/processing.go
@@ -411,6 +411,21 @@ func (w *ProcessingWorkflow) SessionHandler(
 		}
 	}
 
+	// Calculate the SIP checksum if the SIP is not a directory.
+	if !state.sip.isDir {
+		if err := w.calcSIPChecksum(sessCtx, state); err != nil {
+			return err
+		}
+
+		// If duplicate SIPs are not allowed, check if a SIP with the same
+		// checksum has already been ingested or is being processed.
+		if !w.cfg.Ingest.AllowDuplicates {
+			if err := w.checkForDuplicateSIP(sessCtx, state); err != nil {
+				return err
+			}
+		}
+	}
+
 	// Extract the transfer if it's not a directory and a preprocessing child
 	// workflow is not doing the extraction.
 	cfg := w.cfg.ChildWorkflows.ByType(childwf.WorkflowTypePreprocessing)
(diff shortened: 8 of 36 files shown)


djjuhasz:
@jraddaoui this is working and ready for review.  I would recommend reviewing each commit separately, but some of those changes are overwritten later, so it's up to you.

cmd/enduro-a3m-worker/main.go
```diff
@@ -272,6 +272,14 @@ func main() {
 			activities.NewGetSIPExtensionActivity().Execute,
 			temporalsdk_activity.RegisterOptions{Name: activities.GetSIPExtensionActivityName},
 		)
+		w.RegisterActivityWithOptions(
+			activities.NewCalcFileChecksumActivity().Execute,
+			temporalsdk_activity.RegisterOptions{Name: activities.CalcFileChecksumActivityName},
+		)
+		w.RegisterActivityWithOptions(
+			activities.NewCheckDuplicateSIPActivity(perSvc).Execute,
```
jraddaoui: I see why `perSvc` is being passed to this activity. The check is read only, the query already exists in persistence, and calling the Goa shaped `ListSips` method from an activity would not be ideal. I have also argued for this kind of direct use before, so I am not against the idea in principle.

Currently, direct use of `perSvc` does not appear to skip much existing plumbing. `perSvc` is already wrapped with `persistence.WithTelemetry`, so the call still gets a `ListSIPs` span and error recording. Since the check does not change state, it is not skipping an obvious ingest event, audit log entry, or status update side effect right now. That could change if we later add ingest service behavior that should also affect this check, so I would prefer to keep the activity going through `ingestsvc`.

This also lines up with the established activity pattern. Ingest workflow activities are expected to be wired with `ingestsvc`, with persistence reached through that service. There has been deliberate work to keep that shape, even when it means adding some forwarding or duplicate methods to the ingest service.

Could we keep that pattern here and move the duplicate check logic behind an internal method on `ingest.Service`? The activity could still own the workflow step and task handling, while the ingest service owns the SIP lookup rule and the persistence call. For example:

```go
FindDuplicateSIP(ctx context.Context, sipID uuid.UUID, checksum datatypes.Checksum) (*datatypes.SIP, error)
```

I think using `perSvc` directly in activities may be worth a broader discussion, and there may be cases where we decide that is the better direction. I would rather not make that change implicitly as part of this duplicate check, though. For this PR, keeping the activity registered with `ingestsvc` would preserve the existing pattern, and it still leaves us room to improve the query, paging behavior, logging, or telemetry in one place later if needed.

cmd/enduro/main.go
```diff
@@ -508,7 +508,13 @@ func main() {
 
 		// Ingest processing workflow and activities.
 		w.RegisterWorkflowWithOptions(
-			workflow.NewProcessingWorkflow(cfg, rand.Reader, ingestsvc, wsvc).Execute,
+			workflow.NewProcessingWorkflow(
+				cfg,
+				rand.Reader,
+				ingestsvc,
+				perSvc,
```
jraddaoui: We are now passing the persistence service to the workflow, but it's not being used there, right?

internal/ingest/config.go
```diff
@@ -10,6 +10,13 @@ import (
 )
 
 type Config struct {
+	// AllowDuplicates toggles whether a SIP can be ingested more than once.
+	// The default value (false) will stop ingest with a content error when a
+	// SIP archive file (e.g. zip) is submitted that has the same checksum as a
+	// previously ingested SIP. SIPs submitted as a directory are not checked
+	// for duplicate contents.
+	AllowDuplicates bool
```
jraddaoui, line 29: I'd mention the statuses here too.

internal/persistence/ent/schema/sip.go
```diff
@@ -64,6 +64,12 @@ func (SIP) Fields() []ent.Field {
 		field.Int32("file_count").
 			Optional().
 			NonNegative(),
+		field.String("checksum_algorithm").
+			Optional(),
+		// checksum_value is the checksum of the archived SIP (e.g. zip file)
+		// before it is extracted for processing.
+		field.String("checksum_value").
+			Optional(),
```
jraddaoui, line 72: I'm wondering whether we need both `checksum_algorithm` and `checksum_value` for this feature. The current implementation always calculates SHA-256, so the algorithm column does not seem to carry much behavior right now. I can see keeping it if we expect to support multiple checksum algorithms later, but if that is only theoretical, storing just the checksum value may be simpler.

internal/persistence/ent/schema/sip.go
```diff
@@ -101,5 +107,7 @@ func (SIP) Indexes() []ent.Index {
 			StorageKey("sip_uploader_id_idx"),
 		index.Fields("batch_id").
 			StorageKey("sip_batch_id_idx"),
+		index.Fields("checksum_value").
+			StorageKey("sip_checksum_value_idx"),
```
jraddaoui: If we keep `checksum_algorithm` and the duplicate check filters by both algorithm and value, should this be a composite index on both columns?

internal/workflow/activities/calc_file_checksum.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		return nil, fmt.Errorf("calculate file checksum: open file: %v", err)
+	}
+	defer f.Close()
+
+	if fi, err := f.Stat(); err != nil {
+		return nil, fmt.Errorf("calculate file checksum: stat file: %v", err)
+	} else if fi.IsDir() {
+		return nil, fmt.Errorf("calculate file checksum: not a file: %s", params.Path)
+	}
```
jraddaoui: Could we call `os.Stat(params.Path)` before `os.Open`? Stat also returns an error that works with `os.IsNotExist(err)`, so we could handle missing paths and directories before opening the file.

internal/workflow/processing.go
```diff
@@ -411,6 +415,39 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+		}
+
+		// If duplicate SIPs are not allowed, check if a SIP with the same
+		// checksum has already been ingested or is being processed.
+		if !w.cfg.Ingest.AllowDuplicates {
+			if err := w.checkForDuplicateSIP(sessCtx, state); err != nil {
+				return err
+			}
+		}
+	}
```
jraddaoui, line 427: I think we could avoid calculating and persisting the checksum if the configuration is set to allow duplicates. And it would be nice to move it all to its own function, `checkForDuplicateSIP` could do it all.

internal/workflow/processing_setup_test.go
```diff
@@ -89,6 +90,7 @@ func (s *ProcessingWorkflowTestSuite) SetupWorkflowTest(
 
 	ctrl := gomock.NewController(s.T())
 	ingestsvc := ingest_fake.NewMockService(ctrl)
+	persistenceSvc := persistence_fake.NewMockService(ctrl)
```
jraddaoui: We should not need this mock if we are mocking the activities. We probably don't need the ingest service mock either, but we may not be mocking all activities.

enduro.toml
```diff
@@ -329,6 +329,16 @@ url = "file:///home/enduro/internal-storage/sip-source?metadata=skip"
...
 
+# ingest configures the ingest service.
+[ingest]
+
+# allowDuplicates toggles whether a SIP can be ingested more than once.
+# The default value (false) will stop ingest with a content error when a SIP
+# archive file (e.g. zip) is submitted that has the same checksum as a
+# previously ingested SIP. SIPs submitted as a directory are not checked for
+# duplicate contents.
+allowDuplicates = false
```
jraddaoui, line 347: We should mention the statuses here too.

internal/datatypes/checksum.go
```diff
@@ -0,0 +1,10 @@
+package datatypes
+
+type ChecksumAlgo string
+
+const ChecksumAlgoSHA256 ChecksumAlgo = "SHA-256"
+
+type Checksum struct {
+	Algorithm ChecksumAlgo
+	Hash      string
+}
```
jraddaoui, line 10: If we don't care about the algorithm, this could just be the hash string, it would have simplified the implementation a little bit.

jraddaoui (changes requested):
Thanks @djjuhasz!

This is looking great. I also like that you made the configuration flag opt out, which is cleaner than the approach I had used before.

cmd/enduro-a3m-worker/main.go
djjuhasz: I'm okay with using `ingestsvc` and adding a new method, but I'm not a fan of adding single use methods like `FindDuplicateSIP()` to the service API, as I think it leads to an unnecessarily large API surface.  I'm more open to adding a general purpose method analogous to the `persistence.Service` `ListSIPs()` method.  I think the main issue with an ingest service `ListSIPs()` method is what to name it to contrast it with the `ingest.Service` `ListSips()` method; I've suggested the pattern `ListSIPsInternal()` in the past, which is not pretty, but at least clearly indicates that it's returning an internal type (`[]datatypes.SIP`) and not an API type (`goaingest.ListSipsPayload()`).

Despite being open to making this work with `ingestsvc` I really think the better solution here is just to pass the persistence service directly.  The call chain the method doesn't involve the API layer of the application at all, so I think it's better not to use the `ingest.Service` implementation which mixes API and application layer concerns.  I understand that injecting the `persistence.Service` into the activity is breaking the existing application pattern, but I've found that pattern to be limiting, requires creating unnecessary extra code in the `ingest.Service`, and has lead to many conversations similar to this one.  I don't see any harm in using the `persistenceSvc` here, and I think this is a good a time as any to change the de facto standard of not doing so.

cmd/enduro/main.go
djjuhasz: Oops.  I originally implemented `CheckDuplicateSIPActivity` as a local activity, which required passing `perSvc` at run time, but that's no longer necessary.

internal/persistence/ent/schema/sip.go
djjuhasz: I added the algorithm to future-proof the implementation.  I think it's inevitable that we'll have to add more checksum algorithms in the future, either by client request or because "SHA-256" is found to have security weaknesses, or just ages out due to newer crypto innovations.  I figure we might as well add the algorithm now and avoid having to update signatures and the database fields later on.

internal/persistence/ent/schema/sip.go
djjuhasz: Yes, good idea.

internal/workflow/processing.go
djjuhasz: I consciously decided to calculate and persist the checksum even when allowing duplicates so if the institutional policy changes later on to forbid duplicates then new SIPs could be checked against old SIPs.  It might also be helpful to have the checksum if an institution wanted to track duplicates or to provide a duplicate warning, or implement some other mechanism to discourage ingesting duplicates.

cmd/enduro-a3m-worker/main.go
jraddaoui: That makes sense, and I agree that `ingest.Service` already mixes API and application concerns in a way that makes this awkward.

My hesitation with `ListSIPsInternal()` is that it feels like a larger API surface than `FindDuplicateSIP()`. It would expose a general persistence style list operation through the ingest service, while the duplicate check only needs one specific question answered. Even if `FindDuplicateSIP()` is single use today, it keeps the call site narrower and avoids making the activity aware of filters, paging, and which SIP statuses count as duplicates.

I am not opposed to revisiting the pattern and allowing activities to depend on `persistence.Service` directly. I just think that is a broader design decision than this PR. If we want to change that standard, I would rather do it explicitly and consistently instead of introducing it through one activity.

For this change, my preference is still to keep the worker wiring aligned with the current pattern and put the duplicate check behind `ingestsvc`. But I agree the current `ingest.Service` shape is not ideal, and we should probably revisit whether it should keep combining Goa facing methods and internal workflow methods.

internal/persistence/ent/schema/sip.go
jraddaoui: That makes sense, I'm fine with keeping `checksum_algorithm`. It also makes the stored checksum self describing, which is useful even if the current implementation only writes SHA-256.

One distinction: this checksum is currently being used for duplicate detection, not as a security boundary. So I don't think SHA-256 aging out is a direct concern for this specific use case, but I agree that storing the algorithm makes future changes easier.

internal/workflow/processing.go
jraddaoui: I can see the value in storing the checksum as metadata even when duplicates are allowed, especially if we later add reporting or warnings around duplicate content.

One detail worth clarifying: if an institution has `allowDuplicates = true` for a while and later disables it, should new SIPs be checked against SIPs ingested during the earlier "duplicates allowed" period? One could argue the new policy should apply only from the point it is enabled, which is the behavior I would probably expect. I am fine either way, but the current approach seems to assume the stricter behavior: once duplicates are disabled, new SIPs are checked against all known checksums, including SIPs ingested while duplicates were allowed.

There is also a related migration detail: we are not calculating checksums for existing SIPs when the columns are added, so this future policy change would only be able to check against SIPs ingested after this feature is deployed. That is probably fine, but it is worth being explicit about the behavior.

djjuhasz: I would expect that when an institution chooses to disallow duplicate SIPs they will want to check against all previously ingested SIPs, if possible.  I might be wrong in this assumption though, so I'm open to changing the behaviour if the team consensus is that it's better not to record checksums at all when `allowDuplicates = true`.

In the migration case I think there's a high likelihood that Enduro doesn't have access to the old SIPs to calculate checksums during the migration, so I don't think it's worth adding code to calculate legacy checksums at upgrade time.  If we do have checksums for SIPs ingested before this feature was added (like in the SFA case) I think it would be good to migrate the legacy checksums to the SIP table if possible.

internal/persistence/ent/schema/sip.go
djjuhasz: I agree that we don't need a cryptographically secure hash algorithm for a checksum, but I also know that some IT departments are very uncomfortable with insecure algorithms. :sweat_smile:

internal/workflow/processing.go
jraddaoui: I'm not expecting us to add a migration that calculates checksums for previous SIPs, and importing legacy checksums when they already exist seems like a separate best effort case. There also are not many production-like sites yet, so this may be mostly a policy question for now.

The migration gap was more another argument for only recording checksums once duplicate checks are enabled, if we want the policy to apply from that point forward. I do not have a strong opinion either way, but it would be good to check with the team which behavior we want.

sallain: I've polled the analysts and the unanimous vote is to check against all previously ingested SIPs. I think that this makes sense, and the small trade-off of having to compute and store those checksums is reasonable.

I suggested a hypothetical case where an institution sets `allowDuplicates = true` for testing, but then can't ingest the packages they used for testing, but the resounding response was "Don't use prod for testing" so I think, as it currently stands, we can disregard that use case.

cmd/enduro-a3m-worker/main.go
djjuhasz: Okay, done.

djjuhasz:
@jraddaoui this is ready for another look.  I'll have to squash the whole PR into a single commit and then combine the two database migrations into one before it gets merged. :(

internal/workflow/processing.go
```diff
@@ -411,6 +411,39 @@ func (w *ProcessingWorkflow) SessionHandler(
...
+
```
jraddaoui, line 428: I think you missed this in the previous comments, it would be nice to move it all to its own function to avoid the cognitive load, `checkForDuplicateSIP` could do it all. If you still want the separation between the calculation and the check, we could at least persist the SIP inside `calcSIPChecksum`.

enduro.toml
```diff
@@ -329,6 +329,21 @@ url = "file:///home/enduro/internal-storage/sip-source?metadata=skip"
...
+# The default value (false) will stop ingest with a content error when a SIP
+# archive file (e.g. zip) is submitted that has the same checksum as a
+# previously ingested SIP. SIPs submitted as a directory are not checked for
+# duplicate contents.
+# 
+# A SIP is only considered a duplicate if the checksum matches an existing
+# SIP with a status of: "ingested", "pending", "processing", "queued", or
+# "validated". If the SIP status is "error", "failed" or "canceled" the
+# SIP will be ignored when checking for duplicates.
+allowDuplicates = false
```
jraddaoui, line 347: I think we should document the decision we made about the check. We should note that even if system is set to allow duplicates, the checksum is calculated and that if the setting is disabled later, the check will consider all previous checksums.

Also, this should be part of the documentation, not only in this example config comments. What I did before for the buckets config is to add a link to the docs directly from here, to avoid the duplication.

jraddaoui (approved):
Thanks @djjuhasz, looks great to me!

I would be great to document this in the admin manual, but that can happen in another PR.

internal/workflow/processing.go
djjuhasz: I've moved the update of the SIP with the checksum data into the `CalculateFileChecksumActivity`.  I think it makes sense to keep checksum calculation and duplicate check in separate functions because the duplicate check has the additional condition on being run.  Having the two separate functions also makes it easier to handle the creating and updating the preservation tasks.

jraddaoui: Sounds good! I'd have kept the activities separated, I've seen three versions of the calculate checksum activity already, so it should probably go into temporal-activities at some point but. But not a big deal for now.

djjuhasz: @jraddaoui I'm confused, so you just wanted me to move the `updateSIPLocalActivity` call into the `w.calcSIPChecksum()` function, not move it into the `CalculateFileChecksumActivity`?

jraddaoui: Yeah, sorry if that was not clear.

enduro.toml
djjuhasz: @jraddaoui I added a section for "allowDuplicates" to the admin manual and included a description of how the checksums stored even when "allowDuplicates" is true.

I went half-and-half on adding a link to the manual from the config file — I included a short description of the setting and then the link for more details.  I also updated the code comments in the ingest config.

jraddaoui: Looks good, thanks!!

internal/workflow/processing.go
djjuhasz: @jraddaoui I deleted the commit where I moved the checksum save into the `CalculateFileChecksumActivity` and have moved it into the `w.calcSIPChecksum()` method.

internal/workflow/processing_test.go
```diff
@@ -1087,3 +1131,179 @@ func (s *ProcessingWorkflowTestSuite) TestBatchSignalDoNotContinue() {
 		BatchUUID:       batchUUID,
 	}, nil, true)
 }
+
+// TestDuplicateSIP tests:
+// - Archivematica as preservation system.
+// - The "create AIP" workflow type.
+// - duplicate SIP error.
```
jraddaoui: Uppercase "Duplicate...".

internal/workflow/processing_test.go
```diff
@@ -1087,3 +1131,179 @@ func (s *ProcessingWorkflowTestSuite) TestBatchSignalDoNotContinue() {
...
+		Ingest: ingest.Config{
+			AllowDuplicates: true,
+			Storage:         ingest.StorageConfig{DefaultPermanentLocationID: locationID},
+		},
+	}, nil)
+
+	params := defaultParams()
+	downloadExpectations(s, params)
+	calcChecksumExpectations(s, params)
+	expectations["saveChecksum"](s, params)
```
jraddaoui: Why not putting `expectations["saveChecksum"](s, params)` inside `calcChecksumExpectations`? It looks like we are calling both all times.

internal/workflow/processing_test.go
```diff
@@ -1087,3 +1131,179 @@ func (s *ProcessingWorkflowTestSuite) TestBatchSignalDoNotContinue() {
...
+	expectations["archiveExtract"](s, params)
+	expectations["classifySIP"](s, params)
+	CountSIPFilesExpectations(s, params)
```
jraddaoui: I think we missed this in previous work. I doubt `CountSIPFilesExpectations` needs to be a public function.

jraddaoui (approved):
Thanks @djjuhasz! Really minor feedback below, feel free to ignore.

Issue #1666, opened 2026-06-16
Problem: Child workflows lack user context


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Child workflows currently receive SIP, AIP, batch, and custom metadata context, but they do not receive information about the authenticated Enduro user who started the workflow. Some child workflow implementations need a minimal user identifier to perform actions against external systems, attribute downstream actions, or keep their own audit records of who initiated an operation.

This is only relevant for workflows started by a user action. Watched bucket/filesystem workflows are system initiated and should not fabricate or attach user information.

**Describe the solution you'd like**

Pass a minimal user context to every child workflow contract in `pkg/childwf`, including preprocessing, poststorage, and postbatch child workflows.

For the first implementation, include only the user's email address. Do not pass sensitive OIDC values such as `sub`, `iss`, raw claims, access tokens, or any other authentication material. If authentication is disabled, the email claim is unavailable, or the workflow was started by a watched bucket or filesystem event, the child workflow user context should be empty.

**Describe alternatives you've considered**

One alternative is to pass the existing persisted Enduro user record or full authentication claims to child workflows. That would expose more data than child workflows currently need and risks leaking sensitive OIDC identifiers such as `iss` and `sub`.

Another option is to rely on child workflows to query Enduro for user information using SIP or batch identifiers. That makes child workflows more tightly coupled to Enduro APIs and still requires a clear, safe identity boundary. Passing only the email directly keeps the contract explicit and minimal.

**Additional context**

Needed by https://github.com/artefactual-sdps/sfa-enduro-workflows/issues/173#issuecomment-4614652554.

PR #1667, opened 2026-06-16
Pass user email to child workflows

jraddaoui:
Add a user value with only email to child workflow request types. Populate it from auth claims when authentication is enabled and users start ingest from the API upload or a SIP source.

Refs #1666.


djjuhasz (approved):
LGTM! 👍

Issue #1660, opened 2026-06-05
Problem: MinIO is no longer maintained


jraddaoui:
**Is your feature request related to a problem? Please describe.**

Enduro still has several defaults, examples, tests, and runtime assumptions that are tied to [MinIO](https://github.com/minio/minio). MinIO is no longer a stable dependency to build those defaults around: the `minio/minio` repository is archived and its README states that the repository is no longer maintained, with the community edition now distributed only as source code.

This makes MinIO a risky prerequisite for local development, CI, sample deployments, and common production setups that do not otherwise need object storage with S3 support.

**Describe the solution you'd like**

Make Enduro no longer depend on MinIO as the default or required service for internal storage.

Related PRs under this issue may include any work that reduces the default reliance on MinIO, makes local file system storage a better path where it fits, keeps object storage with S3 support available as an optional backend, or clarifies how Enduro should validate and document the storage choices that remain.

The goal is not to remove storage with S3 support from Enduro. The goal is to stop treating MinIO as the default foundation for ordinary Enduro operation.

**Describe alternatives you've considered**

We could keep using a pinned MinIO version, but that leaves Enduro depending on an upstream that is no longer maintained.

We could replace MinIO with another service that supports S3, such as SeaweedFS, Garage, Ceph, or RustFS. That may still be useful for deployments that need object storage or ingress through S3, but it does not by itself solve the broader problem of Enduro being coupled to an object storage service for workflows that can run on a local filesystem.

We may also add SeaweedFS to the dev environment later if using one shared persistent volume proves too awkward or limiting. That should be treated as a practical dev option, not as a requirement for normal Enduro use.

We could remove object storage support entirely, but that would break valid large, distributed, and cloud deployments. Storage with S3 support should remain available where it is the right backend.

**Additional context**

This is intended as a small tracking issue for work that reduces the dependency on MinIO. PRs can reference it when they remove behavior tied to MinIO, make storage on a file system a better default, document supported alternatives, or preserve S3 support in a more optional way.

Detailed background and notes are captured in this Gist: https://gist.github.com/sevein/0cb4b7656a970885abdb98a1c06609bf

PR #1668, opened 2026-06-16
Remove storage AIP submit API and upload workflow

jraddaoui:
Remove the storage submit AIP endpoints, auth scope, service handlers, and obsolete upload workflow. AIPs are now staged between ingest and storage without this public API path.

Refs #1660.


djjuhasz (approved):
LGTM! Thank @jraddaoui. 👍

PR #1669, opened 2026-06-17
Document, fix and enable filesystem watchers

jraddaoui:
**Document filesystem watchers and enable one in dev**

Document filesystem watcher settings and behaviour. Configure a
filesystem watched location in the dev environment. Create the watched
and completed directories in the Docker image and Kubernetes manifests.

**Fix filesystem watcher SIP downloads**

Make filesystem watcher downloads write to the destination path set
by the activity, where the key is already appended to the destination.

**Run SIP cleanup on preservation workers**

Move original SIP deletion and completed directory disposal into the
preservation session so filesystem-backed sources are handled by the
worker that has access to their local paths.

Register the cleanup activities on the AM and A3M preservation workers,
and remove the redundant registrations from the main Enduro workflow
worker. Update workflow and filesystem watcher tests to cover
completed directory disposal for SIPs.

Refs #1660.

### DIFF
--- a/internal/watcher/filesystem_test.go
+++ b/internal/watcher/filesystem_test.go
@@ -163,6 +163,35 @@ func TestFileSystemWatcher(t *testing.T) {
 		)))
 	})
 
+	t.Run("Download copies a file to the destination path", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := t.Context()
+
+		src := fs.NewDir(t, "enduro-test-fswatcher",
+			fs.WithFile("sip.zip", "A test file."),
+		)
+		dest := fs.NewDir(t, "enduro-test-fswatcher")
+
+		w, err := watcher.NewFilesystemWatcher(ctx, &watcher.FilesystemConfig{
+			Name: "filesystem",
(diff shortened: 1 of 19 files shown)


internal/watcher/filesystem_test.go
```diff
@@ -156,10 +158,25 @@ func TestFileSystemWatcher(t *testing.T) {
...
+		assert.NilError(t, err)
+
+		_, err = os.Stat(srcPath)
+		assert.Assert(t, os.IsNotExist(err))
+
+		destPath := filepath.Join(dest.Path(), "example-sip")
+		info, err = os.Stat(destPath)
+		assert.NilError(t, err)
+		assert.Assert(t, info.IsDir())
 		assert.Assert(t, fs.Equal(dest.Path(), fs.Expected(t,
```
djjuhasz, line 172: This tests that `example-sip` exists, is a directory, and has the correct contents, which makes lines 172-175 redundant.

djjuhasz (approved):
LGTM @jraddaoui. 👍 I flagged a few lines that I think are redundant.

PR #1676, opened 2026-06-20
Upgrade temporal kube config to v1.2.0

djjuhasz:
Rebuild `hack/kube/base/temporal.yaml` using the upgraded v1.2.0 temporal helm charts.


jraddaoui (approved):
Nice, thanks @djjuhasz!

PR #1677, opened 2026-06-22
Use `bucket.Attributes` to check if AIP exists

djjuhasz:
Call `bucket.Attributes()` to check if an AIP exists instead of trying to open a blob reader to avoid downloading the entire AIP.

When using `ssblob` as the `blob.Bucket` implementation to open a reader on an AIP file, it immediately downloads the entire file via the HTTP connection. Because `DownloadAipRequest()` was opening a blob reader to confirm if the AIP file existed, the entire AIP was being downloaded when the AIP was stored in the AMSS, which caused the request to time out.

- Implement the `ssblob.Attributes()` method using the AMSS `GET api/v2/file/{UUID}` endpoint
- Return a "Not Found" error from `ssblob.Attributes()` if the AIP doesn't exist in the AMSS database, or if the AIP has been deleted
- Call `bucket.Attributes()` in the storage service `DownloadAipRequest()` method to check if the AIP exists
- Add the storage service `openAIPbucket()` helper method reduce repetition and complexity in the code that opens the bucket where an AIP is stored.
- Remove the `aipLocation()` method because it was only called once, in the `openAIPbucket()` method and wasn't doing that much any more

### DIFF
--- a/internal/storage/download_aip.go
+++ b/internal/storage/download_aip.go
@@ -28,19 +28,24 @@ func (s *serviceImpl) DownloadAipRequest(
 		return nil, goastorage.MakeNotValid(errors.New("AIP is not available for download"))
 	}
 
-	// Check if the failed AIP exists in the location bucket.
-	reader, err := s.AipReader(ctx, aip)
+	bucket, err := s.openAIPBucket(ctx, aip)
+	if err != nil {
+		s.logger.Error(fmt.Errorf("open AIP bucket: %v", err), "AIP UUID", aip.UUID)
+		return nil, goastorage.MakeInternalError(errors.New("error opening AIP storage location"))
+	}
+	defer bucket.Close()
+
+	_, err = bucket.Attributes(ctx, aip.UUID.String())
 	if err != nil {
 		if gcerrors.Code(err) == gcerrors.NotFound {
 			return nil, &goastorage.AIPNotFound{
 				UUID:    aip.UUID,
-				Message: "AIP file not found in the location bucket",
+				Message: "AIP not found",
 			}
 		} else {
-			return nil, goastorage.MakeInternalError(errors.New("error checking AIP file"))
+			return nil, goastorage.MakeInternalError(errors.New("error getting AIP attributes"))
 		}
 	}
-	reader.Close()
 
 	// Request a ticket.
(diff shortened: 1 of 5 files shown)


jraddaoui (commented):
Thanks @djjuhasz!

This looks good overall. My only concern is the recent filesystem source locations that could be used to store AIPs in the a3m path. Would this work without an attributes file alongside the AIP?

I believe we do not have documentation for setting up the default AIP storage location in Enduro, but we create one by default with `metadata=skip` for the a3m setup in the dev environment:

https://github.com/artefactual-sdps/enduro/blob/main/hack/kube/overlays/dev-a3m/mysql-create-a3m-location-job.yaml#L30

So it should be easy to check how this works by testing with a3m. My expectation is that it will not work without the attributes sidecar. If that is the case, we should change the default URL value, and we should probably think about documenting that storage concern.

Right now, creating and configuring a default final AIP storage location is a requirement, but we are not documenting it anywhere.

djjuhasz:
@jraddaoui I just tested downloading an AIP from Enduro using the Tilt dev environment and a3m as the preservation system, and the AIP downloads correctly.  I double checked the `enduro_storage.location` config, and it does use the "metadata=skip" attribute:

```
mysql> select * from location\G
*************************** 1. row ***************************
         id: 1
       name: perma-aips
description: 
     source: filesystem
    purpose: aip_store
       uuid: 48a27da2-9b26-47b2-9aa1-bc572c6cf67f
     config: {"url": {"url": "file:///home/enduro/internal-storage/perma-aips?metadata=skip&no_tmp_dir=true"}}
 created_at: 2026-06-22 19:20:45
1 row in set (0.00 sec)
```

The documentation for the `fileblob` package seems to confirm that `blob.Attributes` will work without the "sidecar" file, it will just use defaults for unknown values:

> By default fileblob stores blob metadata in "sidecar" files under the original filename with an additional ".attrs" suffix. This behaviour can be changed via `Options.Metadata`; writing of those metadata files can be suppressed by setting it to `MetadataDontWrite` or its equivalent "metadata=skip" in the URL for the opener. **In either case, absent any stored metadata many `blob.Attributes` fields will be set to default values.** 

-- https://pkg.go.dev/gocloud.dev/blob/fileblob [emphasis mine]

jraddaoui (approved):

PR #1682, opened 2026-06-24
Log API requests and responses

djjuhasz:
Replace the deprecated Goa middleware logger with
https://github.com/go-chi/httplog, and log all API HTTP requests and responses even when `debug` is false.

- Log success response statuses (100-399, 429) at level 2
- Log error response statuses (400-428, 430-599) at level 0 (highest)
- If the `API.Debug` config is true, then the request and response bodies will also be included in the logs

The log level to HTTP status code mapping is mostly hard coded by `httplog`, with the only difference being the loss of the intermediate "warning" level for 4XX level errors in the conversion between [slog](https://pkg.go.dev/log/slog) and [logr](https://github.com/go-logr/logr) log level semantics.

**Alternatives considered:**
- I tried creating a custom `logr.Logsink` implementation to better map the slog log levels to logr log levels, but didn't have any success
- I tried using a new `slog.JSONHandler` instance just for the API logs, independent of the standard `logr.Logger` instance we are using for the rest of the application, but the difference in log formatting was jarring
- I considered completely replacing our `log.Logger` `logger` instance with a `slog.Logger` instance, but this is a big change that isn't justified by the current need to replace the deprecated Goa logger
- I also experimented with several configurations for `httplog` but settled on the current config as a good balance between logging enough data to help with debugging the request/response cycle and providing some options to increase or decrease the volume of logging as needed


djjuhasz:
Here's what the API logs look like in the Tilt console output:
image

Text output
```
2026-06-24T19:27:13.724Z	V(2)	enduro.api	v3@v3.4.0/middleware.go:169	GET /storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2 => HTTP 200 (1.137667ms)	{"url.full": "[http://backend/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2",](http://backend/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2%22,) "http.request.method": "GET", "url.path": "/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2", "client.ip": "10.42.0.13:52310", "url.domain": "backend", "url.scheme": "http", "http.version": "HTTP/1.1", "http.request.body.bytes": 0, "user_agent.original": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:152.0) Gecko/20100101 Firefox/152.0", "http.request.referrer": "[http://localhost:8080/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2",](http://localhost:8080/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2%22,) "http.response.status_code": 200, "event.duration": 1137667, "http.response.body.bytes": 228}
2026-06-24T19:27:13.731Z	V(2)	enduro.api	v3@v3.4.0/middleware.go:169	GET /storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2/workflows => HTTP 200 (1.499664ms)	{"url.full": "[http://backend/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2/workflows",](http://backend/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2/workflows%22,) "http.request.method": "GET", "url.path": "/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2/workflows", "client.ip": "10.42.0.13:52310", "url.domain": "backend", "url.scheme": "http", "http.version": "HTTP/1.1", "http.request.body.bytes": 0, "user_agent.original": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:152.0) Gecko/20100101 Firefox/152.0", "http.request.referrer": "[http://localhost:8080/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2",](http://localhost:8080/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2%22,) "http.response.status_code": 200, "event.duration": 1499664, "http.response.body.bytes": 3}
2026-06-24T19:27:14.607Z	V(0)	enduro.api	v3@v3.4.0/middleware.go:169	POST /storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2/download => HTTP 404 (17.741859ms)	{"url.full": "[http://backend/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2/download",](http://backend/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2/download%22,) "http.request.method": "POST", "url.path": "/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2/download", "client.ip": "10.42.0.13:36372", "url.domain": "backend", "url.scheme": "http", "http.version": "HTTP/1.1", "http.request.body.bytes": 0, "user_agent.original": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:152.0) Gecko/20100101 Firefox/152.0", "http.request.referrer": "[http://localhost:8080/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2",](http://localhost:8080/storage/aips/b8da7b51-d1a6-421c-9721-d084a50bacf2%22,) "http.response.status_code": 404, "event.duration": 17741859, "http.response.body.bytes": 74}
```

djjuhasz:
@jraddaoui @sevein this is ready for another review. The API logs are now configured and logged separately from the application logs and use the `slog` logger and log handler (sink).  I've also removed the request and response body logging, and changed the API "debug" meaning to match the application log "debug" behaviour.

docs/src/admin-manual/configuration.md
```diff
@@ -152,15 +152,29 @@ port should **not be exposed publicly**.
 ```toml
 [internalapi]
 listen = "0.0.0.0:9002"
-debug = false
+logPath = "stdout"
+logLevel = "warn"
+debug = true
+serviceName = "enduro.api.internal"
```
sevein: What use case requires serviceName to be configurable? If it is only meant to distinguish the internal and external APIs, perhaps Enduro could derive it instead. If configurability is on purpose, could the documentation clarify when an operator should override it?

docs/src/admin-manual/configuration.md
```diff
@@ -152,15 +152,29 @@ port should **not be exposed publicly**.
 ```toml
 [internalapi]
 listen = "0.0.0.0:9002"
-debug = false
+logPath = "stdout"
+logLevel = "warn"
+debug = true
```
sevein: I'd love to revisit this naming at some point. I think `debug` was a poor decision on my end in legacy Enduro that has been carried forward. Something explicit like `logFormat = "json"` or `"text"` would be nicer.

cmd/enduro/main.go
```diff
@@ -352,6 +352,17 @@ func main() {
...
+	// Set up the API logger if LogPath is set. The API logger is used for
+	// logging API HTTP requests and responses.
+	apiLogger, logFile, err := api.NewLogger(cfg.API)
+	if err != nil {
+		logger.Error(err, "Failed to open API log file")
+		os.Exit(1)
+	}
+	if logFile != nil {
+		defer logFile.Close()
+	}
```
sevein: Non-blocking thought: we could return only an `io.Closer`, or, for stronger encapsulation, define an `AccessLogger` that owns the logger and its output and exposes `Close` itself.

docs/src/admin-manual/configuration.md
```diff
@@ -152,15 +152,29 @@ port should **not be exposed publicly**.
 ```toml
 [internalapi]
 listen = "0.0.0.0:9002"
-debug = false
+logPath = "stdout"
```
sevein: Non-blocking thought, but a future improvement maybe? With `LogPath`, `stdout` or `stderr` are magic values. It looks like other projects do this differently:

* [Caddy] exposes `output = stdout | stderr | file <path>`,
* [PostgreSQL] has `log_destination` plus separate `log_directory`/`log_filename` settings.

Could we use something like `logDestination = stdout | stderr | file`, with `logPath` only applying to `file`?

[Caddy]: https://caddyserver.com/docs/caddyfile/directives/log
[PostgreSQL]: https://www.postgresql.org/docs/current/runtime-config-logging.html

internal/api/logger.go
```diff
@@ -0,0 +1,78 @@
...
+	default:
+		f, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o640)
+		if err != nil {
+			return nil, err
+		}
+		lf.File = f
+	}
+
+	return &lf, nil
+}
```
sevein, line 93: This could save us from having to implement `LogFile`:

```go
type nopCloser struct{}

func (nopCloser) Close() error {
    return nil
}

func openLogOutput(path string) (io.Writer, io.Closer, error) {
    switch strings.ToLower(path) {
    case "stdout":
            return os.Stdout, nopCloser{}, nil
    case "stderr":
            return os.Stderr, nopCloser{}, nil
    default:
            f, err := os.OpenFile(
                    path,
                    os.O_APPEND|os.O_CREATE|os.O_WRONLY,
                    0o640,
            )
            if err != nil {
                    return nil, nil, err
            }

            return f, f, nil
    }
}
```

Not needed if you end up building an `AccessLogger` (mentioned earlier).

docs/src/admin-manual/configuration.md
```diff
@@ -152,15 +152,29 @@ port should **not be exposed publicly**.
...
+serviceName = "enduro.api.internal"
 ```
 
 * `listen`: tells Enduro what IP address and port to use for internal API
   communication across components.
-* `debug`: Enables (set to `true`) or disables (set to `false`) debug mode for
-  the internal API. When enabled, debug prints detailed information about
-  incoming requests and outgoing responses, including all headers, parameters,
-  and bodies.
+* `logPath`: sets the path for internal API logs. If `logPath` is set to a file
```
sevein: I guess this should read "path for API logs" instead of "path for internal API logs".

internal/api/config.go
```diff
@@ -1,10 +1,66 @@
...
+
+	"github.com/artefactual-sdps/enduro/internal/auth"
+)
 
 type Config struct {
-	Listen     string
-	Debug      bool
-	Auth       auth.Config
+	// Listen defines the address and port on which the API server will listen
+	// for incoming requests. The default value is "127.0.0.1:9000".
```
sevein: Could we describe this as a "TCP address in host:port form"? "Address and port" sounds like two separate settings and does not make the expected format explicit.

internal/api/logger.go
```diff
@@ -0,0 +1,78 @@
...
+	var h slog.Handler
+	if cfg.Debug {
+		h = slog.NewTextHandler(w, nil)
+	} else {
+		h = slog.NewJSONHandler(w, nil)
+	}
+
+	if cfg.ServiceName != "" {
+		h = h.WithAttrs([]slog.Attr{slog.String("service", cfg.ServiceName)})
+	}
```
sevein: According to the [field reference](https://www.elastic.co/docs/reference/ecs/ecs-service#field-service-name), it should be named `service.name`.

internal/api/logger.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var h slog.Handler
+	if cfg.Debug {
+		h = slog.NewTextHandler(w, nil)
+	} else {
+		h = slog.NewJSONHandler(w, nil)
+	}
```
sevein: Could we pass shared `slog.HandlerOptions` with `Level: cfg.LogLevel` to both handlers? With `nil` options, both default to `INFO`, so `logLevel = "debug"` still drops DEBUG/OPTIONS records even though `logHandler` configures `httplog` at DEBUG. The tests don't exercise DEBUG logging or an OPTIONS request, so they don't catch this.

sevein (approved):
Just one minor issue where I think we're dropping DEBUG/OPTIONS records and small number of suggestions/observations but no blockers.

docs/src/admin-manual/configuration.md
djjuhasz: `serviceName` is essentially a cosmetic choice IMHO, it doesn't change functionality it just makes it possible to distinguish the source of the log messages in the cases where the API logs, internal API logs, and possibly the application logs are all being written to the same output stream.  

I'm fine with hard coding the service name like we have for the various Enduro components in the app logger (using `logger.WithName()`) rather than having it configurable.  

The only use case I can think of for allowing configuration of the name, aside from personal preference, is if both the internal API and API logs are being written to their own private log file in which case it's not necessary to have a service name at all, since there is only one source for the log messages.  The separate log file outlier probably isn't worth the overhead of a config setting though, and even in that case the service name might be beneficial if the log files are being aggregated somewhere.

docs/src/admin-manual/configuration.md
djjuhasz: Yeah, I like `logFormat` better too.  I think I'll just go ahead and use that for the API logs.

I'll have to test whether I can just remove the `debug` setting from the config altogether or if that will cause errors if `debug` is included in the config file.  I think it will be fine to just drop it.  I originally had a deprecation warning, but that's probably overkill.

internal/api/logger.go
djjuhasz: Yes. Good catch!

internal/api/logger.go
djjuhasz: Haha, I had no idea https://www.elastic.co/docs/reference/ecs/ecs-service defined a "service.name" field, I was just grasping for a good name.

internal/api/config.go
djjuhasz: Sure.

docs/src/admin-manual/configuration.md
djjuhasz: When I was working on this PR, I looked at the [Nginx error_log](https://nginx.org/en/docs/ngx_core_module.html#error_log) which supports `error_log = stderr | syslog: | memory: | <filepath>`.

I prefer the current "magic values" approach to requiring two configuration fields (destination and path), with the path being ignored for stderr and stdout destinations.  That's just my preference though.

djjuhasz: Log rotation is another future improvement to consider. It looks like Caddy uses https://github.com/DeRuina/timberjack for log rotation.  I used https://github.com/natefinch/lumberjack when I added an audit log to Enduro. It looks like timberjack is forked from lumberjack, with some nice extra features.

docs/src/admin-manual/configuration.md
djjuhasz: No, this description is specifically for the internal API configuration, and the logPath can be different for the public API and the internal API logs.

docs/src/admin-manual/configuration.md
```diff
@@ -1564,6 +1622,7 @@ workflowName = "custom-postbatch-workflow"
 [Archivematica processing configuration fields]: https://archivematica.org/docs/latest/user-manual/administer/dashboard-admin/#processing-configuration-fields
 [Archivematica transfer with existing checksums]: https://www.archivematica.org/en/docs/latest/user-manual/transfer/transfer/#create-a-transfer-with-existing-checksums
 [Azure]: https://azure.microsoft.com/en-us/products/storage/blobs/
+[ElasticSearch Common Schema (ECS)]: https://www.elastic.co/docs/reference/ecs
```
sevein:
```suggestion
[Elastic Common Schema (ECS)]: https://www.elastic.co/docs/reference/ecs
```

internal/api/logger.go
```diff
@@ -0,0 +1,125 @@
...
+	} else if name != "" {
+		fl.name = name
+	}
+
+	return &fileLogger{
+		Logger: fl.With(slog.String("service.name", fl.name)),
+		f:      fl.f,
+		name:   fl.name,
+	}
+}
```
sevein: This is mutating `fl.name` and adding another `service.name` attribute to the existing logger on every call, e.g.: `WithName("enduro").WithName("api")` produces duplicate `service.name` keys.

I don't think this is a blocker today because the production call sites currently call WithName only once, but one way to avoid the issue would be to pass the complete service name to the constructor and attach it once:

```go
func NewFileLogger(cfg LogConfig, serviceName string) (*fileLogger, error) {
	// Create the handler as usual.

	logger := slog.New(h)
	if serviceName != "" {
		logger = logger.With("service.name", serviceName)
	}

	return &fileLogger{
		Logger: logger,
		f:      f,
	}, nil
}
```

Or I guess you could keep a reference to the base undecorated logger and avoid modifying the original logger, for example:

```go
func (fl *fileLogger) WithName(name string) *fileLogger {
	fullName := fl.name

	if name != "" {
		if fullName != "" {
			fullName += "."
		}
		fullName += name
	}

	logger := fl.base
	if fullName != "" {
		logger = logger.With(
			slog.String("service.name", fullName),
		)
	}

	return &fileLogger{
		Logger: logger,
		base:   fl.base,
		f:      fl.f,
		name:   fullName,
	}
}
```

internal/api/logger_test.go
```diff
@@ -0,0 +1,161 @@
...
+				assert.NilError(t, logger.Close())
+				return
+			}
+
+			logger.Log(context.Background(), tc.cfg.Level, "Test log message")
+
+			data, err := os.ReadFile(tc.cfg.Path)
+			assert.NilError(t, err)
+
+			var msg map[string]any
```
sevein, line 106: Unmarshalling into a map here keeps only the last value, so the test passes without revealing the duplicates I mentioned earlie! An additional check that we could have to ensure that we only have one occurence is:

```go
assert.Equal(t, strings.Count(string(data), `"service.name":`), 1)
```

internal/api/logger.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var h slog.Handler
+	switch cfg.Format {
+	case LogFormatText:
+		h = slog.NewTextHandler(f, &slog.HandlerOptions{Level: cfg.Level})
+	default:
+		h = slog.NewJSONHandler(f, &slog.HandlerOptions{Level: cfg.Level})
```
sevein: I don't know if we want to go this far, but this handler is going to emit `time`, `level` and `msg` fields instead of the ECS fields `@timestamp`, `log.level`, and `message`, so the resulting output is only partially ECS-compatible. We could define the options as:

```go
opts := &slog.HandlerOptions{
    Level:       cfg.Level,
    ReplaceAttr: httplog.SchemaECS.ReplaceAttr,
}
```

That said, I don't think we need to look for ECS-compliance, we could also just ignore this for now.

internal/config/config.go
```diff
@@ -84,7 +93,7 @@ type Configuration struct {
 func (c *Configuration) Validate() error {
 	return errors.Join(
 		c.A3m.Validate(),
-		c.API.Auth.Validate(),
+		c.API.Validate(),
```
sevein, line 96: Could we also call `c.InternalAPI.Validate()` here and add a test for an invalid internal API log format?

sevein (approved):
LGTM overall. I left a few additional observations.
I used GPT-5.6-Sol (xhigh) to review the changes locally.

docs/src/admin-manual/configuration.md
djjuhasz: I've removed this altogether as the log format isn't 100% ECS compliant, as you pointed out below.

internal/api/logger.go
djjuhasz: I've dropped the mention of ECS compliance in the manual because of these non-compliant fields.

internal/api/logger.go
djjuhasz: I've updated `WithName()` to return a new fileLogger with a new handler when the name is changed.

PR #1683, opened 2026-06-24
Use a shared bagit-gython validator

djjuhasz:
Revert commit 362ed76 because [bagit-gython](https://pkg.go.dev/github.com/artefactual-labs/bagit-gython) now (v0.5.0) supports validating multiple bags concurrently via the validator API.


djjuhasz:
I tested this by uploading 10 bags with concurrency 20 and they all successfully processed and validated the bag.

cmd/enduro-am-worker/main.go
```diff
@@ -212,6 +213,19 @@ func main() {
 		})
 	}
 
+	// Set up the BagIt bag validator.
+	bagValidator, err := bagit.NewValidator(bagit.WithPoolSize(1))
```
jraddaoui: Looking at https://pkg.go.dev/github.com/artefactual-labs/bagit-gython#WithPoolSize, I think one is a good default. But it would be nice to make this configurable. Doesn't need to happen in this PR.

enduro.toml
```diff
@@ -372,7 +372,7 @@ url = "file:///home/enduro/internal-storage/sip-source?metadata=skip"
 #
 # See https://enduro.readthedocs.io/admin-manual/configuration/#allowduplicates 
 # for more details.
-allowDuplicates = false
+allowDuplicates = true
```
jraddaoui, line 395: This should go into a different commit.

jraddaoui (approved):
Thanks @djjuhasz!

sevein:
I've just released bagit-gython v0.6.0 with a couple of [additional](https://github.com/artefactual-labs/bagit-gython/commit/222b9f988a36767f22e880e48c7edde5c18dde1a) [improvements](https://github.com/artefactual-labs/bagit-gython/commit/d46e55697f31b5c3d9e894886c41f896b807e11f): should give faster validator startup by reusing the cached runtime, and lower disk use per pool because all pool members now share one runtime copy. It should work out of the box with the existing code, only the bump in go.mod needed.

djjuhasz:
@sevein the "cache dir" option is a nice addition, but the ergonomics of setting it from a config file is awkward, because it you can't do `NewValidator(bagit.WithCacheDir(nil))` to use the default value. Maybe there could be a `bagit.WithTmpCacheDir()` option to use a tmp dir, and `bagit.WithCacheDir("")` would use the default cache dir? Or maybe you have a better idea?

sevein:
> @sevein the "cache dir" option is a nice addition, but the ergonomics of setting it from a config file is awkward, because it you can't do NewValidator(bagit.WithCacheDir(nil)) to use the default value. Maybe there could be a bagit.WithTmpCacheDir() option to use a tmp dir, and bagit.WithCacheDir("") would use the default cache dir? Or maybe you have a better idea?

Right, good point. `NewValidator()` already defaults to the persistent cache dir based on `os.UserCacheDir()`, but I agree the config-file ergonomics were awkward when an empty string meant “use a temporary runtime.”

I've published [v0.6.1](https://github.com/artefactual-labs/bagit-gython/commit/50dda92fdb5e38ab36ed22effdf351bb5b631e70) with this behavior:

- `NewValidator()` uses the default persistent cache dir, same as v0.6.0.
- `NewValidator(WithCacheDir(""))` also uses the default persistent cache dir. This is the v0.6.1 change.
- `NewValidator(WithCacheDir("/path"))` uses the given persistent cache dir.
- `NewValidator(WithTempCacheDir())` uses an ephemeral runtime root that the library removes on `Close()`.

I suspect `NewValidator()` will be enough for your use case, but `WithCacheDir("")` should now also be safe to pass through from config.

djjuhasz:
@sevein awesome, thanks!

enduro.toml
djjuhasz: Oops, testing config commit. :blush:

jraddaoui: I like it, just as another commit. It should be the default for dev.

cmd/enduro-am-worker/main.go
djjuhasz: I've added configuration for the validator pool size and cache directory, as well as documentation for all the BagIt config settings.

docs/src/admin-manual/configuration.md
```diff
@@ -346,6 +346,57 @@ prefix = "enduro"
...
+created.
+
+[a3m] does not support ingest of BagIt bags, so PIPs are structured as
+an [Archivematica transfer with existing checksums] when [a3m] is used for
+preservation and these settings are not relevant.
+
+**Example configuration**:
+
+```toml
+[api.auth.ticket.redis]
```
jraddaoui: Wrong heading.

docs/src/admin-manual/configuration.md
```diff
@@ -346,6 +346,57 @@ prefix = "enduro"
...
+
+* `cacheDir` sets the cache directory used by the [bagit-gython] validator's
+  runtime and runners. If `cacheDir` is an empty string or omitted, the
+  validator will attempt to create a cache directory in the process owner's home
+  directory (e.g. /home/enduro/.cache/bagit-gython). If the user's home
+  directory can not be determined (e.g. because the process has no home
+  directory), the validator will fall back to using a unique temporary directory
+  (e.g. /tmp/bagit-gython-12345) that will be deleted at shutdown.
+* `poolSize` sets the number of available concurrent bag validation runners.
+  `poolSize` must 1 (the default value) or greater. If the number of requested
```
jraddaoui: Must be?

enduro.toml
```diff
@@ -120,6 +120,30 @@ maxSize = 100
...
+# will fall back to using a unique temporary directory
+# (e.g. /tmp/bagit-gython-12345) that will be deleted at shutdown.
+cacheDir = ""
+# poolSize sets the number of available concurrent bag validation runners.
+# poolSize must 1 (the default value) or greater. If the number of requested
+# validation jobs exceeds the available runners, the extra jobs will be queued
+# and run when a runner becomes available. See
+# https://github.com/artefactual-labs/bagit-gython/blob/main/README.md for more
+# details on how the validator pool works.
+poolSize = 1
```
jraddaoui, line 141: It would be nice to link at the documentation directly to avoid duplication and reduce the TOML file size.

jraddaoui (approved):
Thanks @djjuhasz!

docs/src/admin-manual/configuration.md
djjuhasz: Oops :blush:

enduro.toml
djjuhasz: I have mixed feelings about linking out to https://enduro.readthedocs.io/admin-manual/configuration/ from the config file. I'm conscious that the readthedocs site might not be available when someone is reading this file, either because of a temporary connectivity problem, or just because readthedocs.io ceases to exist at some point in the future.  That said I do also dislike having to repeat the same comments here, in the markdown files, and in the Go package. :(

I guess the backup plan if https://enduro.readthedocs.io/ is unavailable is to read the markdown source docs in the repo, but the reader would have to know where to find those files.  Maybe it would be good to add a note to the README.md, or this file (or both), that the doc files are available offline at `docs/src`?

djjuhasz: In any case, I would like to provide at least a brief summary for each config option in this file, so the reader has the basic context needed to change the settings, without having to read the docs.

djjuhasz:
@jraddaoui I've tried to strike a balance in the "enduro.toml" BagIt comments between providing a basic description of the settings, while referring to the docs for a detailed description of fallback behaviours and edge cases.  Let me know what you think.

jraddaoui (approved):

PR #1686, opened 2026-06-27
Move API config to the `api` package

djjuhasz:
- Move the API config to the `api` package and rename it from `config.APIConfig` to `api.Config`
- Move the child workflows config to an internal `childwf` package to resolve a circular import between the `api` package and the `about` package
- Move the `internal/api/auth` package to `interna/auth` to resolve a circular import between the `about` package and the `api` package


jraddaoui (approved):
LGTM, thanks @djjuhasz!

PR #1684, opened 2026-06-26
Use LGTM for dev observability

sevein:
Replace the separate local Grafana, Tempo, and Prometheus manifests with the [Grafana OTEL LGTM](https://github.com/grafana/docker-otel-lgtm) image. Point dev traces at the LGTM OTLP endpoint and keep Grafana connected to the Keycloak dev realm.

Add Alloy log forwarding to Loki and expose Enduro debug listeners for the metrics scrapes. Structured logging is not a goal of this PR; the current "human" log format remains unchanged for now.

LGTM is a good fit for the development stack because it gives us Grafana, Prometheus, Tempo, Loki, Pyroscope, and OTLP ingestion behind one local deployment. For production, this repo should probably stop short of owning a full observability platform; production deployments can point Enduro at whichever managed or separately maintained observability stack the operator already uses.

Tested in comabination with https://github.com/artefactual-sdps/enduro/pull/1687.

PR #1687, opened 2026-06-29
Add Goa API operation timeouts

sevein:
This PR adds centralized timeout and slow-operation handling for normal Goa API requests, with matching log and trace metadata so request stalls are easier to identify from Enduro itself. These operation timeouts also protect overall system health by preventing stuck or slow requests from indefinitely consuming shared resources like goroutines, connections, memory, and worker capacity. When an operation exceeds its budget, the logs and traces now include useful context such as the service, method, duration, timeout, status, and timeout event, without requiring broad request/response body logging.

The timeout policy is applied through [interceptors](https://goa.design/es/docs/1-goa/interceptors/) so it stays consistent across handlers, and long-lived routes such as uploads, downloads, and monitor/WebSocket endpoints are explicitly excluded. Those streaming paths have different failure modes, often involving proxies or clients, so they need separate download/proxy-specific observability rather than a short API operation deadline.

This would not have directly helped diagnose the large-download buffering issue fixed in fd87e8e because that problem lived in the Nginx streaming path, but it could help with other regular API operations that hang, run unexpectedly long, or fail before the caller gets a useful response. In those cases, Enduro will now leave clear application-level breadcrumbs instead of requiring debug/body logging or guesswork from proxy logs alone.

## How to test

> [!IMPORTANT]
> Depends on https://github.com/artefactual-sdps/enduro/pull/1684.

Modify a service method, such as `storage.ListLocations`, so that it blocks until the request context times out. This causes the method to wait for the context deadline to be reached before continuing, allowing timeout behavior to be tested.

```patch
diff --git a/internal/storage/service.go b/internal/storage/service.go
index 799dc1de..f1d82619 100644
--- a/internal/storage/service.go
+++ b/internal/storage/service.go
@@ -215,6 +215,10 @@ func (s *serviceImpl) ListLocations(
        ctx context.Context,
        payload *goastorage.ListLocationsPayload,
 ) (goastorage.LocationCollection, error) {
+       if _, ok := ctx.Deadline(); ok {
+               <-ctx.Done()
+       }
+
        return s.storagePersistence.ListLocations(ctx)
 }
```

In the Enduro dashboard, open the list of storage locations. Then, using the browser developer tools, confirm that the XHR request to `/api/storage/locations` completes with a 500 response and a response body similar to the following:

```
{
    "name": "internal_error",
    "id": "rI20fBxR",
    "message": "API operation timed out",
    "temporary": false,
    "timeout": true,
    "fault": true
}
```

In Tilt, confirm that the Go service logged the timeout error.

In [Grafana / Drilldown / Traces](http://localhost:7490/a/grafana-exploretraces-app/explore) (testing with https://github.com/artefactual-sdps/enduro/pull/1684), set the search range to "Last 30 minutes", and try combining the following filters:

* `span.goa.service=storage`
* `span.goa.method=ListLocations`
* `status=error`
* `duration>=5s`
* `event:name=api.operation.timeout`

image

Click on the Traces tab to inspect the matched traces individually, e.g.:

image

You can also use TraceQL directly in Tempo using the [Grafana Explorer](http://localhost:7490/explore). The following query asks Tempo to retrieve all traces containing the "api.operation.timeout" event and a budget of five seconds:

    { event:name = "api.operation.timeout" && .api.operation.timeout_ms = 5000 }


dashboard/Dockerfile
```diff
@@ -45,13 +45,25 @@ server {
     # add_header Reporting-Endpoints 'csp="https://example.com/csp"' always;
     # (Append to CSP) ... ; report-to csp
 
+    # Keep long-lived transfer routes alive for slow clients. 24h is a
+    # generous upper bound, not an expected transfer duration.
     location ~ "^/api/ingest/sips/([0-9a-fA-F-]{36})/download" {
       proxy_pass http://backend/ingest/sips/\$1/download;
       proxy_buffering off;
+      proxy_read_timeout 24h;
```
djjuhasz: When I was troubleshooting an AIP download timeout issue I came to the conclusion that increasing the `proxy_read_timeout` and `proxy_send_timeout` was unnecessary because it the time is counted between read or write operations, not for the duration of the request or response*.  For instance this timeout would only close the connection if *no* data was read from the connected proxy (the API server in this case) for 24 hours.  I think using the default `proxy_read_timeout` (and `proxy_send_timeout`) of 60 seconds is fine, and will prevent dead connections from lingering for an extended period.

(*) The Nginx docs description of `proxy_read_timeout` is: "The timeout is set only between two successive read operations, not for the transmission of the whole response. If the proxied server does not transmit anything within this time, the connection is closed." (https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout)

internal/api/interceptors_test.go
```diff
@@ -0,0 +1,289 @@
...
+	}{
+		{
+			name: "undeclared internal error",
+			encode: storagesvr.EncodeListAipsError(
+				goahttp.ResponseEncoder,
+				nil,
+			),
+		},
+		{
+			name: "declared internal error",
```
djjuhasz: How is this "declared" internal error different from the previous "undeclared" internal error? It looks like the "internal_error" type is declared in the `t.Run()` below.

internal/api/interceptors_test.go
```diff
@@ -0,0 +1,289 @@
...
+	var serr *goa.ServiceError
+	assert.Assert(t, errors.As(err, &serr))
+	assert.Equal(t, serr.Name, "internal_error")
+	assert.Equal(t, serr.Message, apiOperationTimeoutMsg)
+	assert.Equal(t, serr.Timeout, true)
+	assert.Equal(t, serr.Temporary, false)
+	assert.Equal(t, serr.Fault, true)
+}
+
+func TestOperationTimeoutEncodesAsServerTimeout(t *testing.T) {
```
djjuhasz: Is this test necessary? It seems to be testing the underlying `goa.ServiceError` encoding/decoding which is generated code and not a responsibility of this layer of the application.

internal/api/interceptors_test.go
```diff
@@ -0,0 +1,289 @@
...
+		},
+		newStorageServerInterceptors(logr.Discard()),
+	)
+
+	res, err := endpoint(context.Background(), "payload")
+	assert.NilError(t, err)
+	assert.Equal(t, res, "ok")
+}
+
+func TestOperationTimeoutUsesDeadlineForNormalMethods(t *testing.T) {
```
djjuhasz: This looks the same as the `TestOperationTimeoutUsesDeadline` except with `goastorage.WrapShowAipEndpoint()` instead of `goastorage.WrapDownloadAipRequestEndpoint()`.  I don't think it's actually adding any test coverage at this application layer.  If it does test the new code I'm not sure what the "NormalMethods" means, in contrast to the previous test.

internal/api/interceptors_test.go
```diff
@@ -0,0 +1,289 @@
...
+	assert.Equal(t, spanAttribute(t, spans[0].Attributes(), "goa.service").AsString(), "storage")
+	assert.Equal(t, spanAttribute(t, spans[0].Attributes(), "goa.method").AsString(), "ShowAip")
+	assert.Equal(t,
+		spanAttribute(t, spans[0].Attributes(), "api.operation.timeout_ms").AsInt64(),
+		defaultAPIOperationTimeout.Milliseconds(),
+	)
+	assert.Equal(t, spans[0].Status().Code, codes.Unset)
+}
+
+func noDeadlineEndpoint(t *testing.T) goa.Endpoint {
```
djjuhasz: It would be nice to group the helper functions at the beginning or end of the file rather then mixing them in with the "Test*" functions.

internal/ingest/rollback.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+const rollbackCleanupTimeout = time.Second
+
+// withRollbackCleanupContext runs cleanup with request-scoped values preserved
+// but request cancellation ignored, bounded by the rollback cleanup timeout.
+func withRollbackCleanupContext(ctx context.Context, cleanup func(context.Context) error) error {
+	// Preserve request-scoped values for telemetry/logging while giving
+	// rollback cleanup a fresh, bounded deadline after ingest startup fails.
+	cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), rollbackCleanupTimeout)
```
djjuhasz: I'm not clear why we want to cancel the context if a rollback takes longer than a second.  I don't know if it's very unusual for a normal MySQL rollback operation to take longer than a second, but if a rollback does take longer than a second, what's the advantage of cancelling the context so quickly instead of  waiting for the rollback to complete?  I would expect MySQL has some mechanism for abandoning a stuck rollback operation, and would return an error in that case.

djjuhasz (commented):
Thanks @sevein.  I have some questions and a few notes.

dashboard/Dockerfile
sevein: Oh I see, thank you! Fixed.

internal/api/interceptors_test.go
sevein: Right! Test removed.

internal/api/interceptors_test.go
sevein: Makes sense, test removed.

internal/api/interceptors_test.go
sevein: Thanks, agreed.

internal/ingest/rollback.go
sevein: Sorry, the name was misleading. This is not an actual MySQL transaction rollback; it is compensating cleanup for cases where `AddSip`/`AddBatch` have already persisted state but starting the Temporal workflow failed. The helper intentionally lets cleanup outlive request cancellation introduced by the new operation timeout, so I think it should also give cleanup its own deadline rather than leaving it unbounded. MySQL, the driver, or deployment-level settings may provide their own limits, but those are environment-dependent; this keeps the cleanup budget explicit at the application layer. I agree that `1s` is too aggressive, so I'll rename this away from rollback terminology and increase the timeout to `10s`.

internal/api/interceptors_test.go
sevein: Yeah sorry, that was unclear. I rewrote it as a table-driven test that covers the actual rule boundary: request/ticket operations get the operation timeout but body-returning transfer endpoints skip it. So `DownloadAipRequest` gets a deadline but `DownloadAip` does not, and `AipDeletionReportRequest` gets a deadline but `AipDeletionReport` does not. I've renamed the test as TestOperationTimeoutStorageTransferRules too.

djjuhasz (approved):
Thanks @sevein, this LGTM. 👍

PR #1689, opened 2026-06-29
Replace monitor WebSockets with SSE

sevein:
This PR replaces the dashboard monitor WebSocket transport with Server-Sent Events (SSE).

The monitor streams are browser-consumed, server-to-client-only event feeds. SSE fits that shape better than WebSockets: it uses normal HTTP semantics, works cleanly through standard reverse proxies, and avoids the extra origin/upgrader handling that the current WebSocket implementation needs. The dashboard still keeps the existing authenticated ticket request flow before opening each monitor stream, because native `EventSource` cannot attach an `Authorization` header.

The implementation updates the Goa API design and generated transports, switches the dashboard monitor client from `WebSocket` to `EventSource`, removes WebSocket-specific server/origin code, and updates NGINX/docs guidance for long-lived SSE responses. It also keeps monitor write timeouts disabled so idle event streams are not closed by the API server.


jraddaoui:
Thanks @sevein! Looking great so far, just wondering if we could use the Fetch API to avoid the POST request and ticket cookie. Could that be possible with something like https://github.com/Azure/fetch-event-source?

sevein:
> Thanks @sevein! Looking great so far, just wondering if we could use the Fetch API to avoid the POST request and ticket cookie. Could that be possible with something like https://github.com/Azure/fetch-event-source?

Good thinking! 👀

sevein:
@jraddaoui I've made sure that [29567e0](https://github.com/artefactual-sdps/enduro/pull/1689/commits/29567e03802ded6145bd9d964a9c35c5a05f751f) can be merged on its own, without [2c80bc7](https://github.com/artefactual-sdps/enduro/pull/1689/commits/2c80bc7e74583f1433991aa1b15ef7df54f50a07), in case we decide to include only the first commit.

dashboard/src/monitor.ts
```diff
@@ -28,20 +36,38 @@ function parseMonitorEvent<T extends string>(
...
+  input: string | URL,
+  init: EventSourceFetchInit,
+) {
+  const headers = new Headers(init.headers);
+  const token = useAuthStore().getUserAccessToken;
+  if (token !== "") {
+    headers.set("Authorization", `Bearer ${token}`);
+  }
+
+  return fetch(input, { ...init, headers });
```
jraddaoui, line 50: We used [the client](https://github.com/artefactual-sdps/enduro/blob/main/dashboard/src/client.ts) before to establish the connection, and that client has middleware to handle `401` responses. Now it keeps retrying the connection even if the token has expired.

There are some improvements we could make around token refresh/expiration, but I would consider those outside the scope of this PR.

docs/src/dev-manual/api/openapi3.json
jraddaoui: Probably on Goa's side, but looking at the browser console, this response seems to be a `200` with `text/event-stream`. Maybe we can override those values, I did something similar in https://github.com/artefactual-sdps/sfa-enduro-workflows/blob/main/internal/dips/api/design/design.go#L124-L128.

jraddaoui (changes requested):
Thanks for this @sevein! And sorry for the delay, I thought David reviewed this one too.

Apart from the two comments below, I saw this in the logs, which I think was not happening before:

```
2026-07-06T04:00:19.167Z	V(2)	enduro.ingest-events	event/redis.go:116	Error decoding event.	{"error": "body.started_at must be formatted as a date-time but got value \"\", parsing time \"\" as \"2006-01-02T15:04:05Z07:00\": cannot parse \"\" as \"2006\""}
github.com/artefactual-sdps/enduro/internal/event.(*subscriptionRedisImpl[...]).loop
	github.com/artefactual-sdps/enduro/internal/event/redis.go:116
```

But I could not spot any change in this PR that may have caused it.

internal/ingest/convert.go
```diff
@@ -71,20 +71,16 @@ func workflowToGoa(w *datatypes.Workflow) *goaingest.SIPWorkflow {
 
 // taskToGoa returns the API representation of a task.
 func taskToGoa(task *datatypes.Task) *goaingest.SIPTask {
-	var startedAt string
+	var startedAt *string
 	if !task.StartedAt.IsZero() {
-		startedAt = task.StartedAt.Format(time.RFC3339)
+		startedAt = new(task.StartedAt.Format(time.RFC3339))
 	}
```
jraddaoui: I guess we could use `db.FormatOptionalZeroTime`, same as `CompletedAt`.

jraddaoui:
Thanks @sevein! Make sure you squash, if you can keep the two commits as you said that would be great, but not a big deal if that's a hassle.

jraddaoui (approved):

PR #1691, opened 2026-07-07
Add SeaweedFS object store for dev

jraddaoui:
Add an optional `OBJECT_STORE=seaweedfs` mode for Tilt that renders SeaweedFS dev overlays while keeping filesystem storage as the default.

Provide SeaweedFS manifests with S3 buckets, dev credentials, persistent storage, and the Admin UI. Wire Enduro and the active worker to use the SeaweedFS S3 endpoint in that mode.

Refs #1660.


djjuhasz (approved):
Wow, looks great @jraddaoui!

PR #1698, opened 2026-07-13
Internalize child workflow type enum

jraddaoui:
Move WorkflowType out of the public child workflow contract because only Enduro uses it to configure and dispatch child workflows.

This is a breaking change, but this enum is not being used in any of the existing child workflow projects we have created.


internal/enums/child_workflow_type.go
djjuhasz, line 1: I didn't notice when the child workflow type was in the `pkg` directory, but putting the enum definition and generated file in the `childwf` package breaks with our previous pattern of putting enums in the `enums` package. 🤔 

I don't know that this is necessarily a bad change, but it is different from our established pattern which could be confusing.  Maybe we should put this in the `enums` package as well and call it `enums.ChildWFType` or something similar?

djjuhasz (commented):
@jraddaoui I like the move of the child workflow type from the public interface to the internal one, but I was surprised to find the enum in the `childwf` package instead of the `enums` package.  See the comment on the `childwf/workflow_type.go` file.

internal/enums/child_workflow_type.go
jraddaoui: TBH, I prefer putting them in the related domain/package, and I think we should do the same with activities for example. But you are right, it would be the first one breaking the existing pattern, so I'll move it into the enums package.

The `pkg/childwf` package still contains an enum for task outcomes, but that's shared and I'd like to avoid an extra enums package there.

djjuhasz (approved):
Thanks @jraddaoui, LGTM. 👍

PR #1703, opened 2026-07-21
Replace debug with log format

sevein:
Replace the top-level debug boolean with an explicit `logFormat` option, similar to the work done in 468eb9b87c9a3fa4265b8bf1cbcb52a9c41fc080. Accept text and JSON output formats and keep JSON as the runtime default.

### DIFF
--- a/docs/src/dev-manual/logging.md
+++ b/docs/src/dev-manual/logging.md
@@ -18,7 +18,8 @@ There are two configuration attributes that affect the application logger:
 # structured logs.
 logFormat = "text"
 
-# Show log messages up to V-level 2.
+# Show log messages up to verbosity level 2, where level 0 is the most
+# important and higher levels are less important.
 verbosity = 2
 ```
 
--- a/docs/src/dev-manual/logging.md
+++ b/docs/src/dev-manual/logging.md
@@ -14,14 +14,12 @@ design.
 There are two configuration attributes that affect the application logger:
 
 ```toml
-# When enabled, logs with V-levels greater than zero are shown (debug-y logs).
-# When disabled, only logs with V-level 0 are shown.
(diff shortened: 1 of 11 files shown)


docs/src/dev-manual/logging.md
```diff
@@ -14,14 +14,12 @@ design.
...
-verbose = true
-
-# Use the color-enabled logging encoder.
-# Do not use in production, it disables JSON-encoded structured logging.
-debug = true
+# Use human-readable, colorized text logs. Use "json" for machine-readable
+# structured logs.
+logFormat = "text"
+
+# Show log messages up to V-level 2.
```
djjuhasz: I think it would be helpful to add a bit more context to this description as to what the "V-level" means, e.g. "Show log messages up to verbosity level 2, where level 0 is the most important messages and higher levels are less important."

go.mod
```diff
@@ -46,7 +46,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	go.artefactual.dev/amclient v0.4.1-0.20240705155055-0c5abef5207c
 	go.artefactual.dev/ssclient v0.11.0
-	go.artefactual.dev/tools v0.25.1
+	go.artefactual.dev/tools v0.26.0
```
djjuhasz, line 49: Hmm, I'm not sure where this dev/tools version bump came from, but I think it's fine to keep it.

djjuhasz (approved):
Thanks @sevein. I had one request to add more context about the logging V-level, but it's not a blocker.

go.mod
sevein: That was me! v0.26.0 added `WithFormat` to keep the terminology consistent across the board.

docs/src/dev-manual/logging.md
sevein: Good point, will fix!

PR #1705, opened 2026-07-22
Replace promise dialog dependency

sevein:
> [!IMPORTANT]
> Developed with assistance from gpt-5.6-sol.

[vue3-promise-dialog] made modal interactions conveniently awaitable, but it is
no longer actively maintained and Enduro uses only a narrow part of its API.
Using it also required application plugin registration, a global
DialogWrapper, and an ambient declaration that erased its types.

Replace the dependency with a small runtime built on VueUse
[createTemplatePromise]. Callers now pass concrete Vue components directly to
openDialog, along with a forced-cancellation value and optional props. Infer
the request props and promise result from component-defined props and the
typed resolve event, avoiding registries and per-dialog opener wrappers.

Keep concrete dialogs in components and colocate the shared host, instance,
Bootstrap lifecycle helper, tests, and documentation under dialogs. Reject
overlapping requests instead of sharing results or stacking Bootstrap modals.

Scope the host to authenticated sessions and resolve each request once when
the host is torn down. Hide modals before disposal so Bootstrap restores body
scrolling and removes its backdrop during session expiry and other forced
teardown.

Migrate the About, location, batch review, and AIP deletion flows. Document
the architecture, usage, lifecycle constraints, and process for developing
and testing new dialogs. Cover normal results, user cancellation, forced
cancellation, concurrent requests, Bootstrap cleanup, migrated callers, and
session expiry.

[vue3-promise-dialog]: https://github.com/rlemaigre/vue3-promise-dialog
[createTemplatePromise]: https://vueuse.org/core/createTemplatePromise/

### DIFF
--- a/dashboard/src/dialogs/README.md
+++ b/dashboard/src/dialogs/README.md
@@ -1,0 +1,185 @@
+# Promise dialogs
+
+The dashboard treats a dialog as an asynchronous question:
+
+```ts
+const confirmed = await openDialog(ConfirmDialog, {
+  heading: "Continue?",
+});
+
+if (!confirmed) return;
+```
+
+This keeps the business flow in the component that needs the answer without
+making that component own modal visibility, global placement, or Bootstrap
+cleanup.
+
+Concrete dialogs remain regular application components in
+[`components`](../components). The files in this directory provide only the
+shared machinery needed to open those components as promises.
+
+There is no dialog registry and there are no named opener wrappers. Callers
+pass the concrete component directly to `openDialog`. Vue's generated component
+types let `openDialog` infer the component's props and the value carried by its
+typed `resolve` event.
+
+## Architecture
+
+- [`dialog.ts`](dialog.ts) exposes `openDialog`, holds the pending request, and
+  rejects attempts to open overlapping dialogs.
+- [`DialogHost.vue`](DialogHost.vue) renders a pending request at the
+  application root.
+- [`DialogInstance.vue`](DialogInstance.vue) connects the component's `resolve`
+  event to the pending promise. If the host is removed, it resolves with
+  `undefined`.
+- [`useDialog.ts`](useDialog.ts) carries the chosen result through Bootstrap's
+  hide lifecycle and owns the modal instance.
+
+[`App.vue`](../App.vue) mounts one `DialogHost` while the user session is valid.
+When a session expires, removing the host cancels the pending request and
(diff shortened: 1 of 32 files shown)


dashboard/src/dialogs/README.md
```diff
@@ -0,0 +1,185 @@
+# Promise dialogs
+
+The dashboard treats a dialog as an asynchronous question:
+
+```ts
+const confirmed = await openDialog(ConfirmDialog, {
```
djjuhasz, line 6: Is it a common practice with vue / typescript libraries like this to document the interface in a README?  Personally I would rather have the function interfaces (especially `openDialog()` and `useDialog()`) documented via code comments so the information is handy when reading the code and when examining the function calls in an IDE.

This README is very helpful for understanding how the dialogs "package" works, but I only found it halfway through doing code review.

djjuhasz (approved):
@sevein this looks okay to me, but I don't fully understand a lot of the advanced Typescript "dialogs" package definitions like generics, type inference, and type inheritance.  It all seems to work as a replacement for vue3-promise-dialog, with pretty minor changes required in the existing dialog components, so I'd call it a success. :)

I did add one comment about the documentation of for the dialogs package — the README is very helpful, but I would love to see the functional interface documented in the code rather than in a README.  It took me a while to find the README when I was doing this code review.

dashboard/src/dialogs/README.md
sevein: I thought that the README makes sense because `dialogs` acts as a small internal package and I wanted to describe how to use it but also the architecture, how to test it, etc.

Not sure if there is an established pattern around this, but I often look at VueUse as a reference, and it does both. E.g.: [`createTemplatePromise`](https://vueuse.org/core/createtemplatepromise/) has a dedicated Markdown guide for broader usage and an [IDE-visible documentation comment in the source](https://github.com/vueuse/vueuse/blob/main/packages/core/createTemplatePromise/index.ts) for its exported API.

djjuhasz (approved):
Thanks @sevein. The code comments are very helpful, and the new `code-documentation.md` page is a great guide for how to use comments. Love it! 👍

dashboard/src/dialogs/README.md
djjuhasz: This README does add a lot of valuable context about how to use the dialogs, so I think it's a good addition to the code comments documenting the API.

PR #1714, opened 2026-07-24
Redesign dashboard sign-in experience

sevein:
> [!IMPORTANT]
> Developed with assistance from gpt-5.6-sol.

Replace the minimal sign-in view with a responsive, accessible layout while preserving the existing OIDC redirect flow.

Add reduced-motion-aware pointer tracking, animated gradients, redirect and error states, and an Artefactual attribution link. Cover the sign-in behavior with focused component tests.

---

image


dashboard/src/pages/user/signin.vue
```diff
@@ -1,18 +1,646 @@
...
+        <span class="fs-4 fw-semibold">Enduro</span>
+      </div>
+
+      <div class="signin-visual-copy">
+        <p class="signin-visual-title">
+          Preservation workflows,<br />
+          under control.
+        </p>
+        <p class="signin-visual-description mb-0">
+          Move digital content confidently from ingest to durable storage.
```
djjuhasz: I personally don't like the "confidently" in this section, it sounds too generic to me.  I think it would be better to emphasize the reliability of Enduro, I think that's one of it's best features.

Maybe "Reliably move digital content from ingest to storage"?

I debated whether to strike the "durable" from the storage, but I think it's more honest at this point — Enduro doesn't deal with storage "durability" at all.  Even the Archivematica Storage Service is doesn't provide any guarantee that stored AIPs are "durably" stored, that's left to configuration and the backend storage infrastructure.

dashboard/src/pages/user/signin.vue
```diff
@@ -1,18 +1,646 @@
...
+        <span class="signin-visual-logo">
+          <img src="/logo.png" alt="" />
+        </span>
+        <span class="fs-4 fw-semibold">Enduro</span>
+      </div>
+
+      <div class="signin-visual-copy">
+        <p class="signin-visual-title">
+          Preservation workflows,<br />
+          under control.
```
djjuhasz, line 176: I think this is fine for now, but I think we should come up with a better tagline at some point.

djjuhasz (approved):
Looks pretty swanky @sevein! ✨ 

I like @sallain's suggestion to add a link to the docs.  I also suggested one change to the page text, but that's just a personal opinion.

PR #1716, opened 2026-07-24
Close about dialog on session timeout

djjuhasz:
Fixes #1715.

Developed with assistance from Claude Sonnet 4.6.

- Close the "About" modal when the user login session expires
- Remove Bootstrap 5 modal classes when a user login session expires
- Add unit tests for the "About" modal

### DIFF
--- a/dashboard/src/App.vue
+++ b/dashboard/src/App.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import Modal from "bootstrap/js/dist/modal";
 import { watch } from "vue";
 import { DialogWrapper } from "vue3-promise-dialog";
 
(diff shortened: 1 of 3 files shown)


dashboard/src/App.vue
```diff
@@ -22,6 +22,14 @@ watch(
...
       ingestMonitor.connect();
       storageMonitor.connect();
+    } else {
+      // When the session expires the DialogWrapper is removed from the DOM
+      // immediately via v-if. Bootstrap appends .modal-backdrop directly to
+      // document.body, outside the Vue component tree, so it is never cleaned
+      // up automatically. Remove any lingering backdrop elements and the
+      // modal-open class so the sign-in page is fully interactive.
+      document.querySelectorAll(".modal-backdrop").forEach((el) => el.remove());
+      document.body.classList.remove("modal-open");
```
sevein: I haven't tried but can we see if Bootstrap's `Modal.hide()` can be used instead? It looks like it will perform a bigger cleanup, e.g. triggers `hidden.bs.modal` handler, restores the body scroll style, etc.

```ts
import Modal from "bootstrap/js/dist/modal";

document.
  .querySelectorAll<HTMLElement>(".modal.show")
  .forEach((el) => Modal.getInstance(el)?.hide());
```

sevein (approved):
👍

dashboard/src/App.vue
djjuhasz: Yes, that works. 👍

PR #1721, opened 2026-07-29
Increase timeout for flaky filesystem watcher test

djjuhasz:
Fixes #1720

- Increase test timeout to 500ms for the filesystem watcher polling tests to avoid intermittent test failures in the CI environment
- Also increase the PollInterval to 100ms to avoid overpolling in a high load, high filesystem latency CI environment

### DIFF
--- a/internal/watcher/filesystem_test.go
+++ b/internal/watcher/filesystem_test.go
@@ -10,7 +10,6 @@ import (
 
 	"gotest.tools/v3/assert"
 	"gotest.tools/v3/fs"
-	"gotest.tools/v3/poll"
 
 	"github.com/artefactual-sdps/enduro/internal/watcher"
 )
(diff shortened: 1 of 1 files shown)


internal/watcher/filesystem_test.go
```diff
@@ -83,7 +83,12 @@ func TestFileSystemWatcher(t *testing.T) {
 				t.Fatalf("Couldn't create text.txt in %q", td.Path())
 			}
 
-			poll.WaitOn(t, check, poll.WithTimeout(time.Millisecond*15))
+			poll.WaitOn(
+				t,
+				check,
+				poll.WithTimeout(time.Millisecond*500),
+				poll.WithDelay(time.Millisecond*100),
```
jraddaoui: Why the delay?

jraddaoui (approved):
LGTM!

PR #1725, opened 2026-08-05
Send `/sips POST` data in request body

djjuhasz:
Move SIP source ingest request data from query to request body.


jraddaoui (approved):
LGTM!

PR #1726, opened 2026-08-05
Document the minio watcher `WatchTimeout` setting

djjuhasz:
I spent several hours trying to reduce the time taken by the minioWatcher tests, only to discover that the "PollInterval" setting was actually a timeout value for the redis BLMove() function, and it also doesn't support context cancellation. Change the minioWatcher "pollInterval" setting name and document it's function and limitations.

- Change the name of the minio watcher "PollInterval" setting to "WatchTimeout" and add documentation to describe how it effects the `Watch()` method.
- Add a TODO to the minioWatcher `Watch()` method to add support for context cancellation.
- Explain the minimum "WatchTimeout" value of 1 second.

### DIFF
--- a/internal/watcher/config.go
+++ b/internal/watcher/config.go
@@ -124,8 +124,10 @@ type MinioConfig struct {
 	// a successful ingest. If negative, SIPs will be retained indefinitely.
 	RetentionPeriod time.Duration
 
-	// PollInterval sets the length of time between Redis polls (default: 1s).
-	PollInterval time.Duration
+	// WatchTimeout sets the maximum time the Watch() method will wait for a
+	// Redis event when the queue is empty before returning a timeout error
+	// (default: 1 minute).
+	WatchTimeout time.Duration
 
 	// WorkflowType specifies which workflow this watcher should execute
 	// (default: "create aip").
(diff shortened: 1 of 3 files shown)


internal/watcher/config.go
```diff
@@ -124,8 +124,9 @@ type MinioConfig struct {
 	// a successful ingest. If negative, SIPs will be retained indefinitely.
 	RetentionPeriod time.Duration
 
-	// PollInterval sets the length of time between Redis polls (default: 1s).
-	PollInterval time.Duration
+	// WatchTimeout sets the maximum time the Watch() method will wait for a
+	// Redis event message before returning a timeout error (default: 1 minute).
+	WatchTimeout time.Duration
```
sevein, line 130: Perhaps clarify that this timeout applies when the queue is empty: "wait for a Redis event when the queue is empty before returning a timeout error".

sevein (commented):
> I spent several hours trying to reduce the time taken by the minioWatcher tests...

Opsss, the classic multi-hour rabbit hole! 😅 Coding agents have saved me from a few of those lately.

jraddaoui (approved):
LGTM!

PR #1727, opened 2026-08-06
Prevent duplicate Archivematica transfers

sevein:
Archivematica 1.19 [adds support for idempotency keys](https://github.com/artefactual/archivematica/pull/2352) during transfer submission. A client can use a key to identify one specific submission request. If the request is retried, Archivematica can recognize it and return the original transfer instead of creating a duplicate.

This protects against failures that happen midway through submission. For example, Archivematica may create the transfer but Enduro may lose the HTTP response. Enduro cannot know whether the request succeeded, so Temporal retries it. Sending the same idempotency key makes that retry safe.

This feature was initially developed for Wellcome's S3/Lambda submission workflow. It is also used by [legacy Enduro](https://github.com/artefactual-labs/enduro/pull/724).

In this project, I suggest to use the Temporal workflow run ID as the idempotency key. Activity and worker-session retries remain in the same workflow run and therefore reuse the key. A user-triggered Enduro Retry starts a new workflow run and gets a new key, allowing it to create a new transfer as intended.

Enduro will check the Archivematica version before every submission attempt and sends the key to Archivematica 1.19.0 and newer. Requests that are still in progress can be retried, while reuse of the same key with a different request stops further retries.

Idempotency keys are a well-established API pattern. The header and its expected behavior are described in the [IETF HTTPAPI working-group draft](https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/). Brandur Leach also provides a useful [high-level overview of idempotency keys](https://brandur.org/idempotency-keys).

### DIFF
--- a/internal/am/start_transfer.go
+++ b/internal/am/start_transfer.go
@@ -1,18 +1,28 @@
 package am
 
 import (
-	context "context"
+	"context"
+	"errors"
+	"net/http"
 	"path/filepath"
 
 	"go.artefactual.dev/amclient"
 	temporal_tools "go.artefactual.dev/tools/temporal"
+	temporalsdk_activity "go.temporal.io/sdk/activity"
 )
 
 const StartTransferActivityName = "start-transfer-activity"
 
 type StartTransferActivity struct {
-	cfg  *Config
-	amps amclient.PackageService
+	cfg                *Config
+	serverInfoProvider serverInfoProvider
+	amps               amclient.PackageService
+}
+
+// serverInfoProvider keeps server discovery separate from PackageService.
+// amclient exposes it directly on Client rather than through a service.
+type serverInfoProvider interface {
+	ServerInfo(context.Context) (*amclient.ServerInfo, *amclient.Response, error)
 }
 
 type StartTransferActivityParams struct {
@@ -30,10 +40,14 @@ type StartTransferActivityResult struct {
 	TransferID string
 }
 
-func NewStartTransferActivity(cfg *Config, amps amclient.PackageService) *StartTransferActivity {
+func NewStartTransferActivity(
+	cfg *Config,
+	client *amclient.Client,
+) *StartTransferActivity {
 	return &StartTransferActivity{
-		cfg:  cfg,
-		amps: amps,
+		cfg:                cfg,
+		serverInfoProvider: client,
+		amps:               client.Package,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


internal/am/start_transfer.go
```diff
@@ -62,16 +76,72 @@ func (a *StartTransferActivity) Execute(
...
+		IdempotencyKey:   idempotencyKey,
 	})
 	if err != nil {
+		// Interpret 409 and 422 as idempotency responses only when a key was
+		// sent. A pending request should be retried; reusing the key with a
+		// different payload is permanent and must stop Temporal retries.
+		if idempotencyKey != "" && resp != nil && resp.Response != nil {
+			switch resp.StatusCode {
+			case http.StatusConflict:
+				return nil, errors.New("transfer request is still in progress")
```
jraddaoui, line 108: What does a conflict mean in AM? Is the transfer already started there? Do we need to retry here if the transfer already started? Does the conflict response include the conflicting transfer ID?

My guess is that the transfer is already started, and that we should be able to get the conflicting transfer ID from the response. If that's the case, we don't need to retry here, just get the ID and return it in the result.

jraddaoui (changes requested):
This is really nice @sevein! Thanks for all the work in multiple projects to get it here!

internal/am/start_transfer.go
sevein: Not quite! A 409 status code is used when Archivematica has reserved the idempotency key and created the transfer record and identifier, but it has not confirmed yet that the transfer was successfully queued for processing.

The response doesn't include the identifier on purpose. If Enduro receives it too early, it may consider the submission successful and start polling a transfer that may never run.

The correct action is to retry using the same idempotency key. This will not create a duplicate. Once the transfer is successfully queued, the same request will return `202 Accepted` with the original transfer identifier. If queuing fails, Archivematica releases the idempotency key so the retry can submit the transfer again.

Disclosure: using AI here mercilessly to help me understand this. Fully understanding every protocol-level detail would take more time than I have, but I traced the Archivematica flow and checked it against the IETF draft. This looks correct to me.

Related:
- [Section 2.6 — Idempotency Enforcement](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-idempotency-key-header-07#section-2.6) says a retry made before the original request completes should receive a conflict response.
 - [Section 2.7 — Error Handling](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-idempotency-key-header-07#section-2.7) specifies 409 Conflict for that case and says the client does not need to correct the request before retrying.

Note: not returning the transfer identifier in the `409` response is an implementation detail caused by the current non-transactional enqueueing. Archivematica saves the transfer and idempotency records first, and submits the work to the executor afterwards. Between these two operations, the identifier exists in the database, but there is no guarantee that the transfer was queued. Returning the identifier during this window would incorrectly tell Enduro that the submission succeeded. Archivematica would need to implement [transactional enqueueing](https://riverqueue.com/docs/transactional-enqueueing), so the transfer record and its processing job could be committed together: either both exist or neither exists. This would remove the failure window that requires Archivematica to hide the pending transfer identifier.

sevein: Remember: retrying will not create a duplicate because Enduro uses the same stable idempotency key (the workflow run ID), allowing Archivematica to recognize the request and return the original transfer identifier.

internal/am/start_transfer.go
```diff
@@ -62,16 +76,72 @@ func (a *StartTransferActivity) Execute(
...
+			case http.StatusUnprocessableEntity:
```
djjuhasz, line 109: How does Archivematica check that the idempotency key was used with a different request payload?  Does it store the payload itself or a checksum with the key?

djjuhasz (approved):
LGTM, just one question about how AM checks whether an idempotency key has been used with a previous (different) request payload.

internal/am/start_transfer.go
sevein: The latter, it stores a [fingerprint](https://github.com/artefactual/archivematica/blob/306db6773216e607cf5687f84fb0be353949ddb7/src/archivematica/MCPServer/server/packages.py#L307-L320).

djjuhasz: Cool! :sunglasses:

internal/am/start_transfer.go
jraddaoui: I see, thanks for the detailed response @sevein!

I thought the initial request would complete "immediately", which may still be the case and we are just talking about a really small time frame for the 409 to happen.

So, if we loose the 202 but that request completed, the following request will also return a 202, right?

jraddaoui (approved):

internal/am/start_transfer.go
sevein: Yes, as long as you're using the same idempotency key.

Quick example showing that the retry with the same key returned the same transfer identifier, but then in the third request the key changed so Archivematica interpreted that as a different request.

image

Issue #1723, opened 2026-08-04
Problem: the user menu in Enduro is currently in an unintuitive location


fiver-watson:
**Describe the bug**

Good usability design often relies on the use of established de facto conventions and best practices, so that the familiarity of users with these patterns leads them to predictable outcomes. 

One of these long-standing design conventions on the modern web is tha application user menus are almost ALWAYS located in the top right of the screen. 

Enduro's user menu is currently not there, meaning that users often miss it when initial looking for where to logout. 

**Proposed solution**

Move the current user menu from the bottom left of the navigation sidebar, to the top right side of the application header bar. Make it a drop-down menu instead of a drop-up.

PR #1736, opened 2026-08-11
Move user menu to header

jraddaoui:
Move user menu to header, and move about button into user menu.

image

Refs #1723.

### DIFF
--- a/dashboard/src/components/Header.vue
+++ b/dashboard/src/components/Header.vue
@@ -2,7 +2,7 @@
 import Dropdown from "bootstrap/js/dist/dropdown";
 import { onMounted, ref } from "vue";
 
-import IconInfo from "~icons/clarity/info-standard-solid";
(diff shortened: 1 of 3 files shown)


dashboard/src/components/Header.vue
```diff
@@ -66,14 +78,49 @@ const institution: { logo: string; name: string; url: string } = {
...
+            </button>
+          </li>
+          <li v-if="authStore.isEnabled">
+            <a
+              class="dropdown-item d-flex align-items-center gap-3 text-decoration-none"
+              href="#"
+              @click.prevent="authStore.signoutRedirect()"
+            >
+              <IconLogout aria-hidden="true" />
+              <span>Logout</span>
```
sevein: It should read `Log out`, but you may prefer to maintain the label we were using before?

Image

sevein (approved):
LGTM

PR #1737, opened 2026-08-12
Polish dashboard styles, layouts, and interactions

jraddaoui:
I'm currently working on a dark theme, which will be submitted in a separate PR. Along the way, I identified several minor dashboard issues and opportunities for refinement that this PR addresses.

- Normalize dashboard labels, cards, legends, and task styling.
- Improve responsive layouts and spacing.
- Theme upload and date picker components.
- Fix workflow accordion interactions and visual inconsistencies.

See the individual commit messages for more details.

### DIFF
--- a/dashboard/src/components/WorkflowCollapse.vue
+++ b/dashboard/src/components/WorkflowCollapse.vue
@@ -164,7 +164,6 @@ const showTasks = computed(() => {
         <div class="workflow-summary">
           <div class="workflow-summary-title">
             <span>{{ $filters.getWorkflowLabel(workflow.type) }}</span>
-            <StatusBadge :status="workflow.status" type="workflow" />
           </div>
           <div class="workflow-summary-meta">
             <span v-if="workflow.completedAt">
@@ -182,8 +181,11 @@ const showTasks = computed(() => {
               Started {{ $filters.formatDateTime(workflow.startedAt) }}
             </span>
           </div>
-          <div v-if="workflow.tasks" class="workflow-summary-count">
-            {{ tasks.length }} {{ tasks.length === 1 ? "task" : "tasks" }}
+          <div class="workflow-summary-details">
+            <StatusBadge :status="workflow.status" type="workflow" />
+            <span v-if="workflow.tasks" class="workflow-summary-count">
+              {{ tasks.length }} {{ tasks.length === 1 ? "task" : "tasks" }}
+            </span>
           </div>
         </div>
       </button>
@@ -295,10 +297,10 @@ const showTasks = computed(() => {
   display: grid;
   flex: 1;
   grid-template-areas:
-    "title count"
-    "meta count";
+    "title details"
+    "meta details";
   grid-template-columns: minmax(0, 1fr) max-content;
-  gap: 0.125rem 1rem;
+  gap: 0.25rem 1rem;
   min-width: 0;
   margin-right: 0.75rem;
   text-align: left;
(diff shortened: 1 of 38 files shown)


dashboard/src/components/WorkflowCollapse.vue
```diff
@@ -356,11 +355,11 @@ const showTasks = computed(() => {
 }
 
 .workflow-task-list-item {
-  background: var(--bs-tertiary-bg);
+  background: var(--bs-body-bg);
 }
 
 .workflow-task-list-item:nth-child(even) {
-  background: var(--bs-body-bg);
+  background: var(--bs-tertiary-bg);
```
sevein, line 379: I find the zebra striping a bit visually noisy here, don't you?

Image

sevein (approved):
LGTM.

dashboard/src/components/WorkflowCollapse.vue
jraddaoui: I kind of like it, specially with the dark theme.

image

@fiver-watson @sallain??

sallain: I think rather than light vs dark, I like it when the row contains multiple lines (as in @jraddaoui's screenshot) but less so when there is only one line of text (as in @sevein's). Which really means I could go either way -- though someday, hopefully, we'll see notes for many more tasks. Overall I think it's fairly subtle and not too noisy, but I also don't really feel like I need it, so I'll leave the decision up to others!

sevein: Maybe one option to try is reducing the intensity of the stripes around 20-30%, because right now they almost look like they are competing visually with other elements, like the sidebar right? Is there maybe a more subtle color we could use? The separators could also be a bit more subtle, maybe using a softer row-border constant, so the user can still get the benefit of the stripes without them calling so much attention.

In any case, this shouldn't be a blocker, you have my 👍 to merge!

jraddaoui: Thanks both! I'll wait to hear from @fiver-watson.

I've also changed the light purple background from selected items in the upload from SIP source table, using the same background color. Just a note, for consistency.

image

image

Happy to go on any direction.

jraddaoui: @sevein, I've also added three minor commits more, see below when you have some time.

fiver-watson: I gotta go with @jraddaoui here, I think I like the zebra striping too. Though, I can see what @sevein is saying about it potentially competing visually - perhaps we could try out JGC's suggestion of toning down the stripe contrast a bit, so it is still present but slightly less strong?

I think it looks fine when used for showing checkbox selections - in such a case, you want it to stand out a bit more to confirm your selection. When browsing a long table of results, the zebra striping should be subtle, so your brain can use it to better distinguish between rows but not so much that it becomes the focal point of your attention.

sevein (approved):
👍

dashboard/src/components/WorkflowCollapse.vue
jraddaoui: 50% intensity in background and borders. Considering Sara's comment I have increased a little the padding and reduced the font weight from 600 to 500 in the workflow/task titles.

image
image

sallain: I like this a lot, subtle but still visually helpful!

PR #1738, opened 2026-08-13
Re-enable Dependabot update entries

sevein:
Re-enable Dependabot version updates now that related dependency changes can be grouped into a single pull request per ecosystem. This keeps routine upgrades current without the high PR volume that led us to disable them previously, while preserving focused updates for the application, dashboard, infrastructure modules, documentation, containers, and GitHub Actions.

Dependabot also keeps its open PRs reconciled with the default branch: after a dependency is upgraded elsewhere, the corresponding grouped PR is refreshed to remove or adjust that update. Its branches can be edited like normal PR branches, and Dependabot will rebase them as dependencies or the target branch change.

Example: https://github.com/artefactual/archivematica/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies


jraddaoui (approved):
LGTM! I kind of remember a conversation about [dependency cooldowns](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-), have you consider it here?

PR #1739, opened 2026-08-13
Add dark theme support to the dashboard

jraddaoui:
Initialize the dashboard from the browser color scheme and persist explicit user choices in local storage. Expose the toggle through the user menu and sign-in page.

Adopt Bootstrap color-mode variables across dashboard components, status badges, workflow details, date pickers, and the Uppy uploader. Keep links, focus states, selected rows, and long content readable in both themes.

Allow separate light and dark institution logos with legacy fallback, and document the new settings and CSP requirements.


dashboard/src/theme.ts
```diff
@@ -0,0 +1,52 @@
+import { readonly, ref } from "vue";
```
sevein: Could we implement this with VueUse's [useColorMode](https://vueuse.org/core/useColorMode/)? @vueuse/core is already a dependency, and it has system-theme detection, persistence, reactivity, etc. It would keep changes synchronized across tabs too because it handle storages events.

[useDark](https://vueuse.org/core/useDark/) is another option, but since the rest of the application needs the actual color-mode string rather than only a boolean, useColorMode seems like a better option.

dashboard/src/components/SIPUploadLocal.vue
```diff
@@ -64,6 +69,10 @@ const uppy = new Uppy({
   },
 });
 
+watch(themeController.theme, (theme) => {
+  uppy.getPlugin("Dashboard")?.setOptions({ theme });
+});
+
```
sevein, line 75: I find the uppy's dark theme inner borders so hard to see.

It uses `#333`, tried `#444` and looks better to me.

### `#333`

Image

### `#444`

Image

sevein (approved):
Love it! Two suggestions, but neither of them is blocking.

dashboard/src/components/SIPUploadLocal.vue
jraddaoui: Good catch! I've used the same external border.

dashboard/src/theme.ts
jraddaoui: Thanks, a lot simpler! See ac14e576d62c75d59f47f014a5a574b1b8438668.

PR #1747, opened 2026-08-14
Expose theme state with a composable

sevein:
The former module-level controller exposed VueUse implementation details to each consumer. A conventional useTheme API now provides the shared reactive state through a focused, recognizable interface.

A single useColorMode instance serves every consumer, and initialization before the app mounts prevents a flash of the default theme.


jraddaoui (approved):
Neat, thanks!

PR #1748, opened 2026-08-14
Bump Goa to 3.29.2

sevein:
Goa 3.29.1 adds OpenAPI 3.2 generation, strengthens tagged unions, and fixes invalid parameter metadata.

Regenerated contracts make Enduro's domain safer and easier to consume. Event unions for SIPs, batches, AIPs, workflows, tasks, and locations now bind each discriminator to its matching payload. Location configuration unions do the same for AMSS, S3, SFTP, and URL back ends. Generated decoders reject missing or null values and report missing or unknown discriminators as field validation errors.

The API design now emits 2.0, 3.0, and 3.2 descriptions. The 3.2 description gives the ingest and storage monitor SSE streams typed items whose JSON data points to the corresponding event schema. Regenerated descriptions also remove invalid allowEmptyValue metadata from the upload header and download-ticket cookies. The documentation publishes both 3.0 and 3.2 and uses the latter for Swagger UI. Deterministic normalization also keeps later generated diffs stable.

The dependency refresh advances OpenTelemetry semantic conventions so resource schema URLs remain aligned.


jraddaoui (approved):
Thanks @sevein! Please, add `docs/src/dev-manual/api/openapi3.2.json` to `.gitattributes`.

jraddaoui:
Also, what about the dashboard? Does it need to be updated? Are you planning to use the new spec to generate the client? Will that come in a different PR?

sevein:
Thanks, fixed now. Took the opportunity to bump openapi-generator from 7.21.0 to 7.24.0.

jraddaoui (approved):
👌 thanks!!

sevein:
> Are you planning to use the new spec to generate the client?

openapi-generator doesn't seem to like the 3.2 spec generated by Goa yet. 🤷‍♂️

PR #1752, opened 2026-08-18
Upgrade dashboard dependencies

jraddaoui:
Move the dashboard to Node 26.7.0 and refresh compatible runtime and development dependencies, including Pinia 4, Vue Router 5, Vue Datepicker 14, EventSource 5, TypeScript 6, and current build and test tooling. Update the lockfile, Node preset and types, generated route declarations, dependency docs, and formatter output to match.

Use TypeScript 6 because typescript-eslint 8.67 supports versions below 6.1; TypeScript 7 remains outside its supported peer range. Keep ESLint and @eslint/js on version 9 because eslint-plugin-import 2.32 supports only through ESLint 9, preventing npm peer-resolution failures.

Import ./vite.config.ts explicitly from Vitest to eliminate Vite's native config-loader extension warning. Enable no-emit TypeScript extension imports for config files and provide isolated happy-dom Web Storage globals required by Node 26 test workers.

Migrate the Vue Datepicker custom selectors from the v12 dp__ names to the v14 dp-- names so theme variables and action-button styling apply again.


sevein (approved):
LGTM

PR #1756, opened 2026-08-19
Update Go to 1.27.0

sevein:
Experiments with go 1.27rc1, adopting generic methods and using the new goroutine leak detector to identify and fix three concurrency leaks in SFTP uploads and filesystem watchers.


jraddaoui (approved):
LGTM, lots of good stuff, thanks @sevein!

Issue #1757, opened 2026-08-19
Problem: preservation workflow does not preserve file timestamps


jraddaoui:
### Describe the bug

The Enduro preservation workflow does not preserve the modification times of existing payload files when it extracts an archived SIP.

Enduro's processing workflow invokes the shared `archive-extract` activity for archived SIPs. That activity creates each extracted file and copies its contents, but it does not restore the modification time recorded in the archive entry. The payload files therefore receive the extraction time before the preservation engine sees them.

The unzipped Archivematica transfer path introduces a second loss: Enduro's SFTP client creates each remote file and copies its contents without restoring the source modification time. Archivematica consequently receives a PIP whose file timestamps have already changed.

### To Reproduce

1. Create a ZIP SIP containing a payload file with a known modification time, such as `2001-02-03T04:05:06Z`. Confirm that the ZIP entry records that value before ingest.
2. Ingest the SIP normally, allow the preservation workflow to finish, and retrieve the resulting AIP METS file.
3. In `mets:fileGrp[@USE='original']`, find the `mets:file` for the test file and resolve its `ADMID` to the corresponding `mets:amdSec`.
4. Inspect the PREMIS file object under that administrative metadata section. Compare `premis:dateCreatedByApplication` and the ExifTool `System:FileModifyDate` extension with the timestamp recorded in the input ZIP.
5. Observe that both fields contain the workflow processing time instead of the original modification time.

A focused test against the current `archive-extract` implementation produced:

```text
input=2001-02-03T04:05:06Z
extracted=2026-08-19T17:21:21.263499953Z
```

The alternative zipped transfer path is also affected. After resetting the test file to the original time and running `archive-zip`, the output ZIP member had this timestamp:

```text
input=2001-02-03T04:05:06Z
rezipped=1979-11-30T00:00:00Z
```

In an example completed AIP, all six files in the `original` file group report the same values despite representing different pre-existing files:

```text
premis:dateCreatedByApplication=2026-08-14T18:52:26Z
System:FileModifyDate=2026:08:14 18:52:26+00:00
```

### Expected behavior

The modification time recorded for each original payload file in the submitted archive should be retained through Enduro's extraction, BagIt creation, and transfer to Archivematica. In the completed AIP METS, the original file's `premis:dateCreatedByApplication` and `System:FileModifyDate` should represent that retained value.

#### Data flow and mutation points

- `internal/workflow/processing.go` invokes the shared `archive-extract` activity directly for an archived transfer.
[...]

PR #1762, opened 2026-08-21
Preserve SIP modification times

jraddaoui:
Preserve source mtimes while bundling, moving, downloading, and uploading SIP/PIP content.

Update temporal-activities so archiving, extraction and BagIt creation retain source modification times.

Refs artefactual-sdps/enduro#1757.

### DIFF
--- a/internal/sftp/goclient.go
+++ b/internal/sftp/goclient.go
@@ -114,7 +114,31 @@ func (c *GoClient) UploadDirectory(ctx context.Context, srcPath string) (string,
 func uploadFile(ctx context.Context, src io.Reader, remotePath string, upload *AsyncUploadImpl) {
 	defer upload.Close()
 
-	remoteCopy(ctx, upload, src, remotePath)
+	var info fs.FileInfo
+	if file, ok := src.(fs.File); ok {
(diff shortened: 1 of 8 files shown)


internal/sftp/goclient.go
```diff
@@ -161,6 +207,21 @@ func uploadDirectory(ctx context.Context, srcPath, remoteDir string, upload *Asy
 	})
 	if err != nil {
 		upload.Err() <- fmt.Errorf("goclient: %v", err)
+		return
+	}
+
+	// Creating files and subdirectories updates their parent directories' mtimes,
+	// so restore directory mtimes only after the walk is complete.
+	for _, directory := range directories {
```
sevein, line 224: If we want to check the context before the remote calls:

```go
        if err := ctx.Err(); err != nil {
                upload.Err() <- fmt.Errorf(
                        "goclient: restore directory modification times: %v",
                        err,
                )
                return
        }
```

sevein (approved):
Looks good to me! I wonder if SFTP timestamps using `uint32` could be an issue for dates before 1970. Probably an edge case.

PR #1774, opened 2026-08-28
Add docs for uploading SIPs as a batch, plus glossary term

sallain:
These are docs for the process of uploading SIPs as part of a batch, plus some screenshots and a glossary term.

I'll add docs for searching and browsing the Batches page in another PR.


docs/src/user-manual/ingest/submitting-content.md
```diff
@@ -170,6 +178,40 @@ The deposit or upload process will depend on the source location used.
...
+1. To process a selection of SIPs as a batch, follow steps 1-3 above but before
+   clicking on Start Ingest, click on the toggle next to "Treat this ingest as a
+   batch?"
+
+    ![The batch toggle](../screenshots/sip-source-add-batch-id.png)
+
+2. If you wish, you can add a custom batch identifier. This identifier will be
+   used to link the SIPs together and to search for the batch on the Batches
+   page. If you do not enter a custom batch identifier, Enduro will generate one
+   using a 32-digit UUID.
```
fiver-watson: Just a thought: is it worth including a TIP admonition or similar here to mention where, if at all, this batch ID ends up in the final AIP metadata? (I recall us discussing preserving it as an accession ID but honestly can't remember if we did anything like that or not... but if it does end up in the AIP METS somewhere, we should probably call this out)

fiver-watson (approved):
One  minor comment added for consideration (I honestly couldn't remember the answer myself, sorry), but otherwise it looks great!

docs/src/user-manual/ingest/submitting-content.md
sallain: It does not end up in the METS! Only used in Enduro. That might be helpful to note anyway, though.

